PR #15, opened 2015-04-28
I need to be able to add a new line

jpdriver:
- [x] fix ViewportScrollMixin to render new rows


bakesteve:
no way to get a (would have been) failing test in here?

PR #19, opened 2015-06-30
Allow Grid to use Immutable JS objects as rows and columns to increase performance

malonecj:
- [x] Move BaseGrid state.columns to ReactDataGrid level
 - [x] Rename state.columns to state.columnMetrics
 - [x] Allow Grid to take Immutable JS columns and props
 - [x] Immutable Data tests
 - [x] Immutable Data example 
 - [x] Fix failing Tests


bakesteve:
needs a catchup

src/Cell.js
```diff
@@ -229,12 +233,23 @@ var Cell = React.createClass({
...
+  },
+
+  applyUpdateClass() {
+    var updateCellClass = this.getUpdateCellClass();
+    // -> removing the class
+    if(updateCellClass != null) {
+      this.getDOMNode().classList.remove(updateCellClass);
+      // -> triggering reflow /* The actual magic */
+      // without this it wouldn't work. Try uncommenting the line and the transition won't be retriggered.
+      this.getDOMNode().offsetWidth = element.offsetWidth;
```
bakesteve: from a perf point of view, do we actually want this to do a reflow??

src/ColumnMetrics.js
```diff
@@ -27,60 +27,68 @@ type Column = {
...
  */
-function calculate(metrics: ColumnMetricsType): ColumnMetricsType {
-  var width = 0;
-  var unallocatedWidth = metrics.totalWidth;
-
-  var deferredColumns = [];
-  var columns = metrics.columns.map(shallowCloneObject);
-
-  var i, len, column;
+function recalculate(metrics: ColumnMetricsType): ColumnMetricsType {
```
bakesteve: dont we need to take account of hidden columns here?

src/ColumnMetricsMixin.js
```diff
@@ -51,41 +47,63 @@ module.exports = {
...
+    if(Array.isArray(columns)){
+      return columns[idx];
+    }else if (typeof Immutable !== 'undefined') {
+      return columns.get(idx);
+    }
+  },
+
+  getColumnCount() {
+    var columns = this.state.columnMetrics.columns;
+    if(Array.isArray(columns)){
```
bakesteve: wonder if we should ship a flag for this to avoid the type checks?
either a high level state to indicate my cols are immutable, or even a code flag?

src/Grid.js
```diff
@@ -57,26 +57,27 @@ var Grid = React.createClass({
...
           headerRows={headerRows}
           sortColumn={this.props.sortColumn}
           sortDirection={this.props.sortDirection}
+          onSort={this.props.onSort}
           />
         <div ref="viewPortContainer" onKeyDown={this.props.onViewportKeydown} onDoubleClick={this.props.onViewportDoubleClick}   onDragStart={this.props.onViewportDragStart} onDragEnd={this.props.onViewportDragEnd}>
             <Viewport
               ref="viewport"
-              width={this.state.columns.width}
+              width={this.props.columnMetrics.width}
```
bakesteve: where does `this.props.columnMetrics`.? its not in propTypes and cant see it in the diff?

src/HeaderRow.js
```diff
@@ -26,9 +34,13 @@ var HeaderRow = React.createClass({
     height: PropTypes.number.isRequired,
     columns: PropTypes.arrayOf(React.PropTypes.shape(ExcelColumn)).isRequired,
     onColumnResize: PropTypes.func,
+    onSort: PropTypes.func.isRequired,
     style: PropTypes.shape(HeaderRowStyle)
+
   },
 
+  mixins: [ColumnUtilsMixin],
```
bakesteve: is there much point this being a mixin, rather than just a lib? ie `this.getColumn(this.props.columns, i);` as a mixin vs `ColumnUtils.getColumn(this.props.columns, i);` as a lib? means we are 1 step cloiser to es6 classes

src/__tests__/ColumnMetrics.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    describe('Using ImmutableJs Lists', () => {
+
+      beforeEach(() => {
+        var helpers = require('./GridPropHelpers');
+        this.prevColumns = new Immutable.List(helpers.columns);
+        this.nextColumns = this.prevColumns;
+      });
+
+      it('columns with same memory reference are equal', () => {
+                debugger;
```
bakesteve: no debugger statements (we can lint that, right?)

src/addons/__tests__/Grid.spec.js
```diff
@@ -131,80 +131,80 @@ describe('Grid', () => {
...
-      var baseGrid = TestUtils.findRenderedComponentWithType(component, BaseGridStub);
-      var selectRowCol = baseGrid.props.columns[0];
-      debugger;
-      var fakeEvent = {stopPropagation : function(){}};
-      selectRowCol.onRowSelect(3, fakeEvent);
-      expect(component.state.selectedRows[3]).toBe(false);
-    });
-  });
-
+  // describe("When row selection enabled", () => {
```
bakesteve: how come we are getting rid of this?

src/addons/grids/ReactDataGrid.js
```diff
@@ -446,7 +448,8 @@ var ReactDataGrid = React.createClass({
   handlePaste(){
     if(!this.copyPasteEnabled()) { return; }
       var selected = this.state.selected;
-      var cellKey = this.getColumns()[selected.idx].key;
+      debugger;
```
bakesteve: debugger

bakesteve:
ok - basically all :+1: 
be nice to think of a way to avoid the checks for immutable - what do you think of a state var? or just use feature flag and ship 2 versions?

src/Cell.js
malonecj: performance hit is negligible here since this is only applied to those cells that are updating.

Still, it would be preferable to calculate this class on render and apply the highlight animation class rather than after componentDidUpdate  

This is what I was trying to do for a demo, but I didnt get it to work quickly so went with this option instead. 

Another related problem with this, is that once the highlight has been applied to an updated cell. If we try to apply the highlight class again, nothing happens. Even if the class is first removed. 

I will create a github issue in case I forget

malonecj: https://github.com/adazzle/react-data-grid/issues/26

src/ColumnMetrics.js
malonecj: reactDataGrid has no concept of hidden columns. Another one for the issues list methinks

https://github.com/adazzle/react-data-grid/issues/27

src/Grid.js
malonecj: It comes from ColumnMetricsMixin in ReactDataGrid. I forgot to add it to the propTypes. Will fix

We should implement the AirBnB linting file with babel eslint so that we never miss a propType

malonecj: https://github.com/adazzle/react-data-grid/issues/28

src/HeaderRow.js
malonecj: I dont think it needs to be a mixin, no. It doesn't use any internal component attributes or lifecycle events so it can be a separate file

src/__tests__/ColumnMetrics.spec.js
malonecj: yep

src/addons/__tests__/Grid.spec.js
malonecj: ah I think I branched off some of Mo's work. This should be fixed in master. I'll do a catchup

aqilzeeshan:
@bakesteve - Conor said everything is done. If you are happy with it, can you please merge it.

malonecj:
I just need to do a catchup first

src/Cell.js
bakesteve: or feels like some of teh react animation stuff that was big back in teh day might help? feels like render isnt the right option for something we want to be a temporal flash/highlight?

src/Grid.js
bakesteve: taht forces propTypes - winner. had looked at doing that at runtime by hooking into the reactPerf stuff that lets you touch in at various points in the chain, but as-you-hit-save time is way better!

PR #20, opened 2015-06-30
Select individual and all check boxes 

momaci:
- [x] Select all checkboxes 
- [x] select row checkbox 
- [x] update tests

### DIFF
--- a/src/addons/editors/CheckboxEditor.js
+++ b/src/addons/editors/CheckboxEditor.js
@@ -8,7 +8,6 @@ var React                   = require('react');
 
 var CheckBoxEditor = React.createClass({
 
-
   PropTypes : {
     value : React.PropTypes.bool.isRequired,
     rowIdx : React.PropTypes.number.isRequired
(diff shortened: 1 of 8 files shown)


src/addons/editors/CheckboxEditor.js
```diff
@@ -15,17 +15,14 @@ var CheckBoxEditor = React.createClass({
   },
 
   render(): ?ReactElement{
-    return (<input className="react-grid-CheckBox" type="checkbox" checked={this.props.value} onChange={this.handleChange} />);
+    var checked = this.props.value != null ? this.props.value : false;
+    return (<input className="react-grid-CheckBox" type="checkbox" checked={checked} onClick={this.handleChange} />);
   },
 
   handleChange(e: Event){
+    e.stopPropagation();
```
bakesteve: do we always want to stop propagation? wouldnt it be better to possibly pass the event through to onRowSelect?

src/addons/grids/ReactDataGrid.js
```diff
@@ -341,7 +341,7 @@ var ReactDataGrid = React.createClass({
   },
 
   handleRowSelect(row: Row){
-    var selectedRows = this.state.selectedRows;
+    var selectedRows = this.state.selectedRows.slice(0);
```
bakesteve: what if state or state.selectedRows is null or empty?

src/addons/editors/CheckboxEditor.js
momaci: of yeah, that was there while we were trying things, I'll try it without it

PR #21, opened 2015-07-01
checkboxes fix continued

momaci:
- [x] set initial selected rows state
- [x] add more tests for checkbox comp

### DIFF
--- a/src/addons/editors/__tests__/CheckboxEditor.spec.js
+++ b/src/addons/editors/__tests__/CheckboxEditor.spec.js
@@ -35,20 +35,14 @@ describe('CheckboxEditor', () => {
        var fakeEvent = {stopPropagation : function(){}};
        expect(testColumn.onRowSelect.mostRecentCall.args[0]).toEqual(1, fakeEvent);
      });
- });
 
- describe('Unchecked tests', () => {
-   beforeEach(() => {
-     component = TestUtils.renderIntoDocument(<CheckboxEditor
-       value={false}
-       rowIdx={1}
-       column={testColumn}/>);
+     it('should not be selected if value prop is false', () => {
+       component.setProps({value: false});
+       var Input = TestUtils.findRenderedDOMComponentWithTag(component, 'input');
+       expect(Input.props.checked).toBe(false);
      });
-
-   it('should not be selected if value prop is false', () => {
-     var Input = TestUtils.findRenderedDOMComponentWithTag(component, 'input');
-     expect(Input.props.checked).toBe(false);
-   });
  });
+
+
 
 });
--- a/src/addons/editors/__tests__/CheckboxEditor.spec.js
+++ b/src/addons/editors/__tests__/CheckboxEditor.spec.js
@@ -5,23 +5,36 @@ var TestUtils      = require('react/lib/ReactTestUtils');
 var CheckboxEditor = require('../CheckboxEditor');
 
 describe('CheckboxEditor', () => {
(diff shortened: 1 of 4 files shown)


src/addons/editors/__tests__/CheckboxEditor.spec.js
```diff
@@ -5,23 +5,44 @@ var TestUtils      = require('react/lib/ReactTestUtils');
...
+     });
+
+     it('should create a new CheckboxEditor instance', () => {
+       expect(component).toBeDefined();
+     });
+
+     it('should be selected if value prop is true', () => {
+       var Input = TestUtils.findRenderedDOMComponentWithTag(component, 'input');
+       expect(Input.props.checked).toBe(true);
+     });
```
malonecj: we should be looking at what we're rending here instead of its props

so 
`
var checkboxNode = Input.getDOMNode();
expect(checkBoxNode.checked).toBe(true)
`

syntax might be slightly wrong

src/addons/editors/__tests__/CheckboxEditor.spec.js
```diff
@@ -5,23 +5,44 @@ var TestUtils      = require('react/lib/ReactTestUtils');
...
+
+     it('should call onRowSelect with correct RowIdx when checkbox is clicked', () => {
+       var Input = TestUtils.findRenderedDOMComponentWithTag(component, 'input');
+       TestUtils.Simulate.click(Input.getDOMNode());
+       expect(testColumn.onRowSelect).toHaveBeenCalled();
+       var fakeEvent = {stopPropagation : function(){}};
+       expect(testColumn.onRowSelect.mostRecentCall.args[0]).toEqual(1, fakeEvent);
+     });
```
malonecj: :+1:

src/addons/grids/ReactDataGrid.js
```diff
@@ -93,12 +93,21 @@ var ReactDataGrid = React.createClass({
     var initialState = {selectedRows : [], copied : null, expandedRows : [], canFilter : false, columnFilters : {}, sortDirection: null, sortColumn: null, dragged : null}
     if(this.props.enableCellSelect){
       initialState.selected = {rowIdx: 0, idx: 0};
+      initialState.selectedRows = this.getInitialSelectedRows();
     }else{
```
malonecj: this should be outside check for enableCellSelect
That is something different, and not always set
Should be  var `initialState = {selectedRows : this.getInitialSelectedRows(),...`

momaci: yeah, I forgot to move that

src/addons/editors/__tests__/CheckboxEditor.spec.js
```diff
@@ -5,23 +5,47 @@ var TestUtils      = require('react/lib/ReactTestUtils');
...
+       var checkboxNode = Input.getDOMNode();
+       expect(checkboxNode.checked).toBe(true);
+     });
+
+     it('should not be selected if value prop is false', () => {
+       component.setProps({value: false});
+       var Input = TestUtils.findRenderedDOMComponentWithTag(component, 'input');
+       var checkboxNode = Input.getDOMNode();
+       debugger;
+       expect(checkboxNode.checked).toBe(false);
```
malonecj: can you remove the debugger ref please

malonecj:
Looks good Mo. Just remove that debugger statement and I will merge in

PR #22, opened 2015-07-02
update version

momaci:
- [x] rename onRowSelect to onCellChange
- [x] pass column key to onCellChange
- [x] export checkbox editor 
- [x] update version

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ shallow_clone: true
 # Install scripts. (runs after repo cloning)
 install:
   # Get the latest stable version of Node.js or io.js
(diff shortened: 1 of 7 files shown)


appveyor.yml
```diff
@@ -11,7 +11,7 @@ shallow_clone: true
 # Install scripts. (runs after repo cloning)
 install:
   # Get the latest stable version of Node.js or io.js
-  - ps: Install-Product node $env:nodejs_version
+  - ps: Install-Product node 0
   - npm -g install npm@2.11.3
```
momaci: the builds on appveyor were failing because some packages were failing to build with the latest version of io.js so changed it to use node.

PR #23, opened 2015-07-06
Cell event bubbling

bakesteve:
Only prevent events bubbling where we actually handle them
Previously we did preventDefault before checking if this cell had a click event.
It also removes stopPropagation as others may want to grab hold of this event, so unless there is a good reason not to - we should not swallow.
Where a good reason exists, we can swallow at a lower level (or add an option to do that via cellMetaData?)

That means you cant do things like a simple row based onClick in an overriden rowRenderer - This fixes that


malonecj:
closing this due as has been merged into other branch which is going into master

PR #31, opened 2015-07-10
RW AutoCompleteEditor Fix & Test

richardware:
(Also @malonecj removed some bad practice tests)


malonecj:
failing test. Can you have a look please

PR #33, opened 2015-07-13
use babel es6 not jsx

bakesteve:
babel-loader not deprecated jsx (also means we get es6 sugar)
fix some issues with column metrics transpiling (cant use this inside a jasmine spec)
create a GridRunner class that encapsulates a bunch of test logic to make new integration tests way easier


bakesteve:
this closes #12

PR #38, opened 2015-07-22
fix broken column resize

jpdriver:
- [x] use .get() if it exists

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-data-grid",
-  "version": "0.12.14",
+  "version": "0.12.15-alpha1",
   "description": "Data grid for React",
   "main": "src/index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "moment": "^2.8.1",
     "pretty-hrtime": "^0.2.1",
     "react-highlight": "^0.4.1",
-    "react-router": "^0.11.6",
+    "react-router": "0.11.6",
     "react-tools": "^0.11.2",
     "rewire": "^2.1.3",
     "rewire-webpack": "^1.0.0",
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -234,9 +234,6 @@ var Cell = React.createClass({
     // -> removing the class
     if(updateCellClass != null && updateCellClass != "") {
       this.getDOMNode().classList.remove(updateCellClass);
-      // -> triggering reflow /* The actual magic */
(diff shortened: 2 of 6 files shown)


src/ColumnMetrics.js
```diff
@@ -99,7 +99,8 @@ function setDefferedColumnWidths(columns, unallocatedWidth, minColumnWidth) {
  * @param {number} width
  */
 function resizeColumn(metrics: ColumnMetricsType, index: number, width: number): ColumnMetricsType {
-  var column = metrics.columns[index];
+  // use get if we have it
+  var column = metrics.columns.get ? metrics.columns.get(index) : metrics.columns[index];
```
malonecj: use ColumnUtils wrapper to get an individual column

src/Cell.js
```diff
@@ -234,9 +234,6 @@ var Cell = React.createClass({
     // -> removing the class
     if(updateCellClass != null && updateCellClass != "") {
       this.getDOMNode().classList.remove(updateCellClass);
-      // -> triggering reflow /* The actual magic */
-      // without this it wouldn't work. Try uncommenting the line and the transition won't be retriggered.
```
bakesteve: so this is genuinely no longer needed?

package.json
```diff
@@ -79,7 +79,7 @@
     "moment": "^2.8.1",
     "pretty-hrtime": "^0.2.1",
     "react-highlight": "^0.4.1",
-    "react-router": "^0.11.6",
+    "react-router": "0.11.6",
```
bakesteve: shouldnt this be a ~ or are we saying we ONLY want 0.11.6 (and NOT 0.11.7/8/9?)

src/ColumnMetrics.js
```diff
@@ -99,7 +99,8 @@ function setDefferedColumnWidths(columns, unallocatedWidth, minColumnWidth) {
  * @param {number} width
  */
 function resizeColumn(metrics: ColumnMetricsType, index: number, width: number): ColumnMetricsType {
-  var column = metrics.columns[index];
```
bakesteve: I dont think we should be doing this
inside columnMetrics, surely we already know if a column is an immutable or not?
Looks like the `ColumnMetricsType` definition needs to be update too - but assuming `metrics.columns: ImmutableList<Column>` and not `metrics.columns: ImmutableList<Column> | Aray<Column>`
obviously then means we need to trace back what is changing columns from ImmutableList -> Array?

PR #39, opened 2015-07-23
fix broken column resize 

jpdriver:
- [x] make All Features examples resizable
- [x] use ColumnUtils.getColumn() before performing updates
- [x] add updateMetrics helper to slice in updatedColumn

### DIFF
--- a/src/ColumnUtils.js
+++ b/src/ColumnUtils.js
@@ -8,7 +8,7 @@ module.exports = {
     }
   },
 
-  updateMetrics(metrics, idx, column){
+  spliceColumn(metrics, idx, column){
     if(Array.isArray(metrics.columns)){
       metrics.columns.splice(idx, 1, column);
     }else if (typeof Immutable !== 'undefined') {
--- a/src/ColumnUtils.js
+++ b/src/ColumnUtils.js
@@ -8,6 +8,15 @@ module.exports = {
     }
   },
 
+  updateMetrics(metrics, idx, column){
+    if(Array.isArray(metrics.columns)){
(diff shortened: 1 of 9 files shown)


src/ColumnUtils.js
```diff
@@ -8,6 +8,15 @@ module.exports = {
     }
   },
 
+  updateMetrics(metrics, idx, column){
```
malonecj: we should change the function name to splice as that is what is happening here

src/ColumnUtils.js
```diff
@@ -8,6 +8,15 @@ module.exports = {
...
+    if(Array.isArray(metrics.columns)){
+      metrics.columns.splice(idx, 1, column);
+    }else if (typeof Immutable !== 'undefined') {
```
malonecj: this is fine for now - but we should set a flag when the grid is initialised (most probably on the column metrics object) to say whether we are using immutable data or not

malonecj:
needs tests please

aqilzeeshan:
@malonecj please review it.

src/__tests__/Header.spec.js
```diff
@@ -43,11 +43,44 @@ describe('Header Unit Tests', () => {
...
+      testProps.columnMetrics.columns = new Immutable.List(helpers.columns);
+      header = TestUtils.renderIntoDocument(<Header {...testProps}/>);
+    });
+
+    it('should render a default header row', () => {
+      shouldRenderDefaultRow();
   });
 
   it('header row drag start should set resize column state ', () => {
+                  debugger;
```
malonecj: remove debugger statement

PR #45, opened 2015-08-05
can't build on darwin

jpdriver:
- [x] enable flow for *nix
- [x] update karma to 0.13.3
- [x] deprecate jsx-loader (see https://github.com/petehunt/jsx-loader#dont-use-this)

Before
---------
screen shot 2015-08-05 at 11 25 04 am

After
------
screen shot 2015-08-05 at 6 38 59 pm


gulp/tasks/flow.js
```diff
@@ -1,11 +1,30 @@
...
+gulp.task('flow', function() {
+  if (process && process.platform === 'win32') {
+    //due to flow-bin not working on windows (because there isnt an official build https://github.com/facebook/flow/issues/6)
+    execFile('./flow/flow.exe', ['check','--lib','./flow/libs','--strip-root','./flow'],function (err, stdout, stderr) {
+      if(err) console.log('Error:' + err);
+      if(stdout) console.log(stdout);
+    });
+  } else {
+    //this is how we WANT to run flow
+    var flow = require('gulp-flowtype');
```
malonecj: does this actually have flow installed, or does it require flow to be setup on local machine before?

jpdriver: it's an optionalDependency which means AppVeyor can still build even though it tries and fails.

for os x flow gets installed normally and doesn't require the manual npm install step :)

gulp/tasks/flow.js
```diff
@@ -1,11 +1,30 @@
 var gulp = require('gulp');
 var execFile = require('child_process').execFile;
 
-gulp.task('flow-win', function() {
-  execFile('./flow/flow.exe', ['check','--lib','./flow/libs','--strip-root','./flow'],function (err, stdout, stderr) {
-    if(err) console.log('Error:' + err);
-    if(stdout) console.log(stdout);
-  });
+gulp.task('flow', function() {
+  if (process && process.platform === 'win32') {
```
bakesteve: or win64? or is there an npm package for this?

jpdriver: win64 isn't valid. platform will be one of:-

'darwin', 'freebsd', 'linux', 'sunos', 'win32'

https://nodejs.org/api/process.html#process_process_platform

bakesteve:
Needs a catchup

bakesteve:
failed on AV on flow-build...

jpdriver:
true but for completely unrelated tests... :(

PR #56, opened 2015-08-25
return the underlying cell value

momaci:


bakesteve:
shouldn't this just be falsey? so null ~= false => unchecked?
also needs a test to assert whichever we go on this

momaci:
no, becasue if the value it's false we want to show as un-ticked but with the ability to tick it, and if we don't have a value at all (e.g no pre-fetch tag) that just 'n/a'

bakesteve:
Don’t think this should happen in the root of the grid, that feels like something for a specific use case?

So where you are assigning your editor, you just need a wrapper, or make the column.editable a func that can look at the property in question and return true / false

But if you ask for a CheckboxEditor, I think you should get an editor, and not a CheckboxEditorUnlessItsNullValue?

momaci:
Yeah, I was aware it was a bit hacky..  I added it there to see it working, and to get some feedback from conor.  But a TagsCheckboxEditor sounds like a better way to go!?

Sent from mobile
On Tue, Aug 25, 2015 at 6:23 pm, Steven Baker <notifications@github.com<mailto:notifications@github.com>> wrote:

Don't think this should happen in the root of the grid, that feels like something for a specific use case?

So where you are assigning your editor, you just need a wrapper, or make the column.editable a func that can look at the property in question and return true / false

But if you ask for a CheckboxEditor, I think you should get an editor, and not a CheckboxEditorUnlessItsNullValue?



-

bakesteve:
Yeah, probably one for adazzleUi vs the core grid I'd say. Though having the ability to make column.Editable a function, that can vary by row, seems the ideal? Conor?

malonecj:
Yeah exactly. We should have a check in ReactDataGrid.js here 
https://github.com/adazzle/react-data-grid/blob/master/src/addons/grids/ReactDataGrid.js#L323

that checks if column.editable is a function first. If so call it with the selected row and return the value

malonecj:
@bakesteve agreed that it would be good to make column.editable a function

but for the purpose of this use case the checkbox is actually a formatter so behaves slightly differently. Was thinking that this should be handled in a custom formatter in adazzle ui so that the formatter displays the checkbox on one condition and n/a in another?

momaci:
So , I'll go with `TagsCheckbox` for now and we'll implement col.editable later.

bakesteve:
Ok – and as conor said, move it -> adazzleUI

src/Row.js
```diff
@@ -104,15 +104,14 @@ var Row = React.createClass({
...
     } else if (typeof this.props.row.get === 'function') {
       val = this.props.row.get(key);
+    } else {
+      val = this.props.row[key];
     }
-    else {
-      var val = this.props.row[key];
-    }
-    return !val ? '' : val;
+    return val;
```
momaci: @malonecj  as we discussed it on slack , returning the underlying value here, so that if the value is false is not replaced with an empty string.

PR #67, opened 2015-09-18
Fix grid width issues

bor3ham:
Grid was self-referential and did not adjust properly on window resize.

Also changed grid to occupy width minus scrollbar, so with ample horizontal space there will be no horizontal scroll.

Some of the original settings seemed like weird choices, so maybe I'm missing something and the actual cause of these problems are related to my project that uses the grid.

Note that these changes have only been tested in chrome on osx.

### DIFF
--- a/src/ColumnMetricsMixin.js
+++ b/src/ColumnMetricsMixin.js
@@ -30,7 +30,7 @@ module.exports = {
 
   DOMMetrics: {
     gridWidth(): number {
-      return React.findDOMNode(this).offsetWidth - 2;
+      return React.findDOMNode(this).parentElement.offsetWidth;
     }
   },
 
--- a/src/ColumnMetrics.js
+++ b/src/ColumnMetrics.js
@@ -10,6 +10,7 @@ var shallowCloneObject            = require('./shallowCloneObject');
 var isValidElement = require('react').isValidElement;
 var sameColumn = require('./ColumnComparer');
 var ColumnUtils = require('./ColumnUtils');
+var getScrollbarSize  = require('./getScrollbarSize');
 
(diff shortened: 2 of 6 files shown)


src/ColumnMetricsMixin.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
 
   DOMMetrics: {
     gridWidth(): number {
-      return React.findDOMNode(this).offsetWidth - 2;
+      return React.findDOMNode(this).parentElement.offsetWidth;
```
malonecj: This makes sense. We should rename `DOMMetrics.gridWidth` to `DOMMetrics.gridContainerWidth`

malonecj:
Hey, thanks for the pull request. I've tried this out on other browsers and looks good. 

We would like some degree of unit testing for each PR we accept. So I think if we had a test in ```src/addons/__tests__/Grid.spec.js``` that rendered the grid into a container of a set width and verified that the width of the ```<BaseGrid/>``` component = ```containerWidth - scrollbarSize``` then we would be happy to merge this. 

Also need to address the one failing test and make sure the build is passing. Cheers

SerialVelocity:
Hey, is anything being done about this? It's been sitting here for about a month when it's addressing quite a big bug. Any chance you could write a simple test and get it merged @malonecj?

SerialVelocity:
Also, before the scrollbar was always offscreen but now only 4 pixels out of 15 are off the screen on Mac/Chrome. So it does partially fix another bug in your library too.

malonecj:
I wont get time to work on react data grid this week, but I've tested this pull request and agree it is a needed change. I will merge in and get a test for it later

Issue #78, opened 2015-09-30
Double Event triggered on Editing Cells


gtkovas:
I'm attempting to use the grid as an interactive view into a database, allowing the user to edit/add entries by directly editing cells in the grid. However, I'm getting some strange behaviour apparently dependent on the key that is used to complete the edit:

Simply clicking away from the cell works fine.
Pressing 'Enter' fires two copies of the update event (this is an issue due to the asynchronicity making catching "repeat" events almost impossible)
Pressing 'Tab' fires two events AND fails to re-render the grid/cell (refocusing makes it show the correct info)
Using the arrow keys presents the same issues as the 'Tab' key.

Is there some particular part of the grid definition I might be doing wrong that could cause these errors? Or is there any other advice/help to offer? 

I can't seem to attach a .txt of the code file ("something went really wrong"...), so here is the part defining the grid:

var DataGrid = React.createClass({

  getInitialState: function()
  {
      console.log("Getting initial state");
      var rows = getRows();
      //setGrid(this);
      return {rows : rows};
  },

  rowGetter: function(rowIdx) 
  {
      return this.state.rows[rowIdx];
  },

  handleRowUpdated: function(e)
  {   
      if(this.state.rows[e.rowIdx][e.cellKey] != e.updated[e.cellKey]/* && e.key == 'Enter'*/)
      {//table doesn't seem to update properly (though using forceUpdate below...) if the user
	  //uses a key other than "Enter". Using Tab or the arrow keys doesn't do it.
	  //The database updates, but the shown data doesn't change - refocusing on the element
	  //changes it. For now, restricting to the "Enter" key.
	  console.log(JSON.stringify("Last Event is: " + this.state.lastEvent));
	  console.log(JSON.stringify(e));
	  console.log(JSON.stringify(e) == JSON.stringify(this.state.lastEvent));
	  this.setState({lastEvent: e});
	  var boxNumber = this.state.rows[e.rowIdx]['box_num'];
	  var key = e.cellKey;
	  var data = e.updated[e.cellKey];
	  var sendString = '{"key":"' + key + '", "data":"' + data + '"}';
	  e.boxNumber = boxNumber;
	  console.log("Sending: " + sendString);
	  jQuery.ajax({ 
	      url: ("/CollectionBox/edit/" + boxNumber), 
		  type: 'POST', 
	      dataType: 'text', 
	      data: sendString, 
	      success: function(rMessage) 
	      {
		  changeNotice(rMessage);
		  
[...]

PR #79, opened 2015-10-01
Fix double commit event

malonecj:


malonecj:
@bakesteve here's a fix for https://github.com/adazzle/react-data-grid/issues/78

bakesteve:
Closes #78

PR #81, opened 2015-10-02
bug fixes for column header filtering

bor3ham:
Note: still has bug with focus when typing into filter text input.

Focus is stolen by the Cell's checkFocus() from componentDidUpdate after the grid is re-rendered with changes.

Haven't figured out how to fix it yet, although it seems as though the issue has already been encountered with the filterableHeaderCell's componentDidUpdate method. However, the hack solution existing here does not always work (especially when the render method of the grid cells take a long time).

### DIFF
--- a/examples/scripts/example09-filterable-grid.js
+++ b/examples/scripts/example09-filterable-grid.js
@@ -37,37 +37,45 @@ var columns = [
 {
   key: 'id',
   name: 'ID',
-  width: 80
+  width: 80,
+  filterable: true,
+  filterKey: 'filter_id'
 },
 {
   key: 'task',
   name: 'Title',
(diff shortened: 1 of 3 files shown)


bor3ham:
Updated example 09 - strangely, focus bug mentioned above is not manifesting in the example.

Note: Filters do not currently work properly in conjunction with sorting.

matt-d-rat:
I'm still seeing this issue. The grid I have implemented is based on the [basic use](http://adazzle.github.io/react-data-grid/examples.html#/basic) example. It appears to manifest when an event listener causes a change in state, and if the data set happens to be bound to the state of the component, irrelevant of whether the data has changed or not.

PR #88, opened 2015-10-13
copy value ranges

jpdriver:
- [x] make SimpleTextEditors textareas
- [ ] intercept commits containing \r\n and \t
- [ ] get the col + row co-ords of the current cell
- [ ] use those to calculate corresponding col + row co-ords of the other cells in the range
- [ ] generate the other commits


bakesteve:
closing this off for now

PR #91, opened 2015-10-23
focus hack patch, to solve incorrect focus on re-rendering

bor3ham:
This patch fixes cells focusing when they shouldn't - for example:

- When the initial grid rendering is delayed, it can steal focus from other elements on the page that are in use by the user (eg. form input boxes completely unrelated to this react component)
- In some circumstances when re-rendering the grid (for example when a header filter changes)

It's definitely not elegant (relies on hard coded class name rendered in parent element, traverses dom elements with loops), but maybe can serve as inspiration for a more long-term fix.

### DIFF
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -203,7 +203,31 @@ var Cell = React.createClass({
 
   checkFocus: function() {
     if (this.isSelected() && !this.isActive()) {
-      React.findDOMNode(this).focus();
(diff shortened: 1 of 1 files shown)


malonecj:
thanks for the PR. Yes it would be good to prevent the data grid from stealing the focus from other inputs. At the moment there are some failing tests around cell navigation so these will need to be addressed, cheers

Issue #82, opened 2015-10-02
Last column's idx is off when enableRowSelect is true


mikeastock:
When you set `enableRowSelect` to true a select column is prepended to the the columns stored in state. However the columns that are stored in props remain the same. 

Having the props columns being different than the state columns causes the `idx` to be off by one.  This makes getting the last column return `undefined` for [this line](https://github.com/adazzle/react-data-grid/blob/master/src/addons/grids/ReactDataGrid.js#L324).

```js
var col = this.getColumn(this.props.columns, idx);
```

This issue can also be found on the examples page. If try to edit the furthest right column you can see an exception thrown in the console. http://adazzle.github.io/react-data-grid/examples.html#/all-features-immutable.

mikeastock:
I'm open to any suggestions on how to fix it. Let me know what I can do to help.

bakesteve:
Ah – sounds like a case where we are just not properly propagating the state through. We’ll have a look at the best way to fix this, but fairly sure it will be pretty trivial now you’ve shown exactly how to reproduce

PR #96, opened 2015-11-05
Fix - Last column's idx is off when enableRowSelect is true

malonecj:


malonecj:
fixes this issue https://github.com/adazzle/react-data-grid/issues/82

PR #99, opened 2015-11-09
Support even and odd row classes.

omarish:
Currently, the `react-grid-Row--even` and `react-grid-Row--odd` classes are not appearing properly due to a quoting bug. Here's a screenshot of the classes that are being rendered before the patch.

image

### DIFF
--- a/src/Row.js
+++ b/src/Row.js
@@ -46,7 +46,7 @@ var Row = React.createClass({
   render(): ?ReactElement {
     var className = joinClasses(
       'react-grid-Row',
-      "react-grid-Row--${this.props.idx % 2 === 0 ? 'even' : 'odd'}"
+      'react-grid-Row--' + this.props.idx % 2 === 0 ? 'even' : 'odd'
     );
 
     var style = {


omarish:
Also, not sure what's going on here with CI, but is the CI server functioning properly?

image

bakesteve:
hmm - looks odd - that build timed out. 
I've just kicked it off to check it was just a temporary thing

omarish:
Nope, looks like it happened again

bakesteve:
@malonecj AV timing out - looks like node-gyp issue to me? Assume someone released a broken on windows package again?

bakesteve:
ok @omarish this looks like a windows build issue with node-gyp. probably to do with max directory paths from what I can tell..
but nothing to do with your PR. We'll get that sorted and then merge this in - thanks for the fix

omarish:
Sounds good - let me know if there's anything I can do to help.

src/Row.js
```diff
@@ -46,7 +46,7 @@ var Row = React.createClass({
   render(): ?ReactElement {
     var className = joinClasses(
       'react-grid-Row',
-      "react-grid-Row--${this.props.idx % 2 === 0 ? 'even' : 'odd'}"
+      'react-grid-Row--' + this.props.idx % 2 === 0 ? 'even' : 'odd'
```
alistairjcbrown: A more simple solution would be to replace the double quotes `"` with backticks `` ` `` to make use of the new ES6 string interpolation

malonecj: nice, yes we'll be releasing a style guide for contributing soon, and we'll enforce that with a linting build task. For the moment I will merge this in and will review all string interpolation shortly

PR #109, opened 2015-11-13
Allow columns to be removed and added after initial render

alistairjcbrown:
This updated to have `createColumnMetrics` use the current props from `componentWillReceiveProps` so that columns can be added and removed. In tandem with this is an update to the cell key, which now needs to make reference to the column id so that it is re-rendered when a column is removed.

Row also gets updated to use backticks so that the odd / even classes are set as expected.

### DIFF
--- a/config/karma.js
+++ b/config/karma.js
@@ -134,7 +134,10 @@ module.exports = function (config) {
 
     autoWatch: false,
 
-    frameworks: ['jasmine'],
+    frameworks: [
+      'jasmine',
+      'jasmine-matchers'
+    ],
 
     browsers: getBrowsers(),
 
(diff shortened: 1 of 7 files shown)


malonecj:
@alistairjcbrown this needs a catchup from master

malonecj:
Thanks for the pull request. Would like a unit test in https://github.com/adazzle/react-data-grid/blob/master/src/addons/__tests__/Grid.spec.js verifying that a column can be added and another one that a column can be removed after render. If you don't have time for this let me know and I will try to pick it up at some stage

alistairjcbrown:
I've added tests for columns being added and removed. While I was in adding tests, I took some time to clean up the grid test file.
The diff for it is a little bigger, but it's mostly moving code into `beforeEach` blocks, moving shared functions into the root `beforeEach` and keeping the Jasmine context so that `this` can be used to set values in `beforeEach` and use those values in assertions. I've tried to keep each change in a single commit, so you can walk through them if you need.

The diff is best viewed with whitespace diffing turned off: https://github.com/adazzle/react-data-grid/pull/109/files?w=1

PR #111, opened 2015-11-16
[changed] Reduce renders by only requesting new rows after scroll

alistairjcbrown:
On each scroll, the canvas is rendered and `this.getRows` is called to retrieve the visible rows. However, much of the time, these rows have already been rendered and are not required. Additionally, users may be scrolling quickly in order to reach another part of the list, and may have no requirement for the intermediary rows to be requested and rendered.

In the examples, row data is retrieved directly from an in memory object with very quick read access. In reality,requesting this data may be more computationally expensive. This leads to stuttered scrolling and an inability to skim down a list, say to the mid point.

To help, this PR reduces the number of requests by caching the row data in state and rendering it on each scroll. Requests for new row data is debounced and completed when scrolling stops. Users can configure the debounce wait and provide a "loading" component to be rendered in place of a populated row.

### DIFF
--- a/src/Canvas.js
+++ b/src/Canvas.js
@@ -113,6 +113,8 @@ var Canvas = React.createClass({
   },
 
   renderLoadingRow(position, rowHeight) {
+    if (this.scrollingHighVerticalDistance()) return null;
+
     const LoadingRow = this.props.loadingRowRenderer;
     if (!_.isUndefined(LoadingRow)) {
       return <LoadingRow height={rowHeight} key={position} idx={position} />;
@@ -151,7 +153,7 @@ var Canvas = React.createClass({
 
   _currentRowsLength : 0,
   _currentRowsRange : { start: 0, end: 0 },
-  _scroll : { scrollTop : 0, scrollLeft: 0 },
+  _scroll : { scrollTop : 0, scrollLeft: 0, changeTop: 0, changeLeft: 0 },
 
   getInitialState() {
     return {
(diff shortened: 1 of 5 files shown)


src/Canvas.js
```diff
@@ -4,6 +4,7 @@
  */
 "use strict";
 
+var _               = require('lodash');
```
malonecj: I would prefer not to pull in the entire Lodash library here just for a few utility methods. Would be better to either just require the functions we need or use something like https://github.com/megawac/babel-plugin-lodash and using es6 imports

src/Grid.js
```diff
@@ -23,6 +23,7 @@ var Grid = React.createClass({
     headerRows: PropTypes.oneOfType([PropTypes.array, PropTypes.func]),
     rowHeight: PropTypes.number,
     rowRenderer: PropTypes.func,
+    debounceRowRequestWait: PropTypes.number,
```
malonecj: we should also pass this in as prop from ReactDataGrid component https://github.com/adazzle/react-data-grid/blob/master/src/addons/grids/ReactDataGrid.js which is the main wrapper grid. 
The future plan is to merge this `Grid` and `ReactDataGrid` components to remove confusion

malonecj:
Thanks a lot for the pull request and sorry for not getting back sooner. Really like the idea here. There are a couple of things to sort out first before we can think about merging. 

There are 21 failing unit tests that will need to be addressed.
Some of the functionality in the examples is also broken, for example the cell navigation in https://github.com/adazzle/react-data-grid/tree/master/examples - you can no longer scroll past about 40 rows with keyboard navigation. Also this example https://github.com/adazzle/react-data-grid/blob/master/examples/scripts/example10-one-million-rows.js crashes the browser on scroll

Let me know if I can help out on anything here as improving the scroll performance is one of the key objectives for us

PR #112, opened 2015-11-19
re-render non-selected cells

jpdriver:
if updating a cell in one column also updates the value of a cell on the same row but in a different column, we need to check that as part of shouldComponentUpdate.

at the moment only the selected cell will re-render, although each cell goes through shouldComponentUpdate.

- [x] fire shouldComponentUpdate if non-selected cell values change


bakesteve:
nice. needs a test though

jpdriver:
tagged in GitHub as https://github.com/adazzle/react-data-grid/releases/tag/v0.12.26
and released to npm as an old version

will do another PR for master so we don't lose the change

PR #113, opened 2015-11-20
Add classes to the sortable header cell for sort state

alistairjcbrown:
Allow new styles to be attached to the sortable heading cells by providing classes for the different states.

For example, this allows us to use font-awesome icons for the sorting indicators, and use a faded [sort icon](http://fontawesome.io/icon/sort/) to provide affordance to the user on which columns can be sorted.


malonecj:
This looks good. We need to add a section to the docs on custom styling but will merge this in for now

alistairjcbrown:
@malonecj Sounds good - as part of custom styling it might be worth teasing some of the style attributes out into a base stylesheet and leaving as few inline styles as possible.

PR #117, opened 2015-11-27
Pass in table width as prop or default to responsive

alistairjcbrown:
This change provides a `minWidth` prop which is used to control the table width. If no prop is provided, the existing `gridWidth` calculation is used.
Our use case is that the we want to be table to trigger a resize of the table when a menu is visible on the left, but the window size remains unchanged. This allows us to fully control the table width.

### DIFF
--- a/src/addons/__tests__/Grid.spec.js
+++ b/src/addons/__tests__/Grid.spec.js
@@ -619,4 +619,24 @@ describe("Grid", function () {
       })
     });
   })
+
+  describe("Table width", function () {
+    beforeEach(function () {
+      this.tableElement = React.findDOMNode(this.component);
+    });
+
+    it("should generate the width based on the container size", function () {
+      expect(this.tableElement.style.width).toEqual('0px');
+    });
+
+    describe("providing table width as prop", function () {
+      beforeEach(function () {
+        this.component.setProps({ minWidth: 900 });
+      });
+
+      it("should set the width of the table", function () {
+        expect(this.tableElement.style.width).toEqual('900px');
+      });
+    });
+  });
 });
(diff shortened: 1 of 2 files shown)


jpdriver:
hey @alistairjcbrown -- do you have PhantomJS installed on your path? by default, `gulp test` will try and use that, but alternately you can also use `gulp test --debug true` which will launch Chrome and run the tests there. you can see the individual results in the Dev Tools:

image

alistairjcbrown:
@jpdriver Not sure what the issue was, but was able to get it working with the phantom environment variable set as you said.
I'll try and get this updated with tests this week.

jpdriver:
Cool, thanks! I'll have a chat with people here -- sounds like we might need more documentation around testing / setup and stuff? cc @bakesteve

alistairjcbrown:
@jpdriver I've still no idea why the launchers were stalling (both Chrome and PhantomJS) - I'm going to chalk that one up to my system being weird.
However, it looks like the reason why the phantom environment variable is needed is because you're using `karma-phantomjs-launcher-nonet` which "requires you to set the path to PhantomJS explicitly" - this has only been done for `win32` in the karma file, which is why it fails to find PhantomJS on OSX.
I'll get a PR up for setting paths for this as well to make life easier for devs on linux and OSX.

alistairjcbrown:
@jpdriver Tests added and done in such a way as to hopefully allow easy conflict resolution when rebasing after PR #109 is merged.

jpdriver:
@alistairjcbrown ahh yeah, apologies for that. I'm trying to encourage more cross-platform support but we're predominantly a Microsoft software house so you'll probably see a slight win32 bias in our code! :smile:

thanks very much for adding the tests -- I'll see if I can get this merged in later on this week.

alistairjcbrown:
@jpdriver Cheers - if possible PR #109 should go in first, as I cleaned up the test file as part of that PR

alistairjcbrown:
@bakesteve Updated for new master

bakesteve:
cheers

PR #118, opened 2015-11-28
[fixed] Canvas cdu method to not make extra call to setScrollLeft in always truth condition

ptkach:
if (this._scroll !== {start: 0, end: 0}) is always true and it causes extra call to setScrollLeft which iterates through the columns. And when there're huge amount of columns it hits the performance.

### DIFF
--- a/src/Canvas.js
+++ b/src/Canvas.js
@@ -144,9 +144,6 @@ var Canvas = React.createClass({
   },
 
   componentDidUpdate(nextProps: any) {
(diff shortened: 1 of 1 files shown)


bakesteve:
rather than removing this check, we need to change this to `if (this._scroll !== { scrollTop: 0, scrollLeft: 0 }` so we'll make that change
thanks for reporting the bug though. I'll update here when the fix is up

Issue #119, opened 2015-11-29
Resizable Columns?


Haemoglobin:
The resizable columns example does not work for me?  
http://adazzle.github.io/react-data-grid/examples.html#/resizable  

Hovering over the column divider in the header does not allow me to resize the columns. Is it done another way?

jpdriver:
hey there @Haemoglobin, it looks like we're missing that missing the `react-grid-HeaderCell__draggable` class the component wants :-1:

i'll have a look at that and get it updated, but in the meantime the draggable components are actually there and work correctly. just click and drag near the right edge of the header cell :smile:

PR #120, opened 2015-11-30
missing HeaderCell draggable style

jpdriver:
@malonecj not sure when we lost this?

image


jpdriver:
closes https://github.com/adazzle/react-data-grid/issues/119

PR #127, opened 2015-12-02
re-render non-selected cells

jpdriver:
if updating a cell in one column also updates the value of a cell on the same row but in a different column, we need to check that as part of shouldComponentUpdate.

at the moment only the selected cell will re-render, although each cell goes through shouldComponentUpdate.

- [x] fire shouldComponentUpdate if non-selected cell values change

**THIS WAS RELEASED AS v0.12.26 -- THIS IS A DUPLICATE PR SO WE DON'T LOSE THE CHANGE**


bakesteve:
57 tests failing

jpdriver:
replaced by https://github.com/adazzle/react-data-grid/pull/154

PR #131, opened 2015-12-07
Allow percentage and minimum column widths

alistairjcbrown:
In this PR:

1. Fixes the functionality to allow percentage strings to be provided as column widths
2. Allows a `minWidth` property to be provided to prevent column widths from becoming too narrow
3. Removes `=>` notation in `ColumnMetrics.spec.js` as this breaks Jasmine context binding
4. Fixes tests for column metric `recalculate` function which were using `if` instead of `it`
  - There were tests (though not being run) which aimed to support Immutable objects for the column data. These have been removed as supporting Immutable objects caused `Header Unit Tests` to fail: "_When columns are immutable header row drag start should set resize column state_"

### DIFF
--- a/src/ColumnMetrics.js
+++ b/src/ColumnMetrics.js
@@ -24,14 +24,23 @@ type Column = {
   width: number;
 };
 
+function isImmutable(value, structure) {
+  return typeof Immutable !== 'undefined' &&
+         Immutable[structure] !== 'undefined' &&
+         (value instanceof Immutable[structure]);
+}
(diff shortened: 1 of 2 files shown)


bakesteve:
There are a few failing tests, but otherwise this all looks good

jetpackjarrett:
This has been sitting idle for a while. Would love this feature. @alistairjcbrown are you planning on fixing this test or would you mind if I fixed it?

Issue #129, opened 2015-12-02
possible to set width of React Data Grid? 


raptoria:
Is it possible to set the width and height of the table to auto, such that the table scroll bars disappear and I can use the browser's scrollbar to scroll horizontally or vertically instead (if the table is larger than the initial viewport)? 

thanks!!

bakesteve:
I think this will be covered by #133 so maybe worth checking that out? It should get merged soon, so wait till next week if its easier and then you can just npm update

mohak1990:
Which Grid is more complete. Many react grids come up with lots of issues and few customization options.. Which grid should I choose for a business app?

PR #133, opened 2015-12-08
Introduce new can scroll props to prevent grid scrolling

alistairjcbrown:
Introduces two new props, `canScrollX` and `canScrollY` to allow the horizontal and vertical scroll to be disabled independently.
Our use case for this is when the grid is the same size as the window, but at the bottom of the page and must be scrolled to. Locking the y scroll until the bottom of the page has been reached provides a much nicer user experience.

### DIFF
--- a/src/Canvas.js
+++ b/src/Canvas.js
@@ -29,6 +29,8 @@ var Canvas = React.createClass({
       PropTypes.array.isRequired
     ]),
     onRows: PropTypes.func,
+    canScrollX: PropTypes.bool,
+    canScrollY: PropTypes.bool,
     columns: PropTypes.oneOfType([PropTypes.object, PropTypes.array]).isRequired
   },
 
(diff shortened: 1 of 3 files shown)


bakesteve:
Makes sense. We may want to think about the Props here, as we could end up with a fairly substantial prop list at this rate.. we'll chat here (and on slack) and then merge once we come to a consensus

AlanFoster:
@bakesteve What were the outcomes of your discussion?

bakesteve:
We agreed that we didnt like having so many props (and can only see that going one way) but not on the right approach, as still want it to be easy to override just one prop
We've also been focussed on getting react 14 out the door, and full lint coverage as part of tooling, so will be revisit this later this week.
Feel free to jump in if anyone has string views. And if we dont reach consensus / solution, we'll probably take this anyway and then refactor later (with a breaking version change)

txm:
From #129 An auto width would be great..  Manually applying a width to 40 columns is tedious and never looks as good.

bakesteve:
given we havent established how we are going to resolve the noise on props yet, lets get this in as is, and we can refactor those later (or not)
@alistairjcbrown can you catchup and then we will take this in as is

PR #135, opened 2015-12-09
Add PhantomJS paths for Linux and OSX

alistairjcbrown:
This adds the PhantomJS paths for linux and OSX so that it's picked up automatically when running `npm run test`.

__Side note:__ Seems a little weird to bundle the windows executable file for PhantomJS with the repo.
Was `karma-phantomjs-launcher` not working? It's still part of the `package.json` file, so will still be installed when running `npm install`.

### DIFF
--- a/config/karma.js
+++ b/config/karma.js
@@ -76,6 +76,14 @@ module.exports = function (config) {
     return preprocessors;
   }
 
+  function lookupPhantomJS() {
+    try {
+      return require('phantomjs').path;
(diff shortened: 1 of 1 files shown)


bakesteve:
Yeah - I think phantom.exe was a hangover from trying to get Jest running on windows (not.much.fun)
But will revisit that on a fresh install (but with bundling teh exe) - shouldnt be any need for that

PR #136, opened 2015-12-10
change render options (add value)

delef:


src/addons/editors/DropDownEditor.js
```diff
@@ -32,8 +32,12 @@ class DropDownEditor extends EditorBase {
 
   renderOptions(): Array<ReactElement>{
     var options = [];
-    this.props.options.forEach(function(name){
-      options.push(<option key={name} value={name}  >{name}</option>);
+    this.props.options.forEach(function(name, value){
+      if typeof value === 'string' && typeof value === 'number'
```
bakesteve: think we'd prefer a undefined check here, otherwise something that looks like an object (ie `typof new Date() === 'object'` would not set the value)

bakesteve:
Thanks for this. 
Be good to understand the use case. At the moment, value is a required propType. If we were going to fall back to the drop down using the name as the value, I think we'd just map that in the data given to the editor, but might be missing something
If we do take this in though, we'll need a test and to resolve if we should alter the propTypes

delef:
Sorry for my english. I'm understand the use case, but I need use value..

PR #137, opened 2015-12-10
change render options (fix)

delef:


bakesteve:
Ok - understood. 
Thanks for the PR
We'd like to see some tests for this, and adding PropTypes to document what people are expected to pass in as the options. We can help out on that - do you have an example oif the data you would be passing in / a working example with this extension?

bakesteve:
this has now been implemented, but in a slightly different way, to mimic the autocomplete editor. so you can pass in a string, or an object with `{ id: React.PropTypes.string, title: React.PropTypes.string, meta: React.PropTypes.string }`
see https://github.com/adazzle/react-data-grid/commit/3647224403091bc4469533f613c2fb7045207f5d

PR #141, opened 2015-12-14
Up to the react version

delef:
I tested the react-data-grid on latest react versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "react-component"
   ],
   "peerDependencies": {
-    "react": "~0.13.3"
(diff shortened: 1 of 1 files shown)


bakesteve:
thaks of rthis. Because of the levels of incompatibility with react (esp 13 vs 14) and the impact of bundling > 1 version we are keeping to specific versioning on react, but do have work underway for react 14 tracked here: #63 
Once that is complete, we'll also maintain a stable version on npm, so folks can then choose teh version that fits their use. Hopefully once React goes to v1 this will all be a bit easier!

PR #164, opened 2016-01-11
get React 14 off the ground

jpdriver:
strategy here was absolutely as few changes as possible.

that meant things like adding `fbjs` so we had access to `emptyFunction()` and just updating the `require` statements without actually changing the src.

all tests passing, but undoubtedly more to do. i wanna say this is maybe like 80% of the way there? for what it's worth, `gulp dist` totally works without issue so i think we are actually building ok; but the examples are boned. think it might be something to do with our gulp pipeline, i dunno i couldn't figure it out. it loads the example page but is missing most of the components :frowning: 

also lots of warnings that `render` isn't a pure function. we check for `isMounted` somewhere in the call stack. FB says move that logic to either `componentWillMount` or `componentDidMount`, but those are only warnings and don't stop us for now.

couple of other bits too; `Tapable` was complaining about our gulp task callbacks and there's bug in the `react-jasmine-helper` package (see https://github.com/tommyh/jasmine-react/pull/9)...

but feels like this could be a good start?

Screenshots
-----------------
*118 of 118 passing* :+1: 
screen shot 2016-01-11 at 8 59 15 am

result of just doing a `gulp`:
screen shot 2016-01-11 at 9 00 19 am


gulp/tasks/browserSync.js
```diff
@@ -1,7 +1,7 @@
 var browserSync = require('browser-sync');
 var gulp        = require('gulp');
 
-gulp.task('browserSync', ['examples'], function() {
```
jpdriver: `examples` deliberately temporarily disabled because `Tapable` doesn't like our gulp callbacks :-1: 

help!

gulp/tasks/examples.js
```diff
@@ -41,7 +41,7 @@ webpackConfig = {
 		]
 	},
 	plugins: [
-
+		"transform-react-jsx"
```
jpdriver: not 100% sure we actually need this, but I guess React deprecated it's own one circa v13 -- they recommend using this now (see https://facebook.github.io/react/blog/2015/06/12/deprecating-jstransform-and-react-tools.html#jsxtransformer)

src/Canvas.js
```diff
@@ -168,7 +168,7 @@ var Canvas = React.createClass({
                         || nextProps.columns !== this.props.columns
                         || nextProps.width !== this.props.width
                         || nextProps.cellMetaData !== this.props.cellMetaData
-                        || !shallowEqual(nextProps.style, this.props.style);
+                        || shallowCloneObject(nextProps.style) !== shallowCloneObject(this.props.style);
```
jpdriver: `shallowEqual` is still available but moved to a separate add-on package; i figured we could just use `shallowCloneObject` to achieve the same thing here? we already had it and it's one less package...

src/__tests__/Cell.spec.js
```diff
@@ -84,7 +87,8 @@ describe('Cell Tests', () => {
       testElement = TestUtils.renderIntoDocument(<Cell {...testProps}/>);
       var editorContainerInstance = TestUtils.findRenderedComponentWithType(testElement, EditorContainerStub);
       expect(editorContainerInstance).toBeDefined();
-      expect(editorContainerInstance.props).toEqual({
+      var expectedProps = Immutable.Map(editorContainerInstance.props).toJS();
+      expect(expectedProps).toEqual({
```
jpdriver: `props` are Immutable these days, so this comparison was failing. the way `jasmine` does it's equality comparison is by attaching a new property to each object, so it was attaching to the ones we pass in but not the `editorContainerInstance` ones, and failing.

since jasmine apparently can't actually handle equating frozen objects, i decided to map the `editorContainerInstance` ones to a mutable object...? i dunno if that's right but it made sense at the time and does work... :flushed:

src/__tests__/Cell.spec.js
```diff
@@ -100,10 +104,10 @@ describe('Cell Tests', () => {
     it('should append the update cell class to the dom node if present and cell is updated', () => {
       var updateClass = 'highlight-cell';
       testProps.column.getUpdateCellClass = () => updateClass;
-      var cellInstance = TestUtils.renderIntoDocument(<Cell {...testProps}/>);
+      var cellInstance = new PropTestUtils(Cell, testProps);
       // force update
       cellInstance.setProps({rowData: {}, selectedColumn: testProps.column});
```
jpdriver: `renderIntoDocument` is still a thing, but `setProps` isn't anymore. because of the point of this test and the fact I wanted to touch it as little as possible, I put in `PropTestUtils` which is basically boilerplate but renders into document and then allows you to forcibly set props for testing as we do here.

src/__tests__/ReactVersion.spec.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+var React  = require('react');
+var semver = require('semver');
+
+describe('React Version Tests', () => {
+
+  it('should use React 14', () => {
+    var usingReact14 = semver.gt(React.version, "0.14.0");
+    expect(usingReact14).toBe(true);
```
jpdriver: achievement unlocked: TDD

src/addons/editors/widgets/DateRangeFilter.js
```diff
@@ -1110,7 +1111,7 @@ var DateRangeFilter = React.createClass({
 
 	componentDidMount : function(){
 		//initialise jQuery date range widget -
-		var $calendarNode = $(this.refs.calendar.getDOMNode())
+		var $calendarNode = $(ReactDOM.findDOMNode(this.refs.calendar))
```
jpdriver: this is wrong. don't need `findDOMNode()` for `refs` anymore. (the refs point to DOM nodes themselves now)

test/PropTestUtils.js
```diff
@@ -0,0 +1,49 @@
...
+'use strict';
+
+var React    = require('react');
+var ReactDOM = require('react-dom');
+
+class PropTestUtils {
+
+  constructor(klass, props, container, callback) {
+    this.klass = klass;
+    this.container = container || document.createElement('div');
```
jpdriver: default to doing the same thing as `renderIntoDocument` in the FB TestUtils...

test/PropTestUtils.js
```diff
@@ -0,0 +1,49 @@
...
+      console.warn(
+        'setProps(...): Can only update a mounted or ' +
+        'mounting component. This usually means you called setProps() on ' +
+        'an unmounted component. This is a no-op.'
+      );
+      return;
+    }
+    Object.assign(this.props, partialProps);
+    var element = React.createElement(this.klass, this.props);
+    this.component = ReactDOM.render(element, this.container, callback);
```
jpdriver: (just calling `render` again with the same class and new props re-renders the existing element)

src/Canvas.js
jquense: Sorry for the drive by commenting (I'm here because we manage a fork we use internally [here](https://github.com/intljusticemission/react-data-grid)), and just noticed this.

This doesn't work right? cloning two objects is _never_ going to result in them being referentially equal, so this will always be `true`.

jquense:
Hi ya'll, We recently forked this project to make some internally helpful API changes but in the process also upgraded everything to react 0.14

Most of the work was in the examples, since we could use react-code-mod to do most of the API changes in src automatically. 

I would have honestly loved to give back by sending a PR with our work, but some of it is still intertwined with non relevant changes so it would have taken a bit to pull all that out, so appologize. however if you'd like to see what we did as a sanity check for this PR the relevant commit is: https://github.com/intljusticemission/react-data-grid/commit/4b7ad7aa53948f7fc347a8024c5a5769f3cac08d

package.json
```diff
@@ -19,19 +19,22 @@
     "react-component"
   ],
   "peerDependencies": {
-    "react": "~0.13.3"
+    "react": "~0.14.6"
   },
   "dependencies": {
     "classnames": "^1.2.0",
     "es5-shim": "^4.0.3",
+    "fbjs": "^0.6.1",
```
jquense: FYI you might want to pin to the exact version here, fbJS warns that they [may not follow semver](https://github.com/facebook/fbjs#purpose)

src/Canvas.js
malonecj: thats right. shallowCloneObject is not the right thing to use here, we want to compare the objects not copy them. We should just be checking each property of incoming props and check if equal to the current props. There is a `src/addons/utils` folder available. I think we just copy any utility functions in there rather than relying on external files from `react` package

src/__tests__/Cell.spec.js
malonecj: since setProps is being deprecated there is not point having a test for this. Instead we should simulate the element being updated which can be done with ReactDOM.Render on the same container node 
https://facebook.github.io/react/blog/2015/10/01/react-render-and-top-level-api.html

package.json
malonecj: Thanks for the commit guys. We'll take a look through that and hopefully get this merged to master asap

jbrodriguez:
This is a great component, hopefully you guys can get it working under 0.14 soon. :+1: 
It's a showstopper for me right now.

PR #168, opened 2016-01-19
Set sorting state from prop values

alistairjcbrown:
`sortColumn` and `sortDirection` are provided as props, but after user interaction, internal state then dictates which column is sorted and in what direction.
This PR allows setting these props to change the internal state and thus update the table based on external controls.

### DIFF
--- a/src/addons/grids/ReactDataGrid.js
+++ b/src/addons/grids/ReactDataGrid.js
@@ -117,6 +117,12 @@ var ReactDataGrid = React.createClass({
     if(nextProps.rowsCount  === this.props.rowsCount + 1){
       this.onAfterAddRow(nextProps.rowsCount + 1);
     }
+    if(nextProps.sortColumn){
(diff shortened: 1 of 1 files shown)


src/addons/grids/ReactDataGrid.js
```diff
@@ -117,6 +117,12 @@ var ReactDataGrid = React.createClass({
     if(nextProps.rowsCount  === this.props.rowsCount + 1){
       this.onAfterAddRow(nextProps.rowsCount + 1);
     }
+    if(nextProps.sortColumn){
+      this.setState({ sortColumn: nextProps.sortColumn });
+    }
+    if(nextProps.sortDirection){
+      this.setState({ sortDirection: nextProps.sortDirection });
```
jpdriver: hey @alistairjcbrown, looks good but one thing i'll say is it might be good to include either `nextProps.sortDirection` or a default and only call `setState()` once with both properties specified?

please can you add some tests and then we'll get this merged :smile:

Issue #97, opened 2015-11-06
Add classNames in columns 


markmus:
I dont know if this is possible right now but according to documentation there is still no way how I could add classNames or styles by columns.

Currently what I could do is to create a custom formatter, however it will be much more easy if we could add `classNames` upon declaration.

In example:

```
//Columns definition
var columns = [
{
  key: 'id',
  name: 'ID',
  locked : true,
  classNames: 'center'
},
{
  key: 'task',
  name: 'Title',
  width: 200
}
```

malonecj:
Hi, You should be able to use  the `cellClass` property of a column to set the class. I will add this to the documentation

alistairjcbrown:
Is the `cellClass` also set on the header cell? Or is there another prop which can be used for this? I'm hoping to align right the cell text in columns with numeric data, but would also want to include the heading text.

alistairjcbrown:
Added PR #170 to add the `cellClass` value to the class of the header cells as well

txm:
Could cellClass default to the key?

PR #170, opened 2016-01-19
Add cell class to header cells

alistairjcbrown:
This adds the (possibly undocumented, see #97) `cellClass` prop to the column header so that it can be styled as well as data cells.

Headings can then be styled with `.react-grid-HeaderCell.my-custom-cellclass`
and data cells with `.react-grid-cell.my-custom-cellclass`


igoriok1994:
Can i add className only for row cells? (Not for header cell)

PR #177, opened 2016-01-23
More context for custom formatters please

littlebee:
Hi,  I'm replacing [our current datagrid](https://github.com/mleibman/SlickGrid) with react-data-grid.   I'm really happy with it so far.  Nice work!

It's difficult to duplicate some of the behavior of our slickgrid custom formatters without more context.  This pull request provides the column definition and rowData as props to custom formatters.   

Included in the added tests is another use case I see in several of our slickgrid based grids - using the column definition to pass additional metadata to the formatter / editor such as a switch to control minutia in the rendering like number of decimal places on numbers, etc.   The third added expectation tests this to prevent regressing.

Again, that you so much for do this incredible piece of work.  I must say, I was shocked when I passed in an array of Backbone models expecting a fail and it saw the .get() on the rowData and went for it!  Nicely done.  Is there a test to ensure that behavior?


littlebee:
Sorry.  I just realized that I was still working on the pull branch.  I merged in the v14 branch to get edits working.  I'll close and open another in a day or two.

jpdriver:
hey @littlebee -- it's not really documented yet but you can set a function called `getRowMetaData()` on your column to achieve this.

have a look at https://github.com/adazzle/react-data-grid/blob/59c75161ab248b57f2f6054aed4362e8d1a5b6ea/src/Cell.js#L114

```
  getFormatterDependencies() {
    // convention based method to get corresponding Id or Name of any Name or Id property
    if (typeof this.props.column.getRowMetaData === 'function') {
      return this.props.column.getRowMetaData(this.getRowData(), this.props.column);
    }
  },
```

if you define a `getRowMetaData` func on your column, you can access properties within a formatter component in `this.props.dependentValues`

jpdriver:
ps. it's really cool to hear you're replacing SlickGrid -- we used to use that internally and replacing it was actually one of our main reasons for starting this!

littlebee:
I was wondering what that was for :+1:   And I had noticed the similarities and wondered :)

Issue #146, opened 2015-12-19
Filterable and sortable features, don't work properly when both are used at the same time.


vmsoftware:
I'm trying to implement all features (or almost all) but when I try to add filterable and sortable features to the same data grid... the behavior is strange and it don't work properly.

Any idea?

Thanks in advance

vmsoftware:
<img width="1219" alt="captura de pantalla 2015-12-19 a las 4 46 50 p m" src="https://cloud.githubusercontent.com/assets/16278865/11915340/2c8592cc-a670-11e5-86dd-57cdbdbc16b9.png">

qosmy-michael:
It would be great to have both - otherwise it turns into an opinionated UX, a lot more limiting.

bakesteve:
Yep - this is one we are aware of (and have hit) - its certainly not by design and is being looked at now.
Will add a link back here once we get the fix done

cheahkhing:
in fact, it is still not working as expected. When you sort 1 column, type some filter, and then remove the filter, you'll see that the previous sorting is not being respected. Basically every filtering will reset the existing 'sort', which is probably an issue for those using sort and filter at the same time.

jbgarr:
+1 Still seems broken.

vmsoftware:
Any news?

diogofcunha:
This is fixed after `1.0.23`

avremel:
Still seems to be an issue for me: https://github.com/adazzle/react-data-grid/issues/846

Can anyone comment?

PR #180, opened 2016-01-25
Fixed HeaderRow so that the same grid can be sortable and filterable at the same time 

diogofcunha:
Also added a example of a filterable sortable grid and some validation tests

Fixes issue https://github.com/adazzle/react-data-grid/issues/146


malonecj:
@diogofcunha can you catch this branch up with master, so we can merge

diogofcunha:
@malonecj will do that on the evenning

Issue #188, opened 2016-01-27
Where is the vertical scrollbar ?


chprome:
I am testing this component, and **I can't see the vertical scrollbar** on any example.

Here is what I get :
![capture d ecran de 2016-01-27 22-23-14](https://cloud.githubusercontent.com/assets/1465463/12628692/ab694938-c544-11e5-9172-19cd05c8090e.png)

The horizontal scrollbar is correctly displayed though.

I have tested with Firefox 44, Chrome 46 and Chromium 48 under Linux (Ubuntu Gnome 15.10) and with Internet Explorer 11 and Chrome 42 under a Windows 7 VM.

Is it a standard behaviour or am I facing an issue ?

Cheers,
Christophe

bakesteve:
hmm.. looks like a bug, getting the same for me on http://adazzle.github.io/react-data-grid/examples.html#/million-rows
we'll check (though I suspect its an issue with the long awaited react 14 version)

txm:
I get the slider one when I slide, then if I mouseover I get the complete vertical scroller.  I assumed this was standard?  react 0.14, osx, ff

The examples show most of the button/slider.  react unknown, OSX, safari.

not sure if this helps...

PR #189, opened 2016-01-28
script + rendering fixes

jpdriver:
Fixes
-------

- [x] PropType validation reported by @dotted in Slack
- [x] Script errors on the examples reported by @skyf , @bakesteve
- [x] Missing vertical scrollbar https://github.com/adazzle/react-data-grid/issues/188
- [x] Undefined NaN widths

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "minimist": "^1.1.0",
     "moment": "^2.10.6",
     "pretty-hrtime": "^0.2.1",
-    "react": "~0.14.06",
+    "react": "~0.14.6",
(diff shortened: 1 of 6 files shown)


src/addons/__tests__/Grid.spec.js
```diff
@@ -626,7 +626,7 @@ describe('Grid', function() {
     });
 
     it('should generate the width based on the container size', function() {
```
bakesteve: should add another test where we explicitly set the size as % and px and validate these are obeyed

nqbao:
This looks good. Can we merge soon?

PR #208, opened 2016-02-11
DropDownEditor to accept a similar data structure as AutoCompleteEditor

txm:
var priorities = [{id:0, value : 'Critical'}, {id:1, value : 'High'}, {id:2, value : 'Medium'}, {id:3, value : 'Low'}]
Note/for discussion: where 'title' is reserved for html5/bootstrap tooltip

PR #214, opened 2016-02-16
DropDownEditor to accept a similar data structure as AutoCompleteEditor

diogofcunha:
Applied #208 to current version of react-data-grid

### DIFF
--- a/src/addons/editors/DropDownEditor.js
+++ b/src/addons/editors/DropDownEditor.js
@@ -26,14 +26,21 @@ class DropDownEditor extends EditorBase {
   renderOptions(): Array<ReactElement> {
     let options = [];
     this.props.options.forEach(function(name) {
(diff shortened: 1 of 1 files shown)


txm:
"react-data-grid": "^0.14.10",

[Error] Warning: Failed propType: typeChecker is not a function. (In 'typeChecker(propValue, key, componentName, location, propFullName + '.' + key)', 'typeChecker' is an instance of Object)

AFAICT it's           objectOf: createObjectOfTypeChecker,

I'll need a few days to update my deps and debug.

For now ignore or delete line:

          options: React.PropTypes.arrayOf(React.PropTypes.oneOfType([React.PropTypes.string, React.PropTypes.objectOf({ id: React.PropTypes.string, title: React.PropTypes.string, meta: React.PropTypes.string })])).isRequired

PR #218, opened 2016-02-17
fix enableRowSelect

JacerOmri:
used as string, but declared as boolean


src/addons/grids/ReactDataGrid.js
```diff
@@ -43,7 +43,7 @@ const ReactDataGrid = React.createClass({
     headerRowHeight: React.PropTypes.number,
     minHeight: React.PropTypes.number.isRequired,
     minWidth: React.PropTypes.number,
-    enableRowSelect: React.PropTypes.bool,
+    enableRowSelect: React.PropTypes.string,
```
momaci: Thanks for picking this @JacerOmri! This is being addressed on another PR here(https://github.com/adazzle/react-data-grid/commit/ab49459f566a04b6e801f021bd7fbac66e21f390) and will be merged in soon.

Issue #235, opened 2016-03-03
Add `Select all` checkbox in the selection column header


kgunbin:
Is it possible to have 'Select all' functionality on the table?

jpdriver:
hey @kgunbin -- we should have a checkbox in the header row but it looks like it's stopped working.. i'll have a look now because i need it for a feature :smile:

jpdriver:
@kgunbin just opened a PR to fix. thanks for flagging it! :smile:

kgunbin:
Great, thank you!

jpdriver:
@kgunbin that's been both merged into `master` and published on npm now. update to v0.14.15 :smile:

PR #237, opened 2016-03-03
fix missing select all

jpdriver:
image


jpdriver:
closes https://github.com/adazzle/react-data-grid/issues/235

@malonecj `enableRowSelect` being either one of `true`, `false`, "single" or "multi" feels a bit weird. i know we want to keep the # of PropTypes down but to me that should be `rowSelectEnabled` & `rowSelectType`?

anyways this follows the other things that have been done here for now, so please can you merge?

PR #239, opened 2016-03-04
[added] Each column can have its custom header cell renderer

kgunbin:
Added `column.headerRenderer` tag processing, so each column may use custom renderer for a header cell.
Added tests.
Added the custom header in the 'All feature' example.


malonecj:
This looks good. Can you catch up this branch with master so we can merge

kgunbin:
Done

PR #264, opened 2016-04-01
[added] Added in the ability to define cell level click handlers via …

Lianite:
…the columns declaration. This gives you full access to all the data of the row when clicking on the cell of your choosing.

This allows you to create functionality based on the id of the row of data from any cell in the row.

### DIFF
--- a/examples/scripts/example13-all-features.js
+++ b/examples/scripts/example13-all-features.js
@@ -214,6 +214,16 @@ var AllFeaturesExample = `
     },
 
     render : function() {
+      var self = this;
+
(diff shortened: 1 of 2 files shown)


Lianite:
I guess I do not understand what you mean.

bakesteve:
looks useful enough, but does need a test to assert that if you have onClick defined it will be called, and also that it handles null

malonecj:
Thanks for the pull request. This functionality can already be provided like in this example

http://adazzle.github.io/react-data-grid/examples.html#/column-events

PR #265, opened 2016-04-04
specify height in style of viewport to maintain scrollbar on filter

JamesPortelli:


npm-debug.log
```diff
@@ -0,0 +1,26 @@
+0 info it worked if it ends with ok
```
malonecj: can you remove this file, and add an entry to npm ignore file so it wont be included again

PR #271, opened 2016-04-07
Add editor based on react-select

juanda99:
Similar to Autocomplete edition but supporting several tokens. Based on https://github.com/JedWatson/react-select

### DIFF
--- a/src/addons/editors/AutoCompleteTokensEditor.js
+++ b/src/addons/editors/AutoCompleteTokensEditor.js
@@ -1,0 +1,58 @@
+import React, {Component} from 'react';
+import Select from 'react-select';
+import ReactDOM from 'react-dom';
+
+const ExcelColumn = {
+  name: React.PropTypes.string.isRequired,
+  key: React.PropTypes.string.isRequired,
+  width: React.PropTypes.number.isRequired,
+  filterable: React.PropTypes.bool
+};
+
+class AutoCompleteTokensEditor extends Component {
+
+  static propTypes = {
+    options: React.PropTypes.array.isRequired,
+    column: React.PropTypes.shape(ExcelColumn),
+    value: React.PropTypes.array
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: []
+    };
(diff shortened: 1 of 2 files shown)


juanda99:
upppps, two different pr

malonecj:
Great stuff. We were planning on replacing the existing autocomplete editor with React Select. So this could be a good intermediate step. Before merging, we would like to see an example added to the docs in this file to make sure everything is working along with some unit tests

https://github.com/adazzle/react-data-grid/blob/master/examples/scripts/example06-built-in-editors.js

src/addons/editors/AutoCompleteTokensEditor.js
```diff
@@ -0,0 +1,58 @@
...
+  getValue() {
+    let updated = {};
+    updated[this.props.column.key] = this.state.value;
+    return updated;
+  }
+
+  handleSelectChange(value) {
+    this.setState({value});
+  }
+
```
malonecj: I realise this is not documented anywhere yet, but a select based autocomplete editor needs to implement a `hasResults` method similar to https://github.com/adazzle/react-data-grid/blob/master/src/addons/editors/AutoCompleteEditor.js#L66

so that the editorContainer can handle keyboard navigation correctly. (eg pressing down button will focus a select option instead of navigating down one cell if the Select component has results)

juanda99:
Ok, i'll try to send you a complete PR in the next couple of days.

juanda99:
I've made a fork and a [new branch react-select](https://github.com/juanda99/react-data-grid/tree/react-select) implementing my code as a new column in the table from example06-built-in-editors.js

Can't make it work though, I got an error:
`react-with-addons.js:20263 Uncaught Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object. Check the render method of `EditorContainer`.`

Similar code is working in my own dev environment. @malonecj , could you  have a look?

malonecj:
Thanks @juanda99 I'lll take a look at this

peterpme:
:wave: any change we can merge this in? Thanks

PR #276, opened 2016-04-11
Explicitly allow column.name to render anything React can render 

littlebee:
Hey did you guys know you built a kickass datagrid that will allow you to pass a React component through column.name and it will work?!?   

I've been using this for a while for things like checkboxes in the header, custom rendered data, etc.   Maybe worth considering explicitly allowing it?    Might remove the burden of adding a bunch of different prebaked column header types?

PR just removes the React proptype warning when running in debug mode.

### DIFF
--- a/src/addons/cells/headerCells/SortableHeaderCell.js
+++ b/src/addons/cells/headerCells/SortableHeaderCell.js
@@ -9,7 +9,7 @@ const DEFINE_SORT = {
 const SortableHeaderCell = React.createClass({
   propTypes: {
     columnKey: React.PropTypes.string.isRequired,
-    column: React.PropTypes.shape({ name: React.PropTypes.string }),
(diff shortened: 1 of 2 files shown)


jpdriver:
hey @littlebee -- i think HeaderCells definitely have a lot more potential than what we're currently using them for (see https://github.com/adazzle/react-data-grid/issues/101)... :smile: 

Presumably `node` as you've put in your change here still works fine for either string or an actual component instance, right?

excellent branch name btw :ok_hand:

littlebee:
Sorry for not following up sooner.   According to the React docs:
    // Anything that can be rendered: numbers, strings, elements or an array
    // (or fragment) containing these types.
    optionalNode: React.PropTypes.node,

Not sure what the use case would be for arrays or fragments.

PR #282, opened 2016-04-12
ReactDataGrid component update columnMetrics when column changes in props

JamesPortelli:
- [x] Added a componentWillRecieveProps to ReactDataGrid to handle the updating of the columnMetrics when the columns change in ReactDataGrid.


JamesPortelli:
No changes were needed

PR #286, opened 2016-04-15
[fixed] support for touch events for column resizing

alefherrera:
Hi!,another thing that i modified on my local proyect, is the class ResizeHandle, adding a width of 16px, for an easy way to resize on mobile devices, it can be a good improvement too, let me know what do you think about this and i can add it.
Cheers!

### DIFF
--- a/src/Draggable.js
+++ b/src/Draggable.js
@@ -35,8 +35,8 @@ const Draggable = React.createClass({
     }
 
     window.addEventListener('mouseup', this.onMouseUp);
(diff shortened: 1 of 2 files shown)


alefherrera:
Hi!,another thing that i modified on my local proyect, is the class ResizeHandle, adding a width of 16px, for an easy way to resize on mobile devices, it can be a good improvement too, let me know what do you think about this and i can add it.
Cheers!

PR #291, opened 2016-04-18
[added] Column resize callback

lesliev:
Allow grid to report on column resize events so that new column sizes
can be persisted on the back-end.

This exposes an onColumnResize prop for ReactDataGrid.

### DIFF
--- a/examples/scripts/example02-resizable-cols.js
+++ b/examples/scripts/example02-resizable-cols.js
@@ -32,8 +32,8 @@ var columns = [
 {
   key: 'id',
   name: 'ID',
-  resizable : true
-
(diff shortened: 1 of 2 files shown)


lesliev:
I added a note in the resizable columns example. Should I add to the example code itself? Will a callback with a console.log of the column index and new width suffice?

I took a look at the tests and didn't know where to start with making a new one, sorry. I've made a few other changes that you might accept - in time I'll learn how to write the tests!

lesliev:
I just realised the tests are under /src/__tests__ and I thought they were in /test. No wonder I couldn't figure them out.

Issue #298, opened 2016-04-21
"up" dependency?


alvaro1728:
My company needs to approve (whitelist) all npm modules used in production and they're blocking react-data-grid because they don't have licenses for two of its sub-dependencies.  Here's the message they sent me:

> Unfortunately, some subdependencies of this module are not licensed at all, meaning we cannot legally use them. In particular, the following two modules have no license:
> https://www.npmjs.com/package/cloudup-ua
> https://www.npmjs.com/package/cloudup-client
> Until licenses are added to those two modules, we can't really whitelist react-data-grid.

Those are dependencies of "up". I just don't understand why a grid component would require cloud access.  Can you expand on this and is there a way to remove these unusual dependencies in the near future?

Thanks,
Alvaro

PR #299, opened 2016-04-22
remove npm up dependency

malonecj:
'up' npm dependency is not required for this project. Removing from package.json


malonecj:
https://github.com/adazzle/react-data-grid/issues/298

PR #306, opened 2016-05-03
Clear filters fix on toggling of filters [PL-680] [PL-600]

JamesPortelli:
Pull request for https://adazzle.atlassian.net/browse/PL-680, https://adazzle.atlassian.net/browse/PL-600

- [x] Fix an error on toggle filter which was causing the filters not to clear on toggle.

- [x] Fix the removal of the bottom scroll bar when toggling off the filters when the grid is being filtered


lehaininh:
Hey is it working correctly.

When I toggle the filter on and off. The filter values on each column are cleared. 
But the rows are still in their filtered form not full without filter?

PR #308, opened 2016-05-04
Add column events

diogofcunha:
<h2>Motivation</h2>
It's useful and necessary to allow different columns to have different behaviours based on the native React events. This can have a huge number of benefits like logging, statistic gather, fast editor opening...

<h2>Technical Side</h2>

- <h3>Bound one or more events to a column</h3>
This should be simple and work by convention, my suggestion in this one is to use the same [convention used by react](https://facebook.github.io/react/docs/events.html#supported-events) for the naming, of course that **not every react event would be supported in the first release but we should aim to cover as much events as possible in the next major release**.
<h6>API:</h6>
For every event implementation in the column context the following convention will be adopted 
`function onXxx(ev :SyntheticEvent, args: {rowIdx :number, idx: number, column :ColumnShape}) { ... }`. The correct usage of the args is a user responsibility.
<h6>Usage example:</h6>
```
columns = [{
   key: 'title',
   ...,
   events: {
      onDoubleClick: function() {
         console.log("The user double clicked on title column");
      },
      onClick: function(ev, args) {
         doStuff(ev, args);
      }
   }
}]
```
- <h3>Event execution on the Grid</h3>
To execute the event on the Grid side the best thing to do would be have something similar to a dispatcher or a afterEach in jasmine. The main responsibility for it would be to make sure that after a n event being fired in the grid context it would look at the column where the event was fired and fire the column event when necessary.

<h2>Progress</h2>
- [x] Pass an event to the grid should trigger that event (click and doubleClick for now)
- [x] Handle events on the grid
- [x] Efficient event execution on the grid 
- [x] Tests

### DIFF
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -331,9 +331,16 @@ const Cell = React.createClass({
     return (this.props.column.editor != null) || this.props.column.editable;
   },
 
-  createEventCallBack(onColumnEvent, info) {
+  createColumEventCallBack(onColumnEvent, info) {
     return (e) => {
       onColumnEvent(e, info);
+    };
+  },
+
+  createCellEventCallBack(gridEvent, columnEvent) {
+    return (e) => {
+      gridEvent(e);
+      columnEvent(e);
     };
   },
 
@@ -344,14 +351,11 @@ const Cell = React.createClass({
       if (columnEvents.hasOwnProperty(eventKey)) {
         let event = columnEvents[event];
         let eventInfo = {rowIdx: this.props.rowIdx, idx: this.props.idx, name: eventKey};
-        let eventCallback = this.createEventCallBack(onColumnEvent, eventInfo);
+        let eventCallback = this.createColumEventCallBack(onColumnEvent, eventInfo);
 
         if (allEvents.hasOwnProperty(eventKey)) {
           let currentEvent = allEvents[eventKey];
-          allEvents[eventKey] = (e) => {
-            currentEvent(e);
-            eventCallback(e);
-          };
+          allEvents[eventKey] = this.createCellEventCallBack(currentEvent, eventCallback);
         } else {
           allEvents[eventKey] = eventCallback;
         }
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -75,17 +75,17 @@ const Cell = React.createClass({
     || this.props.value !== nextProps.value;
   },
 
-  onCellClick() {
+  onCellClick(e) {
     let meta = this.props.cellMetaData;
     if (meta != null && meta.onCellClick != null) {
-      meta.onCellClick({rowIdx: this.props.rowIdx, idx: this.props.idx});
+      meta.onCellClick({rowIdx: this.props.rowIdx, idx: this.props.idx}, e);
     }
   },
 
-  onCellDoubleClick() {
+  onCellDoubleClick(e) {
     let meta = this.props.cellMetaData;
     if (meta != null && meta.onCellDoubleClick != null) {
-      meta.onCellDoubleClick({rowIdx: this.props.rowIdx, idx: this.props.idx});
+      meta.onCellDoubleClick({rowIdx: this.props.rowIdx, idx: this.props.idx}, e);
     }
   },
 
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -318,8 +318,8 @@ const Cell = React.createClass({
     return (this.props.column.editor != null) || this.props.column.editable;
   },
 
-  createEventCallBack(e, onColumnEvent, info) {
-    return () => {
+  createEventCallBack(onColumnEvent, info) {
+    return (e) => {
       onColumnEvent(e, info);
     };
   },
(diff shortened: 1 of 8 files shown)


src/addons/grids/ReactDataGrid.js
```diff
@@ -86,31 +86,46 @@ const ReactDataGrid = React.createClass({
...
-            && rowIdx < this.props.rowsCount
-          ) {
-          this.setState({selected: selected});
-        }
+  hasSelectedCellChanged: function(selected: SelectedType) {
+    let previouslySelected = Object.assign({}, this.state.selected);
+    return previouslySelected.rowIdx !== selected.rowIdx || previouslySelected.idx !== selected.idx || previouslySelected.active === false;
+  },
+
+  onSelect: function(selected: SelectedType, associatedEvent: string) {
```
diogofcunha: To implement a generic behaviour and don't violate SRP associated events should not be bound with `onSelect`. 
The problem comes when the state needs to change, how can we make sure outside the `setState` callback that when `onColumnEvent `is triggered the state will contain the last selected column/Cell

src/addons/grids/ReactDataGrid.js
```diff
@@ -86,31 +86,46 @@ const ReactDataGrid = React.createClass({
...
+        });
+      }
+    } else {
+      this.onColumnEvent(associatedEvent);
+    }
+  },
+
+  onColumnEvent: function(columnEvent: string) {
+    if (columnEvent) {
+      let column = this.getColumn(this.state.selected.idx);
```
bakesteve: shouldnt we pass in the column (or column index) rather than assuming that onColumnEvent fires only on teh selectedColumn? also falls into the trap @JamesPortelli highlighted that `setState` is batched (though reading back up, you are accounting for that above in your setState)

src/addons/grids/ReactDataGrid.js
```diff
@@ -86,31 +86,46 @@ const ReactDataGrid = React.createClass({
...
+
+      if (column && column.events && column.events[columnEvent] && typeof column.events[columnEvent] === 'function') {
+        column.events[columnEvent]();
```
bakesteve: we should pass some useful args here. probably the column and row index? though for row, would we need to differentiate between a header click and data ele click?
while we can sniff that from grid.state, that always feels like we are being naughty and looking under the covers.

bakesteve:
ok, with a column shae, this seems a reasonable extension point, though think we should pass some data out to the func by default (as a single object, we can then use destructuring if you only want 1 prop)

diogofcunha:
@bakesteve you mean passing the useful information mentioned on the previous comment as a object for `onColumnEvent` instead of the simple string?

diogofcunha:
any ideas about event execution? hard-coding the event args like in [here](https://github.com/adazzle/react-data-grid/blob/dc-column-events/src/addons/grids/ReactDataGrid.js#L124) doesn't look like the best way to do it, left some ideas in the PR description.

bakesteve:
for the event args, ideally we'd have reference to the `SyntheticEvent` still. is that possible to pass up?

for the signature of the `col.events.onXXX` I think that shoudl be something like `function onClick(ev :SyntheticEvent, args: {rowIdx :number, column :ColumnShape}) { ... }`

diogofcunha:
`onCellClick` and `onCellDoubleClick` are not using the `SyntheticEvent ` but pass that up is straight forward for those 2 events, the rest of the supported events are already using it.

I agree with the signature like that.

src/Cell.js
```diff
@@ -348,7 +386,7 @@ const Cell = React.createClass({
     let dragHandle = (!this.isActive() && this.canEdit()) ? <div className="drag-handle" draggable="true" onDoubleClick={this.onDragHandleDoubleClick}><span style={{display: 'none'}}></span></div> : null;
 
     return (
-      <div {...this.props} className={className} style={style} onClick={this.onCellClick} onDoubleClick={this.onCellDoubleClick} onDragOver={this.onDragOver}>
+      <div {...this.props} className={className} style={style} onClick={this.columnEventDispatcher} onDoubleClick={this.columnEventDispatcher} onDragOver={this.columnEventDispatcher}>
```
malonecj: If we use {...column.events} spread syntax into the div we can dynamically bind each column event rather than hardcode them as above. Although, we should leave onClick and onDoubleClick, and make sure default functionality of those two events is not overwritten

src/addons/grids/ReactDataGrid.js
```diff
@@ -444,6 +469,17 @@ const ReactDataGrid = React.createClass({
     this.onSelect({idx: idx, rowIdx: rowIdx});
   },
 
+  openCellEditor(rowIdx, idx) {
+    let selected = {rowIdx, idx};
+    if (this.hasSelectedCellChanged(selected)) {
```
malonecj: We should check for canEdit here as well

src/__tests__/Cell.spec.js
```diff
@@ -110,5 +111,68 @@ describe('Cell Tests', () => {
...
+        // Arrange.
+        let cellInstance = TestUtils.renderIntoDocument(<Cell {...testProps} />);
+        cellInstance.createCellEventCallBack = jasmine.createSpy();
+
+        // Act.
+        cellEvents = cellInstance.getEvents();
+
+        // Assert.
+        expect(cellInstance.createCellEventCallBack).toHaveBeenCalled();
+      });
```
malonecj: :+1:

PR #311, opened 2016-05-05
Stop the grid from autofocusing

peterpme:
If a grid live below the fold, the page will scroll down to the selected area.

A grid can _still_ be selectable without autofocusing it.

Thanks!

### DIFF
--- a/src/addons/grids/ReactDataGrid.js
+++ b/src/addons/grids/ReactDataGrid.js
@@ -1,18 +1,19 @@
-const React                 = require('react');
-const ReactDOM = require('react-dom');
-const BaseGrid              = require('../../Grid');
-const Row                   = require('../../Row');
-const ExcelColumn           = require('./ExcelColumn');
-const KeyboardHandlerMixin  = require('../../KeyboardHandlerMixin');
-const CheckboxEditor        = require('../editors/CheckboxEditor');
+const React                = require('react');
+const ReactDOM             = require('react-dom');
+const BaseGrid             = require('../../Grid');
+const Row                  = require('../../Row');
+const ExcelColumn          = require('./ExcelColumn');
+const KeyboardHandlerMixin = require('../../KeyboardHandlerMixin');
+const CheckboxEditor       = require('../editors/CheckboxEditor');
 const DOMMetrics           = require('../../DOMMetrics');
-const ColumnMetricsMixin      = require('../../ColumnMetricsMixin');
-const RowUtils = require('../../RowUtils');
-const ColumnUtils = require('../../ColumnUtils');
+const ColumnMetricsMixin   = require('../../ColumnMetricsMixin');
+const RowUtils             = require('../../RowUtils');
+const ColumnUtils          = require('../../ColumnUtils');
 
 if (!Object.assign) {
   Object.assign = require('object-assign');
-}
+};
+
 type SelectedType = {
   rowIdx: number;
   idx: number;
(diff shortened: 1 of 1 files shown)


jpdriver:
mm this doesn't feel quite right though. having the first cell selected with the checkboxes active makes sense. i see this more as preventing focus being stolen than just never having a cell selected?

that `checkFocus()` method does actually attempt to take focus on `body` into account, but doesn't have anything to detect whether or not the grid is actually visible within the browser viewport.

https://github.com/adazzle/react-data-grid/blob/f5656244c18c287aee5c7a14e73d56f11365e7f9/src/Cell.js#L295

so that's probably where the change needs to go? be good to have some official input here too @malonecj 😄

src/addons/grids/ReactDataGrid.js
```diff
@@ -77,13 +77,22 @@ const ReactDataGrid = React.createClass({
 
   getInitialState: function(): {selected: SelectedType; copied: ?{idx: number; rowIdx: number}; selectedRows: Array<Row>; expandedRows: Array<Row>; canFilter: boolean; columnFilters: any; sortDirection: ?SortType; sortColumn: ?ExcelColumn; dragged: ?DraggedType;  } {
     let columnMetrics = this.createColumnMetrics();
-    let initialState = {columnMetrics, selectedRows: [], copied: null, expandedRows: [], canFilter: false, columnFilters: {}, sortDirection: null, sortColumn: null, dragged: null, scrollOffset: 0 };
-    if (this.props.enableCellSelect) {
```
bakesteve: we may want to look at this from avoiding the `.focus()` on having a selectedCell, as even if you did have a grid and explictly wanted to start with another cell selected (say 1,4), but its below the fold, we dont want to auto scroll in view

jpdriver: so probably something like a check to see if the grid's parentOffset is > window.innerHeight within checkFocus?

peterpme:
Thanks for the feedback!

I think `enableCellSelect` is over-stepping its bounds from a naming perspective because it does more than just enable row selection, it enables: 

- the ability to select rows
- automatically selecting the first `[0, 0]`

I can see something like this (and I'll happily PR it) making a bit more sense:

`enableCellSelect: true` -> enables the ability to select a cell,
`initialSelectedCell: [0, 0]` -> focuses the desired cell

This also makes sense:

`enableCellSelect: true`,
`enableCellEdit: true` -> this doesn't select a cell, it just gives you the ability to edit one
`initialSelectedCell: [0, 0]`

jpdriver:
yeah fair point about `enableCellSelect` doing more than it should, and like @bakesteve said you might want to start on something else like 1,4 so taking in props for that could be a nice addition.

but i still think `checkFocus()` should be a bit more intelligent itself. you could absolutely just pass in -1, -1 and avoid the focus problem, but then anyone else who runs into this is going to have to implement the same hack / workaround.. the less of those we have the better 😅 

definitely feels like two issues to me? one for focus-stealing and another for default selected cell?

peterpme:
I'd be happy to PR it all in one go to be honest :smile:

- `enableCellSelect` (bool: true/false, default: false) - ability to select a cell
- `enableCellEdit` (bool: true/false, default: false) - ability to edit a cell (no selection method required)
- `initialSelectedCell` (array, default: [-1, -1]) - ability to set an initially selected cell if you choose

I think `checkFocus()` can be improved on as well, but I do agree, THAT should be handled in a separate PR :)

What do you think?

bakesteve:
yes, that sounds sensible enough and gives more fine grained props control

jpdriver:
don't think we need `enableCellEdit` though? that should be controlled by the `editable = true / false` prop on the column? looks good other than that

peterpme:
If you set `enableCellSelect` to `false` but have `editable = true` in your COLUMN, you aren't capable of 

You need to have `editable` set to `true` AND `enableCellSelect` set to `true` if you want to edit a cell.

bakesteve:
True, essentially for edit, we need cellSelect. We’ll have a chat here about the best approach tomorrow and comment back here

peterpme:
I probably jumped the gun here, but I made a few adjustments:

- cleaned up component spacing
- normalize methods by removing extraneous `: function()` (there were more w/o it)
- added logic to handle editing & selecting a cell

peterpme:
@bakesteve hey, any other discussions that took place? I'd love to make this happen!

My PR might be a little ambitious with the additional changes, but I can strip those :)

eddanger:
Was a solution ever found for this autoselecting issue?

peterpme:
The solution exists in this PR, but it looks like the team didn't believe this was a problem and closed the issue.

Because of that, I've sinced moved on from react-data-grid. You're welcome to take a look through this, open up a new PR and give it another shot 😄.

There are only a couple changes required to make it happen.

cwiese:
What react grid do you use now??

peterpme:
Take a look at https://bvaughn.github.io/react-virtualized/#/components/Grid

If you look through the tickets and search across GitHub for examples, you'll find exactly what you're looking for! It's very extensible and HIGHLY performant and definitely worth the time to set up!

The owner of the repo actually got hired at Facebook not too long ago.

My schedule is pretty tied up right now, but if I can find some time over the next couple of weeks, I'd be happy to try and abstract my code out and open source it! 😄

Issue #307, opened 2016-05-04
move to next/previous row when selection moves beyond last/first cell in row


chadi-kazan:
right now the only way to move to other rows is with the up and down arrow keys. 
can we have it automatically jumping to first cell in the next row/last cell in previous row if the user tries to navigate beyond the end/start of a row?
it should be something fairly simple to implement by changing moveSelectedCell along the lines below
```
moveSelectedCell(e: SyntheticEvent, rowDelta: number, cellDelta: number) {
	    // we need to prevent default as we control grid scroll
	    // otherwise it moves every time you left/right which is janky
	    e.preventDefault();
			let rowIdx = this.state.selected.rowIdx;
			let idx = this.state.selected.idx + cellDelta;
			if (cellDelta > 0){
				if (this.state.selected.idx === this.props.columns.length - 1 && this.state.selected.rowIdx < this.props.rowsCount -1) {
					rowDelta = rowDelta + 1;
					idx = 0;
				}
			} else if (cellDelta < 0){
				if (this.state.selected.idx === 0 && this.state.selected.rowIdx > 0) {
					rowDelta = rowDelta - 1;
					idx = this.props.columns.length - 1;
				}
			}
	    rowIdx = rowIdx + rowDelta;
	    this.onSelect({ idx: idx, rowIdx: rowIdx });
	  }
```

bakesteve:
Nice idea, though possible that people might want to just loop round the same row? So right on the last cell takes me to the 1st cell on the same row?

chadi-kazan:
then in this case it can be parameterized to do one or the other with props

PR #316, opened 2016-05-13
move to previous/next row when selection moves beyond first/last cell in row

chadi-kazan:
pull request for [https://github.com/adazzle/react-data-grid/issues/307](https://github.com/adazzle/react-data-grid/issues/307). 

- move selection to next/previous row when user navigates beyond last/first cell in a row.

- added a property "enableCellSelectionLoop" to enable/disable looping around the same row's cells when navigating with keyboard. when set to true, moving the selection beyond the last cell will select the first cell in the same row.

### DIFF
--- a/examples/scripts/example20-cell-navigation.js
+++ b/examples/scripts/example20-cell-navigation.js
@@ -78,7 +78,8 @@ ReactDOM.render(<ReactDataGrid
   rowGetter={rowGetter}
   rowsCount={_rows.length}
   minHeight={500}
-  enableCellSelect={true}/>, mountNode);
+  enableCellSelect={true}
+  cellNavigationMode="changeRow" />, mountNode);
 `;
 
 module.exports = React.createClass({
--- a/examples/scripts/example20-cell-navigation.js
+++ b/examples/scripts/example20-cell-navigation.js
@@ -78,8 +78,7 @@ ReactDOM.render(<ReactDataGrid
   rowGetter={rowGetter}
   rowsCount={_rows.length}
   minHeight={500}
-  enableCellSelect={true}
-  enableCellSelectionLoop={false} />, mountNode);
+  enableCellSelect={true}/>, mountNode);
 `;
 
 module.exports = React.createClass({
@@ -87,9 +86,10 @@ module.exports = React.createClass({
   render:function(){
     return(
       <div>
-        <h3>Column Navigation Example</h3>
-        <p>By setting enableCellSelectionLoop={true}, you enable looping round the same row when navigation goes beyond the first/last cells.</p>
-        <p>The default behavior is to jump to the next/previous row.</p>
+        <h3>Column Navigation Modes Example</h3>
+        <p>By setting <code>cellNavigationMode = 'loopOverRow'</code>, you enable looping round the same row when navigation goes beyond the first/last cells.</p>
+        <p>Setting <code>cellNavigationMode = 'changeRow'</code>, would make the selection jump to the next/previous row.</p>
+        <p>The default behavior is to do nothing.</p>
         <ReactPlayground codeText={NavigationExample} />
       </div>
     )
--- a/examples/scripts/example20-cell-navigation.js
+++ b/examples/scripts/example20-cell-navigation.js
@@ -1,0 +1,98 @@
+var ReactGrid             = require('../build/react-data-grid');
+var QuickStartDescription = require('../components/QuickStartDescription')
+var ReactPlayground       = require('../assets/js/ReactPlayground');
+
+var NavigationExample = `
+
+//helper to generate a random date
+function randomDate(start, end) {
+  return new Date(start.getTime() + Math.random() * (end.getTime() - start.getTime())).toLocaleDateString();
+};
+
+var _rows = [];
+for (var i = 1; i < 1000; i++) {
+  _rows.push({
+    id: i,
+    task: 'Task ' + i,
+    complete: Math.min(100, Math.round(Math.random() * 110)),
+    priority : ['Critical', 'High', 'Medium', 'Low'][Math.floor((Math.random() * 3) + 1)],
+    issueType : ['Bug', 'Improvement', 'Epic', 'Story'][Math.floor((Math.random() * 3) + 1)],
+    startDate: randomDate(new Date(2015, 3, 1), new Date()),
+    completeDate: randomDate(new Date(), new Date(2016, 0, 1))
+  });
+};
+
+//function to retrieve a row for a given index
+var rowGetter = function(i){
+  return _rows[i];
+};
+
+
+//Columns definition
+var columns = [
+{
+  key: 'id',
+  name: 'ID',
+  locked : true
+},
+{
+  key: 'task',
+  name: 'Title',
+  width: 200
+},
+{
+  key: 'priority',
+  name: 'Priority',
+  width: 200
+},
+{
+  key: 'issueType',
+  name: 'Issue Type',
+  width: 200
+},
+{
+  key: 'complete',
+  name: '% Complete',
+  width: 200
+},
+{
+  key: 'startDate',
+  name: 'Start Date',
+  width: 200
+},
+{
+  key: 'completeDate',
+  name: 'Expected Complete',
+  width: 200
+},
+{
+  key: 'completeDate',
+  name: 'Expected Complete',
+  width: 200
+}
+]
+
+
+ReactDOM.render(<ReactDataGrid
+  columns={columns}
+  rowGetter={rowGetter}
+  rowsCount={_rows.length}
+  minHeight={500}
+  enableCellSelect={true}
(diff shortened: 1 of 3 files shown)


diogofcunha:
Hi chadi, nice piece of work, will be reviewed very soon.

Could you please add this feature to [all the features example](https://github.com/chadi-kazan/react-data-grid/blob/master/examples/scripts/example13-all-features.js)?

If you're up to it you could also create a new example where cell selection and this new feature are both active?

Regards

chadi-kazan:
Hi Diogo,

I just created a separate example. 

cheers

src/addons/grids/ReactDataGrid.js
```diff
@@ -480,9 +482,27 @@ const ReactDataGrid = React.createClass({
     // we need to prevent default as we control grid scroll
     // otherwise it moves every time you left/right which is janky
     e.preventDefault();
-    let rowIdx = this.state.selected.rowIdx + rowDelta;
+    let rowIdx = this.state.selected.rowIdx;
     let idx = this.state.selected.idx + cellDelta;
-    this.onSelect({idx: idx, rowIdx: rowIdx});
+    let _rowDelta = rowDelta;
+    const { enableCellSelectionLoop } = this.props;
+    if (cellDelta > 0) {
```
diogofcunha: Think you complicated a little bit in this one don't you? 

1. Why run this piece of code when `enableCellSelectionLoop` is false in the first place? It will just make the if/else chain grow which equals more complexity 
3. For the sake of readability can you please put the if conditions that check state in a new method? For example replacing `this.state.selected.idx === this.props.columns.length - 1 && this.state.selected.rowIdx < this.props.rowsCount - 1` for a method called `isFinalCellInTheRowAndNotLastRow()`?
2. The enable/disable is not working right and we want to keep this turned off by default, replacing that piece of code for something like this will do the trick

```
    if (enableCellSelectionLoop) {
      if (cellDelta > 0 && this.isFinalCellInTheRowAndNotLastRow()) {
        _rowDelta = rowDelta + 1;
        idx = 0;
      } else if (cellDelta < 0 && this.isFirstCellInTheRowAndNotFirstRow()) {
        _rowDelta = rowDelta - 1;
        idx = this.props.columns.length - 1;
      }
    }
```

examples/scripts/example20-cell-navigation.js
```diff
@@ -0,0 +1,98 @@
...
+  enableCellSelect={true}
+  enableCellSelectionLoop={false} />, mountNode);
+`;
+
+module.exports = React.createClass({
+
+  render:function(){
+    return(
+      <div>
+        <h3>Column Navigation Example</h3>
```
diogofcunha: You forgot to change the example header name

examples/scripts/example20-cell-navigation.js
```diff
@@ -0,0 +1,98 @@
...
+        <p>By setting enableCellSelectionLoop={true}, you enable looping round the same row when navigation goes beyond the first/last cells.</p>
```
diogofcunha: `enableCellSelectionLoop={true}` the `{true}` will not be displayed in the page, wrap this with `<code></code> `tag please

src/addons/__tests__/Grid.spec.js
```diff
@@ -231,6 +231,58 @@ describe('Grid', function() {
     });
   });
 
+  describe('Cell Navigation', function() {
+    describe('when cell selection looping is disabled', function() {
+      beforeEach(function() {
+        this.component = this.createComponent({enableCellSelectionLoop: false, enableCellSelect: true});
```
diogofcunha: shouldn't `enableCellSelect `be false for this set of tests?

src/addons/__tests__/Grid.spec.js
```diff
@@ -231,6 +231,58 @@ describe('Grid', function() {
...
+      });
+
+      describe('when on last cell', function() {
+        beforeEach(function() {
+          this.component.setState({ selected: { idx: 3, rowIdx: 1 } });
+        });
+        it('selection should move to first cell in next row', function() {
```
diogofcunha: Cell Navigation when cell **selection looping is disabled** when on last cell **selection shouldn't move to first cell in next row**

src/addons/__tests__/Grid.spec.js
```diff
@@ -231,6 +231,58 @@ describe('Grid', function() {
...
+          this.simulateGridKeyDown('Tab');
+          expect(this.component.state.selected).toEqual({ idx: 0, rowIdx: 2 });
+        });
+      });
+      describe('when on first cell', function() {
+        beforeEach(function() {
+          this.component.setState({ selected: { idx: 0, rowIdx: 2 } });
+        });
+        it('selection should move to last cell in previous row', function() {
```
diogofcunha: Cell Navigation when cell **selection looping is disabled** when on first cell **selection shouldn't move to last cell in previous row**

src/addons/__tests__/Grid.spec.js
```diff
@@ -231,6 +231,58 @@ describe('Grid', function() {
...
+          this.simulateGridKeyDown('ArrowLeft');
+          expect(this.component.state.selected).toEqual({ idx: 3, rowIdx: 1 });
```
diogofcunha: the expect in this case needs to be 
`expect(this.component.state.selected).toEqual({ idx: 0, rowIdx: 2 });`

src/addons/__tests__/Grid.spec.js
```diff
@@ -231,6 +231,58 @@ describe('Grid', function() {
...
+        this.component = this.createComponent({enableCellSelectionLoop: false, enableCellSelect: true});
+      });
+
+      describe('when on last cell', function() {
+        beforeEach(function() {
+          this.component.setState({ selected: { idx: 3, rowIdx: 1 } });
+        });
+        it('selection should move to first cell in next row', function() {
+          this.simulateGridKeyDown('Tab');
+          expect(this.component.state.selected).toEqual({ idx: 0, rowIdx: 2 });
```
diogofcunha: the expect in this case needs to be 
expect(this.component.state.selected).toEqual({ idx: 3, rowIdx: 1 });

src/addons/grids/ReactDataGrid.js
chadi-kazan: because the default behavior would be for it to move to first cell in next row, last cell in previous row. i think the prop name might be confusing, by enableCellSelectionLooping, i mean to make it loop around to the beginning/end of the row. Otherwise, when set to false, it would move to next/previous row rather than do nothing

chadi-kazan: will extract the check methods from that code

diogofcunha:
Good job and thanks for the contribution, just left a couple of comments in the code that need to be addressed, after that it will be fine for merge.

cheers

src/addons/__tests__/Grid.spec.js
chadi-kazan: I think you are confusing the intended functionality this new prop will unlock.
when selection looping is disabled, it would move to next row. When it's enabled, it would move to the first cell in the row. i guess i should give the user 3 rather than 2 navigation options:

1. move to next/previous row
2. move to first/last cell
3. do nothing

Currently it defaults to 1 and 3. does not exist

diogofcunha: Didn't understood the main purpose on the first look I'm sorry for that.

Basicly we don't want to completely drop the current functionality, we want to be able to:

1. do what the grid does now,
2. loop in a row
3. move to next/previous row when selection moves beyond last/first cell in row.

The default behaviour should be the current behaviour so that people using the grid at the moment don't end up having unexpected behaviours.

Sorry for the misunderstood.

chadi-kazan: no worries! i absolutely agree! i was thinking the same thing too!

I will change the implementation to bring the default behavior back

diogofcunha: Maybe put the logic for the new behaviours in a new method so that `moveSelectedCell`doesn't start to grow out of control? Something like:

```
if (isCellSelectionRowLoopEnabled || isCellSelectionGridLoopEnabled) {
   let {rowDelta, idx} = this.calculateNextPosition(...);
}
```

examples/scripts/example20-cell-navigation.js
```diff
@@ -0,0 +1,98 @@
...
+  width: 200
+}
+]
+
+
+ReactDOM.render(<ReactDataGrid
+  columns={columns}
+  rowGetter={rowGetter}
+  rowsCount={_rows.length}
+  minHeight={500}
```
diogofcunha: can set this one to be `cellNavigationMode = 'changeRow'`?

chadi-kazan: ok

diogofcunha:
Great work and contribution, nice set of tests.

Cheers

PR #318, opened 2016-05-13
Column editable to cater for functions.

JamesPortelli:
Add functionality for column.editable to allow functions.

### DIFF
--- a/src/addons/grids/ReactDataGrid.js
+++ b/src/addons/grids/ReactDataGrid.js
@@ -517,9 +517,14 @@ const ReactDataGrid = React.createClass({
       this.setState({selected: selected});
     }
   },
-
+    
+  // Logic extented to allow for functions to be passed down in column.editable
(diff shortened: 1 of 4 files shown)


src/addons/grids/ReactDataGrid.js
```diff
@@ -517,9 +517,14 @@ const ReactDataGrid = React.createClass({
       this.setState({selected: selected});
     }
   },
-
+    
+  // Logic extented to allow for functions to be passed down in column.editable
+  // this allows us to deicde whether we can be edting from a cell level
   canEdit(idx: number): boolean {
     let col = this.getColumn(idx);
```
malonecj: to remove duplication you can put this function into columnUtils.js and create a test for it there, using a fake column with an editable function

malonecj: instead of referencing props in the function, pass the row and column as parameters

PR #319, opened 2016-05-16
Fix for Drag double click feature

diogofcunha:
Feature was broken [in here](https://github.com/adazzle/react-data-grid/commit/d8b3e36de47bd0e1285175a62dfd707d9ed26f00). 

To fix it it will be a matter of fixing the if clause [in here](https://github.com/adazzle/react-data-grid/blob/master/src/Cell.js#L102) to be` meta != null && meta.onDragHandleDoubleClick && typeof(meta.onDragHandleDoubleClick) === 'function'`

Need to write some tests for this meta functions to make sure no one breaks this any more.

- [x] Failing test for onDragHandleDoubleClick
- [x] Fix if clause
- [x] Write tests to avoid further similar breaking changes

@willgates Will take this one

### DIFF
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -99,7 +99,7 @@ const Cell = React.createClass({
   onDragHandleDoubleClick(e) {
     e.stopPropagation();
     let meta = this.props.cellMetaData;
-    if (meta != null && typeof(meta.onDragHandleDoubleClick != null) === 'function') {
+    if (meta != null && typeof(meta.onDragHandleDoubleClick) === 'function') {
       meta.onDragHandleDoubleClick({rowIdx: this.props.rowIdx, idx: this.props.idx, rowData: this.getRowData(), e});
     }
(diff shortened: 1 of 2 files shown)


src/Cell.js
```diff
@@ -99,7 +99,7 @@ const Cell = React.createClass({
   onDragHandleDoubleClick(e) {
     e.stopPropagation();
     let meta = this.props.cellMetaData;
-    if (meta != null && typeof(meta.onDragHandleDoubleClick != null) === 'function') {
+    if (meta != null && typeof(meta.onDragHandleDoubleClick) === 'function') {
```
diogofcunha: we should also check if `meta.onDragHandleDoubleClick` is null or undefined (by doing `..&& meta.onDragHandleDoubleClick && ..`

diogofcunha:
Looks good to me, nice job

PR #322, opened 2016-05-18
configure default formatter and editor for boolean columns

chadi-kazan:
optionally enable immediate editing of checkbox cells upon selection.

I created simple default editor and formatter for checkboxes. well, it basically renders checkboxes for columns marked as boolean columns with `isBoolean: true`.

the idea is to have excel like behavior where a boolean value is rendered as a checkbox and if editable, one won't have to activate edit mode to edit it, just like typing a char immediately starts editing text values. the edit mode is immediately activated upon cell selection if one chooses too by setting `checkRadioEditOnSelect={true}`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-data-grid",
-  "version": "0.14.36",
+  "version": "0.14.37",
   "description": "Data grid for React",
   "main": "src/index.js",
   "scripts": {


diogofcunha:
Hi @chadi-kazan.

Regarding the `checkRadioEditOnSelect` I don't think it makes much sense to add that functionality because of the recently added feature that allows you to associate [events to columns](http://adazzle.github.io/react-data-grid/examples.html#/column-events). Seems to me that what you are looking for is something like what is implemented on the second and fourth column of that example, Am I right?

The grid is supposed to act as a powerful core that you can extend, and in my opinion `checkRadioEditOnSelect` is not generic enough to worth to effort on implement/maintain, but let's involve @bakesteve and @malonecj in this matter.

malonecj:
Hi @chadi-kazan. Thanks for your pull request

I agree with @diogofcunha . We are trying to simplify the configuration options of the grid. I feel like this behavior can already be achieved with a formatter and using custom column events. However, perhaps it might be desirable to have an option to display an editor by default without having to activate

chadi-kazan:
hey guys, I honestly haven't checked column events. Can we setup an onCellSelected event? this way we can maybe achieve the same result in an event handler. It is desirable to have the editor activated by default without an action to trigger it. What if one wants to navigate and edit the grid using keyboard only?

chadi-kazan:
hey guys
any news on this? i checked and it seems only after selection events are currently supported. I am gonna add cell selection and deselection event handler props. that i guess would give us the desired flexibility. would that be ok?

chadi-kazan:
i will ditch this and go for cell selection events. it makes more sense

PR #329, opened 2016-05-24
implement cell selection/deselection events

chadi-kazan:
add props to enable firing events for when a cell is selected/deselected. This would be a step towards full keyboard navigation functionality support. It could be used to activate the cell editor for example without the need to trigger via a click or a key down on the cell.


diogofcunha:
Great contribution once again @chadi-kazan.
LGTM and tested on local.

Cheers.

PR #333, opened 2016-05-25
default checkbox editor and formatter for boolean columns

chadi-kazan:
Checkbox editor and formatter to use with boolean columns. it is important to have the right editor/formatter for this type of data as it does not make sense to edit boolean values with text fields.
we could use the default formatter which would show true/false but i find it does not make sense either.

### DIFF
--- a/examples/scripts/example22-checkbox-editor-formatter.js
+++ b/examples/scripts/example22-checkbox-editor-formatter.js
@@ -1,0 +1,135 @@
+var QuickStartDescription = require('../components/QuickStartDescription')
+var ReactPlayground       = require('../assets/js/ReactPlayground');
+
+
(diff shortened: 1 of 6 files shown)


examples/scripts/example22-checkbox-editor-formatter.js
```diff
@@ -0,0 +1,135 @@
...
+
+ReactDOM.render(<Example />, mountNode);
+`;
+
+  module.exports = React.createClass({
+
+    render:function(){
+      return(
+        <div>
+          <h3>Built-In Cell Editor Example</h3>
```
malonecj: Can you change the description of this example. This has been copied from another example. Thanks

PR #334, opened 2016-05-25
fix: cell navigation broken when row selection is enabled

chadi-kazan:
This is to fix a bug with the cell navigation. When row selection is enabled, a new column -the row selection checkbox- is added to the columns list and we end up with `props.columns.length` not returning the correct number of columns.


chadi-kazan:
sorry the commit message is wrong, it's not only when it's set to multi but always when it's enabled.

chadi-kazan:
hello, any news on this?

malonecj:
Thanks for the PR. Will merge now

PR #365, opened 2016-07-05
Declare peer deps against React 15

cburgmer:
We would like to use react-data-grid with React 15.2.0, but npm keeps complaining about unmet peer dependencies.

Especially using `npm shrinkwrap` seems to blow up with react-data-grid pointing to an older version of `react`.

The alternative, hosting react-data-grid locally inside a vendor dir within the Git repository does not help, as react-data-grid seems to have a non-trivial packaging setup, which fails `npm outdated` for example.


cburgmer:
It looks like the same issue exists with dependency `ron-react-automcomplete`. As this is a fork of `react-autocomplete` I don't know what's the prefered way of resolving this. Revert back to the original package (expecting it to have resolved the reason for the fork), or updating the fork, which seems to be lagging behind.

yarivkenan:
is there a way  around it for the time being?

gnestor:
How about using "^0.14.0 || ^15.0.0-0"? I've taken a sample of React components and this seems to be the most conventional semver to depend on:

https://github.com/alexkuz/react-json-tree/blob/master/package.json#L61-L63
https://github.com/facebook/fixed-data-table/blob/master/package.json#L6-L9
https://github.com/ReactTraining/react-router/blob/master/package.json#L41-L43
https://github.com/reactjs/react-chartjs/blob/master/package.json#L29-L33

cburgmer:
Could you please clarify why that was rejected? The latest version still hasn't seen any update on this.

jpdriver:
hey @cburgmer -- the update to React 15 is being worked on as we speak. you can follow it's progress here https://github.com/adazzle/react-data-grid/issues/558

PR #366, opened 2016-07-05
Fixed rows container plugins related bug

diogofcunha:
<h3>Bug description</h3>
`getContextMenuContainer` was running either if the container was using `react data grid plugins` or not, obviously a error was thrown in [RowsContainer Component constructor](https://github.com/adazzle/react-data-grid/blob/master/src/RowsContainer.js#L20) when the plugins were not in use and not accessible in the window. that is wrong.

### DIFF
--- a/src/RowsContainer.js
+++ b/src/RowsContainer.js
@@ -14,20 +14,22 @@ SimpleRowsContainer.propTypes = {
 };
 
 class RowsContainer extends React.Component {
-
-  constructor() {
-    super();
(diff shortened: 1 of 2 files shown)


aqilzeeshan:
@diogofcunha this build is failing https://ci.appveyor.com/project/adazzle/react-data-grid/build/1.0.1482 can you look into that.

diogofcunha:
@aqilzeeshan failed because of npm publish, it's ok

PR #371, opened 2016-07-10
Fixed Warning: Unknown prop

TheLus:
https://facebook.github.io/react/warnings/unknown-prop.html

I couldn't use spread operator, so I used Array reduce and indexOf.

If I can use spread operator, rewrite this.
```
// example Cell.js
const {rowIdx, idx, selectedselectedColumn, ref, column, isExpanded, isRowSelected, cellMetaData, handleDragStart, cellControls, rowData, forceUpdate, ...divProps} = this.props;
```

### DIFF
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -430,6 +430,13 @@ const Cell = React.createClass({
   },
 
   render() {
+    const divProps = Object.keys(this.props).reduce((props, key) => {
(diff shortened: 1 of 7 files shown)


ro-savage:
Would like to this or another update that gets read of the UnknownProp types warning merged.

malonecj:
Hi, there are 24 failing tests in this PR which need to be addressed

Issue #416, opened 2016-08-08
Uncaught InvalidStateError: Failed to read the 'selectionStart' property when using non text editor


chadi-kazan:
**I'm submitting a ...**  (check one with "x")

```
[ x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

when navigating with the arrows, if one of the cells has, for example, a checkbox editor, you get the following error

> Uncaught InvalidStateError: Failed to read the 'selectionStart' property from 'HTMLInputElement': The input element's type ('checkbox') does not support selection.

while it tries to check for the caret position in EditorContainer.

**Expected/desired behavior**
It should ignore the caret test, and anything related to text input if something other than a text element is used.

**Reproduction of the problem**
use the following editor

```
class CellCheckboxEditor extends React.Component {

    getStyle() {
        return {
            width: '100%'
        };
    }

    getValue() {
        let updated = {};
        const node = this.getInputNode();
        updated[this.props.column.key] = node.checked;
        return updated;
    }

    getInputNode() {
        let domNode = ReactDOM.findDOMNode(this._input);
        if (domNode.tagName === 'INPUT') {
            return domNode;
        }

        return domNode.querySelector('input:not([type=hidden])');
    }

    inheritContainerStyles() {
        return true;
    }
    render() {
        const {value} = this.props;
        let checked = value === null ? false : value;
        return <div><input type="checkbox" ref={c => {
            this._input = c;
        }} defaultChecked={checked} style={{maxHeight: '18px', maxWidth: '18px'}}/></div>;
    }
}

CellCheckboxEditor.propTypes = {
    value: React.PropTypes.bool,
    column: React.PropTypes.object
};
```

and activate it on the cell in the onCellSelectedEvent for that cell

```
onCellSelected(cell) {
    if (cell.idx === checkboxIndex) {
            this._grid.openCellEditor(cell.rowIdx, cell.idx);
     }
  },
```

enter edit mode in one of the adjacent cells and navigate **with the arrows** to the checkbox cell

PR #421, opened 2016-08-10
fix for issue #416

mttomorowicz:
## Description

Only 'text' type input elements have the selection start / end checks done on them (in different keyDown event handlers).

```
[ x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/adazzle/react-data-grid/issues/416

What is the new behavior?**

Things are good now.

```
[ ] Yes
[ x] No
```


src/addons/editors/CheckboxEditor.js
```diff
@@ -12,6 +13,14 @@ const CheckboxEditor = React.createClass({
     dependentValues: React.PropTypes.object
   },
 
+  getValue(){
+    return this.getInputNode().value;
```
diogofcunha: We shouldn't get the value from the DOM, that's not a reactive approach, the checkbox value should be added to state and changed accordingly, in the `getValue` we just need to return the value saved in state.

diogofcunha:
Aside from the way we get the value on `CheckboxEditor` it looks good to me, as this sits on the shell for some time now and it fixes an actual issue I will find time to do the change myself during this week.

Thanks @mttomorowicz

PR #427, opened 2016-08-12
Row reordering

malonecj:
## Description
Allows drag and drop reordering of rows via addons module

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

**What is the new behavior?**

```
[ ] Yes
[ ] No
```

**Other information**:


src/Cell.js
```diff
@@ -69,48 +69,32 @@ const Cell = React.createClass({
     }
   },
 
-  shouldComponentUpdate(nextProps: any): boolean {
```
diogofcunha: We shouldn't remove this as it can damage performance.

src/ReactDataGrid.js
```diff
@@ -102,7 +102,7 @@ const ReactDataGrid = React.createClass({
     onRowClick: React.PropTypes.func,
     onGridKeyUp: React.PropTypes.func,
     onGridKeyDown: React.PropTypes.func,
-    rowGroupRenderer: React.PropTypes.func
```
diogofcunha: Deleting this will break the `rowGroupRenderer`, it is injected as a prop in [BaseGrid](https://github.com/adazzle/react-data-grid/blob/0b05b5303b4d3c7e630b54a22c5571b8b60b9f6a/src/ReactDataGrid.js#L874) using spread syntax.

src/addons/draggable/RowActionsCell.js
```diff
@@ -0,0 +1,68 @@
+import React, { PropTypes } from 'react';
+import { DragSource } from 'react-dnd';
+import CheckboxEditor from '../editors/CheckboxEditor';
+
+class RowActionsCell extends React.Component {
+
+  constructor(props) {
```
diogofcunha: This is redundant, react will do it by himself without need for constructor

src/addons/draggable/RowDragLayer.js
```diff
@@ -0,0 +1,122 @@
...
+      rows.push(rowGetter(j));
+    }
+    return rows;
+  }
+
+  getDraggedRows() {
+    let draggedRows;
+    let {rowSelection} = this.props;
+    if (rowSelection && rowSelection.selectBy.keys) {
+      let rows = this.getRows();
```
diogofcunha: Will he get the rows every time we render?
Can the `rowCount` be big?
Can it damage the performance? If not looks ok to me, if it can do (quick profile to figure out?) I guess we should put the rows in state and update only when needed.

src/addons/draggable/RowDragLayer.js
```diff
@@ -0,0 +1,122 @@
...
+  renderDraggedRows() {
+    return this.getDraggedRows().map((r, i) => {
+      return <tr key={`dragged-row-${i}`}>{this.renderDraggedCells(r, i) }</tr>;
+    });
+  }
+
+  renderDraggedCells(item, rowIdx) {
+    let cells = [];
+    if (item != null) {
+      for (let c in item) {
```
diogofcunha: Shouldn't this be a map? Looks like the length for the original and result array are the same

diogofcunha:
@malonecj for now LGTM, just left a couple of comments

src/addons/draggable/RowDragLayer.js
malonecj: yep, lets cache this in state. Its used to give the row preview when you drag so we just need to call it once

Issue #463, opened 2016-09-22
Draggable rows also does a drag to copy


michaeljacobdavis:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

**Current behavior**
The selected cells value is duplicated across all rows that are dragged over.
![](http://g.recordit.co/Rmr9mqoDZk.gif)

**Expected/desired behavior**
The row is dragged and the cell is deselected.

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem.
This [Gist](https://gist.github.com/michaeljacobdavis/bac988bee0603db970f0cbb2cf961d15) recreates the problem when pasted into the examples editor.

Basically it's when you have editable/draggable cells and draggable rows enabled.

**Solution**
My first thought is to add a check in [`onDragStart`](https://github.com/adazzle/react-data-grid/blob/master/src/ReactDataGrid.js#L312) that checks if *what* is being dragged, but I'm not super familiar with the internals of react-data-grid, so I'd love some guidance on fixing this. I have no problem doing a PR for it.

diogofcunha:
Hi @michaeljacobdavis, thanks for the feedback.

That bug was fixed in 1.0.43.

michaeljacobdavis:
@diogofcunha I can reproduce this in master currently. Please checkout this [Gist](https://gist.github.com/michaeljacobdavis/bac988bee0603db970f0cbb2cf961d15) and paste it into one of the live examples http://adazzle.github.io/react-data-grid/examples.html#/row-reordering.

michaeljacobdavis:
@diogofcunha This is still an issue. Can you reopen?

PR #467, opened 2016-09-23
Check for existing drag and drop context before creating

michaeljacobdavis:
## Description

This allows consumers of `react-data-grid` to use the draggable rows on the same page as another component that uses `react-dnd`.

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently react-data-grid adds a react-dnd backend/context in the `DragDropContainer` for draggable rows. This works if the rest of the page does not use `react-dnd` however you get errors if you're including it elsewhere (see https://github.com/gaearon/react-dnd/issues/186).

For future reference, these are some of the errors because of this:
- `Cannot have two HTML5 backends at the same time`
- `Uncaught Invariant Violation: Cannot call beginDrag while dragging.` - (Because events hit both instances)

**What is the new behavior?**
The preferred way to use `react-dnd` is to [wrap the root component of your application](http://gaearon.github.io/react-dnd/docs-drag-drop-context.html) and it then uses `context` to pass the manager down to children. I've added a `dnd-context.js` helper that checks the given context for that and if one exists, uses it (to avoid the two backends error) and if none exists creates one (for backward compatibility).

```
[ ] Yes
[x] No
```

**Other information**:

Just a note, this is really just a workaround to this library including `react-dnd`. Normally this should probably be a peerDependency, but because `react-data-grid` publishes `dist` to npm rather than source, peer dependencies are not used. Any thoughts on changing the distribution process to either compile on install or just provide source files?

Side note, you could speed up consumer install time by moving everything from `dependencies` to `devDependencies` as only `dist` is used.

### DIFF
--- a/src/addons/draggable/DragDropContainer.js
+++ b/src/addons/draggable/DragDropContainer.js
@@ -1,11 +1,16 @@
 import React, {Component} from 'react';
-import HTML5Backend from 'react-dnd-html5-backend';
-import { DragDropContext } from 'react-dnd';
 import DraggableHeaderCell from './DraggableHeaderCell';
 import RowDragLayer from './RowDragLayer';
 import isColumnsImmutable from '../../isColumnsImmutable';
+import getDndContext from './dnd-context';
 
 class DraggableContainer extends Component {
+
+  getChildContext() {
+    return {
+      dragDropManager: getDndContext(this.context)
+    };
+  }
 
   getRows(rowsCount, rowGetter) {
     let rows = [];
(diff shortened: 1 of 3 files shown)


malonecj:
Hi @michaeljacobdavis. Could you please make this pull request mergable and we will get it in. Also, we will take a look at making react-dnd a peer dependency as well as including the source files

michaeljacobdavis:
@malonecj should be good. Could you also check out https://github.com/adazzle/react-data-grid/issues/463? It's somewhat related (has to do with DND) and was closed but is still reproducible on your examples site.

Issue #418, opened 2016-08-09
React throwing Warning: Unkown Props


FauzanKhan:
**I'm submitting a ...**  (check one with "x")

```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

**Current behavior**
I'm on React version 0.14.8 and on initializing the data grid as given [here](http://adazzle.github.io/react-data-grid/examples.html#/editable), I get the following errors in the console: 

```
VM3181:27 Warning: Unknown props `enableCellSelect`, `columns`, `rowGetter`, `rowsCount`, `minHeight`, `onRowUpdated`, `minColumnWidth`, `columnEquality`, `rowHeight`, `enableRowSelect`, `rowKey`, `rowScrollTimeout`, `cellNavigationMode`, `headerRows`, `columnMetrics`, `cellMetaData`, `selectedRows`, `rowSelection`, `expandedRows`, `rowOffsetHeight`, `sortColumn`, `sortDirection`, `onSort`, `totalWidth`, `onViewportKeydown`, `onViewportKeyup`, `onViewportDragStart`, `onViewportDragEnd`, `onViewportDoubleClick`, `onColumnResize` on <div> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by Grid)
    in Grid (created by ReactDataGrid)
    in div (created by ReactDataGrid)
    in div (created by ReactDataGrid)
    in ReactDataGrid (created by UploadCsv)
    in div (created by UploadCsv)
    in UploadCsv

```

This is just one of the many similar warnings I'm getting

Here's how I'm initializing the grid -

```
<ReactDataGrid
              enableCellSelect={true}
              columns={columns}
              rowGetter={this.rowGetter}
              rowsCount={this.state.rows.length}
              minHeight={500}
              onRowUpdated={this.handleRowUpdated} />
```

**Expected/desired behavior**

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem.

**What is the expected behavior?**
The error is thrown multiple times for different props. None of the errors should be there.

**What is the motivation / use case for changing the behavior?**

malonecj:
@FauzanKhan  Thanks for that. We'll take a look

masterfung:
I am getting the same error but on react 15.2.1

[...]

PR #528, opened 2016-11-20
Update components to only pass known properties when rendering divs (fix for #418)

supamanda:
## Description
Only pass known properties when rendering divs

- [x ] The commit message follows our guidelines: https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
- [x ] Tests for the changes have been added (for bug fixes / features)
- [n/a ] Docs have been added / updated (for bug fixes / features)

```
[x ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

See #418 

**What is the new behavior?**
Only known properties will be passed to divs, so will not print warnings about unknown props to the console.

```
[ ] Yes
[x ] No
```

**Other information**:

### DIFF
--- a/src/Cell.js
+++ b/src/Cell.js
@@ -8,6 +8,10 @@ const isFunction = require('./addons/utils/isFunction');
 const CellMetaDataShape = require('./PropTypeShapes/CellMetaDataShape');
 const SimpleCellFormatter = require('./addons/formatters/SimpleCellFormatter');
 const ColumnUtils = require('./ColumnUtils');
+const createObjectWithProperties = require('./createObjectWithProperties');
+
+// The list of the propTypes that we want to include in the Cell div
+const knownDivPropertyKeys = ['height', 'tabIndex', 'value'];
 
 const Cell = React.createClass({
 
@@ -409,6 +413,10 @@ const Cell = React.createClass({
     return this.createEventDTO(gridEvents, columnEvents, onColumnEvent);
   },
 
+  getKnownDivProps() {
+    return createObjectWithProperties(this.props, knownDivPropertyKeys);
+  },
+
   renderCellContent(props) {
     let CellContent;
     let Formatter = this.getFormatter();
@@ -448,7 +456,7 @@ const Cell = React.createClass({
     let events = this.getEvents();
 
     return (
-      <div {...this.props} className={className} style={style}   {...events}>
+      <div {...this.getKnownDivProps()} className={className} style={style}   {...events}>
         {cellContent}
         {dragHandle}
       </div>
--- a/src/Draggable.js
+++ b/src/Draggable.js
@@ -1,12 +1,17 @@
 const React         = require('react');
 const PropTypes     = React.PropTypes;
+const createObjectWithProperties = require('./createObjectWithProperties');
+
+// The list of the propTypes that we want to include in the Draggable div
+const knownDivPropertyKeys = ['onDragStart', 'onDragEnd', 'onDrag', 'style'];
 
 const Draggable = React.createClass({
   propTypes: {
     onDragStart: PropTypes.func,
     onDragEnd: PropTypes.func,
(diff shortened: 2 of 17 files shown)


malonecj:
@supamanda  Thanks for the great contribution. I will merge this now.

Also, this pr seems to introduce a very beneficial side effect.  When scrolling downwards using the keyboard, the scrolling is now really smooth (in chrome). A quick test in the timeline shows an increase in the framerate ~30fps. Did you notice this? I assume it is because we are not passing so many unnecessary props to the div. However it only improves scrolling downwards. Scrolling horizontally, and upwards exhibits the same janky behavior. I'll need to investigate this. Let me know if you have any thoughts on that

malonecj:
Actually, that side effect seemed to be because my browser window was zoomed out to 80%. I don't know why this should affect the scrolling performance though. I will look into that one

supamanda:
@malonecj  that would have been an impressive side effect! I would take a guess that to improve scrolling performance would have something to do with caching data so it's ready to display, but haven't looked into it at all.

gaurav-quasar:
@supamanda I was using below code for some necessary calculation of mouse up/down for row.
Is there any other way to achieve the same in latest release.

`        <Row
          ref={(c) => { this.row = c; }}
          onMouseDown={e => this.props.handleRowMouseDown(e, this.props)}
          onMouseUp={e => this.props.handleRowMouseUp(e, this.props)}
          {...this.props}
        />`

gaurav-quasar:
Is it somehow possible to keep knownDivPropertyKeys customizable?

supamanda:
hi @gaurav-quasar ,

The knownDivPropertyKeys are currently hardcoded in the components, and they only determine what gets put in the html elements. I'm not sure how that affects your mouse up/down calculations - if you set up a gist or some running code that shows your problem, I'd be happy to take a look

gaurav-quasar:
@supamanda ,

We cannot pass events in it anymore which are not mentioned in knownDivPropertyKeys,
Please have a look at Pull request#731. @maximecharron tried to add 'onDoubleClick' to knownDivPropertyKeys to bring back his functionality.

supamanda:
I see. Well in answer to your second question - currently knownDivPropertyKeys is hardcoded. In answer to your first question, I have personally been using the column events, row selection/deselection events, grid events to configure events throughout the grid. I'm pretty sure you could get one of these to do the calculations you want.

Maybe a good solution for #731 would be to allow configuration for other row events, similar to column events - although it's possible these would produce conflicting events so whoever implemented it would need to check what impact it has.

maximecharron:
@supamanda However, there is a side effect to what you introduced in this PR. Since all the props linked to React's events are not in the hardcoded list, it becomes impossible to use them. I opened #731 for the doubleClick event, but that's not the only one known by React that gets removed by your changes here.

supamanda:
hi @maximecharron  - all the events I mentioned above are still working (you can test them out on the demos). As I said above, I would be happy to have a look if you set up some sample running code or a gist.

maximecharron:
@supamanda the events you mention are the Datagrid's events. However, we should still be able to use native event like onDoubleClick. You can take a look to #697 for an example that was working before the changes introduced by this PR.

gaurav-quasar:
Ideally knownDivPropertyKeys should filter out custom events/props which were specific to parent components not native events.

supamanda:
Previous to this PR, the onDoubleClick would be on the Row div as well as the Cell div - in my opinion, this is too loose. I think it is better to set them up in column events or grid events so that you can get the row and cell index, and I could (& would) use these to implement the onDoubleClick event that you want.

Regarding #697: passing onDoubleClick via the row renderer looks to me like a workaround because you can't define row events. But as you say, that is what you want to do. Maybe you should submit a pull request? Best of luck :)

Issue #531, opened 2016-11-22
Clicking to sort column does not update the sort direction icon, and only the first click sorts the data, introduced between (inclusive) 1.0.57, before 1.0.62


supamanda:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

**Current behavior**
Clicking on the header cell of a sortable column does not update the sort direction icon, and only the first click sorts the data

**Expected/desired behavior**
When grid column is sortable, you should be able to click on the column header and change the sort direction, NONE -> ASC -> DESC -> NONE, which was the previous behaviour

**Reproduction of the problem**

Go to http://adazzle.github.io/react-data-grid/examples.html#/sortable-cols
Click on the Title column once

Expected: Title data is sorted ascending, Title column should show ASC arrow
Actual: Title data is sorted ascending. Title column shows no arrow

Click on the Title column again

Expected: Title data is sorted descending, Title column should show DESC arrow
Actual: Title data is still sorted ascending. Title column shows no arrow

ocularrhythm:
Just noticed this behavior as well since updating to 1.0.64

PR #533, opened 2016-11-23
Fix for #531

supamanda:
## Description
Fixes column sorting issue #531 

- [x ] The commit message follows our guidelines: https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

```
[x ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

See #531 

**What is the new behavior?**
When grid column is sortable, you should be able to click on the column header and change the sort direction, NONE -> ASC -> DESC -> NONE, which was the previous behaviour

```
[ ] Yes
[x ] No
```

**Other information**:
* While investigating this issue, I noticed HeaderRow redefined DEFINE_SORT as an array and was using it incorrectly (ie calling DEFINE_SORT.NONE)
* I considered different ways of implementing this - eg pass sortColumn & sortDescription as props to HeaderCell and then checking if these changed, to compare in shouldComponentUpdate method whether the renderers had changed but this always returned true. In the end I thought the best way to do this was to remove the shouldComponentUpdate method because we are allowing users to define different renderers which might change without us being able to accurately detect it in the HeaderCell method.

### DIFF
--- a/src/HeaderCell.js
+++ b/src/HeaderCell.js
@@ -4,7 +4,6 @@ const joinClasses    = require('classnames');
 const ExcelColumn    = require('./PropTypeShapes/ExcelColumn');
 const ResizeHandle   = require('./ResizeHandle');
 const PropTypes      = React.PropTypes;
-import shallowCompare from 'react-addons-shallow-compare';
 
 function simpleCellRenderer(objArgs: {column: {name: string}}): ReactElement {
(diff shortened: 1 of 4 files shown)


malonecj:
Fixes #531

PR #572, opened 2016-12-19
Pass headerRenderer to SortableHeaderCell when column is sortable

computerlove:
## Description
Currently it is not possible that a column both is sortable and has a custom renderer. This PR is a suggestion for how this can be solved.

Will update with tests if this is a sane way to solve this.

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`column.headerRenderer` takes precedence of column.sortable, so the internal state for `sortColumn` and `sortDirection` is in effect hidden and the header is not updated if handling sorting of these columns is attempted outside ReactDataGrid.

**What is the new behavior?**
Passing the header renderer to `SortableHeaderCell` makes it possible to have a sortable column with custom renderer.

```
[ ] Yes
[ ] No
```

I am not quite sure whether this would break anything.


talcohen100:
Is this going to be merged? it's an important fix that isn't available.

nevace:
This is really needed. Didn't realise I'd lose functionality by having a custom header.

nevace:
@computerlove you're probably long gone by now but I manually added your changes above. A couple of things in case anyone else stumbles across this:

The column properties aren't passed to the headerRenderer this way, so you don't have any context of the cell you're rendering (name, keyword, key, etc.). This is pretty hacky but I passed the column props to the headerRenderer like so:

`getHeaderRenderer(column) {
    let renderer;
    
    if (column.headerRenderer) {
    const columnProps = Object.assign({}, column);
    delete columnProps.headerRenderer;

    column.headerRenderer = React.cloneElement(column.headerRenderer, {column: columnProps});
    }`

Sofiane000:
When will this be merged  and @nevace can this fix my issue of headerRenderer error #933?

packages/react-data-grid/src/cells/headerCells/SortableHeaderCell.js
```diff
@@ -13,6 +13,7 @@ class SortableHeaderCell extends React.Component {
     column: PropTypes.shape({ name: PropTypes.node }),
     onSort: PropTypes.func.isRequired,
     sortDirection: PropTypes.oneOf(Object.keys(DEFINE_SORT)),
+    headerRenderer: React.PropTypes.node,
```
amanmahajan7: Deprecated `React.PropTypes`

packages/react-data-grid/src/cells/headerCells/SortableHeaderCell.js
```diff
@@ -52,13 +53,13 @@ class SortableHeaderCell extends React.Component {
       'react-grid-HeaderCell-sortable--ascending': this.props.sortDirection === 'ASC',
       'react-grid-HeaderCell-sortable--descending': this.props.sortDirection === 'DESC'
     });
-
+    let content = this.props.headerRenderer ? this.props.headerRenderer :  this.props.column.name;
```
amanmahajan7: `const` instead of `let`

packages/react-data-grid/src/cells/headerCells/SortableHeaderCell.js
```diff
@@ -52,13 +53,13 @@ class SortableHeaderCell extends React.Component {
...
     return (
       <div className={className}
         onClick={this.onClick}
         style={{cursor: 'pointer'}}>
+        {content}
```
amanmahajan7: Is changing the order intentional?

malonecj: This is an old PR. It seems that the order has been changed in master since it was created. I'll keep it to what it is in master

PR #577, opened 2016-12-21
Fix for header scroll when container/grid width cannot be determined

sonny-chan:
## Description

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When the container width cannot be determined the header will not scroll.

This is because this.props.totalWidth will be 100%.

and when the headerRowStyle gets calculated the code will attempt to do:
"100%" - scrollbarSize
which will result in NaN

**What is the new behavior?**
The header should scroll correctly

```
[ ] Yes
[x] No
```

**Other information**:

### DIFF
--- a/src/Header.js
+++ b/src/Header.js
@@ -92,11 +92,12 @@ const Header = React.createClass({
         rowHeight = '500px';
       }
       let scrollbarSize = getScrollbarSize() > 0 ? getScrollbarSize() : 0;
+      let updatedWidth = isNaN(this.props.totalWidth - scrollbarSize) ? this.props.totalWidth : this.props.totalWidth - scrollbarSize;
       let headerRowStyle = {
         position: 'absolute',
         top: this.getCombinedHeaderHeights(index),
         left: 0,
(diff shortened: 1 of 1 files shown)


src/Header.js
```diff
@@ -92,11 +92,12 @@ const Header = React.createClass({
         rowHeight = '500px';
       }
       let scrollbarSize = getScrollbarSize() > 0 ? getScrollbarSize() : 0;
+      let updatedWidth = isNaN(this.props.totalWidth - scrollbarSize) ? this.props.totalWidth : this.props.totalWidth - scrollbarSize;
```
richardware: @malonecj @sonny-chan `Number.isNaN(value)` might have been a marginally more robust approach here. It's not exactly a major issue, but it's worth using the new version wherever we would normally use `isNaN`.

More: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

Issue #600, opened 2017-01-12
enableDragAndDrop doesn't seem to be used


thomas-shares:
**WHICH VERSION OF REACT ARE YOU USING?**
```
Officially Supported:
[X] v0.14.x
```
```
Community Supported:
[ ] v15.0.x
```
**WHICH BROWSER ARE YOU USING?**
```
Officially Supported:
[ ] IE 9 / IE 10 / IE 11
[ ] Edge
[ ] Chrome
```
```
Should work:
[X] Firefox
[ ] Safari
```
--------
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report
[X] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```
**Current behavior**
We are currently experiencing issue #225 and as a work around we were thinking about disabling dragging completely. A investigation of the docs revealed the enableDragAndDrop option and setting this to 'false' didn't give us the expected behaviour. We did a grep over the code and it looks like that the enableDragAndDrop flag isn't used anywhere in the code and thus not functional.

**Expected/desired behavior**
If react-data-grid is initialised with enableDragAndDrop set to false one would expect that dragging would not be possible. And this in turn would work as a work around of issue #225.

**Reproduction of the problem**
NA

**What is the expected behavior?**
NA

**What is the motivation / use case for changing the behavior?**
Initial motivation for us would be to workaround issue #225, but also to make the code match the expectation as set in the documentation.

thomas-shares:
ps. I am willing to help with this, but don't have enough JS experience for this at the moment.

thomas-shares:
I have added a PR that allows the user to turn off DraggingAndDropping. Unfortunately this doesn't address issue #225 .

thomas-shares:
@diogofcunha  as issue #225 has now been fixed (issue #603) do we still want to pursue this issue and its associated PR? 

Please advice.

Thomas

Issue #225, opened 2016-02-24
Dragging to fill down redirects Mac OS Firefox to dummy.com


moodysalem:

moodysalem:
Actually it even happens in the demo
http://adazzle.github.io/react-data-grid/examples.html#/editable

moodysalem:
And it happens with or without a drag handler

ibrahima:
Happens on Firefox on Linux too, this is really bizarre!

adeboni:
It also happens on Firefox on Windows

chris-brannen-alps:
Just wondering if anyone had been able to fix this, if not is there a way to disable the drag functionality?

LiranV:
I've encountered the same behavior on Firefox (Linux).

thomas-shares:
I am seeing this as well and I can also recreate it at http://adazzle.github.io/react-data-grid/examples.html#/cell-drag-down with the latest FireFox 50.1.0 on Mac OS and also on Linux (with the same FireFox version).

Would there be a way to work around this, for instance by disabling dragging for instance?

thomas-shares:
I have made a PR for issue #600, but that, unfortunately, doesn't work around this issue. Any advice on how to deal with this would be greatly appreciated.
Thank you

PR #602, opened 2017-01-13
Take enableDragAndDrop into account.

thomas-shares:
## Description
Take the enableDragAndDrop flag into account.

**Please check if the PR full fills these requirements**
- [X] The commit message follows our guidelines: https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

In issue #600 I pointed out that the enableDragAndDrop flag isn't taken into account. This is code that enables that. Please note that issue #225 isn't fixed with this fix.

**What is the new behavior?**
When enableDragAndDrop is set to false you can't DragAndDrop. If it is undefined or set to true it will work.

```
[ ] Yes
[X] No
```

**Other information**:
I haven't update the unit tests for this. Please advice how to do this. Thank you

### DIFF
--- a/src/ReactDataGrid.js
+++ b/src/ReactDataGrid.js
@@ -118,6 +118,7 @@ const ReactDataGrid = React.createClass({
       minHeight: 350,
       rowKey: 'id',
       rowScrollTimeout: 0,
+      enableDragAndDrop: true,
       cellNavigationMode: 'none',
       overScan: {
         colsStart: 5,
@@ -857,7 +858,7 @@ const ReactDataGrid = React.createClass({
   },
 
   dragEnabled: function(): boolean {
-    return (this.props.onGridRowsUpdated !== undefined || this.props.onCellsDragged !== undefined) && this.props.enableDragAndDrop !== false;
+    return this.props.enableDragAndDrop && (this.props.onGridRowsUpdated || this.props.onCellsDragged);
   },
 
   renderToolbar(): ReactElement {
--- a/src/ReactDataGrid.js
+++ b/src/ReactDataGrid.js
@@ -857,7 +857,7 @@ const ReactDataGrid = React.createClass({
   },
 
   dragEnabled: function(): boolean {
-    return this.props.onGridRowsUpdated !== undefined || this.props.onCellsDragged !== undefined;
+    return (this.props.onGridRowsUpdated !== undefined || this.props.onCellsDragged !== undefined) && this.props.enableDragAndDrop !== false;
   },
 
   renderToolbar(): ReactElement {


src/ReactDataGrid.js
```diff
@@ -857,7 +857,7 @@ const ReactDataGrid = React.createClass({
   },
 
   dragEnabled: function(): boolean {
-    return this.props.onGridRowsUpdated !== undefined || this.props.onCellsDragged !== undefined;
+    return (this.props.onGridRowsUpdated !== undefined || this.props.onCellsDragged !== undefined) && this.props.enableDragAndDrop !== false;
```
diogofcunha: Can we check `enableDragAndDrop` first and only then check the callbacks? I guess most times it won't need to check the callbacks, checking if the drag and drop is enable first sounds like a reasonable thing to do.

Aside from that, having a falsy check in here is safe and would make the code much more readable.

Would end up being

```
return this.props.enableDragAndDrop && (this.props.onGridRowsUpdated || this.props.onCellsDragged);

```

thomas-shares: Hi @diogofcunha,

The down side of your proposal would be that if enableDragAnDrop is not explicitly set to true (ie. undefined) DragAndDrop is suddenly disabled. It does make sense though to put the enable test at the beginning though.

And would it make sense to turn of the "drag" box when DragAndDrop is disabled (at the moment it isn't)

Please advice.

Thomas

diogofcunha: Hey @thomas-shares.

Enabling by default won't do any warm as we also check for the callbacks.

You can set `enableDragAndDrop` to true in the `defaultProps`

thomas-shares: Just testing that at the moment ;)

thomas-shares:
Please review the this PR again and please advice on whether Cell.js line 469 should be updated as well with respect of the dragHandle (not display the dragHandle if dragging is disabled).

thomas-shares:
@diogofcunha  Could you please have a look again, at your convenience of course. Thank you.

diogofcunha:
Hi @thomas-shares, we are making some changes in https://github.com/adazzle/react-data-grid/pull/603.

I was looking at the git history and I think that prop was added by mistake, we need to decide if the overload of having an extra prop (when you get the same result when you don't  inject the callback to handle the action is a good idea)

thomas-shares:
Hi @diogofcunha,

 Issue #603 has fixed our issue. Thank you very much for that. Feel free to discard this PR if you think that is the correct thing. If you want to proceed with it, or remove the enableDragAndDrop flag please advice. I'll try and help where I can.

Issue #547, opened 2016-12-01
Set default sorting column and sorting direction while initializing


JosephGu:
Hi team,

I'd like to set a arrow up icon for specified column while initializing,(i.e., We can handle the sorting in the backend service, and just need to set the icons to the specified column)

How could I implement it?

tdzienniak:
Hey, I've just made PR resolving this issue. I have some questions about this functionality:
- should setting sort column and direction initially trigger `onGridSort` callback?
- or maybe user should provide already sorted data?
- shoud there be a check that column in `sortColumn` prop is sortable?

mgscreativa:
Hi! In my case I'm using meteor to fetch a paginated list. Triggering onGridSort handler calls a subscription that sorts the data server side then publishes the paginated data back to the client. To get that subscription data I must use **createContainer** HOC.

In my scenario, it will be just cool to have those props to tweak them in my render function, so after subscription is ready, the render function should get the sort order and column and set that on the ReactDataGrid component because the data got sorted on the server.

Please merge this!

Thanks!

GlebDolzhikov:
Why version 3 doesn't contain it? plss merge it

amanmahajan7:
This feature has been added in #649

PR #649, opened 2017-02-08
Set sort column and direction with props

tdzienniak:
## Description

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

#547 

**What is the new behavior?**

Sorting direction can be set with props during grid initialization.

```
[ ] Yes
[x] No
```

**Other information**:


packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -66,6 +66,8 @@ const ReactDataGrid = React.createClass({
     onCellsDragged: React.PropTypes.func,
     onAddFilter: React.PropTypes.func,
     onGridSort: React.PropTypes.func,
+    sortColumn: React.PropTypes.string,
+    sortDirection: React.PropTypes.oneOf(['ASC', 'DESC', 'NONE']),
```
jpdriver: hey @tdzienniak -- this looks good, but one suggestion: can we use the `DEFINE_SORT` const from here

https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid/src/cells/headerCells/SortableHeaderCell.js#L3

rather than having multiple arrays of possible values?

There's probably also an argument to be made for moving the existing array out of `SortableHeaderCell`.

My two cents are that `sortColumn` and `sortDirection` are in the right place here, but the existing definitions should be moved here too? Sorting should conceptually apply to the whole grid, not just HeaderCells..

tdzienniak: I agree. Maybe we can move these definitions to `Constants.js` or `AppConstants.js` file?

stevewillard:
What's the status of this PR? I would really like this feature.

packages/react-data-grid/src/ReactDataGrid.js
lakshmi-ammu: Hi, i am facing one issue if i want to keep " link" inside the column may i know the process in react data grid.waiting your replay.

jasonpatt:
Any update on the status of this pull request? This is a needed feature that my team is waiting for.

packages/react-data-grid/src/ReactDataGrid.js
tdzienniak: @jpdriver now prop type check is using `DEFINE_SORT` constants from `SortableHeaderCell.js` file. I think that moving these definition to other file needs separate PR, maybe global refactor of dealing with constant values as these.

GlebDolzhikov:
Please add this feature to package, this three lines of code, but very important to have it! 
I can't install forked version because of multi packages structure of repo(

tdzienniak:
@jpdriver I can correct any issues with this PR. If there are none, why not merge it? I see that people are interested in this change.

mgscreativa:
Hi! In my case I'm using meteor to fetch a paginated list. Triggering onGridSort handler calls a subscription that sorts the data server side then publishes the paginated data back to the client. To get that subscription data I must use **createContainer** HOC.

In my scenario, it will be just cool to have thoose props to tweak them in my render function, so after sibscription is ready, the render function should get the sort order and columns and set that on the ReactDataGrid component because the data got sorted on the server.

Please merge this!

Thanks!

hiredgunhouse:
What is the status of this feature?

mrgurdeep:
@jpdriver @tdzienniak Should you have a suggestion on what change you are expecting. I can make those changes in this PR.

Otherwise we can accept this PR.

mlmassey:
Please pull this in.  Minor change and is a needed feature for my usage

kozaxinan:
What is missing in this pr? Why is it not merged already? Does it need any change or update?

jimmed:
For those looking to work around this issue without forking the project, you can use React's `ref` API along with ReactDataGrid's `handleSort` method, like so:

```js
class MyAwesomeGridWrapper extends Component {

  // ...

  componentDidMount() {
	const { sortColumn, sortDirection } = this.props;
    if (!this._grid || !sortColumn || !sortDirection) return;
	this.grid.handleSort(sortColumn, sortDirection);
  }

  // ...

  render() {
    return (<DataGrid ref={grid => (this.grid = grid)} ... />)
  }
}
```

PR #666, opened 2017-02-16
Autocomplete Filter does not work with number and undefined values

diegomurakami:
## Description
During the implementation of this filter type, It was not included tests for other value types. Now I included more tests and now the filter can compare other types of data.
Also included a minor fix on the header renderer.

Please check if the PR fulfills these requirements**

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When filtering number or undefined values, and exceptions happens. The actual code is only expecting strings to filter.
The simpleCellRenderer for the HeaderRow renders as default the column name even when it is a filter or sortable row.

**What is the new behavior?**
The filter now handles other values types.
The simpleCellRendererdefault now renders column name only for rowType 'header'.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
+++ b/packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
@@ -38,12 +38,11 @@ class AutoCompleteFilter extends React.Component {
         continue;
       }
 
-      if (typeof columnValue !== 'undefined' && typeof filterTerms[key].value !== 'undefined') {
+      if (columnValue !== undefined && filterTerms[key].value !== undefined) {
+        let strColumnValue = columnValue.toString();
         let filterTermValue = filterTerms[key].value.toString();
-
-        let checkValueIndex = columnValue.toString().trim().toLowerCase().indexOf(filterTermValue.trim().toLowerCase());
-        let columnMatchesSearch = checkValueIndex !== -1 && (checkValueIndex !== 0 || columnValue.toString() === filterTermValue);
-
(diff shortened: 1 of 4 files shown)


packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
```diff
@@ -38,14 +38,16 @@ class AutoCompleteFilter extends React.Component {
         continue;
       }
 
-      let filterTermValue = filterTerms[key].value;
+      if (typeof columnValue !== 'undefined' && typeof filterTerms[key].value !== 'undefined') {
```
diogofcunha: We not `columnValue !== undefined`? It's less verbose

packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
```diff
@@ -38,14 +38,16 @@ class AutoCompleteFilter extends React.Component {
...
+        let filterTermValue = filterTerms[key].value.toString();
 
-      let checkValueIndex = columnValue.trim().toLowerCase().indexOf(filterTermValue.trim().toLowerCase());
-      let columnMatchesSearch = checkValueIndex !== -1 && (checkValueIndex !== 0 || columnValue === filterTermValue);
+        let checkValueIndex = columnValue.toString().trim().toLowerCase().indexOf(filterTermValue.trim().toLowerCase());
```
diogofcunha: call `columnValue.toString()` only once as this can change. put this on a var named `columnValueText` for example.

PR #683, opened 2017-02-27
Fix default heigth for header row

joao-inacio:
## Description
Currently if no prop is given for header filters row heigth, it will assume the regular rows' height which is not enough  to render the filters properly. 
We also changed the behaviour of not proving a `headerRowHeight` - from just using the same height as all the others row, to have a default of 45px.

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Header row filter's default height is not big enough 

**What is the new behavior?**
When no `headerFiltersHeight` is given, we use the previously hard coded value of 45px. We also don't assume a different value for `headerRowHeight`, which will continue to have the same height as a regular row if no prop for it is given.

```
[ ] Yes
[x] No
```

**Other information**:
n/a

### DIFF
--- a/packages/react-data-grid-addons/src/__tests__/Grid.spec.js
+++ b/packages/react-data-grid-addons/src/__tests__/Grid.spec.js
@@ -105,7 +105,7 @@ describe('Grid', function() {
   });
 
   // Set of tests for the props that defined the height of our rows
-  describe('when defininig heigths on props', function() {    
+  describe('when defininig heigths on props', function() {
     describe('for defaults props', function() {
       beforeEach(function() {
         const ToolBarStub = new StubComponent('Toolbar');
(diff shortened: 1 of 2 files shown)


packages/react-data-grid-addons/src/__tests__/Grid.spec.js
```diff
@@ -104,6 +104,76 @@ describe('Grid', function() {
...
   });
 
+  // Set of tests for the props that defined the height of our rows
+  describe('when defininig heigths on props', function() {    
+    describe('for defaults props', function() {
+      beforeEach(function() {
+        const ToolBarStub = new StubComponent('Toolbar');
+        this.component = this.createComponent({ toolbar: <ToolBarStub /> }).node;
+        this.toolbarInstance = TestUtils.findRenderedComponentWithType(this.component, ToolBarStub);
+        this.toolbarInstance.props.onToggleFilter();
```
malonecj: This is fine. In future though, take a look at http://airbnb.io/enzyme/
You'll find the tests much easier to write, and its what we should be using for React component tests going forward

joao-inacio: Yeah I just followed on on what was there already - what would be a better way of doing this?

PR #704, opened 2017-03-20
Don't autofocus if the entire react-data-grid is offscreen

theunixbeard:
## Description
Due to the current autofocus behavior, when the react-data-grid is rerendered the scroll bar 'jumps' up to put the focused cell in view, even if it was a completely different section of the UI that caused the underlying data to change. This would more often occur in applications with complex UIs such as the one I am working on.

I created a short 1-minute video demonstrating the issue & explaining my fix:
https://www.youtube.com/watch?v=xejQdVGJX6w

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When the react-data-grid is rerendered the scroll bar 'jumps' up to put the focused cell in view, even if it was a completely different section of the UI that caused the underlying data to change.

**What is the new behavior?**
In the isFocusedOnBody() method in Cell.js, we simply check if the dataGridDOMNode is currently on screen or off screen. If it's off screen, then don't trigger the autofocus.

```
[ ] Yes
[x] No
```

**Other information**:
Tests have been added. I also ran the linter and that gave a clean output too.

Thanks for taking a look at this PR!

### DIFF
--- a/packages/react-data-grid/src/Cell.js
+++ b/packages/react-data-grid/src/Cell.js
@@ -347,7 +347,23 @@ const Cell = React.createClass({
     return !this.isSelected() && this.isDraggedOver() && this.props.rowIdx > dragged.rowIdx;
   },
 
-  isFocusedOnBody() {
+  isTableOnScreen(dataGridDOMNode) {
+    // http://stackoverflow.com/a/11193613
+    const docViewTop  = (window.pageYOffset !== undefined) ? window.pageYOffset : (document.documentElement || document.body.parentNode || document.body).scrollTop;
+    // http://stackoverflow.com/a/28241682
+    const docViewBottom = docViewTop + (window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight);
+
+    const elemTop = dataGridDOMNode.offsetTop;
+    const elemBottom = elemTop + dataGridDOMNode.offsetHeight;
+
+    if ((elemTop >= docViewTop && elemTop <= docViewBottom) || (elemBottom >= docViewTop && elemBottom <= docViewBottom)) return true;
+    return false;
+  },
+
+  isFocusedOnBody(dataGridDOMNode) {
+    // Make sure the table is on-screen as prerequisite
+    if (!this.isTableOnScreen(dataGridDOMNode)) return false;
+
     return document.activeElement == null || (document.activeElement.nodeName && typeof document.activeElement.nodeName === 'string' && document.activeElement.nodeName.toLowerCase() === 'body');
   },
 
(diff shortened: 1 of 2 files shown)


jpdriver:
@theunixbeard thanks for opening such a detailed PR!

supamanda:
Good idea. I think it would make more sense to call the method 'isTableOnScreen` or something like that, rather than tying it in with the isFocusedOnBody method

Also see issue https://github.com/adazzle/react-data-grid/issues/714 - i think it should not autofocus when it's focused on the body because chrome & firefox set focus to body during blur which means it may steal focus from other parts of the page, even though it has no business doing so.

packages/react-data-grid/src/Cell.js
```diff
@@ -347,7 +347,23 @@ const Cell = React.createClass({
...
+    const docViewBottom = docViewTop + (window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight);
+
+    const elemTop = dataGridDOMNode.offsetTop;
+    const elemBottom = elemTop + dataGridDOMNode.offsetHeight;
+
+    if ((elemTop >= docViewTop && elemTop <= docViewBottom) || (elemBottom >= docViewTop && elemBottom <= docViewBottom)) return true;
+    return false;
+  },
+
+  isFocusedOnBody(dataGridDOMNode) {
```
supamanda: hi @theunixbeard , it would be better if you changed this back to the original isFocusedOnBody function, and then...

packages/react-data-grid/src/Cell.js
```diff
@@ -363,7 +379,7 @@ const Cell = React.createClass({
       // Only focus to the current cell if the currently active node in the document is within the data grid.
       // Meaning focus should not be stolen from elements that the grid doesnt control.
       let dataGridDOMNode = this.props.cellMetaData && this.props.cellMetaData.getDataGridDOMNode ? this.props.cellMetaData.getDataGridDOMNode() : null;
-      if (this.isFocusedOnCell() || this.isFocusedOnBody() || (dataGridDOMNode && dataGridDOMNode.contains(document.activeElement))) {
+      if (this.isFocusedOnCell() || this.isFocusedOnBody(dataGridDOMNode) || (dataGridDOMNode && dataGridDOMNode.contains(document.activeElement))) {
```
supamanda: ... change this if statement to include ` || this.isFocusedOnBody() || this.isTableOnScreen(dataGridDOMNode) `

(Or better still remove this.isFocusedOnBody() altogether.)

malonecj:
Closing this issue, as auto focus will be removed in v 4.0 of the grid. The plan is to move the responsibility of cell section out of the Cell component and into a dedicated SelectionMask that sits at canvas level. This will provide the following benefits
 - Increased performance as navigating will result in minimal state tree changes
 - Prevent grid stealing focus from other elements
 You can see progress [here](https://github.com/adazzle/react-data-grid/pull/1123)

PR #797, opened 2017-05-18
make react-datagrid able to read json recursively

matrixlice:
Current grid control unable to read json data into rows recursively, add new data-read logic to read data (valueB) like following: 
[
{"propA":{"propB":"valueB"}
{...}
]

```
[ ] Bugfix
[X ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

unable to get cell value if the row JSON object has more than 2 levels

**What is the new behavior?**

able to get values from any kind of JSON object 

```
[ ] Yes
[X] No
```

**Other information**:

### DIFF
--- a/packages/react-data-grid/src/Row.js
+++ b/packages/react-data-grid/src/Row.js
@@ -138,7 +138,20 @@ const Row = React.createClass({
       return this.props.isSelected;
     } else if (typeof this.props.row.get === 'function') {
       val = this.props.row.get(key);
-    } else {
+    } else if (key.indexOf('.') >= 0) {
+      val = (function getVal(valchain, subkey) {
+        const tokenIdx = subkey.indexOf('.');
+        if (tokenIdx >= 0) {
+          const token = key.slice(0, tokenIdx);
+          const remain = key.slice(tokenIdx + 1, key.length);
+          if (valchain[token]) {
+            return getVal(valchain[token], remain);
+          }
+          return null;
+        }
+        return valchain[key] || null;
+      })(this.props.row, key);
+    }else {
       val = this.props.row[key];
     }
     return val;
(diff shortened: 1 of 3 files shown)


packages/react-data-grid/src/Row.js
```diff
@@ -138,7 +138,20 @@ const Row = React.createClass({
...
       val = this.props.row.get(key);
-    } else {
+    } else if (key.indexOf('.') >= 0) {
+      val = (function getVal(valchain, subkey) {
+        const tokenIdx = subkey.indexOf('.');
+        if (tokenIdx >= 0) {
+          const token = key.slice(0, tokenIdx);
+          const remain = key.slice(tokenIdx + 1, key.length);
+          if (valchain[token]) {
+            return getVal(valchain[token], remain);
```
jpdriver: hey @matrixlice -- thanks for the contribution. a couple of quick thoughts:-

1 - please can you extract `getVal()` to a separate utility function so we can include tests against it?

2 - rather than using indexes, slices, and recursively calling `getVal()`, could we instead do `subkey.split('.')` and use `Array.map()` to iterate over that?

here's a similar example:-

```
function getValueForKey (key, model) {
  let returnVal = model
  const keyArray = key.split('.')
  keyArray.map(key => {
    if (returnVal[key]) {
      returnVal = returnVal[key]
    }
  })
  return returnVal
}
```

this avoids the inline fn definition, is a single call, and will safely handle cases with and without subkeys.

what do you think? thanks!

hedikasmanto:
has this PR been implemented ?

packages/react-data-grid/src/Row.js
```diff
@@ -138,7 +139,9 @@ const Row = React.createClass({
       return this.props.isSelected;
     } else if (typeof this.props.row.get === 'function') {
       val = this.props.row.get(key);
-    } else {
+    } else if (key.indexOf('.') >= 0) {
+      val = utils.getValueForKey(this.props.row, key);
```
malonecj: The signature for getValueForKey is `getValueForKey(key, model)`. Looks like it is being called with the wrong parameters here

PR #824, opened 2017-06-07
Extract CSS assets separately during build phase

bovee:
## Description
Modify the webpack build process to split out CSS files instead of inlining them along with a loader into the JS. This allows downstream much more control over how and where to include them.

```
[ ] Bugfix
[ ] Feature
[X] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

CSS is inlined along with webpack's style loader into the npm JS assets. When imported into another project and executed this generates new `<style>` tags in the rendered HTML. Downstream users have no control over this process and it's therefore impossible to override some style properties without using an `!important` because the load order isn't fully under downstream's control.

**What is the new behavior?**

_CSS is split into it's own file during the build phase. This file can then be imported separately._

```
[X] Yes
[ ] No
```

_Applications needs to import CSS assets separately._

**Other information**:

It's might be possible to modify the `index.js` files in each package to also import the CSS and load it dynamically (like what happens before this PR) so this change would be entirely unbreaking. I'm not familiar enough with the rest of the build process (I think `lerna` updates that?) or embedding something like `style-loader` outside of `webpack` to know how feasible that is though.

### DIFF
--- a/config/webpack.common.config.js
+++ b/config/webpack.common.config.js
@@ -1,11 +1,15 @@
 const webpack = require('webpack');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const argv = require('minimist')(process.argv.slice(2));
 const RELEASE = argv.release;
+
+var extractCSS = new ExtractTextPlugin('[name].css');
 
 function getPlugins() {
   const nodeEnv = RELEASE ? '"production"' : '"development"';
   var pluginsBase =  [
-    new webpack.DefinePlugin({'process.env.NODE_ENV': nodeEnv, 'global': 'window'})
+    new webpack.DefinePlugin({'process.env.NODE_ENV': nodeEnv, 'global': 'window'}),
+    extractCSS
   ];
 
   if (RELEASE) {
@@ -41,7 +45,7 @@ const config = {
   module: {
     loaders: [
       { test: /\.js$/, exclude: /node_modules/, loader: 'babel-loader' },
-      { test: /\.css$/, loader: 'style-loader!css-loader' }
+      { test: /\.css$/, loader: extractCSS.extract('style-loader', 'css-loader') }
     ]
   },
   plugins: getPlugins(),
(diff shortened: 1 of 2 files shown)


cstjohn-bhg:
@bovee It would be great to have this included (I'd like to use it in one of my projects) but looks like CI checks failed, which is probably why the maintainers didn't merge.  Any plans to continue work on this?

`
Chrome 57.0.2987 (Windows 8.1 0.0.0) ERROR
  Uncaught Error: Cannot find module "../../../themes/react-data-grid-row.css"
  at C:/projects/react-data-grid/test/FullTests.jsx:4960
`

bovee:
@cstjohn-bhg I would guess that this won't be merged until a major version release (if at all) because it's a breaking change; i.e. users will have to import the CSS in addition to the JS, and I'm reluctant to spend the time changing all the tests over to do that unless there's some sign this is an architectural direction the project maintainers are happy with. So, chicken-egg?

cstjohn-bhg:
@bovee Makes sense, thanks for the response.  

@adazzle In regards to css-loader issues like [this](https://github.com/webpack-contrib/css-loader/issues/240) and  Wouldn't it be best to *not* inject CSS into <style> tags in the DOM automatically?  Perhaps provide a css sheet with the distribution that can be linked to in the consuming application and easily overridden?

malonecj:
First of all apologies for neglecting this issue. I agree it would be best to not inject the CSS into the style tags. We are preparing for the next major release, so could include this there if the test is fixed

mungojam:
Please could this be looked at again, I think if you switch from `require` to `import` for your CSS imports, then they will end up in our create-react-app's bundled CSS file which would then fix the issue since that is imported before the main.css file with our own CSS in.

As you said, at present we have to resort to !important

yasotnik:
Can someone take a look again at this PR? react-data-grid still injects css and we do not have any control against it except the `!important` which I found a very BAD pattern.

PR #835, opened 2017-06-16
Fixed column causes editors to be obscured

sonny-chan:
## Description
Locked columns can sometimes cause editors to appear behind the grid

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Locked columns will sometimes cause editors to appear behind the grid

**What is the new behavior?**
Editors opened in locked columns should appear normally.

```
[ ] Yes
[x] No
```

**Other information**:

### DIFF
--- a/packages/react-data-grid/src/Cell.js
+++ b/packages/react-data-grid/src/Cell.js
@@ -307,7 +307,7 @@ const Cell = React.createClass({
         node.style.transform = transform;
       }
     }
-  }
+  },
 
   isCopied(): boolean {
     let copied = this.props.cellMetaData.copied;
--- a/packages/react-data-grid/src/Cell.js
+++ b/packages/react-data-grid/src/Cell.js
@@ -293,6 +293,7 @@ const Cell = React.createClass({
         let transform = `translate3d(${scrollLeft}px, 0px, 0px)`;
         node.style.webkitTransform = transform;
         node.style.transform = transform;
+        node.style.transformStyle = 'preserve-3d';
       }
     }
   },
--- a/packages/react-data-grid/src/Cell.js
+++ b/packages/react-data-grid/src/Cell.js
@@ -94,7 +94,8 @@ const Cell = React.createClass({
       || this.props.forceUpdate === true
       || this.props.className !== nextProps.className
       || this.props.expandableOptions !== nextProps.expandableOptions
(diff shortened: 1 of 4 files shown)


packages/react-data-grid/src/Cell.js
```diff
@@ -297,6 +297,18 @@ const Cell = React.createClass({
     }
   },
 
+  removeScroll() {
+    let ctrl: any = this; // flow on windows has an outdated react declaration, once that gets updated, we can remove this
+    if (ctrl.isMounted()) {
```
amanmahajan7: This is an anti pattern, any reason for using it?
https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html

packages/react-data-grid/src/Cell.js
```diff
@@ -297,6 +297,18 @@ const Cell = React.createClass({
...
+      let node = ReactDOM.findDOMNode(this);
```
amanmahajan7: We should add a ref on the container [element](https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid/src/Cell.js#L502)

https://facebook.github.io/react/docs/refs-and-the-dom.html
"React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts."
In theory, this should eliminate the need to check isMounted.

packages/react-data-grid/src/Cell.js
```diff
@@ -297,6 +303,14 @@ const Cell = React.createClass({
     }
   },
 
+  removeScroll() {
+    let node = ReactDOM.findDOMNode(this);
```
diogofcunha: 😱 isn't there another way to do this? when using react mutating the DOM directly is dangerous and not adviced at all

sonny-chan: @diogofcunha i'm not sure of another way to do this. if you have any good ideas, please let me know.

I was only following the pattern used in setScrollLeft: https://github.com/adazzle/react-data-grid/blob/9d0e0bff012f9cdedf21c87a19314d2e14bee2e1/packages/react-data-grid-addons/src/draggable/DraggableHeaderCell.js#L18

PR #857, opened 2017-06-29
group by with object array and string array

d4nj0nes:
## Description

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When using grouping you need to override the default toolbar and row renderer to display user friendly names for the columns. 

Could only render a custom toolbar by rendering inline in the property declaration rather than via an import

**What is the new behavior?**
You can pass in the groupBy prop as an object array. 
It assumes current behaviour - that groupBy is an array of strings.
If it is not it expects the objects to have a key and name property.
It will use name to render the groupBy button and the row grouping header.

The toolbar property can only be a React Element not a function. When defining a Generic custom toolbar that enables filtering and grouping in a separate file in TypeScript, nothing was rendered as it failed this validation. 

```
[ ] Yes
[x] No
```

**Other information**:


packages/react-data-grid-addons/src/data/RowGrouper.js
```diff
@@ -21,7 +21,7 @@ class RowGrouper {
 
   groupRowsByColumn(rows, columnIndex = 0) {
     let nextColumnIndex = columnIndex;
-    let columnName = this.columns[columnIndex];
+    let columnName = this.columns.length > 0 && typeof this.columns[columnIndex] === 'string' ? this.columns[columnIndex] : this.columns[columnIndex].key;
```
diogofcunha: Is the length ever `<=0`?

packages/react-data-grid-addons/src/toolbars/GroupedColumnsPanel.js
```diff
@@ -31,7 +31,10 @@ class GroupedColumnsPanel extends Component {
 
   renderGroupedColumns() {
     return this.props.groupBy.map(c => {
-      return (<GroupedColumnButton key={c} name={c} onColumnGroupDeleted={this.props.onColumnGroupDeleted}/>);
```
diogofcunha: Would rather not repeat the jsx assignments, it can be:

```
const groupedColumnButtonProps = {
   columnKey: c,
   name: c,
   onColumnGroupDeleted: this.props.onColumnGroupDeleted,
   key: typeof c === 'string' ? c : c.key
}

return <GroupedColumnButton {...groupedColumnButtonProps} />;
```

packages/react-data-grid/src/RowGroup.js
```diff
@@ -95,7 +95,7 @@ const  DefaultRowGroupRenderer = (props) => {
   return (
     <div style={style} onKeyDown={onKeyDown} tabIndex={0}>
       <span className="row-expand-icon" style={{float: 'left', marginLeft: marginLeft, cursor: 'pointer'}} onClick={props.onRowExpandClick} >{props.isExpanded ? String.fromCharCode('9660') : String.fromCharCode('9658')}</span>
-      <strong>{props.columnGroupName} : {props.name}</strong>
```
diogofcunha: What does this change in the actual render? can you show?

packages/react-data-grid-addons/src/data/RowGrouper.js
d4nj0nes: yeah - if you render the grid before grouping which I want to - columns here are the grouped columns not all grid columns

packages/react-data-grid-addons/src/toolbars/GroupedColumnsPanel.js
d4nj0nes: got it 👍

packages/react-data-grid/src/RowGroup.js
d4nj0nes: so it would say
`sellerName: ITV`
I just want it to say the value of the group not the column name i.e. `ITV`
We could make that optinal - but to me it just looks messy

packages/react-data-grid-addons/src/toolbars/GroupedColumnsPanel.js
d4nj0nes: But the props change - c is either a string or an object. I'll work it out

diogofcunha: it's doing that check in the key assignment in my snippet

d4nj0nes: just needed to do it for the other props too name and key
Done.

packages/react-data-grid/src/RowGroup.js
diogofcunha: My main concern would be breaking this for people that are already using it, because they are expecting the previous and we will change it in our core and break their functionality

d4nj0nes: sure - then I'll look at making that optional

PR #907, opened 2017-08-21
As add column to on cell click

alanschapira:


packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -203,7 +203,7 @@ const ReactDataGrid = React.createClass({
     this.onSelect({rowIdx: cell.rowIdx, idx: cell.idx});
 
     if (this.props.onRowClick && typeof this.props.onRowClick === 'function') {
-      this.props.onRowClick(cell.rowIdx, this.props.rowGetter(cell.rowIdx));
+      this.props.onRowClick(cell.rowIdx, this.props.rowGetter(cell.rowIdx), this.getColumn(cell.idx));
```
diogofcunha: Add a test to assert that the correct column is injected to onRowClick

PR #925, opened 2017-09-10
Cell.renderCellContent: pass column and rowIdx to formatter

jo-so:
Passing the rowIdx to the formatter makes it possible to have a column
with the row index, without the need to include the index in the row
data. The column properties might be helpful, too.

### DIFF
--- a/packages/react-data-grid-examples/src/scripts/example05-custom-formatters.js
+++ b/packages/react-data-grid-examples/src/scripts/example05-custom-formatters.js
@@ -26,7 +26,8 @@ const Example = React.createClass({
       {
         key: 'id',
         name: 'ID',
-        width: 80
+        width: 80,
+        formatter: props => <span>{props.rowIdx + 1}</span>
       },
       {
         key: 'task',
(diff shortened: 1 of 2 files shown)


urbinopescada:
I've been researching this issue because I've several formatters and everyting was working in perfection, until I need to implement row reordering.

I found that there is a difference in props received just by passing the formatter diferently.
If "formatter =<myformater/>" it will receive different props than if "formatter= myformatter".

Why is this different? Couln't the code be just

    if (React.isValidElement(Formatter)) {
      props.dependentValues = this.getFormatterDependencies();
      CellContent = React.cloneElement(Formatter, props);
    } else if (isFunction(Formatter)) {
      // CellContent = (
      //  <Formatter value={this.props.value} dependentValues={this.getFormatterDependencies()}
      //    column={props.column} rowIdx={props.rowIdx} />);
      props.dependentValues = this.getFormatterDependencies();
      CellContent = (<Formatter {....props} />);
    } else {
      CellContent = <SimpleCellFormatter value={this.props.value} />;
    }

jo-so:
The code for `<Formatter ...` came in with 6434d53 from @malonecj. It looks like the line was inspired by the `<SimpleCellFormatter ...` below. Maybe it would be better to clean up all three cases and pass everytime the same arguments.

jamesajones:
if I pass in the rowmetadata as the whole row or just one column value what if the static propTyes{} block like. This assumesthat the formatter is a separate class

Issue #840, opened 2017-06-22
Values not showing up for right-hand column cells when the screen size is small. 


tylerbecks:
> **HAVE YOU ALREADY SEARCHED FOR SIMILAR ISSUES? PLEASE HELP US OUT AND DOUBLE-CHECK FIRST!**

> **ALSO, PLEASE DON'T BE THAT PERSON WHO DELETES THIS TEMPLATE. IT'S HERE FOR A REASON.**

> **THANKS!**

**WHICH VERSION OF REACT ARE YOU USING?**
```
Officially Supported:
[ ] v0.14.x
```

```
Community Supported:
[x ] v15.0.x
```

**WHICH BROWSER ARE YOU USING?**

```
Officially Supported:
[ ] IE 9 / IE 10 / IE 11
[ ] Edge
[x] Chrome
```

```
Should work:
[ ] Firefox
[ ] Safari
```

--------

**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

**Current behavior**
Values not showing up for right-hand columns when the screen size is small.  The smaller the screen, the less columns show values.  If I drag the size of the browser to be smaller or larger, the values all show up and never go away until a hard refresh.

**Expected/desired behavior**
For all the values to show initially.

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem.
Values in right-hand columns not showing up: <img width="696" alt="screen shot 2017-06-21 at 6 28 08 pm" src="https://user-images.githubusercontent.com/16826230/27413591-36c70cfa-56b1-11e7-8013-8fd9a63922e8.png">
HTML showing adjacent cells showing and not showing values: 
<img width="482" alt="screen shot 2017-06-21 at 6 23 30 pm" src="https://user-images.githubusercontent.com/16826230/27413604-4af35de6-56b1-11e7-8d60-baf599f54225.png">

**What is the expected behavior?**
Expected behavior (after I manually adjust the browser size): 
<img width="777" alt="screen shot 2017-06-21 at 6 28 13 pm" src="https://user-images.githubusercontent.com/16826230/27413627-61cbcc6a-56b1-11e7-985b-357dbde9c1ab.png">

**What is the motivation / use case for changing the behavior?**

[...]

PR #946, opened 2017-09-21
Added fix for #840

regentcid434:
## Description

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

#840 

**What is the new behavior?**
`getRenderedColumnCount` now accounts for the first visible column being partially offscreen and adjusts its calculation to only include to visible width of the column

```
[ ] Yes
[x] No
```

**Other information**:

### DIFF
--- a/packages/react-data-grid/src/ColumnUtils.js
+++ b/packages/react-data-grid/src/ColumnUtils.js
@@ -43,5 +43,20 @@ module.exports = {
       value = column[property];
     }
     return value;
+  },
+
+  getWidthOfColumns: function getSize(columns, startIdx, endIdx) {
+    let totalWidth = 0;
+    if (Array.isArray(columns)) {
+      for (let idx = startIdx; idx <= endIdx; idx++) {
+        totalWidth += columns[idx].width;
+      }
+      return totalWidth;
+    } else if (typeof Immutable !== 'undefined') {
+      for (let idx = startIdx; idx <= endIdx; idx++) {
+        totalWidth += columns.get(idx).width;
+      }
+      return totalWidth;
+    }
   }
 };
--- a/packages/react-data-grid/src/ViewportScrollMixin.js
+++ b/packages/react-data-grid/src/ViewportScrollMixin.js
@@ -62,13 +62,15 @@ module.exports = {
     };
   },
 
-  getRenderedColumnCount(displayStart, width) {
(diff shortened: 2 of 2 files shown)


packages/react-data-grid/src/ColumnUtils.js
```diff
@@ -43,5 +43,20 @@ module.exports = {
       value = column[property];
     }
     return value;
+  },
+
+  getWidthOfColumns: function getSize(columns, startIdx, endIdx) {
+    let totalWidth = 0;
+    if (Array.isArray(columns)) {
+      for (let idx = startIdx; idx <= endIdx; idx++) {
+        totalWidth += columns[idx].width;
```
jpdriver: I don't think we can rely on columns always having an explicit `width` here.

Can you add some unit tests -- where all columns have widths, only some have widths, and none have widths?

regentcid434: @jpdriver I've tested this locally (still need to make unit tests) and width always has a value. The library has a hard-coded minWidth of 80, so a column won't be smaller than that unless you specify one explicitly. Columns without explicit widths set act as if they were each `flex: 1`.

In the current implementation of `getRenderedColumnCount`, the column widths are already being accessed in the same way my new util function does. Each iteration, the column metrics for the given index are retrieved and the width is subtracted from `remainingWidth`.
```
getRenderedColumnCount(displayStart, width) {
    let remainingWidth = width && width > 0 ? width : this.props.columnMetrics.totalWidth;
    if (remainingWidth === 0) {
      remainingWidth = ReactDOM.findDOMNode(this).offsetWidth;
    }
    let columnIndex = displayStart;
    let columnCount = 0;
    while (remainingWidth > 0) {
      let column = ColumnUtils.getColumn(this.props.columnMetrics.columns, columnIndex);

      if (!column) {
        break;
      }

      columnCount++;
      columnIndex++;
      remainingWidth -= column.width;
    }
    return columnCount;
  }
```

kevyu:
can this pr be merged?

jrymer:
What is the status on this PR?

regentcid434:
I've just been using my branch since it was a production issue that needed to be addressed, and I've not gotten around to any unit tests. They said that my calculation uses a width that might not have a value. However, the current implementation uses that same value. I'm just changing the math around it.

PR #996, opened 2017-10-26
Fix double-click outside cell

willgates:
## Description
When double clicking in the viewport but not _on_ any row/cell (e.g. in the empty area when there are only a few rows), you can edit the previously selected. This PR changes this behaviour so nothing happens when double-clicking in an empty area of the grid.

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Select a cell. When double clicking in the viewport but not _on_ any row (e.g. in the empty area when there are only a few rows), you can edit the previously selected.

**What is the new behavior?**
You shouldn't be able to edit a previously selected cell when double-clicking in an empty area. Nothing should happen.

```
[ ] Yes
[x] No
```

**Other information**:

### DIFF
--- a/packages/react-data-grid/src/ReactDataGrid.js
+++ b/packages/react-data-grid/src/ReactDataGrid.js
@@ -219,10 +219,6 @@ const ReactDataGrid = createReactClass({
 
   onCellDoubleClick: function(cell: SelectedType) {
     this.onSelect({rowIdx: cell.rowIdx, idx: cell.idx});
-    this.setActive('DoubleClick');
-  },
-
-  onViewportDoubleClick: function() {
     this.setActive();
   },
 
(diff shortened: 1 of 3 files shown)


packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -964,7 +972,8 @@ const ReactDataGrid = createReactClass({
             onViewportKeyup={this.onKeyUp}
             onViewportDragStart={this.onDragStart}
             onViewportDragEnd={this.handleDragEnd}
-            onViewportDoubleClick={this.onViewportDoubleClick}
+            onViewportClick={this.deselect}
+            onViewportDoubleClick={this.deselect}
```
willgates: @malonecj Clicking outside a row or cell will now deselect the currently selected cell. Alternatively we could get rid of `onViewport(Double)Click` all together since we already have a `onCellDoubleClick`.

PR #1009, opened 2017-11-04
Navigate with enter

### LINKED ISSUE #942: Enter navigation

✅ Officially supported ✅
- [ X ] v15.4.x

- [ ] v15.5.x
- [ ] v15.6.x

- [ ] v16.x.x

---

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [ X ] Chrome

- [ X ] Firefox
- [ X ] Safari

---

### I'm submitting a ...

- [ ] 🐛 Bug Report
- [ X ] 💡 Feature Request

---

### Issue Details

```
Please include:
- What the current behavior is
    After a cell is expanded and edited, enter key is used to close the cell and trigger 'onRowUpdate'. Currently there is no way to control which cell is focused aside from tab navigation. 

- What the desired behavior is
    It would be nice be able to control which cell is focused or allow for the user to be automatically navigated to a specific cell after hitting 'enter' to finish editing the current cell.

- (If Feature) The motivation / use case for the feature
    I have a table where the user may need to enter in hundreds of weights for data collection; each possibly unique from the last. The weights all exist in the same column and as a result the down arrow must be used to navigate to the next row. As all inputs are entered via num pad, our clients would prefer to enter in a weigh, hit enter to be navigated to the next row, and begin typing.

Demonstration of use case:
![image](https://user-images.githubusercontent.com/5068500/30500391-e73184fe-9a12-11e7-8652-33eb19b90591.png)


The Best Issue Is A Pull Request™
```

jeffcarbs:
## Description
Resolves #942

In the most common spreadsheets (I'm on a Mac so I tested in Google Sheets and Numbers), pressing:
- `Tab` moves you right
- `Shift + Tab` moves you left
- `Enter` puts you into edit mode (if editable) then pressing again moves you down.
- `Shift + Enter` puts you into edit mode (if editable) then pressing again then moves you up.

This component already handled the `Tab`-based navigation. This PR adds the `Enter`-based navigation. 

**Google Sheets (pressing enter then shift+enter):**
nov-04-2017 00-43-07

**This PR (pressing enter then shift+enter on editable and non-editable cells):**
nov-04-2017 01-11-35

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Pressing `Enter` does not navigate.

**What is the new behavior?**
Pressing `Enter` will navigate down and pressing `Shift + Enter` will navigate up.

```
[ ] Yes
[X] No
```

**If this PR contains a breaking change, please describe the impact and migration path for existing applications:**
This change won't break anything but it is new default behavior. We could consider putting it behind a config prop like `navigateOnEnter` that defaults to false.

**Other information**:

### DIFF
--- a/packages/react-data-grid/src/ReactDataGrid.js
+++ b/packages/react-data-grid/src/ReactDataGrid.js
@@ -251,7 +251,11 @@ const ReactDataGrid = createReactClass({
   },
 
   onPressEnter(e: SyntheticKeyboardEvent) {
-    this.setActive(e.key);
+    const didSetActive = this.setActive(e.key);
+
+    if (!didSetActive) {
+      this.moveSelectedCell(e, e.shiftKey ? -1 : 1, 0);
+    }
   },
 
   onPressDelete(e: SyntheticKeyboardEvent) {
@@ -313,9 +317,6 @@ const ReactDataGrid = createReactClass({
   onCellCommit(commit: RowUpdateEvent) {
     let selected = Object.assign({}, this.state.selected);
     selected.active = false;
-    if (commit.key === 'Tab') {
-      selected.idx += 1;
-    }
     let expandedRows = this.state.expandedRows;
     // if(commit.changed && commit.changed.expandedHeight){
     //   expandedRows = this.expandRow(commit.rowIdx, commit.changed.expandedHeight);
@@ -808,7 +809,7 @@ const ReactDataGrid = createReactClass({
     this.setState({selected});
   },
 
-  setActive(keyPressed: string) {
+  setActive(keyPressed: string): boolean {
     let rowIdx = this.state.selected.rowIdx;
     let row = this.props.rowGetter(rowIdx);
 
@@ -829,8 +830,11 @@ const ReactDataGrid = createReactClass({
           this.setState({selected}, () => { this.scrollToColumn(idx); });
         }
         this.handleCancelCopy();
+        return true;
       }
     }
+
+    return false;
   },
 
   setInactive() {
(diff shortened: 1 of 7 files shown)


packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -251,7 +251,11 @@ const ReactDataGrid = createReactClass({
   },
 
   onPressEnter(e: SyntheticKeyboardEvent) {
-    this.setActive(e.key);
+    const didSetActive = this.setActive(e.key);
+
+    if (!didSetActive) {
```
jeffcarbs: If we didn't set active it means the cell was either currently active or the cell is not editable. In both cases we want to navigate.

packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -313,9 +317,6 @@ const ReactDataGrid = createReactClass({
   onCellCommit(commit: RowUpdateEvent) {
     let selected = Object.assign({}, this.state.selected);
     selected.active = false;
-    if (commit.key === 'Tab') {
-      selected.idx += 1;
```
jeffcarbs: When `onCellCommit` it triggered by `Tab` (or `Enter` for that matter) the event is picked up by `onPressTab` (or `onPressEnter`) which already handles the navigation. They both are trying to move the `selected` state but since the `onPress` handler is called last it's state change "wins".

By only having the movement from a given key handled in one spot it makes it a little easier to follow. Also, this code wasn't respecting the boundaries of the grid so if you pressed `Tab` while editing the last cell in the row it would move the index to a non-existent column.

PR #954, opened 2017-09-27
add PropTypes as a separate package

suxxes:
Fixes #744

## Description
Because @GlebDolzhikov did not respond since May, I decided to fork, fix, and pull request his original changes for PropType validations.

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x")
```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

**What is the current behavior?** (You can also link to an open issue here)

With recent React 16 release application does not work with it, because original PropTypes were removed from the core library into a separate.

**What is the new behavior?**

Application uses separate external prop-types library to validate incoming props.

**Does this PR introduce a breaking change?** (check one with "x")
```
[ ] Yes
[x] No
```

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:

Issue #744, opened 2017-04-14
Getting warnings with react 15.5+


jrgleason:
**WHICH VERSION OF REACT ARE YOU USING?**
```
Officially Supported:
[ ] v0.14.x
```

```
Community Supported:
[X] v15.0.x
```

**WHICH BROWSER ARE YOU USING?**

```
Officially Supported:
[X] IE 9 / IE 10 / IE 11
[X] Edge
[X] Chrome
```

```
Should work:
[X] Firefox
[X] Safari
```

--------

**I'm submitting a ...**  (check one with "x")
```
[X] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

**Current behavior**

When you try to run with React 15.5.X you get the following warnings...

```
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
warning.js:36 Warning: Draggable: React.createClass is deprecated and will be removed in version 16. Use plain JavaScript classes instead. If you're not yet ready to migrate, create-react-class is available on npm as a drop-in replacement.
```

**Expected/desired behavior**

No warnings

**Reproduction of the problem**

Just use the right version of React

**What is the motivation / use case for changing the behavior?**

React upgrage

Both stack traces show it coming from RDG, I double checked and didn't see any open issues.

dzwiedziu-nkg:
Confirm, I get following warings on React v15.5.4:

`Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.`

and

`Warning: Draggable: React.createClass is deprecated and will be removed in version 16. Use plain JavaScript classes instead. If you're not yet ready to migrate, create-react-class is available on npm as a drop-in replacement.`

dmk255:
I am also receiving the above errors.

I am also having issues where the rowGetter method is being called with a different state than I would otherwise expect. See code below:
```
const ReactDataGrid = require('react-data-grid');
import React, { Component } from 'react';
import { connect } from 'react-redux';

@connect(
  state => ({
    rows: state.rows
  })
)
export default class Test extends Component {

  constructor(props) {
    super(props);

    this._columns = [
      {
        key: 'id',
        name: 'ID',
        width: 80
      },
      {
        key: 'task',
        name: 'Title',
        editable: true
      },
      {
        key: 'priority',
        name: 'Priority',
        editable: true
      },
      {
[...]

PR #1027, opened 2017-11-20
Convert `createReactClass` components to functional or class components

### LINKED ISSUE #978: Getting getDefaultProps warnings with react 15.6+

✅ Officially supported ✅
- [ ] v15.4.x

- [ ] v15.5.x
- [x] v15.6.x

- [ ] v16.x.x

---

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [x] 🐛 Bug Report
- [ ] 💡 Feature Request

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---

### Issue Details
current behavior:
```
Warning: getDefaultProps is only used on classic React.createClass definitions. Use a static property named `defaultProps` instead.
```
image

desired behavior: no warnings

motivation: less noise in console warnings and compatibility to upgrade React

amanmahajan7:
## Description
Convert `createReactClass` components to functional or class components

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`React 15.5` deprecated `React.createClass` and `React data grid` switched to `create-react-class` package as recommended by the Facebook team.

**What is the new behavior?**
Components that can be converted to `class` or `functional` components are migrated. Remaning components that are still using `mixins` will be migrated in a separate PR

```
[ ] Yes
[x] No
```

**Other information**:


jpdriver:
@amanmahajan7 will this be tested on React 15.5+ and React 16 too, or only React 15.4?

amanmahajan7:
@jpdriver I was planning on testing with React 15.5+ but if we decide to keep `createReactClass` then I will test with `React 16` also. Apart from the already fixed files, there are additional 9 files that use `mixins` and cannot be converted easily
I think there are two ways to fix these warnings temporarily
1. Import the `dev` version of `createReactClass`. `const createReactClass = require('create-react-class/create-react-class');` Bundling the `dev` version would increase the bundle size but this can be ignored for now
As we are bundling `create-react-class` package with `react-data-grid`, webpack strips out the code that sets the`isReactClassApproved` [flag](https://github.com/facebook/react/blob/v15.5.4/addons/create-react-class/create-react-class.js#L713) and as `React` does not find this flag, it logs `development` time warnings (false positive). This warning is logged every time a new `Element` is created or cloned so we are seeing a lot of these warnings. 
Both the `dev` and `prod` versions of `react-data-grid` are built with the `--release` [flag](https://github.com/adazzle/react-data-grid/blob/master/package.json#L11), I think this flag should only be set for `prod` build,

2. Revert back to `React.createClass`. This will log a single dev time deprecation warning for `createClass`. For `React 16` this [patch](https://github.com/adazzle/react-data-grid/issues/744#issuecomment-333890937) will need to used. 

@jpdriver @malonecj what are you thoughts? Any other ideas? Also for the next major version we should set `prop-types` and `create-react-class` (if still used) as _peer_ dependencies

jpdriver:
Remember though -- this issue is directly caused by React 16. Facebook put deprecation warnings in the release before React 16, and then deliberately made this throw in React 16. Whatever the ultimate fix is, I think it needs to be tested on 15.4, 15.5, and 16. We need to ensure the latter two don't log warnings and errors respectively.

I maintain that the cleanest, quickest, and most likely best solution is to revert the original PRs. Trying to screw about with different dev and prod versions / bundling an entire Facebook library alongside our own source / deliberately overriding React internal flags, etc. are all just messy solutions.

We weren't ready to take the PRs in the first place. I could see the script errors on the branch, so I'm not sure how that was missed or how it was checked internally before merging. But even so, the underlying grid architecture fundamentally isn't compatible right now, and we should've been aware of that. What we have dates from the React 11 / Prometheus Research days.

I strongly suggest we ship that patch as an invisible wrapper for the Grid, and then it wouldn't matter which version of React anyone was on, nobody would see warnings, and we'd have enough time to properly sort our internal architecture out for full compatibility.

I'd be happy to set that up in a branch if you guys want to play around with the result and confirm it, but I'll only bother doing that if you're open to considering it as the solution.

Also for what it's worth, at one point I separately started working on removing / converting the Mixins. Some of them are straight forward because they're only consumed by single components (the `ScrollMixins` are an example of that) but the ones we need to properly migrate to higher order components are mostly the `DOM` / `Column` ones.

amanmahajan7:
@jpdriver, I completely agree with you that we should not be bundling Facebook libraries (`prop-types` and `create-react-class`) and this is the root cause of the problem. IMO removing these libraries from the bundle and specifying them as _peer_ dependencies would solve all the issues (except the `prop-types` warning you mentioned) but this change would probably require a major bump. I wasn't sure how to handle the versioning in this case.

It seems like we have two options
1) Like you mentioned, revert the original PRs and provide a wrapper. IMO this wrapper is also a bit hacky and to use this wrapper one must install `prop-types` and `create-react-class` packages.

```js
// temporaryHackForReactDataGrid.js: import this file before react-data-grid

require('react').PropTypes = require('prop-types');
require('react').createClass = require('create-react-class');
```
2) Keep the original PRs and specify `prop-types` and `create-react-class` as externals so webpack does not bundle them with `react-data-grid`. This also requires installing `prop-types` and `create-react-class` packages so in that regard it is similar to the first approach. `create-react-class`  was specifically designed for the projects that are not ready to migrate to `React 16` which `react-data-grid` is due to `mixins`. We this approach we can incrementally work on removing 
- Mixins
- [PropTypes warning](https://github.com/adazzle/react-data-grid/pull/954#issuecomment-332809999) 

I am personally inclined towards the 2nd option as a lot or work was put in to migrate `React.PropTypes` and `React.createClass` but I am fine with the first option as well. I would like to hear @malonecj  take on this issue.

jpdriver:
@amanmahajan7 @malonecj ok -- how about we validate option 2 first, and confirm that there are no issues with that on React 15.4 / 15.5 / 16?

if there are, we'll fall back to option 1.

jpdriver:
to be honest I don't think it would, but I'm happy to be proven wrong and right now we don't know for certain.

so the next move should be do a spike, investigate / validate, and if I'm wrong -- great! -- and if not then we fall back to option 1.

amanmahajan7:
The build process bundles `createReactClass` package with `process.env.NODE_ENV` flag set to `production`. This strips out any `development` time code from the `createReactClass` package which includes the [block that sets the `isReactClassApproved` flag](https://github.com/facebook/react/blob/v15.5.4/addons/create-react-class/create-react-class.js#L719). 

The `development` version of `React` [logs a warning if it does not find this flag](https://github.com/facebook/react/blob/7d27851bf4aa8129276614b62edf9ade4aaa4cbd/packages/react/src/ReactElementValidator.js#L233). 

If `create-react-class` is specified as a _peer_ dependency then it becomes the responsibility of individual projects to include the correct files
- `development` version of `React` and `create-react-class` minified with webpack or any other build tool
or
- `production` version of `React` and `create-react-class` 

Again this theory must be tested and it case it fixes the warnings should the major version be bumped?

Swagetha:
Any news on this?

amanmahajan7:
I published a [new repository](https://github.com/amanmahajan7/react-data-grid/pull/1) with `create-react-class` and `prop-types` specified as `peer` dependencies and there are no more `getDefaultProps` warnings

Original (create-react-grid)
https://codesandbox.io/s/3z3lq9zn1

Fork (react-data-grid-aman)
https://codesandbox.io/s/409mmpynp4 react 15.6.2
https://codesandbox.io/s/k0j07n81nv react 16.2.0

@jpdriver @malonecj

achillesrasquinha:
HURRAY!

PR #1038, opened 2017-11-28
Add onBeforeEdit event

azhangstrata:
## Description
Add onBeforeEdit event

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

**What is the new behavior?**
It provide callback when trying to set active editor

```
[ ] Yes
[x ] No
```


packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -830,6 +831,7 @@ const ReactDataGrid = createReactClass({
         } else {
           this.setState({selected}, () => { this.scrollToColumn(idx); });
         }
+        this.props.onBeforeEdit();
```
malonecj: This will throw an error if not passed in as a props. Either add it as a default prop or check that it exists first

Issue #1016, opened 2017-11-10
cellClick can not be forced from react data grid component


Aldizh:
- Have you already searched for similar issues? Please help us out and double-check first!
_Yes there was one issue that came close to this which suggested that we have our own implementation of onRowClick handler. But this is more specific to cellClick event_

- Also, please don't be that person who deletes this template. It's here for a reason.

- Thanks!

---

### Which version of React JS are you using?

✅ Officially supported ✅
- [x] v15.4.2

⚠️ Not officially supported, expect warnings ⚠️
- [ ] v15.5.x
- [ ] v15.6.x

☣️ Not officially supported, expect warnings and errors ☣️
- [ ] v16.x.x

---

### Which browser are you using?

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

⚠️ Not officially supported, but "should work" ⚠️
- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [ ] 🐛 Bug Report
- [x] 💡 Feature Request

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---

### Issue Details
First i would like the thank you guys for the great work you have done.
I was wondering if there is a way to force a click event on the cell via focus/cell select instead of a mouse click. This could be very helpful for input elements such as date picker or dropdown list.
```
Please include:
- What the current behavior is
- What the desired behvaior is
- (If Bug) Steps to reproduce the issue
- (If Feature) The motivation / use case for the feature

We especially love screenshots / videos of problems, and remember
The Best Issue Is A Pull Request™
```
![dropdown](https://user-images.githubusercontent.com/2282120/32672254-fbccb55e-c60f-11e7-94ae-215682c43aa4.png)
<img width="434" alt="focus" src="https://user-images.githubusercontent.com/2282120/32672496-e4361664-c610-11e7-8eed-43d41661aa1d.png">

Currently when you focus on a cell that has a datepicker as an input element nothing happens until you actually click on it.
Would be nice if we can actually trigger the datepicker/dialog expansion on focus.
This could be very practical if a user wants to tab through the whole table and use keyboard only to enter data as quickly and easily as possible.
I have tried two approaches so far but no success.
[...]

PR #1053, opened 2017-12-14
Added autofocus feature by utilizing and exposing enableAutoFocus flag

Aldizh:
- Have you already searched for similar issues? Please help us out and double-check first!

1. Yes, raised [autofocus issue ](https://github.com/adazzle/react-data-grid/issues/1016) since i didn't find a similar issue or feature request

- Also, please don't be that person who deletes this template. It's here for a reason.

- Thanks!

---

✅ Officially supported ✅
- [x] v15.4.2

- [ ] v15.5.x
- [ ] v15.6.x

- [ ] v16.x.x

---

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [ ] 🐛 Bug Report
- [x] 💡 Feature Request

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---

### Issue Details
First i would like the thank you guys for the great work you have done.
I was wondering if there is a way to enable auto focus event on a cell. This could be very helpful for input elements such as date picker or dropdown list. It will enable the user to tab through the whole grid without having to use the mouse.
```
Please include:
- What the current behavior is
- What the desired behvaior is
- (If Bug) Steps to reproduce the issue
- (If Feature) The motivation / use case for the feature

We especially love screenshots / videos of problems, and remember
The Best Issue Is A Pull Request™
```
dropdown
focus

Currently when you focus on a cell that has a dropdown/date-picker as an input element nothing happens until you actually click on it.
Would be nice if we can actually trigger the dialog expansion on cell focus.
This could be very practical if a user wants to tab through the whole table and use keyboard only to enter data as quickly and easily as possible.
I have tried two approaches so far but no success.
1. Use enableCellSelect flag and onCellSelected but the problem here is that we can only access the cell coordinates not the event.

        handleCellFocus = (args) => {
             this.refs.reactDataGrid.openCellEditor(args.rowIdx, args.idx)
             // Here i need a way to force the cellClick event
         }
        

        <ReactDataGrid
              ref={'reactDataGrid'}
              enableCellSelect
              onCellSelected={(args) => { this.handleCellFocus(args) }}
              ...
        

2. Use the formatter itself and go up the DOM tree until we hit the 'react-Cell-Grid' div and force a click event there.
      `this.datePicker.refs.input.getInputNode().parentNode.parentNode.parentNode.click()`

So instead we could modify the cell element to utilize enableCellAutofocus flag to control the behavior of the focus event.


Aldizh:
Could you guys provide some insight to this ci error:
`An error was thrown in afterAll
  Unable to get property 'getBoundingClientRect' of undefined or null reference thrown`

When i run npm test in local i get something similar:
`TypeError: this.menu is null thrown`

Seems to be caused by this method:
`getWidthFromMouseEvent = (e: SyntheticMouseEvent): number => {
    let right = e.pageX || (e.touches && e.touches[0] && e.touches[0].pageX) || (e.changedTouches && e.changedTouches[e.changedTouches.length - 1].pageX);
    let left = ReactDOM.findDOMNode(this).getBoundingClientRect().left;
    return right - left;
  };
`
but would like to get your input as modifying this will break other tests.

malonecj:
@Aldizh - that error should be fixed now, and was a result of the third party context menu we were using. The test was quite brittle to be honest

Do you think you could catch your branch up and resolve conflicts

Aldizh:
Thank you for fixing that.

I resolved the  conflicts and ci test suite are passing now.
As far as the feature is concerned this is an optional enhancement for smooth keyboard navigation (with cells that have a popup like a datepicker or an autoselect). It is turned off by default abs can be turned on by passing in enableCellAutofocus flag as true to react-data-grid
I have documented this and provided an example as well.

Let me know if you have any questions/concerns.

PR #1081, opened 2018-01-15
Context menu upgrade from v1.6.2 to v2.9.2

qili26:
## Description

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

react data grid + react v16 + current react context menu v1.6.2 doesn't display context menu. 

**What is the new behavior?**
the context menu will be displayed.

```
[x] Yes
[ ] No
```

If this PR contains a breaking change, please describe the impact and migration path for existing applications: 

Some css has to be changed due to react context menu upgrade (breaking changes were introduced from react-contextmenu)

Original `ContextMenuLayer` HoC is converted to `ContextMenuTrigger` as a component. Hence the export (wrapper of context menu in RDG) has to be updated as well.

**Other information**:


packages/react-data-grid/src/RowsContainer.js
```diff
@@ -1,35 +1,33 @@
...
   constructor(props) {
     super(props);
     this.plugins = props.window ? props.window.ReactDataGridPlugins : window.ReactDataGridPlugins;
     this.hasContextMenu = this.hasContextMenu.bind(this);
     this.renderRowsWithContextMenu = this.renderRowsWithContextMenu.bind(this);
-    this.getContextMenuContainer = this.getContextMenuContainer.bind(this);
-    this.state = {ContextMenuContainer: this.getContextMenuContainer(props)};
+    this.validateContextMenu = this.validateContextMenu.bind(this);
+
+    this.validateContextMenu();
```
amanmahajan7: I wonder if `constructor` is the right place for validation. Did not find any recommendation on where to throw the errors but I think `render` method would probably be a better place. It is recommended to avoid any side effects in constructor 
https://reactjs.org/docs/react-component.html#constructor

```
render() {
    if (this.hasContextMenu())  {
        if (!this.plugins) { throw new Error('...') }
        return this.renderRowsWithContextMenu();
    }
    return <SimpleRowsContainer {...this.props} />;
}
```

@malonecj @diogofcunha thoughts?

packages/react-data-grid/src/RowsContainer.js
```diff
@@ -1,35 +1,33 @@
...
 };
 
+export const getNewContextMenuProps = ({ contextMenuId, rowIdx, idx }) => ({
+  rowIdx, idx, id: contextMenuId || DEFAULT_CONTEXT_MENU_ID
+});
+
 class RowsContainer extends React.Component {
   constructor(props) {
     super(props);
     this.plugins = props.window ? props.window.ReactDataGridPlugins : window.ReactDataGridPlugins;
```
amanmahajan7: This is an old code but is there a use case of `props.window`? I wonder if anyone is actually using it, I do not think it can even be populated. This can be cleaned up (probably in a separate PR)

https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid/src/Canvas.js#L284

packages/react-data-grid/src/__tests__/RowsContainer.spec.js
```diff
@@ -1,17 +1,67 @@
...
+  describe('with context menu', () => {
     it('should create a new RowsContainer instance', () => {
-      expect(componentWithoutContextMenu).toBeDefined();
+      const wrapper = shallow(<RowsContainer {...props} />);
+      expect(wrapper.find(FakeContextMenuTrigger).length).toBe(1);
+    });
+
+    it('should throw exception for no context menu plugin', () => {
+      const newProp = Object.assign({}, props, { window: {}});
+      try {
```
amanmahajan7: `toThrow/toThrowError` can be used instead
```
expect(shallow(<RowsContainer {...newProp} />)).toThrow()
```

packages/react-data-grid/src/RowsContainer.js
diogofcunha: @amanmahajan7 can it be server side rendering?
I think this was added back in the days due to IE8 support problems, having the context menu added directly would break IE8 apart

packages/react-data-grid/src/RowsContainer.js
diogofcunha: @amanmahajan7 we shouldn't add any side effects in the constructor even when not using react, creating a new object should never affect other parts of an application, you are correct, in react the constructor is used only to set the initial state

qili26:
@amanmahajan7 
Updated, could you take another look?

packages/react-data-grid/src/RowsContainer.js
```diff
@@ -1,52 +1,61 @@
...
+          <SimpleRowsContainer {...this.props} />
+        </ContextMenuTrigger>
+        {contextMenu}
+      </div>
+    );
   }
 
   render() {
-    return this.hasContextMenu() ? this.renderRowsWithContextMenu() : <SimpleRowsContainer {...this.props} />;
+    return this.getRowsContainer();
```
amanmahajan7: `getRowsContainer` method is probably not required and the `render` method can just return the container 

```
render () {
    if (this.hasContextMenu()) {
      this.validatePlugin();
      return this.renderRowsWithContextMenu();
    }

    return <SimpleRowsContainer {...this.props} />;
}
```

packages/react-data-grid/src/RowsContainer.js
amanmahajan7: I am not sure but that does seem like a workaround for SSR. We can revisit this code once RDG officially supports server side rendering.

amanmahajan7:
This upgrade is also required to support `React 16` as `react-contextmenu 1.6.x` is using the deprecated APIs
https://github.com/vkbansal/react-contextmenu/releases/tag/v2.5.0

hutber:
Should we fix the conflicts on this and get it merged? Makes sense to merge it

amanmahajan7:
@hutber `ContextMenu 2.0` has a [breaking change](https://github.com/vkbansal/react-contextmenu/releases/tag/v2.0.0-alpha.1) which is preventing us from merging this PR.

sontek:
What is the breaking change?  Can this PR address it?

qili26:
@sontek 
see PR description.
or you can reference this link https://github.com/adazzle/react-data-grid/pull/1081/files#diff-84ed4e0dd4f5ae1391f59e46c9c02175R11

sontek:
@qili26 Ahh, is that all there is?   I understand that upgrade is backwards incompatible but I think a version bump would suffice.  Or are you taking extra steps to make it compatible?

qili26:
@sontek 
I think it would be great to update this with the react v16 update together, and that's what's this PR for.
https://github.com/adazzle/react-data-grid/pull/1116

PR #1096, opened 2018-01-26
Proof of Concept: Improve cell rendering with RxJs

malonecj:
- [x] Benchmark test to navigate downwards for twenty rows
 - [x] Connect Higher Order Component using RxJS
 - [x] Cell Connect HOC listens to selected cell and only renders the cells within one cell of the selected cell


packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -1,1114 +1,14 @@
-const React                 = require('react');
```
bakesteve: if we go ahead with this approach, when rebasing this, should use a git move  here to preserve the history

packages/react-data-grid/src/Cell.js
```diff
@@ -139,6 +141,22 @@ class Cell extends React.Component {
     }
   };
 
+  onKeyDown = (e) => {
+      // we need to prevent default as we control grid scroll
+    // otherwise it moves every time you left/right which is janky
+    e.preventDefault();
```
bakesteve: should only prevent default if we are handling, otherwise we could prevent other expected beahviour to stop working, though in most cases, taht will still be ok through propogation (thinking tab / shift tab, etc)

PR #1109, opened 2018-02-02
Improve Cell navigation Performance (using rxjs proof of concept)

malonecj:
## Description

## Results
image

## Tasks
 - [x] SelectionMask component
 - [x] Wrap cell selector in React Motion wrapper for smooth cell animation
 - [ ] Scrolling with keyboard - now that selected cells do not automatically focus, the grid no longer scrolls with keyboard (rudimentary proof of concept for scroll down working)

## Further Work
 - [ ] Cell editing - this should be trivial, and is close to working
 - [ ] Frozen columns

 
This pr was based off [a previous pull request](https://github.com/adazzle/react-data-grid/pull/1096), that uses a state container with rxjs to both publish actions, and subscribe to them using a Connect function, similar to how Redux works. Whereas this proof of concept could easily be done, without this wrapper, it made it very simple to add the connected SelectionMask component, as all its necessary props were already in shared state. 

 

## Use of rxjs
 Pros 
- Easily subscribe to state change from any component
- Greater control of what and when we render
- Allows all components to be stateless and simple, with complexity hidden away in state management folder
- No more need for refs - having connected components mean that we will not need to query down the component tree to access props and state of child components

## Cons
 - More overhead - we can import just the parts of the library that we need but is it overkill?
 - Not trivial to understand - however, if we keep this part to a core section or even extract out to another package with a very simple api, that way any component can be connected
Any others?


packages/react-data-grid/src/Cell.js
```diff
@@ -71,43 +71,21 @@ class Cell extends React.Component {
...
-      || this.props.isCellValueChanging(this.props.value, nextProps.value)
-      || this.props.forceUpdate === true
-      || this.props.className !== nextProps.className
-      || this.props.expandableOptions !== nextProps.expandableOptions
-      || this.hasChangedDependentValues(nextProps)
-      || this.props.column.locked !== nextProps.column.locked;
-    return shouldUpdate;
-  }
-
   onCellClick = (e) => {
```
malonecj: using rxjs should allow us to reduce the complexity of our shouldComponentUpdate functions

packages/react-data-grid/src/Cell.js
```diff
@@ -21,9 +21,9 @@ class Cell extends React.Component {
   static propTypes = {
     rowIdx: PropTypes.number.isRequired,
     idx: PropTypes.number.isRequired,
-    selected: PropTypes.shape({
-      idx: PropTypes.number.isRequired
-    }),
+    isSelected: PropTypes.bool,
```
malonecj: code no longer necessary now that using SelectionMask

packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -150,13 +159,7 @@ class ReactDataGrid extends React.Component {
...
     super(props, context);
     let columnMetrics = this.createColumnMetrics();
-    let initialState = {columnMetrics, selectedRows: [], copied: null, expandedRows: [], canFilter: false, columnFilters: {}, sortDirection: null, sortColumn: null, dragged: null, scrollOffset: 0, lastRowIdxUiSelected: -1};
-    if (props.enableCellSelect) {
-      initialState.selected = {rowIdx: 0, idx: 0};
-    } else {
-      initialState.selected = {rowIdx: -1, idx: -1};
-    }
-    this.state = initialState;
+    this.state = {columnMetrics, selectedRows: [], copied: null, expandedRows: [], canFilter: false, columnFilters: {}, sortDirection: null, sortColumn: null, dragged: null, scrollOffset: 0, lastRowIdxUiSelected: -1};
```
malonecj: Have removed state.selected - this now lives in single store

packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -331,26 +329,14 @@ class ReactDataGrid extends React.Component {
...
-            this.props.onCellSelected(selected);
-          }
-        });
-      } else if (rowIdx === -1 && idx === -1) {
-        // When it's outside of the grid, set rowIdx anyway
-        this.setState({selected: { idx, rowIdx }});
-      }
+  onSelect = (selected) => {
+    const {selectCell, onCellDeSelected, onCellSelected} = this.props;
+    selectCell(selected);
```
malonecj: no more setting state on each cell selection, this now triggers an action, and any component can subscribe to state change if necessary

packages/react-data-grid/src/connectedComponents/SelectionMask.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { Motion, spring } from 'react-motion';
+import { connect } from '../stateManagement/state/RxState';
+import initialState from '../stateManagement/state/initialState';
+
+/* MapStateToProps */
+
+const Mask = ({ selectedPosition }) => {
+  const { idx, rowIdx } = selectedPosition;
+  const [width, height] = [150, 35];
```
malonecj: assuming fixed cell dimensions to calculate position. Need to figure out the optimal way to obtain this info

hutber:
Would love to be able to test this out as well :)

PR #1114, opened 2018-02-06
Improve Cell navigation Performance (using provider pattern with pub/sub)

malonecj:
## Description

## Results
image

## Tasks
 - [x] SelectionMask component
 - [x] Wrap cell selector in React Motion wrapper for smooth cell animation
 - [ ] Scrolling with keyboard - now that selected cells do not automatically focus, the grid no longer scrolls with keyboard (rudimentary proof of concept for scroll down working)

## Further Work
 - [ ] Cell editing - this should be trivial, and is close to working
 - [ ] Frozen columns


hutber:
Gunna sound like a stupid question... but how can I test this out, just out of curiosity, not that I will contribute to it :) but `npm link` inside of the `packages/` folders will give me errors when npm linking back inside my project as I normally would to test a external package.

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -0,0 +1,94 @@
...
+
+  componentDidMount() {
+    const {cellEvents} = this.props;
+    cellEvents.subscribe(CellEventTypes.onClick, (cell) => {
+      this.selectCell(cell);
+      this.node.focus();
+    });
+  }
+
+  onKeyDown = (e) => {
```
diogofcunha: A decision to move events to a single source of true means that we need to have a single pattern to attach events, all events need to be registered by subscription

packages/react-data-grid/src/cellEvents/Provider.js
```diff
@@ -0,0 +1,24 @@
...
+export class CellEventsProvider extends React.Component {
+  static propTypes = {
+    children: PropTypes.isRequired
+  };
+  static childContextTypes = {
+    cellEvents: PropTypes.object.isRequired
+  };
+  constructor() {
+    super();
+    this.cellEvents = new EventDispatcher();
```
diogofcunha: The pattern looks good, but we probably want to be more generic for the actual implementation and avoid creating new components as much as possible, this should work a little bit like redux, where we initially register a set of modules that can handle events for certain entities, and then we subscribe to events using a generic wrapper around our components. It could work like `combineReducers` in the end

packages/react-data-grid/src/stateManagement/Connect.js
```diff
@@ -0,0 +1,40 @@
...
+const sharedState = {
+  selectedPosition: {
+    idx: -1,
+    rowIdx: -1
+  },
+  isEditorEnabled: false,
+  firstEditorKeyPress: null
+};
+
+function connect(mapStateToProps, mapDispatchToProps, subscribeToEvents = () => {}) {
```
malonecj: @diogofcunha was trying to make this more generic, and now it feels like I'm just making a basic version of redux

packages/react-data-grid/src/cellEvents/Provider.js
amanmahajan7: This makes me wonder if we should simply use `redux` instead of a custom solution. Another reason is that `context` is [going to be deprecated soon](https://dev.to/kentcdodds/reacts--new-context-api-dpi) and having a 3rd party library taking care of that would be nice. Although, it is not a big change to switch to the new context api.

diogofcunha: They made changes on the context api so that people don't always need to use third parties for simple jobs.

The main problem with redux will always be the number of things that will re-render when something simple changes, the only way we would be able to stop that from happening would be adding scu in the connected components and that is one of our main design problems atm.

Adding it also means a significant increase in our bundle size and pulling a big library when we only need a fraction of the functionality

malonecj:
@hutber I'm not sure how worthwhile that would be as quite a few features of the grid are broken with this proof of concept. I have not used it yet myself but we use Lerna to publish our packages to npm so you could find some documentation on their website for local publishing.
https://github.com/lerna/lerna/pull/263

packages/react-data-grid/src/cellEvents/Provider.js
amanmahajan7: I am not sure if this is entirely true unless I am missing something. `react-redux` only renders the wrapped component when the state returned by `mapStateToProps` changes so `mapStateToProps` in a way becomes the new `render` and as long this function is simple, there should not be any performance issues. [Here](https://gist.github.com/gaearon/1d19088790e70ac32ea636c025ba424e) is a simplified version of the connect function without any performance optimizations. This is very similar to our custom pub/sub approach and the only difference is we are using `setState` instead of `forceUpdate`. In terms of bundle size, minified `redux` is 5.7 KB and minified `react-redux` is 12.8 KB so it should not increase the bundle size by much (~19 KB in total). Here the [FAQ](https://redux.js.org/docs/faq/Performance.html#how-well-does-redux-scale-in-terms-of-performance-and-architecture) on performance

hutber:
Ye, thanks @malonecj I jumped on to other things and am waiting patiently for this to hopefully get finished.

PR #1123, opened 2018-02-12
Improve Cell navigation Performance

malonecj:
## Description

## Results
image

## Tasks
 - [x] SelectionMask component
 - [x] Cell editing 
     - [x] Move `enableCellSelect` check to interaction layer
     - [x]  Move `onCheckCellIsEditable` check to interaction layer
 - [x] Scrolling with keyboard - now that selected cells do not automatically focus, the grid no longer scrolls with keyboard (rudimentary proof of concept for scroll down working)
    - [x] Fix cell navigation mode
    - [x] Fix horizontal scrolling
 - [x] Frozen columns
 - [x] Migrate cell drag down to Interaction Layer
 - [x] Migrate copy/paste behavior to Interaction Layer
 - [x] Migrate React Data Grid component keyboard behavior to Interaction Layer
    - [x] Fix tab behavior
 - [x] Add test coverage for new components 
 - [x] Fix Existing Unit Tests
 - [x] Delete OverflowCell/Row components
 - [x] Ensure `refs` work with container components
 - [x] Ensure `CellRenderer` and `RowRenderer` works with container components
 - [x] Ensure connecting grid/cell to the store does not affect cell rendering performance
 - [x] State management
    - [x] Decide if `redux` should be used instead
    - [x] Remove singleton store/listeners
 - [x] Fix Cell Actions- Fixed using `pointer events`
 - [x] Fix Grid Events
 - [x] Fix Column Events
 - [x] Fix ContextMenu.
 - [x] Fix Filter Row focus
 - [ ] Decide zIndex strategies

## Bugs
 - [x] Cell double click editing does not work
 - [x] Clicking on a cell scrolls the canvas to the top
 - [x] Fix SelectedCell position in the grouping example
image
 - [x] Fix zIndex for filters 
image


packages/react-data-grid/src/masks/EventBus.js
```diff
@@ -0,0 +1,28 @@
+class EventBus {
```
amanmahajan7: `Cell` component uses event bus to interact with the `InteractionMasks` component. The idea is to remove the local state from the `ReactDataGrid` component so that the whole tree is not rendered on every keystroke. 

- Click `Cell`
- Fire `SelectCell` event
- `InteractionMasks` is subscribed to the event and handle the selection logic

packages/react-data-grid/src/RowGroup.js
```diff
@@ -67,7 +72,8 @@ RowGroup.propTypes = {
   isExpanded: PropTypes.bool.isRequired,
   treeDepth: PropTypes.number.isRequired,
   name: PropTypes.string.isRequired,
-  renderer: PropTypes.func
+  renderer: PropTypes.func,
+  eventBus: PropTypes.isRequired
```
diogofcunha: is this valid? what's the shape of event bus?

amanmahajan7: No, this should be an object or an object with `{ dispatch, subscribe }` methods but I am just using `object` for now

Issue #1133, opened 2018-03-04
Selecting rectangular ranges


rowanhill:
### Which version of React JS are you using?

None (yet)!

---

### Which browser are you using?

N/A

---

### I'm submitting a ...

- [ ] 🐛 Bug Report
- [x] 💡 Feature Request

---

#### Current behaviour
_As far as I can tell_ (please correct me and close this if I'm wrong), react-data-grid currently only supports selecting a single cell, or multiple entire rows (via `rowSelection` and associated checkboxes).

#### Desired behaviour
Clicking and dragging in the grid selects the rectangular range described by the drag (i.e. the cells receiving the mousedown and mouseup are the diagonal extremes of the selected area). The selected range can be programatically interacted with (e.g. as a property that can be read, or raised to an event handler).

#### Motivation
A couple of use cases include:
 - Allowing the user to select a group of cells which can then be programatically altered (e.g. select an area, then "bulk update" the cells, or some subset of the, in one go from an external control).
- On-the-fly reporting on the selected range (e.g. counts, max, min, average, or whatever application specific logic makes sense).

---

As above, apologies if this is already possible, but react-data-grid looks great, and this feature is the only thing (other than time!) stopping me from migrating to it.

I'm assuming this is a relatively involved change, if it's not already supported, but if anyone thinks it's simple, please shout (and ideally point me in the right direction).

Thanks!

rowanhill:
I've taken a quick look at how implementing this might go, and I'm currently thinking, it'd be something like:
- Add `selectedRange` to the state of `ReactDataGrid` (in addition to the existing `select`), of the form `{topLeft:{idx, rowIdx}, bottomRight:{idx, rowIdx}}`
- Add cell `mousedown` and `mouseenter` event handlers for tracking starting selecting and moving over cells, plus a `mouseup` handler on `window` for tracking selecting ending (in case you mouseup off the grid)
- Update `Row.shouldComponentUpdate` to return true if the row is included in the currently in-progress selection range, and similarly for `Cell`
- Add CSS classes to cells to draw the selection outline, instead of relying on `:focus`
- Prevent text in the grid from being selected when you drag (see below, I have some questions about this)
- Probably needs some tweaks to how deselecting works (see below, I also have questions here)
- Hide/disable the drag handle when the selection range is bigger than 1x1
- Support keyboard range selection via shift+arrow keys (involves modifying `ReactDataGrid.onKeyDown` and friends)
[...]

PR #1158, opened 2018-04-01
Rectangular selection range support

rowanhill:
## Description
As per my feature request and rough outline in #1133, this PR allows users to select rectangular ranges with both the mouse and keyboard (and to subscribe to events about such ranges, so that their application can react appropriately).

```
```

I don't (yet) understand what the approach to documentation is on this project. If someone could let me know what they'd expect to see, how docs are generated, etc, I'll by happy to add a commit to this PR.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Clicking & dragging selects the text in the cells / shift+arrows just moves the selected cell.

**What is the new behaviour?**
Clicking & dragging selects a rectangular range, with the selected/"cursor" cell following the mouse. The range is displayed with a thin border; the selected cell retains a (slightly thicker) border. When the range is greater than 1x1, the selected cells also have a semi-transparent overlay. Text in the grid is no longer selectable except when editing content (in modern browsers; this behaviour won't be supported by IE9 or lower).

The selected range can also be manipulated by holding shift and using the arrow keys. (An existing range is updated by keeping the same start point and moving the "cursor" cell)

Clicking in the window but not on a cell clears the selection.

New callbacks (onCellRangeSelectionStarted, onCellRangeSelectionUpdated, and onCellRangeSelectionCompleted) are called (if they're present) when starting a drag, moving whilst dragging, and ending a drag. Modifying the range via the keyboard calls only the latter two callbacks.

There were a few questions I asked in #1133. I've taken the below approaches:
- Text selection is prevented in browsers other than IE9 and below. For now, I'm just assuming that's okay, and that this project doesn't support IE9 or lower; I can't see anything official on what's supported, but shout if that's not true.
- onViewportDoubleClick remains. I don't understand how it's ever actually invoked by a user (given if you click a cell, it stops the event from propagating), but leaving it in doesn't seem like a problem for this PR.
- I've removed onViewportClick, but I believe the behaviour is preserved: if you click in the viewport but outside a cell, when you mouseup the window's event handler will see that but also see there's no selection drag in process, and so call deselect. As with onViewportDoubleClick, I don't actually understand how you trigger this as a user, but I've tried to preserve behaviour regardless.
- I haven't done anything about onCellDeSelect not being called in some situations when a cell is, in fact, deselected. I do think that's probably a bug, but I thought it was best left for a different PR.
- Clicking outside the grid calls deselect() now. Previously the cell just lost focus (so it looked deselected, because the CSS relied on :focus, but the selected state remained). Now the actual selected state is changed. I'm not sure this is an externally observable difference, though, because of the point above about onCellDeSelect not being fired when you might expect it to.
- This feature is always on - i.e. I haven't provided any setting to disable it. I considered providing one, or tying it to enableCellSelect, but as far as I can tell enableCellSelect doesn't actually prevent cell selection, but merely stops there being a cell selected by default when the grid is first rendered.

```
[ ] Yes
[x] No
```
Okay, so there are a few small differences noted above, but I don't believe they're "breaking." The part I'm most concerned by is preventing text selection, but I strongly suspect that's not behaviour that anyone would be relying on.

**Other information**

The first commit in this PR is essentially unrelated to this feature - it's tidying up and isolating various tests. It's a prerequisite of the next commit, though: without doing that tidy up, some state was leaking between tests (such that something at some point triggered a render of Canvas, which had fewer rows than specified, so ultimately generated an exception that killed the test runner).

### DIFF
--- a/packages/react-data-grid-addons/src/__tests__/Grid.spec.js
+++ b/packages/react-data-grid-addons/src/__tests__/Grid.spec.js
@@ -473,130 +473,120 @@ describe('Grid', function() {
   });
 
   describe('When selection enabled and using rowSelection props', function() {
-    beforeEach(function() {
-      let self = this;
-      this._selectedRows = [];
-      this._deselectedRows = [];
-      this.rows = [{id: '1', isSelected: true}, {id: '2', isSelected: false}, {id: '3', isSelected: false}, {id: '4', isSelected: false}];
+    let rows;
+    let selectRowCol;
+    let onRowsSelectedSpy;
+    let onRowsDeselectedSpy;
+
+    function createComponentWithSpies() {
+      onRowsSelectedSpy = jasmine.createSpy('onRowsSelected');
+      onRowsDeselectedSpy = jasmine.createSpy('onRowsDeselectedSpy');
+
       let columns = [{name: 'Id', key: 'id'}];
       let rowGetter = function(i) {
-        return self.rows[i];
+        return rows[i];
       };
-      this.component = this.createComponent({ rowsCount: this.rows.length, rowGetter: rowGetter, columns: columns, rowSelection: {enableShiftSelect: true, selectBy: {isSelectedKey: 'isSelected'},
-        onRowsSelected: function(selectedRows) {
-          self._selectedRows = selectedRows;
-        },
-        onRowsDeselected: function(deselectedRows) {
-          self._deselectedRows = deselectedRows;
+
+      const props = {
+        rowsCount: rows.length,
+        rowGetter: rowGetter,
+        columns: columns,
+        rowSelection: {
+          enableShiftSelect: true,
+          selectBy: {isSelectedKey: 'isSelected'},
+          onRowsSelected: onRowsSelectedSpy,
+          onRowsDeselected: onRowsDeselectedSpy
         }
-      }}).node;
-      this.baseGrid = this.getBaseGrid();
-      this.selectRowCol = this.baseGrid.props.columnMetrics.columns[0];
+      };
+
+      this.component = this.createComponent(props).node;
+      const baseGrid = this.getBaseGrid();
+      selectRowCol = baseGrid.props.columnMetrics.columns[0];
+    }
+
+    beforeEach(function() {
+      rows = [{id: '1', isSelected: true}, {id: '2', isSelected: false}, {id: '3', isSelected: false}, {id: '4', isSelected: false}];
+
+      createComponentWithSpies.apply(this);
     });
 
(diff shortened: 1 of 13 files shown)


spchuang:
would love to have this

rowanhill:
Me too.

It'd be great if the maintainers could give some kind of indication of when they might be able to take a look at this. I appreciate that might not be for a while, but it would be useful to set my expectations - if it's months away from being merged, I might want to change my plans.

amanmahajan7:
Hi @rowanhill, thank you for the PR and apologies for the late reply. We are currently working on moving the keyboard navigation logic into a separate component to improve performance. Here is the [PR](https://github.com/adazzle/react-data-grid/pull/1123), the range selection logic should also be moved to the new [`InteractionMasks`](https://github.com/adazzle/react-data-grid/blob/c8ed8cb413913cbb7dfb865d2a17c29acd2b315b/packages/react-data-grid/src/masks/InteractionMasks.js) component when it is ready. Someone from the team will review this PR and provide the next steps. This might take some time as some of our team member are on vacations.

rowanhill:
Oh, wow. I haven't read through properly, but I'm guessing that might be a pretty significant change.

Sounds like the plan will be to get #1123 merged before we take a look at this one - is that right? And if so, any timelines on when the former might make it in?

Also, looks like that's on top of next, as opposed to master - does that mean there's no chance of this getting through as a non-breaking change? And if not, any ideas when the next release might be?

Sorry for all the questions!

amanmahajan7:
No problem. It is big change as we are rewriting the internal of `react-data-grid`. This is mainly needed to improve the keyboard navigation performance and make future development easier. I do not expect this to be a breaking change other than removing some unused props. We have changed the release process and we no longer publish a new  version on every commit and instead we will do more planned releases in the future. Here is the road map for the next few months
- Cut the next major release to support `React-16` ASAP. `next` branch fully supports `react-16` and we just need to publish a new version
- Merge #1123 (probably in a couple of weeks)
- Cut a new minor release
- Handle long outstanding issues, we will compile a list soon
- We can start looking into your PR once #1123 is merged or even work in parallel

rowanhill:
@amanmahajan7 I see #1123 was just merged to `next` - nice!

What are the next steps for this PR? Is anyone on your team expecting to look into rebasing this, or is it something that'll have to wait until I can find time to do it?

Totally understood if your team won't have time for it, I just want to understand your process.

rowanhill:
I think this PR is now obsolete - see my new PR at #1210 which adds the same functionality, but on top of next (including the refactoring from #1123).

PR #1183, opened 2018-05-01
Fix editor refs

JamesPortelli:
Fixed an issue whereby incorrect binding of functions was causing editor commits to throw an undefined error.


packages/react-data-grid-addons/src/editors/ContainerEditorWrapper.js
```diff
@@ -7,16 +7,20 @@ import React, { Component } from 'react';
 // to the required methods.
 module.exports = (ContainerEditor) => {
   return class ContainerEditorWrapper extends Component {
-    getInputNode() {
+    getInputNode = () => {
```
amanmahajan7: We do not need this change, `getValue` and `getInputNode` methods are always used like `this.ref.getValue` so _this_ value will be correct.

Issue #920, opened 2017-09-05
Allow Other Backends for DraggableContainer (and fix dnd)


seurimas:
### Which version of React JS are you using?

✅ Officially supported ✅
- [x] v15.4.x

⚠️ Not officially supported, expect warnings ⚠️
- [ ] v15.5.x

☣️ Not officially supported, expect warnings and errors ☣️
- [ ] v16.x.x

---

### Which browser are you using?

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

⚠️ Not officially supported, but "should work" ⚠️
- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [ ] 🐛 Bug Report
- [x] 💡 Feature Request

---

### Issue Details

Currently, DraggableContainer [installs its own drag-and-drop backend](https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid-addons/src/draggable-header/DraggableContainer.js#L29). This both:
* Forces the use of HTML5 Backend, even though another backend may be desirable for many projects, although there exists [a separate issue](https://github.com/yahoo/react-dnd-touch-backend/issues/83) with using the Touch backend due to some differences in the backends. 
* (!) Instantiates multiple backends. Backends are suggested to be instantiated at the root of the entire project, not the component. This results in [some related issues](https://github.com/adazzle/react-data-grid/issues/863).

Both issues can be worked around by referencing the `DecoratedComponent` field of DraggableContainer currently.

Ideally, the DraggableContainer should simply not install the backend at all, and the docs should be updated to indicate that a backend should be installed by the user.

PR #1195, opened 2018-05-17
Fix Allow Other Backends for DraggableContainer (#920)

jdu3ds:
## Description

- [ x] The commit message follows our guidelines: https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
- [ x] Tests for the changes have been added (for bug fixes / features)

```
[x ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Presently, DraggableContainer always creates a React-dnd context when used, regardless of whether such a context has already been created by another component higher in the component tree. Indeed, the DraggableContainer class is wrapped in a React-dnd DragDropContext(HTML5Backend)(DraggableContainer) HOC before being exported. If another component in the same app also uses React-dnd, or if two ReactDataGrids are used in the same app, an error will occur at HTML5Backend initialization time ('Cannot have two HTML5 backends at the same time') causing the app the fail.

**What is the new behavior?**
The fix tests whether a React-dnd context is already there when rendering the DraggableContainer (to do that it checks if the 'dragDropManager' react context property). If not, it inserts a DragDropContextProvider. This way, existing code should not break, because of the DragDropContextProvider which replaces the previous DragDropContext(HTML5Backend). And code which mixes react-data-grid with another React-dnd-enabled component should work provided they both live under the umbrella of a DragDropContextProvider defined as a common ancestor of these components.

```
[ ] Yes
[ x] No
```

**Other information**:

### DIFF
--- a/packages/react-data-grid-addons/src/draggable-header/DraggableContainer.js
+++ b/packages/react-data-grid-addons/src/draggable-header/DraggableContainer.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import html5DragDropContext from '../shared/html5DragDropContext';
 import DraggableHeaderCell from './DraggableHeaderCell';
+import { DragDropContextProvider } from 'react-dnd';
+import HTML5Backend from 'react-dnd-html5-backend';
 
-class DraggableContainer extends React.Component {
+export default class DraggableContainer extends React.Component {
   renderGrid() {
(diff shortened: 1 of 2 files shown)


laaglu:
Any news as to when this pull request could be reviewed ?

Hybueno09:
Any updates now? I also have this problem...

PR #1209, opened 2018-05-26
Fix some bugs introduced by Interaction Mask component

amanmahajan7:
- [ ] Fix cell styles
- [ ] Unselect selected cell when clicked outside the grid
- [x] Fix [drag exception](http://mereskin.github.io/dnd/) in Internet explorer 
> Internet Explorer only supports 'text' and 'URL' for the 'type' argument. If you try setting type to anything else, IE will throw an exception.
- [ ] Fix performance regression in firefox.  When the grid is scrolled to the middle then the drag fill operation is extremely slow. This is caused by [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1443146) bug.
> The problem comes from the fact that the item's container is shifted right with a CSS transformation (translateX). The ghost is positioned like there was no translation.

### DIFF
--- a/packages/react-data-grid/src/Row.js
+++ b/packages/react-data-grid/src/Row.js
@@ -44,12 +44,20 @@ class Row extends React.Component {
   }
 
   handleDragEnter = (e) => {
-    e.dataTransfer.dropEffect = 'move';
+    // Prevent default to allow drop
+    e.preventDefault();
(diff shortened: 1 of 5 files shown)


amanmahajan7:
Replaced by 
https://github.com/adazzle/react-data-grid/pull/1228
https://github.com/adazzle/react-data-grid/pull/1227

PR #1210, opened 2018-05-26
Rectangular selection range support (on top of 'next')

rowanhill:
## Description
Now that #1123 is complete, I've taken another run at resolving my feature request in #1133 - this PR obsoletes #1158.

This PR adds the ability to select (rectangular) ranges of cells, rather than just a single cell at a time. Both cursor (click and drag) and keyboard (shift + arrow keys) are supported.

I've also added a new example to demonstrate responding to range selection events from a parent component of a ReactDataGrid.

```
```

I have added a new example page, but I'm not sure if that's sufficient - do I need to update some documentation (e.g. the markdowns in react-data-grid-examples/src/docs)? If so, how are they generated?

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Clicking & dragging selects the text in the cells / shift+arrows just moves the selected cell.

**What is the new behavior?**
(The below text is similar to, but different from, the text of my old PR.)

Clicking & dragging selects a rectangular range, with the selected/"cursor" cell following the mouse. The range is displayed with a thin border; the selected cell retains a (slightly thicker) border. When the range is greater than 1x1, the selected cells also have a semi-transparent overlay. Text in the grid is no longer selectable except when editing content (in modern browsers; this behaviour won't be supported by IE9 or lower).

The selected range can also be manipulated by holding shift and using the arrow keys. (An existing range is updated by keeping the same start point and moving the "cursor" cell.)

New callbacks (onCellRangeSelectionStarted, onCellRangeSelectionUpdated, and onCellRangeSelectionCompleted) are called (if they're present) when starting a drag, moving whilst dragging, and ending a drag. Modifying the range via the keyboard calls only the latter two callbacks.

There were a few questions I asked in #1133. I've taken the below approaches:

- Text selection is prevented in browsers other than IE9 and below. For now, I'm just assuming that's okay, and that this project doesn't support IE9 or lower; I can't see anything official on what's supported, but shout if that's not true.
- onViewportDoubleClick doesn't exist (except for outdated docs and specs) on next; it did on master.
- onViewportClick has also been removed on next.
- Deselecting in general seems to behave slightly differently on next, so I think my original comments on onCellDeSelect aren't relevant anymore.
- This feature is always on - i.e. I haven't provided any setting to disable it. (I haven't looked into the behaviour of enableCellSelect on next.)

```
[ ] Yes
[x] No
```
As before, I think the closest thing to a breaking change is that text isn't selectable any more.

**Other information**:

My previous PR had a first commit that refactored some tests (because without that refactoring, I couldn't get them to run reliably). I haven't needed them this time around, but I did rebase them, so I can easily put up another PR to add them to next if anyone's interested.

### DIFF
--- a/packages/react-data-grid/src/masks/InteractionMasks.js
+++ b/packages/react-data-grid/src/masks/InteractionMasks.js
@@ -2,6 +2,7 @@ import React, { isValidElement, cloneElement } from 'react';
 import PropTypes from 'prop-types';
 
 import SelectionMask from './SelectionMask';
+import SelectionRangeMask from './SelectionRangeMask';
 import CopyMask from './CopyMask';
 import DragMask from './DragMask';
 import DragHandle from './DragHandle';
@@ -14,7 +15,8 @@ import {
   getSelectedColumn,
   getNextSelectedCellPosition,
   canExitGrid,
-  isSelectedCellEditable
+  isSelectedCellEditable,
+  selectedRangeIsSingleCell
 } from '../utils/SelectedCellUtils';
 import isFunction from '../utils/isFunction';
 import * as AppConstants from '../AppConstants';
@@ -54,6 +56,9 @@ class InteractionMasks extends React.Component {
     onCommitCancel: PropTypes.func,
     onCellSelected: PropTypes.func,
     onCellDeSelected: PropTypes.func,
+    onCellRangeSelectionStarted: PropTypes.func,
+    onCellRangeSelectionUpdated: PropTypes.func,
+    onCellRangeSelectionCompleted: PropTypes.func,
     onCellsDragged: PropTypes.func,
     onDragHandleDoubleClick: PropTypes.func.isRequired,
     onBeforeFocus: PropTypes.func.isRequired,
@@ -64,6 +69,14 @@ class InteractionMasks extends React.Component {
     selectedPosition: {
       idx: -1,
       rowIdx: -1
+    },
+    selectedRange: {
+      topLeft: {
+        idx: -1, rowIdx: -1
+      },
+      bottomRight: {
+        idx: -1, rowIdx: -1
+      }
     },
     copiedPosition: null,
     draggedPosition: null,
@@ -99,6 +112,9 @@ class InteractionMasks extends React.Component {
     const { eventBus, enableCellAutoFocus } = this.props;
 
     this.unsubscribeSelectCell = eventBus.subscribe(EventTypes.SELECT_CELL, this.selectCell);
+    this.unsubscribeSelectStart = eventBus.subscribe(EventTypes.SELECT_START, this.selectStart);
+    this.unsubscribeSelectUpdate = eventBus.subscribe(EventTypes.SELECT_UPDATE, this.selectUpdate);
+    this.unsubscribeSelectEnd = eventBus.subscribe(EventTypes.SELECT_END, this.selectEnd);
     this.unsubscribeDragEnter = eventBus.subscribe(EventTypes.DRAG_ENTER, this.handleDragEnter);
 
     if (enableCellAutoFocus && this.isFocusedOnBody()) {
@@ -108,6 +124,9 @@ class InteractionMasks extends React.Component {
 
   componentWillUnmount() {
     this.unsubscribeSelectCell();
+    this.unsubscribeSelectStart();
+    this.unsubscribeSelectUpdate();
+    this.unsubscribeSelectEnd();
     this.unsubscribeDragEnter();
   }
 
@@ -129,7 +148,7 @@ class InteractionMasks extends React.Component {
     const { enableCellSelect, columns, rowGetter, onCheckCellIsEditable } = this.props;
     const { selectedPosition } = this.state;
     return isSelectedCellEditable({ enableCellSelect, columns, rowGetter, selectedPosition, onCheckCellIsEditable });
-  }
+  };
 
   openEditor = ({ key } = {}) => {
     if (this.isSelectedCellEditable() && !this.state.isEditorEnabled) {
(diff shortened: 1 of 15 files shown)


rowanhill:
@amanmahajan7 Any news on this / when someone could take a look?

Sorry to pester, but I don't want the code to move on again and mean I have to rework and resubmit this PR yet again! :)

rowanhill:
@amanmahajan7 Is this really wontfix? It seemed like there was interest in getting it merged, before?

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -296,23 +315,60 @@ class InteractionMasks extends React.Component {
...
+
+  changeCellFromArrowKeyAction(e) {
+    const currentPosition = this.state.selectedPosition;
+    const { cellNavigationMode } = this.props;
+    const next = this.getNextPositionAndFireOnHitBoundaryIfNeccesary(e, currentPosition, cellNavigationMode);
+
+    this.selectCell(next);
+  }
+
+  changeCellFromTabAction(e) {
```
malonecj: changeCellFromArrowKeyAction and changeCellFromTabAction are practically the same function. It should be consolidated to changeCellFromArrowKeyAction, passing in the relevant cellNavigationMode as a parametre

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -296,23 +315,60 @@ class InteractionMasks extends React.Component {
...
+      this.changeSelectedRangeFromArrowKeyAction(e);
+    } else {
+      this.changeCellFromArrowKeyAction(e);
+    }
+  }
+
+  changeCellFromArrowKeyAction(e) {
+    const currentPosition = this.state.selectedPosition;
+    const { cellNavigationMode } = this.props;
+    const next = this.getNextPositionAndFireOnHitBoundaryIfNeccesary(e, currentPosition, cellNavigationMode);
```
malonecj: The name of this function suggests it has too many responsibilities
Better change this to 
```javascript
const next = this.getNextPosition(e, currentPosition, cellNavigationMode);
this.checkCellIsAtGridBoundary();
```

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -296,23 +315,60 @@ class InteractionMasks extends React.Component {
...
+
+  changeSelectedRangeFromArrowKeyAction(e) {
+    const { cellNavigationMode } = this.props;
+    const currentPosition = this.state.selectedRange.cursorCell || this.state.selectedPosition;
+    const next = this.getNextPositionAndFireOnHitBoundaryIfNeccesary(e, currentPosition, cellNavigationMode);
+
+    this.selectUpdate(next, true, () => { this.selectEnd(); });
+  }
+
+  getNextPositionAndFireOnHitBoundaryIfNeccesary(e, currentPosition, cellNavigationMode) {
```
malonecj: change function name to getNextPosition as described above

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -352,11 +408,85 @@ class InteractionMasks extends React.Component {
...
+          bottomRight: next,
+          startCell: next,
+          cursorCell: next,
+          isDragging: false
+        }
       }, callback);
     }
   };
 
+  selectStart = (cellPosition) => {
```
malonecj: Should be more meaningful name like onCellRangeSelectionStarted

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -352,11 +408,85 @@ class InteractionMasks extends React.Component {
...
+        startCell: cellPosition,
+        cursorCell: cellPosition,
+        isDragging: true
+      },
+      selectedPosition: cellPosition
+    }, () => {
+      if (isFunction(this.props.onCellRangeSelectionStarted)) {
+        this.props.onCellRangeSelectionStarted(this.state.selectedRange);
+      }
+    });
```
malonecj: function could be simplified using shorthand notation 
```javascript
onCellRangeSelectionStarted(selectedPosition) {
  this.setState({
    selectedRange: getSelectedRange(selectedPosition),
    selectedPosition
  }, () => {
    if (isFunction(this.props.onCellRangeSelectionStarted)) {
      this.props.onCellRangeSelectionStarted(this.state.selectedRange);
    }
  })
};
```

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -452,23 +582,38 @@ class InteractionMasks extends React.Component {
             columns={columns}
           />
         )}
-        {!isEditorEnabled && this.isGridSelected() && (
-          <SelectionMask
-            selectedPosition={selectedPosition}
+        {selectedRangeIsSingleCell(this.state.selectedRange) ? (
```
malonecj: I think it could be nice to split this chunk of jsx into simple sub components for readability and to simplify the conditional logic.

```javascript
{selectedRangeIsSingleCell(this.state.selectedRange) ?
<SingleCellSelectView {...relavantProps/> : <CellRangeSelectView {...otherRelevantProps}/>
```

rowanhill:
Thanks for the comments - I've made some changes.

Re "Also, I would like to see some tests around new functionality as well" - there are quite a number of tests in InteractionMasks.spec.js that I think exercise the fiddliest bits of the code, could you let me know more specifically what sort of tests you're thinking of?

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -452,23 +627,10 @@ class InteractionMasks extends React.Component {
...
-              <DragHandle
-                onDragStart={this.handleDragStart}
-                onDragEnd={this.handleDragEnd}
-                onDoubleClick={this.onDragHandleDoubleClick}
-              />
-            )}
-          </SelectionMask>
-        )}
+        {selectedRangeIsSingleCell(this.state.selectedRange) ?
+          this.getSingleCellSelectView(rowData) :
```
malonecj: Our preference is to break render method up into smaller components, rather than splitting into smaller render methods. Even if the two functions here wont be reused outside of this component, splitting into smaller components  is a good way to decompose fat components, and allows us to easily identify which props and state the rendered components use. 

It also allows to easily isolate those units for independent testing.

https://medium.com/dailyjs/techniques-for-decomposing-react-components-e8a1081ef5da

rowanhill: Did you see my previous comment about pulling out components rather than render methods?

I'll do it if you're sure, but I don't think it's a great idea: the components themselves end up being extremely simple (so there's little value in testing them), and take so many props the JSX ends up looking just as bad, if not worse, than before.

Additionally, one of these render methods returns an array of elements, which isn't possible with a single component in anything but the most recent versions of React.

malonecj: As a rule of thumb I would keep the new components in the same file if they are not going to be reused. I thought with destructuring it would look clean to pick the relevant props, and then pass them in with spread syntax. 

But if you tried this and it didn't look so neat, I'm happy to go with this solution. I'm not sure if anyone is using this with earlier versions of React. Unofficially, only React 16 is supported in v4, but the docs should be explicit about this. 

Anyway, this looks good to me

rowanhill:
Okay, I've pushed those changes to the API you requested, plus I've also refactored the selection masks into full-on components. I'm still not 100% convinced by that change, but now you can take a look for yourself and decide which you prefer! I've isolated that change in a commit, so we can just revert it if you'd prefer.

rowanhill:
Oh, and just to be super explicit: that selection mask component refactoring does include a component that returns an array of elements, so it does definitely make this library React 16+ only - so you'll need to be sure that's okay!

malonecj:
I see what you mean. I thought it would be less versbose to pass the props in. Really appreciate your effort on all this. Let's remove the last commit and its good to merge in.

rowanhill:
Okay, I've pushed a revert of that commit, so hopefully that's all sorted now.

xdev-x:
What is the status on this PR? indicated as merged, but it seems it has never been merged to main :(

Issue #1232, opened 2018-06-25
React 16 support


pie6k:
Spent 20+ hours trying to figure out 

```
Error: Invariant Violation: Failed exporting HTML for URL / (XYZ): Element type is invalid: expected a string (for built-in components) or a class/  function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default   and named imports.
```

when trying to build website with `react-static` only to realize `react-data-grid` is not supporting React 16.

It should be red-text'ed on main readme as first line of text. **Damn I'm angry!**

Reproduce:
1. Use `react-static`
2. Create any project with `react-static create`
3. Add `react-data-grid`
4. Use it in any form 
5. Try to run production build (it works in dev build! SIC!) with `react-static build`

Expected result: 
Build works as in development mode

Actual result:
`react-data-grid` is breaking the build silently with error message as above. Removing all instances of `react-data-grid` useage makes build work again. Note `react-static` is based on React 16+.

---

### Which version of React JS are you using?

✅ Officially supported ✅
- [ ] v15.4.x

⚠️ Not officially supported, expect warnings ⚠️
- [ ] v15.5.x
- [ ] v15.6.x

☣️ Not officially supported, expect warnings and errors ☣️
- [x] v16.x.x!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

---

### Which browser are you using?

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

⚠️ Not officially supported, but "should work" ⚠️
- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [x] 🐛 Bug Report
- [ ] 💡 Feature Request

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---

malonecj:
Hi
Yes ReactDataGrid is React 16 compatible as of version 4. I believe your issue is with React static integration.

PR #1233, opened 2018-06-25
Update README.md about React 16

pie6k:
https://github.com/adazzle/react-data-grid/issues/1232


MickaToast:
Sorry, but it does work with React 16.

PR #1244, opened 2018-07-20
Added cell drag side-to-side functionality for editing. 

colindjk:
## Description
Added code to support toggle-able drag side to side functionality. This functionality is accessed by setting the optional ReactDataGrid prop `<enableDragHorizontal={true}`. A toggle to control drag down functionality was also added. 

```js
// These are the defaultProp values, which allow for current grid functionality.
ReactDataGrid.props += { enableDragHorizontal: false, enableDragVertical: true };
```

Added tests in [Grid.spec.js](https://github.com/colindjk/react-data-grid/blob/master/packages/react-data-grid-addons/src/__tests__/Grid.spec.js) in the package `react-data-grid-addons`. 

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Use is only able to drag up and down when `enableCellSelect=true`. 

**What is the new behavior?**
Adds the toggles `enableDragVertical` and `enableDragHorizontal` which allow the user to control how they can drag values across the grid. The `enableDragHorizontal` toggle accesses new functionality which allows the user to copy values across the grid horizontally. 

```
[ ] Yes
[x] No
```

**Other information**:
The `handleDragEnterRow` is essentially deprecated with this update, but still contains it's old functionality for the sake of avoiding a breaking change. In the future, the `handleDragEnterRow` should be replaced with `handleDragEnterCell`. It also should be noted that `handleDragEnterRow` _currently_ is called every time `handleDragEnterCell` is called (meaning even if the row hasn't been left, the function gets called) meaning it might need to have been renamed anyway.

### DIFF
--- a/docs/api/docs.json
+++ b/docs/api/docs.json
@@ -3689,6 +3689,28 @@
         "required": false,
         "description": ""
       },
+      "enableDragVertical": {
+        "type": {
+          "name": "bool"
+        },
+        "required": false,
+        "description": "Allows user to toggle drag down functionality",
+        "defaultValue": {
+          "value": true,
+          "computed": false
+        }
+      },
+      "enableDragHorizontal": {
+        "type": {
+          "name": "bool"
+        },
+        "required": false,
+        "description": "Allows user to toggle drag side-to-side functionality",
+        "defaultValue": {
+          "value": false,
+          "computed": false
+        }
+      },
       "onGridRowsUpdated": {
         "type": {
           "name": "func"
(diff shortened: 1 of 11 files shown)


colindjk:
Is there anyone that could help me figure out why the tests disconnected? All tests pass when I run them locally. 
Basically: are there any differences between the local tests and the CI tests that I should know about? 

Thanks for any help in advance.

PR #1316, opened 2018-10-24
RDG Tree Cell Expand Styling Issues.

JamesPortelli:
- [x] Modify styling so that Tree view with rows larger than 35px will be styled correctly.

- [x] Modify code and styles so that the Cell Exapnder will not be overlapped by the cell content.


amanmahajan7:
Can you catchup master, we just synced next and master

JamesPortelli:
Done @amanmahajan7

packages/react-data-grid/src/Cell.js
```diff
@@ -310,18 +310,23 @@ class Cell extends React.PureComponent {
...
     let isDeleteSubRowEnabled = this.props.cellMetaData.onDeleteSubRow ? true : false;
 
     if (treeDepth > 0 && isExpandCell) {
       cellDeleter = <ChildRowDeleteButton treeDepth={treeDepth} cellHeight={this.props.height} siblingIndex={this.props.expandableOptions.subRowDetails.siblingIndex} numberSiblings={this.props.expandableOptions.subRowDetails.numberSiblings} onDeleteSubRow={this.onDeleteSubRow} isDeleteSubRowEnabled={isDeleteSubRowEnabled} />;
     }
-    return (<div className="react-grid-Cell__value">{cellDeleter}<div style={{ marginLeft: marginLeft }}><span>{CellContent}</span> {this.props.cellControls} {cellExpander}</div></div>);
+    return (
+      <div className="react-grid-Cell__value">
+        {cellDeleter}
+        <div style={{ marginLeft: marginLeft, position: 'relative', top: '50%', transform: 'translateY(-50%)' }}>
```
amanmahajan7: Shorthand property names can be used
```
<div style={{ marginLeft, position: 'relative', top: '50%', transform: 'translateY(-50%)' }}>
```

packages/react-data-grid/src/__tests__/CellExpand.spec.js
```diff
@@ -1,68 +1,60 @@
...
   };
 
-  const renderComponent = (props) => {
-    const wrapper = mount(<CellExpander {...props} />);
-    return wrapper;
+  return {
+    props,
+    wrapper: mount(<CellExpander {...props} />)
   };
+};
```
amanmahajan7: 👍

packages/react-data-grid/src/__tests__/CellExpand.spec.js
```diff
@@ -1,68 +1,60 @@
 import React from 'react';
```
amanmahajan7: Can we rename the test file to `CellExpander.spec.js` to match the component file

packages/react-data-grid/src/__tests__/CellExpand.spec.js
```diff
@@ -1,68 +1,60 @@
...
   });
 
   it('should render correctly when expanded is true', () => {
-    let fakeProps = getFakeProps(true);
-    testElement = renderComponent(fakeProps);
-    expect(testElement.state('expanded')).toBeTruthy();
-    expect(testElement.find('span.rdg-cell-expand').text()).toBe(CellExpand.DOWN_TRIANGLE);
+    const { wrapper } = setup({ expanded: true });
+
+    expect(wrapper.state('expanded')).toBeTruthy();
```
amanmahajan7: Rather than setting the state, can we use `simulate` users behavior

```
wrapper.simulate('click'); // Not sure if this will work but the idea is to simulate a click event
expect(wrapper.find('span').text()).toBe(CellExpand.DOWN_TRIANGLE);
```

packages/react-data-grid/src/__tests__/CellExpand.spec.js
```diff
@@ -1,68 +1,60 @@
...
   });
 
   it('should render correctly when expanded is false', () => {
-    let fakeProps = getFakeProps(false);
-    testElement = renderComponent(fakeProps);
-    expect(testElement.state('expanded')).toBeFalsy();
-    expect(testElement.find('span.rdg-cell-expand').text()).toBe(CellExpand.RIGHT_TRIANGLE);
+    const { wrapper } = setup({ expanded: false });
+
+    expect(wrapper.state('expanded')).toBeFalsy();
```
amanmahajan7: We should not test state, that is an implementation detail. If `expanded` variable name is changed then the tests should not break

packages/react-data-grid/src/__tests__/CellExpand.spec.js
```diff
@@ -1,68 +1,60 @@
...
-    testElement = renderComponent(fakeProps);
-    testElement.simulate('click');
-    expect(testElement.state('expanded')).toBeTruthy();
-    let secondProps = getFakeProps(false);
-    testElement.setProps(secondProps);
-    expect(testElement.state('expanded')).toBeFalsy();
+    wrapper.find('span').simulate('click');
+    expect(wrapper.state('expanded')).toBeTruthy();
+    wrapper.find('span').simulate('click');
+    expect(wrapper.state('expanded')).toBeFalsy();
```
amanmahajan7: I think we can delete some of these tests and just check the span's text after the click event

PR #1349, opened 2018-10-29
Preserve window scroll position on cell selection

amanmahajan7:


amanmahajan7:
@JamesPortelli I had the same concern about page jumping to the top and then scroll back to the position but it does not seem to be happening. I tested on Chrome 65, FF, and Edge.

Calling focus is causing the jump to the top of the page  The default behavior of focus is to scroll the element into the visible area of the browser and somehow it is causing this issue. We can use [`preventScroll`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#Parameters) option but is only supported in Chrome

PR #1353, opened 2018-10-31
Defining the ref callback as a bound method

amanmahajan7:
> If the ref callback is defined as an inline function, it will get called twice during updates, first with null and then again with the DOM element. This is because a new instance of the function is created with each render, so React needs to clear the old ref and set up the new one. You can avoid this by defining the ref callback as a bound method on the class, but note that it shouldn’t matter in most cases.

https://reactjs.org/docs/refs-and-the-dom.html


packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -558,10 +558,9 @@ class ReactDataGrid extends React.Component {
   };
 
   getHeaderRows = () => {
-    let rows = [{ ref: (node) => this.row = node, height: this.props.headerRowHeight || this.props.rowHeight, rowType: HeaderRowType.HEADER }];
+    const rows = [{ height: this.props.headerRowHeight || this.props.rowHeight, rowType: HeaderRowType.HEADER }];
```
amanmahajan7: `ref: this.row` is not used anywhere. Header row defines its own [ref](https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid/src/Header.js#L107).

packages/react-data-grid/src/Grid.js
```diff
@@ -133,9 +133,17 @@ class Grid extends React.Component {
...
+  };
+
+  setEmptyViewRef = (emptyView) => {
+    this.emptyView = emptyView;
+  };
+
   render() {
-    let headerRows = this.props.headerRows || [{ref: (node) => this.row = node}];
-    let EmptyRowsView = this.props.emptyRowsView;
+    const { headerRows } = this.props;
```
amanmahajan7: This will never be undefined.
https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid/src/ReactDataGrid.js#L560
https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid/src/ReactDataGrid.js#L700

packages/react-data-grid/src/Header.js
```diff
@@ -102,29 +105,30 @@ class Header extends React.Component {
...
-        onHeaderDrop={this.props.onHeaderDrop}
-        sortColumn={this.props.sortColumn}
-        sortDirection={this.props.sortDirection}
-        onSort={this.props.onSort}
-        onScroll={this.props.onScroll}
-        getValidFilterValues={this.props.getValidFilterValues}
-        />);
+      return (
+        <HeaderRow
+          key={row.rowType}
```
amanmahajan7: Not sure why was `key` a function (`row.ref` defined in [`ReactDataGrid.js`](https://github.com/adazzle/react-data-grid/blob/master/packages/react-data-grid/src/ReactDataGrid.js#L561))

PR #1369, opened 2018-11-04
POC: use react portals for cell editors

amanmahajan7:
This approach fixes the following problems
- Editor is never hidden as it is rendered outside the grid
- Currently the editor is using fixed position that requires resetting the editor's position when window is scrolled. This is no longer needed as editor is absolutely positioned.
- Additional padding is automatically added when an autocomplete editor is opened on the last row

This PR however requires a major version as it works with React 16. I think it is about time we drop support for React 15

### DIFF
--- a/packages/common/editors/EditorContainer.js
+++ b/packages/common/editors/EditorContainer.js
@@ -2,9 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import joinClasses from 'classnames';
 import SimpleTextEditor from './SimpleTextEditor';
-import {isFunction} from 'common/utils';
+import { isFunction } from 'common/utils';
 import { isKeyPrintable, isCtrlKeyHeldDown } from 'common/utils/keyboardUtils';
 import zIndexes from 'common/constants/zIndexes';
+import EditorPortal from './EditorPortal';
+
 require('../../../themes/react-data-grid-core.css');
 
 const isFrozen = column => column.frozen === true || column.locked === true;
@@ -22,14 +24,13 @@ class EditorContainer extends React.Component {
     onCommit: PropTypes.func,
     onCommitCancel: PropTypes.func,
     firstEditorKeyPress: PropTypes.string,
-    width: PropTypes.number,
-    top: PropTypes.number,
-    left: PropTypes.number,
+    width: PropTypes.number.isRequired,
     scrollLeft: PropTypes.number,
-    scrollTop: PropTypes.number
+    scrollTop: PropTypes.number,
+    position: PropTypes.object.isRequired
   };
 
-  state = {isInvalid: false};
+  state = { isInvalid: false };
   changeCommitted = false;
   changeCanceled = false;
 
@@ -42,7 +43,6 @@ class EditorContainer extends React.Component {
         inputNode.style.height = this.props.height - 1 + 'px';
       }
     }
-    window.addEventListener('scroll', this.setContainerPosition);
   }
 
   componentDidUpdate(prevProps) {
@@ -53,22 +53,8 @@ class EditorContainer extends React.Component {
 
   componentWillUnmount() {
     if (!this.changeCommitted && !this.changeCanceled) {
-      this.commit({key: 'Enter'});
+      this.commit({ key: 'Enter' });
     }
-    window.removeEventListener('scroll', this.setContainerPosition);
-  }
-
-  setContainerPosition = () => {
-    if (this.container) {
-      this.container.style.transform = this.calculateTransform();
-    }
-  }
-
-  calculateTransform = () => {
-    const { column, left, scrollLeft, top, scrollTop } = this.props;
-    const editorLeft = isFrozen(column) ? left : left - scrollLeft;
-    const editorTop = top - scrollTop - window.pageYOffset;
-    return `translate(${editorLeft}px, ${editorTop}px)`;
   }
 
   isKeyExplicitlyHandled = (key) => {
@@ -132,15 +118,15 @@ class EditorContainer extends React.Component {
       return <CustomEditor ref={this.setEditorRef} {...editorProps} />;
     }
 
-    return <SimpleTextEditor ref={this.setEditorRef} column={this.props.column} value={this.getInitialValue()} onBlur={this.commit} rowMetaData={this.getRowMetaData()} onKeyDown={() => {}} commit={() => {}}/>;
+    return <SimpleTextEditor ref={this.setEditorRef} column={this.props.column} value={this.getInitialValue()} onBlur={this.commit} rowMetaData={this.getRowMetaData()} onKeyDown={() => { }} commit={() => { }} />;
   };
 
   onPressEnter = () => {
-    this.commit({key: 'Enter'});
+    this.commit({ key: 'Enter' });
   };
 
   onPressTab = () => {
-    this.commit({key: 'Tab'});
+    this.commit({ key: 'Tab' });
   };
 
   onPressEscape = (e) => {
@@ -239,13 +225,13 @@ class EditorContainer extends React.Component {
   };
 
   commit = (args) => {
-    const {onCommit} = this.props;
+    const { onCommit } = this.props;
     let opts = args || {};
     let updated = this.getEditor().getValue();
     if (this.isNewValueValid(updated)) {
       this.changeCommitted = true;
       let cellKey = this.props.column.key;
-      onCommit({cellKey: cellKey, rowIdx: this.props.rowIdx, updated: updated, key: opts.key});
+      onCommit({ cellKey: cellKey, rowIdx: this.props.rowIdx, updated: updated, key: opts.key });
     }
   };
 
@@ -257,7 +243,7 @@ class EditorContainer extends React.Component {
   isNewValueValid = (value) => {
     if (isFunction(this.getEditor().validate)) {
       let isValid = this.getEditor().validate(value);
-      this.setState({isInvalid: !isValid});
+      this.setState({ isInvalid: !isValid });
       return isValid;
     }
 
@@ -306,8 +292,8 @@ class EditorContainer extends React.Component {
 
   getRelatedTarget = (e) => {
     return e.relatedTarget ||
-            e.explicitOriginalTarget ||
-            document.activeElement; // IE11
+      e.explicitOriginalTarget ||
+      document.activeElement; // IE11
   };
 
   handleRightClick = (e) => {
@@ -321,7 +307,7 @@ class EditorContainer extends React.Component {
     }
 
     if (!this.isBodyClicked(e)) {
-	    // prevent null reference
+      // prevent null reference
       if (this.isViewportClicked(e) || !this.isClickInsideEditor(e)) {
         this.commit(e);
       }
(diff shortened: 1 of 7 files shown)


packages/react-data-grid/src/Canvas.js
```diff
@@ -218,9 +218,13 @@ class Canvas extends React.PureComponent {
   };
 
   setScrollLeft = (scrollLeft) => {
+    if (this.interactionMasks && this.interactionMasks.setScrollLeft) {
```
amanmahajan7: Using the same same approach to fix mask's position

packages/react-data-grid/src/Canvas.js
```diff
@@ -246,28 +250,23 @@ class Canvas extends React.PureComponent {
     return this.props.rowHeight * rowIdx;
   }
 
-  getSelectedRowHeight = (rowIdx) => {
```
amanmahajan7: I believe this logic was added to support variable height rows. This adds additional complexity and it is currently not used by `Copy` and `Drag` masks

packages/react-data-grid/src/Canvas.js
```diff
@@ -414,9 +414,6 @@ class Canvas extends React.PureComponent {
           onCellRangeSelectionCompleted={this.props.onCellRangeSelectionCompleted}
           scrollLeft={this._scroll.scrollLeft}
           scrollTop={this._scroll.scrollTop}
-          prevScrollLeft={this.props.prevScrollLeft}
```
amanmahajan7: This props do not seem to be used anywhere

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -95,6 +92,39 @@ class InteractionMasks extends React.Component {
...
 
+  saveEditorPosition = () => {
+    if (this.selectionMask) {
+      const { left, top } = this.selectionMask.getBoundingClientRect();
+      const { scrollLeft, scrollTop } = document.documentElement;
+      this.editorPosition = { left: left + scrollLeft, top: top + scrollTop };
+    }
+  };
+
+  setMaskScollLeft = (mask, position, scrollLeft) => {
```
amanmahajan7: This logic adjust the mask position during scroll

packages/react-data-grid/src/masks/SelectionMask.js
```diff
@@ -1,43 +1,15 @@
...
-  const column = columnUtils.getColumn(columns, selectedPosition.idx);
-  const height = getSelectedRowHeight(selectedPosition.rowIdx);
-  const top = getSelectedRowTop(selectedPosition.rowIdx);
-  const frozen = isFrozenColumn(columns, selectedPosition);
-  const zIndex = frozen ? zIndexes.FROZEN_CELL_MASK : zIndexes.CELL_MASK;
-  const left = getLeftPosition(frozen, column.left, props);
-  return { height, top, width: column.width, left, zIndex };
-};
+import CellMask from './CellMask';
+import { getSelectedDimensions } from '../utils/SelectedCellUtils';
```
amanmahajan7: This component can be really simplified if we remove `getSelectedRowHeight` and `getSelectedRowTop`

packages/common/editors/EditorContainer.js
```diff
@@ -22,14 +24,13 @@ class EditorContainer extends React.Component {
...
     onCommitCancel: PropTypes.func,
     firstEditorKeyPress: PropTypes.string,
-    width: PropTypes.number,
-    top: PropTypes.number,
-    left: PropTypes.number,
+    width: PropTypes.number.isRequired,
     scrollLeft: PropTypes.number,
-    scrollTop: PropTypes.number
+    scrollTop: PropTypes.number,
+    position: PropTypes.object.isRequired
```
qili26: I am not that familiar with the really old syntax, should we add the shape of the position?

packages/common/editors/EditorContainer.js
```diff
@@ -132,15 +118,15 @@ class EditorContainer extends React.Component {
       return <CustomEditor ref={this.setEditorRef} {...editorProps} />;
     }
 
-    return <SimpleTextEditor ref={this.setEditorRef} column={this.props.column} value={this.getInitialValue()} onBlur={this.commit} rowMetaData={this.getRowMetaData()} onKeyDown={() => {}} commit={() => {}}/>;
+    return <SimpleTextEditor ref={this.setEditorRef} column={this.props.column} value={this.getInitialValue()} onBlur={this.commit} rowMetaData={this.getRowMetaData()} onKeyDown={() => { }} commit={() => { }} />;
```
qili26: should we lift the `onKeyDown={() => { }} commit={() => { }}` to the component like `this.emptyCall`?

packages/common/editors/EditorContainer.js
```diff
@@ -349,15 +335,22 @@ class EditorContainer extends React.Component {
   };
 
   render() {
-    const { width, height, column } = this.props;
-    const zIndex = isFrozen(column) ? zIndexes.FROZEN_EDITOR_CONTAINER  : zIndexes.EDITOR_CONTAINER;
-    const style = { position: 'fixed', height, width, zIndex, transform: this.calculateTransform() };
+    const { width, height, column, position } = this.props;
```
qili26: This `position` under this context in `render()` might be confusing when the dev first look at the code, maybe we can rename it just in this render function? Because `position` is a valid attr in `style`, while we are assigning `...position` to the style as well. But I am OK if we don't change this.

packages/react-data-grid/src/Canvas.js
qili26: since you are touching this file, maybe we can also add some simple doc above the function? like

```
/**
 * what does scrollLeft do
 * @param scrollLeft number
 */
```

packages/react-data-grid/src/Canvas.js
qili26: by saying variable height, do you mean `row1` is 50 and `row2` is 35 ?

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -95,6 +92,39 @@ class InteractionMasks extends React.Component {
     firstEditorKeyPress: null
   };
 
+  saveEditorPosition = () => {
+    if (this.selectionMask) {
+      const { left, top } = this.selectionMask.getBoundingClientRect();
+      const { scrollLeft, scrollTop } = document.documentElement;
+      this.editorPosition = { left: left + scrollLeft, top: top + scrollTop };
```
qili26: I noticed you didn't use this `window.pageYOffset` in the `transform` function that I added in the merged PR. Is it because of the `this.selectionMask` ref? Because the original PR added a window scroll event listener, so that when the grid is at far bottom with window scroll bar and the user drags the window level scroll bar, the mask position will be refreshed. For this `InteractionMasks`, will the position be refreshed when the window level scroll bar is dragged?

amanmahajan7: `window.pageYOffset` is not needed anymore as the editor is now absolutely positioned instead of fixed so it will just work. We however need to make sure scrollleft works for frozen columns and `setScrollLeft` handles that logic

packages/react-data-grid/src/Canvas.js
amanmahajan7: That is correct. This is a great feature to have but we need to handle all the masks (Copy and Drag). This currently does not work with them and has some edge cases so I removed it

packages/common/editors/EditorContainer.js
amanmahajan7: Good idea, I will do that in a separate PR to keep this simple.

packages/common/editors/EditorContainer.js
amanmahajan7: I have removed `position` and added `left/top` props instead

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -699,19 +704,23 @@ class InteractionMasks extends React.Component {
...
+            firstEditorKeyPress={firstEditorKeyPress}
+            onCommit={this.onCommit}
+            onCommitCancel={this.onCommitCancel}
+            rowIdx={selectedPosition.rowIdx}
+            value={getSelectedCellValue({ selectedPosition, columns, rowGetter })}
+            rowData={rowData}
+            column={getSelectedColumn({ selectedPosition, columns })}
+            scrollLeft={scrollLeft}
+            scrollTop={scrollTop}
+            {...{
```
qili26: maybe it's better to clean this code like
```
const restProps = { ...getSelectedDimensions({ selectedPosition, rowHeight, columns, scrollLeft }), ...this.editorPosition };
isEditorEnabled && (
  <EditorContainer 
     existing codes
     {...restProps} 
  />
);
```

packages/react-data-grid/src/Canvas.js
malonecj: I agree that it adds complexity. When starting the work on the Interaction Mask, I did it under the assumption that RDG only supports fixed row height. However, this is not exactly true as it is possible to have dynamic row height using custom row renderers. I have seen a lot of cases out in the wild where this is happening. Removing this, is going to cause a lot of strange display issues for those cases

Supporting dynamic row height is not trivial, and this current implementation was really just a stop gap to support these non documented edge cases. I would prefer to leave this here and think of a better solution that to completely remove it

packages/react-data-grid/src/masks/InteractionMasks.js
malonecj: 👍

packages/common/editors/EditorContainer.js
```diff
@@ -132,15 +119,23 @@ class EditorContainer extends React.Component {
...
+        value={this.getInitialValue()}
+        onBlur={this.commit}
+        rowMetaData={this.getRowMetaData()}
+      />
+    );
   };
 
   onPressEnter = () => {
-    this.commit({key: 'Enter'});
+    this.commit({ key: 'Enter' });
```
malonecj: Maybe we can refactor this to use constants instead of magic strings

packages/react-data-grid/src/masks/SelectionMask.js
malonecj: Agreed but I can see some users getting upset that their variable row height grids no longer work. We need to think of a better solution to handle dynamic row heights

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -95,6 +92,44 @@ class InteractionMasks extends React.Component {
...
+
+  /**
+   * Sets the position of SelectionMask and CopyMask components when the canvas is scrolled
+   * This is only required on the frozen columns
+   * @param scrollLeft number
+   */
+  setScrollLeft = (scrollLeft) => {
+    const { selectionMask, copyMask, state: { selectedPosition, copiedPosition } } = this;
+    this.setMaskScollLeft(selectionMask, selectedPosition, scrollLeft);
+    this.setMaskScollLeft(copyMask, copiedPosition, scrollLeft);
```
malonecj: Does this handle the case of cell range selection as well?

packages/react-data-grid/src/Canvas.js
malonecj: I think they can be safely removed.

PR #1392, opened 2018-11-21
Cell Tooltip - Focus Issues

JamesPortelli:
- [x] Fix  issues between cell action and cell tooltips whereby when trying to click on a cell-action the tooltip will show making it very difficult to select the options.

- [x] Remove un-referenced editors.

- [x] Clean up Frozen column code to get rid of z-index css.


packages/react-data-grid/src/Cell.js
```diff
@@ -168,13 +168,11 @@ class Cell extends React.PureComponent {
       this.props.column.cellClass,
       'react-grid-Cell',
       this.props.className,
-      columnUtils.isFrozen(this.props.column) ? 'react-grid-Cell--frozen' : null,
```
amanmahajan7: I think we should still add the class for any customizations

packages/react-data-grid/src/Cell.js
```diff
@@ -317,13 +315,21 @@ class Cell extends React.PureComponent {
     if (treeDepth > 0 && isExpandCell) {
       cellDeleter = <ChildRowDeleteButton treeDepth={treeDepth} cellHeight={this.props.height} siblingIndex={this.props.expandableOptions.subRowDetails.siblingIndex} numberSiblings={this.props.expandableOptions.subRowDetails.numberSiblings} onDeleteSubRow={this.onDeleteSubRow} isDeleteSubRowEnabled={isDeleteSubRowEnabled} />;
     }
+
+    const tooltip = this.props.tooltip ? (<span className="cell-tooltip-text">{this.props.tooltip}</span>) : null;
```
amanmahajan7: Minor cleanup

```
const tooltip = this.props.tooltip && (<span className="cell-tooltip-text">{this.props.tooltip}</span>);
```

packages/react-data-grid/src/Cell.js
```diff
@@ -317,13 +315,21 @@ class Cell extends React.PureComponent {
...
+
+    const classes = joinClasses('react-grid-Cell__value',
```
amanmahajan7: I think we need to move the tooltip outside the grid otherwise there will be issues similar to editors. Currently if the cell has tooltip then a new div is rendered and positioned. I propose the following solution instead

```
return (
      <div
        {...this.getKnownDivProps()}
        className={className}
        style={style}
        {...events}
        ref={this.setCellRef}
      >
        {cellActionButtons}
        {cellExpander}
        { tooltip ? (<TooltipTrigger placement="left" overlay={tooltip}> {cellContent}<TooltipTrigger > : cellContent }
      </div>
    );
```

`TooltipTrigger` can render the tooltip in a portal and position it

themes/react-data-grid-cell.css
```diff
@@ -50,6 +42,13 @@
 
 .react-grid-Cell:hover {
   background: #eee;
+  /*
+    A minor hack to get cell tooltips to show correctly.
+    Hover is triggered on a child (react-grid-Cell__value)
+    and to avoid having the tooltip hide under the cell in
+    the next row we need to up the z-index of the cell being hovered.
+  */
+  z-index: 101;
```
amanmahajan7: We can get rid of `zIndex` related code if we use the above mentioned solution

PR #1393, opened 2018-11-21
Cleanup zIndex logic

amanmahajan7:
- Removed unnecessary high values of zIndexes
- Fix `EditorContainer` zIndex as is rendered outside the grid

https://www.useloom.com/share/c2a82eb341dc4a639c72f679d681cc52


wennyben:
Many admin-panel theme already have it's z-index hierarchy, mostly from 1 to 10. This low zIndex logic may failed once we are using react-data-grid inside it.
When I was clicking the cell, the editor will not be rendered correctly. 

I was using .css to override it (e.g. .rdg-editor-container { z-index: 5 !important;} ) but this is not a good solution.

can we add an offset, e.g. 10 to avoid this conflict ?

PR #1398, opened 2018-11-26
Enable the no-class-assign

amanmahajan7:
- [x] Enable the `no-class-assign` rule
- [x] Fix row export. This fixes the row re-order example

### DIFF
--- a/packages/react-data-grid/src/index.js
+++ b/packages/react-data-grid/src/index.js
@@ -1,12 +1,15 @@
 import Grid from './ReactDataGrid';
 import RowComparer from 'common/utils/RowComparer';
 import Cell from './Cell';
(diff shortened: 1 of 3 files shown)


packages/react-data-grid/src/index.js
```diff
@@ -1,12 +1,15 @@
 import Grid from './ReactDataGrid';
 import RowComparer from 'common/utils/RowComparer';
 import Cell from './Cell';
+import Row from './Row';
```
amanmahajan7: `Row` uses [`export default`](https://github.com/adazzle/react-data-grid/blob/am-no-class-assign/packages/react-data-grid/src/Row.js#L205) and `require` syntax returns an object instead of the default import

PR #1404, opened 2018-11-28
Fix outside click logic for committing changes

amanmahajan7:


packages/common/editors/EditorContainer.js
```diff
@@ -51,11 +54,27 @@ class EditorContainer extends React.Component {
...
+    document.removeEventListener('click', this.handleDocumentClick);
     if (!this.changeCommitted && !this.changeCanceled) {
       this.commit({ key: 'Enter' });
     }
   }
 
+  handleDocumentClick = (e) => {
+    const { target } = e;
+    const { container, editor } = this;
+    const isClickedInsideEditorContainer = container !== null && (target === container || container.contains(target));
```
Runi-B: Instead of `container !== null && ...` could we use one of the below?

```suggestion
    const isClickedInsideEditorContainer = container && (target === container || container.contains(target));
```

```suggestion
    const isClickedInsideEditorContainer = !!container && (target === container || container.contains(target));
```

packages/common/editors/EditorContainer.js
```diff
@@ -51,11 +54,27 @@ class EditorContainer extends React.Component {
...
+    let isClickedInsideEditor = false;
+    if (editor !== null) {
```
Runi-B: As above...

```suggestion
    if (editor) {
```
```suggestion
    if (!!editor) {
```

amanmahajan7: > React will assign the current property with the DOM element when the component mounts, and assign it back to null when it unmounts

I usually prefer an explicit check than using the `!!` operator but either works

PR #1413, opened 2018-12-04
Tooltip POC

amanmahajan7:


qili26:
@amanmahajan7 
Do we still need this PR? I remember we had a tooltip fix in the patch.

amanmahajan7:
@qili26 I am trying to use Portals to improve the tooltip logic. With current implementation, we add an additional tooltip div with each cell and it suffers from the same issues as editors like it will be cut off on the last row.

PR #1417, opened 2018-12-04
Fix click outside logic

amanmahajan7:
Check the details [here](https://github.com/adazzle/react-data-grid/pull/1417/files#diff-fdcbe490b776c1eed24c88d4b5f27d66R5)

Also fixes the following issues
https://github.com/adazzle/react-data-grid/issues/1429
https://github.com/adazzle/react-data-grid/issues/1450
https://github.com/adazzle/react-data-grid/issues/1432

### DIFF
--- a/packages/common/editors/ClickOutside.js
+++ b/packages/common/editors/ClickOutside.js
@@ -1,0 +1,90 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * Detecting outside click on a react component is surprisingly hard.
+ * A general approach is to have a global click handler on the document
+ * which checks if the click target is inside the editor container or
+ * not using editorContainer.contains(e.target). This approach works well
+ * until portals are used for editors. Portals render children into a DOM
+ * node that exists outside the DOM hierarchy of the parent component so
+ * editorContainer.contains(e.target) does not work. Here are some examples
+ * of the DOM structure with different types of editors
+ *
+ *
+ * SimpleEditor for example Texbox (No Portals)
+ *   <div react-data-grid>..<div>
+ *   <div portal-created-by-the-grid-for-editors>
+ *      <div editor-container>
+ *        <div simple-editor>..</div>
+ *      </div>
+ *   <div>
+ *
+ * ComplexEditor for example Modals (using Portals)
+ *   <div react-data-grid>..<div>
+ *   <div portal-created-by-the-grid-for-editors>
+ *      <div editor-container>
+ *        // Nothing here
+ *      </div>
+ *   <div>
+ *   <div portal-created-by-the-editor>
+ *     <div complex-editor>..</div>
+ *   </div>
+ *
+ *
+ * One approach to detect outside click is to use event bubbling through
+ * portals. An event fired from inside a portal will propagate to ancestors
+ * in the containing React tree, even if those elements are not ancestors
+ * in the DOM tree. This means a click handler can be attached on the document
+ * and on the editor container. The editor container can set a flag to notify
+ * that the click was inside the editor and the document click handler can use
+ * this flag to call onClickOutside. This approach however has a few caveats
+ * - Click handler on the document is set using document.addEventListener
+ * - Click handler on the editor container is set using onClick prop
+ *
+ * This means if a child component inside the editor calls e.stopPropagation
+ * then the click handler on the editor container will not be called whereas
+ * document click handler will be called. To solve this issue mousedown event
+ * is used instead.
+ *
+ * One assumption in the current fix is that click handler on the editor
+ * container is called before the click handler on the document.
+ */
+
+export default class ClickOutside extends React.Component {
+  static propTypes = {
+    children: PropTypes.element.isRequired,
+    onClickOutside: PropTypes.func.isRequired
+  };
+
+  isClickedInside = false;
+
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handleDocumentClick);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', this.handleDocumentClick);
+  }
+
+  handleDocumentClick = (e) => {
+    if (this.isClickedInside) {
+      this.isClickedInside = false;
+      return;
+    }
+
+    this.props.onClickOutside(e);
+  };
+
+  handleClick = () => {
+    this.isClickedInside = true;
+  };
+
+  render() {
+    return React.cloneElement(
+      React.Children.only(this.props.children), {
+        onMouseDown: this.handleClick
+      }
+    );
+  }
+}
--- a/packages/common/editors/ClickOutside.js
+++ b/packages/common/editors/ClickOutside.js
@@ -44,11 +44,10 @@ import PropTypes from 'prop-types';
  *
  * This means if a child component inside the editor calls e.stopPropagation
  * then the click handler on the editor container will not be called whereas
- * document click handler will be called. To solve this issue mousedown event
- * is used instead.
+ * document click handler will be called.
+ * https://github.com/facebook/react/issues/12518
  *
- * One assumption in the current fix is that click handler on the editor
- * container is called before the click handler on the document.
+ * To solve this issue onClickCapture event is used.
  */
 
 export default class ClickOutside extends React.Component {
@@ -60,11 +59,11 @@ export default class ClickOutside extends React.Component {
   isClickedInside = false;
 
   componentDidMount() {
-    document.addEventListener('mousedown', this.handleDocumentClick);
+    document.addEventListener('click', this.handleDocumentClick);
   }
 
   componentWillUnmount() {
-    document.removeEventListener('mousedown', this.handleDocumentClick);
+    document.removeEventListener('click', this.handleDocumentClick);
   }
 
   handleDocumentClick = (e) => {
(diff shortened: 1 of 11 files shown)


packages/common/editors/ClickOutside.js
```diff
@@ -0,0 +1,90 @@
...
+    this.props.onClickOutside(e);
+  };
+
+  handleClick = () => {
+    this.isClickedInside = true;
+  };
+
+  render() {
+    return React.cloneElement(
+      React.Children.only(this.props.children), {
```
MayhemYDG: What does this render method do exactly?

packages/common/editors/EditorContainer.js
```diff
@@ -336,15 +313,17 @@ class EditorContainer extends React.Component {
     const style = { position: 'absolute', height, width, left, top, zIndex: zIndexes.EDITOR_CONTAINER };
     return (
       <EditorPortal>
-        <div style={style}
-          ref={this.setContainerRef}
-          className={this.getContainerClass()}
-          onKeyDown={this.onKeyDown}
-          onContextMenu={this.handleRightClick}
```
MayhemYDG: Do we not need this anymore?

packages/common/editors/EditorContainer.js
```diff
@@ -336,15 +313,17 @@ class EditorContainer extends React.Component {
...
-        >
-          {this.createEditor()}
-          {this.renderStatusIcon()}
-        </div>
+        <ClickOutside
+          onClickOutside={this.commit}>
```
MayhemYDG: Could keep this on the same line as the opening tag.

packages/common/editors/__tests__/EditorContainer.spec.js
```diff
@@ -172,13 +172,17 @@ describe('Editor Container Tests', () => {
...
       expect(props.onCommit.calls.count()).toEqual(0);
     });
 
     it('should commit if any element outside the editor is clicked', () => {
-      document.querySelector('.container').click();
+      const clickEvent = document.createEvent('MouseEvents');
+      clickEvent.initEvent('mousedown', true, true);
+      const input = document.createElement('input');
+      document.querySelector('.container').appendChild(input);
+      input.dispatchEvent(clickEvent);
```
MayhemYDG: Do we need to dispatch the event on an input? We could just mousedown on the body.

packages/common/editors/ClickOutside.js
amanmahajan7: It does 2 things
- `cloneElement` takes as single element so `React.cloneElement(this.props.children)` will may or may not work.[ `React.Children.only`](https://reactjs.org/docs/react-api.html#reactchildrenonly) will make sure that only one child is provided.
> Verifies that children has only one child (a React element) and returns it. Otherwise this method throws an error.

- Add an `onClickCapture` event on the child. I could have used a container `div` but I wasn't sure if it would affect any existing styles or not

packages/common/editors/EditorContainer.js
amanmahajan7: Done

packages/common/editors/__tests__/EditorContainer.spec.js
amanmahajan7: It works. Previously I was dispatching an event on the container and it was returning a [false positive](https://github.com/adazzle/react-data-grid/pull/1404/files#diff-a4a272b1d00cac994690178e6a7ca36aR180) so I decided to use an explicit sibling node.

amanmahajan7:
This is still WIP

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -454,6 +470,10 @@ class InteractionMasks extends React.Component {
 
   selectCell = (cell, openEditor) => {
     const callback = openEditor ? this.openEditor : () => null;
+    // Close the editor to commit any pending changes
+    if (this.state.isEditorEnabled) {
```
amanmahajan7: This fixes https://github.com/adazzle/react-data-grid/issues/1429

packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -216,7 +216,9 @@ class ReactDataGrid extends React.Component {
     /** Called when a column is resized */
     onColumnResize: PropTypes.func,
     /** Called when the grid is scrolled */
-    onScroll: PropTypes.func
+    onScroll: PropTypes.func,
+    /** The node where the editor portal should mount. */
+    editorPortalTarget: PropTypes.instanceOf(Element)
```
amanmahajan7: This fixes https://github.com/adazzle/react-data-grid/issues/1432

packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -236,9 +238,10 @@ class ReactDataGrid extends React.Component {
       rowsEnd: 2
     },
     enableCellAutoFocus: true,
-    onBeforeEdit: () => {},
+    onBeforeEdit: () => { },
     minColumnWidth: 80,
-    columnEquality: ColumnMetrics.sameColumn
+    columnEquality: ColumnMetrics.sameColumn,
+    editorPortalTarget: document.body
```
amanmahajan7: This fixes https://github.com/adazzle/react-data-grid/issues/1450

packages/common/editors/EditorPortal.js
```diff
@@ -2,24 +2,36 @@ import React from 'react';
...
+    this.setState({ isMounted: true });
   }
 
   componentWillUnmount() {
-    editorRoot.removeChild(this.el);
+    this.props.target.removeChild(this.el);
   }
 
   render() {
+    // Don't render the portal until the component has mounted,
```
amanmahajan7: This is needed for editors that require the editor container to be attached to DOM during `componentDidMount`. For e.g legacy editors using jquery to initialize the component in `componentDidMount`

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -701,21 +721,24 @@ class InteractionMasks extends React.Component {
...
-            rowData={rowData}
-            column={getSelectedColumn({ selectedPosition, columns })}
-            scrollLeft={scrollLeft}
-            scrollTop={scrollTop}
-            {...{
-              ...this.getSelectedDimensions(selectedPosition),
-              ...this.editorPosition
-            }}
-          />
+          <EditorPortal target={this.props.editorPortalTarget}>
```
amanmahajan7: The `componentDidMount()` method of child components is invoked before that of parent components and this caused two issues. 
- If the editor is rendered and the editor container is not attached to the DOM then legacy editors break
- If rendering is deferred using [`isMounted` flag](https://github.com/adazzle/react-data-grid/blob/d447bb45d439428315d9298ec08e3e8e7d50b439/packages/common/editors/EditorPortal.js#L31) then [`componentDidMount`](https://github.com/adazzle/react-data-grid/blob/d447bb45d439428315d9298ec08e3e8e7d50b439/packages/common/editors/EditorContainer.js#L38) throws an error

Moving `EditorPortal` to the parent component fixes both problems. Let me know if more explanation is needed

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -139,11 +140,24 @@ class InteractionMasks extends React.Component {
...
+      const { left: selectionMaskLeft, top: selectionMaskTop } = this.selectionMask.getBoundingClientRect();
+      if (editorPortalTarget === document.body) {
+        const { scrollLeft, scrollTop } = document.scrollingElement || document.documentElement;
+        return {
+          left: selectionMaskLeft + scrollLeft,
+          top: selectionMaskTop + scrollTop
+        };
+      }
+
+      const { left: portalTargetLeft, top: portalTargetTop } = editorPortalTarget.getBoundingClientRect();
```
amanmahajan7: @MayhemYDG I am only checking `editorPortalTarget` for now as checking all the parent components were causing some issues and I do not expect react data grid to be used in multiple nested scrolling parents. If a need arises then we can revisit this logic.

packages/common/editors/EditorPortal.js
```diff
@@ -2,24 +2,36 @@ import React from 'react';
...
+    isMounted: false
   };
 
   el = document.createElement('div');
 
   componentDidMount() {
-    editorRoot.appendChild(this.el);
+    this.props.target.appendChild(this.el);
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState({ isMounted: true });
```
MayhemYDG: Does this need to be in the state? I wonder if we can just do `this.isMounted = true`

packages/react-data-grid/src/ReactDataGrid.js
```diff
@@ -179,9 +179,9 @@ class ReactDataGrid extends React.Component {
         }),
         PropTypes.shape({
           keys: PropTypes.shape({
-             /** The selected unique ids of each row */
+            /** The selected unique ids of each row */
             values: PropTypes.array.isRequired,
-             /** The name of the unoque id property of each row */
+            /** The name of the unoque id property of each row */
```
MayhemYDG: Typo: `unique` instead of `unoque`.

packages/common/editors/EditorPortal.js
amanmahajan7: We need to trigger an update when the element is attached to DOM. With instance variable react would not know that we want to rerender the component

PR #1422, opened 2018-12-05
Cell Tooltip - Focus Issues 

JamesPortelli:
- [x] Fix  issues between cell action and cell tooltips whereby when trying to click on a cell-action the tooltip will show making it very difficult to select the options.

- [x] Remove un-referenced editors.

- [x] Clean up Frozen column code to get rid of z-index css.


packages/react-data-grid/src/Row.js
```diff
@@ -91,7 +91,7 @@ class Row extends React.Component {
     const { colOverscanStartIdx, colOverscanEndIdx, columns } = this.props;
     const frozenColumns = columns.filter(c => columnUtils.isFrozen(c));
     const nonFrozenColumnsToRender = columns.slice(colOverscanStartIdx, colOverscanEndIdx + 1).filter(c => !columnUtils.isFrozen(c));
-    return frozenColumns.concat(nonFrozenColumnsToRender)
+    return nonFrozenColumnsToRender.concat(frozenColumns)
```
qili26: Is it OK to change the `frozen` vs `non-frozen` order?

PR #1425, opened 2018-12-06
Release 5.0.5 -> Master

JamesPortelli:


themes/react-data-grid-cell.css
```diff
@@ -27,6 +27,15 @@
   box-shadow: 2px 0 5px -2px rgba(136, 136, 136, .3) !important;
 }
 
+/* cell which have tooltips need to have a higher z-index on hover so that the tooltip appears above the other cells*/
+.react-grid-Cell.has-tooltip:hover {
+  z-index: 5;
```
amanmahajan7: We need to change the values to match the new zIndex values
https://github.com/adazzle/react-data-grid/blob/master/packages/common/constants/zIndexes.js#L3

so 
```
z-index: 1;

and

z-index: 3
```

docs/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ title: Changelog
 - **Breaking** Use react portals for cell editors ([1369](https://github.com/adazzle/react-data-grid/pull/1369))
 - **TechDebt:** Upgrade build tools to latest versions ([1350](https://github.com/adazzle/react-data-grid/pull/1350))
 
+## 5.0.5 (Dec 6, 2018)
```
amanmahajan7: This can be moved above the last major entry. I think we should sort it by date

Issue #1478, opened 2019-02-28
react-data-grid-addons breaks minification in production build


carbonrobot:
### Which version of React JS are you using?

✅ Officially supported ✅
- [ ] v15.4.x

⚠️ Not officially supported, expect warnings ⚠️
- [ ] v15.5.x
- [ ] v15.6.x

☣️ Not officially supported, expect warnings and errors ☣️
- [x ] v16.x.x

---

### Which browser are you using?

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x ] Chrome

⚠️ Not officially supported, but "should work" ⚠️
- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [x ] 🐛 Bug Report
- [ ] 💡 Feature Request

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---

### Issue Details

The react-data-grid-addons module fails to minify in a production build of create-react-app or plain webpack build.

Error in browser console:
```
react-data-grid-addons.js:2182 Uncaught TypeError: Super expression must either be null or a function, not undefined
    at react-data-grid-addons.js:2182
```

Steps to reproduce:

```
$ create-react-app gridtest
$ cd gridtest
$ npm install react-data-grid react-data-grid-addons immutable
$ npm start
```

- Paste the following code in the App.js file.

```
import React, { Component } from 'react';
import ReactDataGrid from 'react-data-grid';
import { Toolbar } from 'react-data-grid-addons';

const columns = [{ key: 'id', name: 'ID' }, { key: 'title', name: 'Title' }, { key: 'count', name: 'Count' }];
const rows = [{ id: 0, title: 'row1', count: 20 }, { id: 1, title: 'row1', count: 40 }, { id: 2, title: 'row1', count: 60 }];

class App extends Component {
  render() {
    return (
      <div className="App">
        <ReactDataGrid columns={columns} rowGetter={i => rows[i]} rowsCount={3} minHeight={150} toolbar={Toolbar} />
      </div>
    );
  }
}
export default App;
```

- Open browser and confirm the application works normally
- Stop the application, and create a production build

```
$ npm run build
$ npm install -g serve
$ serve -s build
```

- Open browser and confirm the application is broken with error in in console.

slonix:
[...]

PR #1490, opened 2019-03-13
Update packages

amanmahajan7:
Fixed https://github.com/adazzle/react-data-grid/issues/1478

I am not sure if there is a better fix but setting `keep_fnames: false` fixes the problem

Similar issues
https://github.com/terser-js/terser/issues/308
https://github.com/airbnb/react-with-styles/issues/151


config/webpack.common.config.js
```diff
@@ -34,19 +35,55 @@ const config = {
...
+        use: ['style-loader', 'css-loader']
       }
     ]
   },
   resolve: {
     alias: {
       common: path.resolve('packages/common/')
     }
+  },
+  optimization: {
```
amanmahajan7: Copied from https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L186

amanmahajan7:
Not sure why the build is failing. It passes locally 

image

packages/react-data-grid-examples/package.json
```diff
@@ -25,9 +25,7 @@
   "devDependencies": {
     "bootstrap": "^3.2.0",
     "jquery": "^2.1.1",
-    "lodash": "^4.13.1",
-    "lodash.groupby": "^4.5.1",
     "markdown": "^0.5.0",
-    "tape": "^4.9.1"
+    "recharts": "^1.5.0"
```
MayhemYDG: I wonder if we shouldn't just move all the deps under devDependencies for the examples.

packages/react-data-grid/src/__tests__/CellAction.spec.js
```diff
@@ -1,104 +1,103 @@
...
-        const { wrapper } = setup({ isFirst: false });
-        const wrapperProps = wrapper.props();
-
-        expect(wrapper.length).toBe(1);
-        expect(wrapperProps.className).toBe('rdg-cell-action');
-      });
-    });
-  });
-});
+import React from 'react';
```
MayhemYDG: Let's add a gitattributes file in another PR.
https://help.github.com/en/articles/dealing-with-line-endings

packages/react-data-grid/package.json
```diff
@@ -19,5 +19,13 @@
...
     "react": "^16.0.0",
     "react-dom": "^16.0.0"
+  },
+  "devDependencies": {
+    "react-contextmenu": "^2.10.0"
+  },
+  "dependencies": {
+    "object-assign": "^4.1.1",
+    "react-is-deprecated": "^0.1.2",
+    "shallowequal": "^1.1.0"
```
MayhemYDG: Needs some sorting out here as well.

packages/react-data-grid-addons/package.json
```diff
@@ -18,10 +18,21 @@
     "jquery": "^2.1.1",
     "lodash": "^4.13.1",
     "lodash.groupby": "^4.5.1",
-    "react-data-grid": "^6.0.1"
+    "moment": "^2.22.2",
+    "react-data-grid": "^6.0.1",
+    "react-dnd-test-backend": "^2.6.0",
+    "underscore": "^1.9.1"
```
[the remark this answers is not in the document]
amanmahajan7: `moment` is marked as [external ](https://github.com/adazzle/react-data-grid/blob/692aff20942be5f1b83897d30a10ba26f694c2d8/config/webpack.common.config.js#L22)so added it as dev dependency. i do not think it is needed as the components that use it are not exported. I will remove it in a separate PR. I will move underscore

packages/react-data-grid-addons/package.json
```diff
@@ -18,10 +18,21 @@
     "jquery": "^2.1.1",
     "lodash": "^4.13.1",
     "lodash.groupby": "^4.5.1",
-    "react-data-grid": "^6.0.1"
+    "moment": "^2.22.2",
+    "react-data-grid": "^6.0.1",
```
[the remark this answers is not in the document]
amanmahajan7: correct

package-lock.json
```diff
@@ -13,31 +13,89 @@
...
         "debug": "^4.1.0",
         "json5": "^2.1.0",
-        "lodash": "^4.17.10",
+        "lodash": "^4.17.11",
         "resolve": "^1.3.2",
         "semver": "^5.4.1",
         "source-map": "^0.5.0"
       },
       "dependencies": {
+        "@babel/generator": {
```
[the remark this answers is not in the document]
amanmahajan7: Deleted lock files

packages/react-data-grid-examples/package.json
amanmahajan7: I think we should just delete the examples project and migrate all the examples to the new website

MayhemYDG: Let's do that yeah.

PR #1497, opened 2019-03-14
Delete unused code

amanmahajan7:
The following components are never exported so they are safe to delete
- DateRangeEditor
- DateRangeFilter
- DateRangeFormatter
- AutoCompleteTokensEditor


packages/react-data-grid-addons/package-lock.json
```diff
@@ -2558,6 +2546,12 @@
 				"object-assign": "^4.1.0"
 			}
 		},
+		"react-data-grid": {
```
MayhemYDG: Might want to run `npm install` from the root directory to get rid of this.

amanmahajan7: npm install does not fix it

PR #1507, opened 2019-03-18
Add support for component using ForwardRef

amanmahajan7:
Using [`react-is`](https://github.com/facebook/react/tree/master/packages/react-is) to test valid elements and components. I tested a custom editor using `ForwardRef` and everything works as expected

Also check https://github.com/facebook/react/issues/12453

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     "@babel/polyfill": "^7.2.5",
     "classnames": "^2.2.6",
     "keymirror": "^0.1.1",
-    "prop-types": "^15.7.2"
+    "prop-types": "^15.7.2",
+    "react-is": "^16.8.4"
   },
   "devDependencies": {
     "@babel/core": "^7.3.4",
--- a/packages/react-data-grid/src/ColumnComparer.js
+++ b/packages/react-data-grid/src/ColumnComparer.js
@@ -1,11 +1,11 @@
-import { isValidElement } from 'react';
+import { isElement } from 'react-is';
 
 export function sameColumn(a, b) {
   let k;
 
   for (k in a) {
     if (a.hasOwnProperty(k)) {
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -33,7 +33,8 @@
     "@babel/polyfill": "^7.2.5",
     "classnames": "^2.2.6",
     "keymirror": "^0.1.1",
-    "prop-types": "^15.7.2"
+    "prop-types": "^15.7.2",
+    "react-is": "^16.8.4"
```
MayhemYDG: This should be a dependency of rdg once we move common.

packages/react-data-grid/src/ColumnComparer.js
```diff
@@ -1,11 +1,11 @@
-import { isValidElement } from 'react';
+import { isElement } from 'react-is';
 
 export function sameColumn(a, b) {
   let k;
 
   for (k in a) {
     if (a.hasOwnProperty(k)) {
-      if ((typeof a[k] === 'function' && typeof b[k] === 'function') || (isValidElement(a[k]) && isValidElement(b[k]))) {
+      if ((typeof a[k] === 'function' && typeof b[k] === 'function') || (isElement(a[k]) && isElement(b[k]))) {
```
MayhemYDG: Shouldn't these typeof checks be replaced with `isValidElementType`?

package.json
amanmahajan7: That's correct. For now I added it here

packages/react-data-grid/src/ColumnComparer.js
amanmahajan7: That makes sense. For this PR I just replaced `React.isValidElement` with `ReactIs.isElement` to be on  the safer side. I will revisit this once we clean up column metrics logic.

PR #1512, opened 2019-03-18
Enable various react/jsx-* rules

amanmahajan7:
All auto fixed except the `indent` rule. It was mostly auto-fixed but had to manually make some [changes](https://github.com/adazzle/react-data-grid/pull/1512/commits/b43c25ade00f398837303de39a9e6eb47d54bbed)

### DIFF
--- a/packages/react-data-grid-addons/src/draggable/DropTargetRowContainer.js
+++ b/packages/react-data-grid-addons/src/draggable/DropTargetRowContainer.js
@@ -19,13 +19,15 @@ const rowDropTarget = (Row) => class extends React.Component {
     const overlayTop = this.props.idx * this.props.height;
     return connectDropTarget(<div>
       <Row ref={(node) => this.row = node} {...this.props} />
-      {isOver && canDrop && <div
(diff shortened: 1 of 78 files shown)


packages/react-data-grid-addons/src/draggable/DropTargetRowContainer.js
```diff
@@ -19,10 +19,13 @@ const rowDropTarget = (Row) => class extends React.Component {
...
       <Row ref={(node) => this.row = node} {...this.props} />
-      {isOver && canDrop && <div className="rowDropTarget" style={{
+      {isOver && canDrop && <div
+        className="rowDropTarget"
+        style={{
         top: overlayTop,
         height: this.props.height
-      }} /> }
+      }}
+                            /> }
```
amanmahajan7: Which eslint rule can fix this?

MayhemYDG: `jsx-wrap-multilines` iirc.

PR #1514, opened 2019-03-19
Update faker

MayhemYDG:


MayhemYDG:
Yep, looks good, no errors.

PR #1520, opened 2019-03-19
Deprecate performance utils

amanmahajan7:
`whyDidYouUpdate` is not used in any projects
`rowComparer` does not need to be re-export from the `addons` package


amanmahajan7:
Good point! We should look into removing `lodash`

PR #1529, opened 2019-03-22
Jasmine migration

amanmahajan7:


packages/react-data-grid-addons/src/editors/__tests__/AutoCompleteEditor-integration.spec.js
```diff
@@ -41,7 +41,7 @@ describe('AutoCompleteEditor integration', () => {
       TestUtils.Simulate.click(textInputEle.nextSibling.children[0]);
       // Expect value to have been committed
       expect(commitSpy).toHaveBeenCalled();
-      expect(commitSpy.calls.count()).toEqual(1);
+      expect(commitSpy.mock.calls.length).toEqual(1);
```
MayhemYDG: I think we can replace these two exceptions with
```js
expect(commitSpy).toHaveBeenCalledTimes(1);
```

PR #1532, opened 2019-03-22
Update reselect

MayhemYDG:
https://github.com/reduxjs/reselect/blob/master/CHANGELOG.md


packages/react-data-grid-addons/package.json
```diff
@@ -26,7 +26,7 @@
     "react-dnd": "^2.6.0",
     "react-dnd-html5-backend": "^2.6.0",
     "react-select": "^1.3.0",
-    "reselect": "^2.5.1",
+    "reselect": "^4.0.0",
```
amanmahajan7: Did you check if `AutoCompleteFilter` works as expected?

amanmahajan7:
Needs catchup

packages/react-data-grid-addons/package.json
MayhemYDG: Just tried and it seems like it's working fine. Dunno if this is a component impacted by this change though.

PR #1537, opened 2019-03-25
Bundle stylesheets separately

MayhemYDG:
- Moved the stylesheets under `packages/react-data-grid{-addons,}/styles/`
  - It's possible we might have some addons-only css in the base rdg stylesheet now, we can clean up later.
- Renamed them, changed extensions to `.less`
- Removed imports in js, effectively stopped bundling the css stylesheets in the webpack bundles
  - This is a necessary step towards removing webpack bundles
- These files are unused afaict so I deleted them:
  - `themes/daterangepicker-bs3.css`
  - `themes/DragColumn.PNG`
  - `themes/drag_column_full.PNG`
  - `themes/react-data-grid-viewport.css`
    - ☝️ This one has a single `.react-grid-Viewport` selector, which is referenced only once in `packages/react-data-grid/src/Viewport.js`, but the stylesheet itself was never imported so I deleted it anyway.


config/webpack.common.config.js
```diff
@@ -36,10 +36,6 @@ module.exports = {
         test: /\.(js|jsx)$/,
         exclude: /node_modules/,
         use: ['babel-loader']
-      },
-      {
```
amanmahajan7: Have you considered using[ `mini-css-extract-plugin`](https://github.com/webpack-contrib/mini-css-extract-plugin)? I think it has a few advantages
- We do not need custom `prebuild` step and custom `buildStylesheets.js` script
- We can combine it with `less-loader`
- It should create separate CSS bundle of each entry point. 
- Each file can import the css it needs which I think is cleaner and we can move the css next to file instead of a separate style folder.

MayhemYDG: I'm considering removing webpack bundles instead, so we don't bundle RDG's dependency in and simplify the workflow, so we wouldn't need that.

MayhemYDG:
I haven't thought about theming support, but if we have consistent class names it might be good enough. It's certainly not a priority right now.

config/webpack.common.config.js
MayhemYDG: We could also just move these files closer to their components, glob them, generate the index file and pass it to Less. Less only needs text as an input, so we don't actually *need* an index file.

tools/buildStylesheets.js
```diff
@@ -0,0 +1,29 @@
...
+const CleanCSS = require('clean-css');
+
+const cleanCSS = new CleanCSS({
+  level: { 1: { specialComments: false }, 2: {} },
+  rebase: false,
+  returnPromise: true
+});
+
+[
+  './packages/react-data-grid/style/rdg.less',
```
amanmahajan7: We should name it `react-data-grid.css` to match the generated js bundle name

PR #1543, opened 2019-04-01
Deprecate EditorBase and ContainerEditorWrapper

amanmahajan7:


liamzhang40:
Is there a reason why ContainerEditorWrapper is being deprecated?

MayhemYDG:
We were using it as band-aid to support wrapped redux components, which isn't an issue anymore with react-redux supporting forwardRef.

PR #1549, opened 2019-04-03
Temporary fix for the selection bug

amanmahajan7:
This issue happens when RowRenderer returns different columns for different rows. 

How to reproduce
- Add 3 rows with 10, 10, and 2 columns
- Select column 5 on row 2. 
- Delete row 2.
- RDG tries to find column 5 on row 2 (which was row 3) and it does not exists.

This is [same behavior as V5](https://github.com/adazzle/react-data-grid/pull/1369/files#diff-04fdd57267f50e49721eb930c70e4059L593) and it needs to be revisited

### DIFF
--- a/packages/react-data-grid/src/masks/InteractionMasks.js
+++ b/packages/react-data-grid/src/masks/InteractionMasks.js
@@ -646,9 +646,9 @@ class InteractionMasks extends React.Component {
     this.copyMask = node;
   };
 
-  getSelectedDimensions = (selectedPosition) => {
-    const { scrollLeft, getRowHeight, getRowTop, getRowColumns } = this.props;
-    const columns = getRowColumns(selectedPosition.rowIdx);
+  getSelectedDimensions = (selectedPosition, useGridColumns) => {
+    const { scrollLeft, getRowHeight, getRowTop, getRowColumns, columns: gridColumns } = this.props;
+    const columns = useGridColumns ? gridColumns : getRowColumns(selectedPosition.rowIdx);
     const top = getRowTop(selectedPosition.rowIdx);
     const rowHeight = getRowHeight(selectedPosition.rowIdx);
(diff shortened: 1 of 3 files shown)


packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -646,9 +646,9 @@ class InteractionMasks extends React.Component {
     this.copyMask = node;
   };
 
-  getSelectedDimensions = (selectedPosition) => {
-    const { scrollLeft, getRowHeight, getRowTop, getRowColumns } = this.props;
-    const columns = getRowColumns(selectedPosition.rowIdx);
+  getSelectedDimensions = (selectedPosition, useGridColumns) => {
+    const { scrollLeft, getRowHeight, getRowTop, getRowColumns, columns: gridColumns } = this.props;
+    const columns = useGridColumns ? gridColumns : getRowColumns(selectedPosition.rowIdx);
```
MayhemYDG: Not familiar with the RDG codebase yet, but why not just always use `this.props.columns`? Is there a benefit to using `getRowColumns()` in this method?

amanmahajan7: I am not 100% sure but I think `getRowColumns`, `getRowTop` methods were added to support variable height rows and grouping. I found [this bug](https://github.com/adazzle/react-data-grid/pull/1289/files) that I fixed but I do not remember the details (always fill in ticket details)

It is probably the same issue where you are copying values form a row with less columns

amanmahajan7: I think we may be able to remove `getRowColumns` as I fixed a different issue with copy mask
https://github.com/adazzle/react-data-grid/pull/1289/files#diff-04fdd57267f50e49721eb930c70e4059R640

PR #1561, opened 2019-04-05
Set up Typescript build

MayhemYDG:
- Dropped published webpack bundles
  - we'll now publish esm files built by tsc, targeting es5 compatibility
  - dependencies won't be bundled anymore, rdg users will have their bundle sizes reduced
- We need to enable `allowJs` for now
  - this means we can't enabled the `declaration` config for now, as they don't work together.
  - this in turns means we can't enable the `composite` config, as it requires `declaration` to be enabled.
  - this in turns means we can't use project references yet
    - https://www.typescriptlang.org/docs/handbook/project-references.html
  - so the root `tsconfig.json` doesn't work yet
  - so we have to run `tsc` in each individual package for now
  - once we get rid of all js files, we'll be able to just do `tsc --build` with the root config to build everything at once.
  - hope it's all clear 🙃 
- Since rdg-addons depends on rdg, we'll need to move rdg to full-ts first, disable `allowJs`, enable `declaration`, and add the `types` field in the rdg package.json.
  - otherwise ts code in rdg-addons will complain about missing types for rdg.
- The examples (`npm start`) still work, but we still need to run `npm build` beforehand.
- ~~We'll need to tweak eslint so it parses/lints ts files properly~~


.vscode/settings.json
```diff
@@ -0,0 +1,18 @@
+{
+  "eslint.enable": true,
+  "eslint.autoFixOnSave": true,
```
qili26: do we need vs code plugin for this?

MayhemYDG: You don't "need" the vscode plugin, but it helps.

packages/react-data-grid/src/HeaderCellType.ts
```diff
@@ -0,0 +1,8 @@
+enum HeaderCellType {
+  SORTABLE = 0,
```
qili26: I don't think the value matters, maybe we should just set the `SORTABLE = 'SORTABLE'`, it will be easier for debugging. as `0, 1, 2, 3` doesn't mean anything when you see the value.

https://www.typescriptlang.org/docs/handbook/enums.html - string enums

MayhemYDG: I was thinking that maybe we could just use enum keys, with no assigned values.

>it will be easier for debugging

Shouln't be an issue with sourcemaps I think.

packages/react-data-grid/src/common/cells/headerCells/FilterableHeaderCell.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface Props {
+  column: Column;
+  onChange(change: { filterTerm: string; column: Column }): void;
+}
+
+export default function FilterableHeaderCell({ column, onChange }: Props) {
+  const [filterTerm, setFilterTerm] = useState('');
+
+  function handleChange(event: React.ChangeEvent<HTMLInputElement>) {
```
qili26: is there any reason you prefer named function vs arrow function?

packages/react-data-grid/src/common/constants/index.ts
```diff
@@ -0,0 +1,28 @@
+import * as CellNavigationMode from './CellNavigationMode';
+import * as EventTypes from './EventTypes';
+import * as HeaderRowType from './HeaderRowType';
+
+enum UpdateActions {
+  CELL_UPDATE,
```
qili26: we need to use string enums here

packages/react-data-grid/src/common/editors/CheckboxEditor.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+interface Column<T> {
```
qili26: maybe we should consider using some of the defs from my PR?
https://github.com/adazzle/react-data-grid/pull/1466/files#diff-bee94f77d0ed359d978babfd95123920R309

packages/react-data-grid/src/common/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface Column {
```
qili26: https://github.com/adazzle/react-data-grid/pull/1561/files#diff-17633e7ac6651e7fd23ad88c079c43efR3

It's duplicated here.

packages/react-data-grid/src/masks/__tests__/DragHandle.spec.tsx
```diff
@@ -20,7 +20,7 @@ describe('DragHandle', () => {
   it('should render a draggable component', () => {
     const { dragHandle } = setup();
 
-    expect(dragHandle.prop('draggable')).toBe('true');
+    expect(dragHandle.prop('draggable')).toBe(true);
```
qili26: `toBeTruthy()` ?

packages/react-data-grid/src/common/cells/headerCells/FilterableHeaderCell.tsx
MayhemYDG: Mostly taste. They're shorter, quicker to parse for humans, easier to type as well.

They're also hoisted, so you can't run into any mutual recursion bugs. You could even place them after the `return` and they'd still work.
image

packages/react-data-grid/src/common/types.ts
MayhemYDG: Resolved.

packages/react-data-grid/src/common/editors/CheckboxEditor.tsx
MayhemYDG: I'd like to define it as needed/incrementally, to make sure we get it 100% right, and not make mistakes ahead of time.

packages/react-data-grid/src/common/constants/index.ts
MayhemYDG: For debugging purposes?

qili26: for example when you are debugging an cell update event, and the event is passed to somewhere else and you only have the instance. e.g. when you see `event.updateType(don't remember the correct field)`, it will give you the string value `CELL_UPDATE` rather than `0, 1, 2, 3`

packages/react-data-grid/src/masks/__tests__/DragHandle.spec.tsx
qili26: I see, we want the exactly `true`

packages/react-data-grid/package.json
```diff
@@ -19,13 +21,16 @@
...
+  },
   "dependencies": {
     "classnames": "^2.2.6",
-    "keymirror": "^0.1.1",
     "prop-types": "^15.7.2",
     "react-is": "^16.8.4",
     "react-is-deprecated": "^0.1.2",
-    "shallowequal": "^1.1.0"
+    "shallowequal": "^1.1.0",
+    "tslib": "^1.9.3"
```
amanmahajan7: Do we need `tslib`?

packages/react-data-grid/package.json
```diff
@@ -19,13 +21,16 @@
...
   },
   "peerDependencies": {
     "react": "^16.0.0",
```
amanmahajan7: We should change the peer dependency as we are using hooks now

tsconfig.base.json
```diff
@@ -0,0 +1,20 @@
...
+    "jsx": "react",
+    "lib": ["es7", "es6", "dom", "dom.iterable"],
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noImplicitReturns": true,
+    "noUnusedLocals": true,
+    "pretty": true,
+    "sourceMap": true,
+    "strict": true,
+    "target": "es5"
```
amanmahajan7: Should we not set the target to `es6` or `esnext`?

config/webpack.common.config.js
```diff
@@ -4,42 +4,6 @@ const TerserPlugin = require('terser-webpack-plugin');
 
 module.exports = {
   mode: RELEASE ? 'production' : 'development',
```
amanmahajan7: May be we should delete this file and move the code to `webpack-dev-server.js`

packages/react-data-grid/src/KeyCodes.ts
```diff
@@ -96,6 +96,6 @@ const keyCodes = {
   BackSlash: 220,
   CloseBracket: 221,
   SingleQuote: 222
-};
+} as const;
```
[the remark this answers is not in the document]
MayhemYDG: It's a nice thing to have I think. We can't accidentally modify the object, and the types will be the actual values rather than `number`.

packages/react-data-grid/package.json
MayhemYDG: Yes, the `importHelpers` compiler option makes the generated code import helpers from tslib, instead of generating them for each file.
Much like `@babel/plugin-transform-runtime` is generating imports to `@babel/runtime`
https://babeljs.io/docs/en/babel-plugin-transform-runtime
https://www.typescriptlang.org/docs/handbook/compiler-options.html

tsconfig.base.json
MayhemYDG: `"target": "es5"` will generate es5 code (so no classes, no template strings, etc).
`"module": "esnext"` will generate es6 imports/exports.
That way we get tree-shaking from bundlers + ie11 compat.

We could just not transpile at all besides ts types/enums, and let users transpile down to the browsers they target, wdyt?

amanmahajan7: That is what I was thinking as well and let the users take care of transpiling. Then we probably not need tslib also. 

Just to be on the safer side, lets just keep ES5 for now and we can revisit.

MayhemYDG: Sounds good.

PR #1569, opened 2019-04-10
tsify SelectionRangeMask, utils, RowsContainer, RowGroup

MayhemYDG:


MayhemYDG:
I think we should just import Immutable instead of relying on `window.Immutable`.

packages/react-data-grid/src/RowGroup.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { forwardRef } from 'react';
+import { last } from './common/utils';
+import { EventTypes } from './common/constants';
+import { Column, CellMetaData } from './common/types';
+import EventBus from './masks/EventBus';
+
+interface Props {
+  height: number;
+  columns: Column[];
+  row: unknown;
```
MayhemYDG: We'll need to overhaul these types at some point.

packages/react-data-grid/src/common/utils/index.ts
```diff
@@ -0,0 +1,51 @@
+// eslint-disable-next-line
+declare var Immutable: any;
```
MayhemYDG: I think we should just import Immutable and not rely on `window.Immutable`.
This file might need an overhaul, don't think we really need this `isFunction`.

amanmahajan7: That would be better and we can add `ImmutableJS` as a peer dependency. I also wonder if we even need to support immutable js or not.

PR #1575, opened 2019-04-11
Add a ToolbarContainer component

MayhemYDG:


packages/react-data-grid/src/ToolbarContainer.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface ToolbarProps {
+  columns: Column[];
+  rowsCount: number;
+  onToggleFilter(): void;
+}
+
+export interface ToolbarContainerProps extends ToolbarProps {
+  /** Component used to render toolbar above the grid */
+  toolbar: React.ReactElement<ToolbarProps> | React.ComponentType<ToolbarProps>;
```
amanmahajan7: Should we change the type to 

```
toolbar: React.ReactElement<ToolbarProps> | React.ComponentType<ToolbarProps>;
```

or change the call to 

```
this.props.toolbar && <ToolbarContainer
```

MayhemYDG: I think we should strive to reduce the complexity in big components, hence why I make this ToolbarContainer in the first place, so I went with your first suggestion.

PR #1576, opened 2019-04-11
tsify ColumnComparer

MayhemYDG:
We might want to revisit the types when we know more about how `sameColumn` is used.


packages/react-data-grid/src/ColumnComparer.ts
```diff
@@ -0,0 +1,24 @@
+import { isElement } from 'react-is';
+
+export function sameColumn<A, B>(a: A, b: B): boolean {
+  for (const k in a) {
+    if (a.hasOwnProperty(k)) {
+      const valA = a[k] as unknown;
+      const valB = b[k as string as keyof B] as unknown;
```
MayhemYDG: Without the `as string as keyof B` part:
image

Without at least one `as unknown`:
image

packages/react-data-grid/src/ColumnComparer.ts
```diff
@@ -0,0 +1,24 @@
+import { isElement } from 'react-is';
+
+export function sameColumn<A, B>(a: A, b: B): boolean {
```
MayhemYDG: We should probably replace the generic types once we know how this function is used.

PR #1577, opened 2019-04-11
tsify SelectedCellUtils, viewportUtils, canvasUtils, Viewport

MayhemYDG:
I've also moved `keyboardUtils.spec.ts` closer to `keyboardUtils.ts` and tsified it.


packages/react-data-grid/src/common/types.ts
```diff
@@ -29,3 +29,23 @@ export interface CellMetaData {
...
+
+export interface Dimension {
+  width: number;
+  height: number;
+  top: number;
+  left: number;
+  zIndex: number;
+}
+
+export type RowGetter = (rowIdx: number) => unknown;
```
qili26: Should we consider a generic type?
```
export type RowGetter = <R = any>(rowIdx: number) => R;
```

packages/react-data-grid/src/masks/InteractionMasks.js
```diff
@@ -29,7 +29,7 @@ import keyCodes from '../KeyCodes';
 const SCROLL_CELL_BUFFER = 2;
 
 export default class InteractionMasks extends React.Component {
-  static dispplayName = 'InteractionMasks';
+  static displayName = 'InteractionMasks';
```
qili26: do we really need this `displayName`?

packages/react-data-grid/src/utils/SelectedCellUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+  return getColumn(columns, selectedPosition.idx);
+}
+
+interface getSelectedCellValueOpts {
+  selectedPosition: Position;
+  columns: Column[];
+  rowGetter: RowGetter;
+}
+
+export function getSelectedCellValue({ selectedPosition, columns, rowGetter }: getSelectedCellValueOpts) {
```
qili26: consider adding return type?

packages/react-data-grid/src/masks/InteractionMasks.js
MayhemYDG: It's not a necessity, but it helps with debugging as class/function names will change with minification.

packages/react-data-grid/src/utils/SelectedCellUtils.ts
MayhemYDG: I need to tsify rowutils first, otherwise I don''t know yet what the return type should be. Should I add a TODO comment?

packages/react-data-grid/src/common/types.ts
MayhemYDG: I'm not sure yet.

I want to avoid `<R = any>` for now, as it's likely users won't set the generic type and `any` isn't useful.

`<R>` alone defaults to an empty object `{}` if the generic isn't defined.
image

I think we should just revisit this later.

packages/react-data-grid/src/utils/__tests__/viewportUtils.spec.ts
```diff
@@ -46,22 +69,22 @@ describe('viewportUtils', () => {
...
-      { key: 'col2', width: 20, left: 20 },
-      { key: 'col3', width: 20, left: 40 },
-      { key: 'col4', width: 20, left: 60 },
-      { key: 'col5', width: 20, left: 80 },
-      { key: 'col6', width: 1, left: 100 },
-      { key: 'col7', width: 1, left: 101 },
-      { key: 'col8', width: 1, left: 102 },
-      { key: 'col9', width: 1, left: 103 }
+    const getCols = (): Column[] => [
+      { key: 'col1', name: 'col1', width: 20, left: 0, onCellChange() {} },
```
amanmahajan7: It seems like `onCellChange` is special prop for `Checkbox` column. It is not used anywhere else

https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/ReactDataGrid.js#L736
https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/common/editors/CheckboxEditor.tsx#L13

May be mark it as an optional prop? We need to revisit `Checkbox` formatter

PR #1582, opened 2019-04-15
Overhaul common/utils/index, tsify ColumnUtils and HeaderRow

MayhemYDG:
We're not importing Immutable directly.


packages/react-data-grid/src/HeaderRow.js
```diff
@@ -1,172 +0,0 @@
...
-      || this.props.sortDirection !== nextProps.sortDirection
-    );
-  }
-
-  getHeaderCellType = (column) => {
-    if (column.filterable) {
-      if (this.props.filterable) return HeaderCellType.FILTERABLE;
-    }
-
-    if (column.sortable && column.rowType !== HeaderRowType.FILTER) return HeaderCellType.SORTABLE;
```
MayhemYDG: `column.rowType` 🙁 Replaced it with a prop.

packages/react-data-grid/src/HeaderRow.js
```diff
@@ -1,172 +0,0 @@
...
-
-    return HeaderCellType.NONE;
-  };
-
-  getFilterableHeaderCell = (column) => {
-    const FilterRenderer = column.filterRenderer || FilterableHeaderCell;
-    return <FilterRenderer {...this.props} onChange={this.props.onFilterChange} />;
```
MayhemYDG: We were not setting the `column` prop here. 🙁

packages/react-data-grid/src/HeaderRow.js
```diff
@@ -1,172 +0,0 @@
...
-  };
-
-  getSortableHeaderCell = (column) => {
-    const sortDirection = this.props.sortColumn === column.key ? this.props.sortDirection : DEFINE_SORT.NONE;
-    const sortDescendingFirst = column.sortDescendingFirst === undefined ? false : column.sortDescendingFirst;
-    return <SortableHeaderCell columnKey={column.key} onSort={this.props.onSort} sortDirection={sortDirection} sortDescendingFirst={sortDescendingFirst} headerRenderer={column.headerRenderer} />;
```
MayhemYDG: `column` prop missing here as well. 🙁

packages/react-data-grid/src/HeaderRow.js
```diff
@@ -1,172 +0,0 @@
...
-      case HeaderCellType.SORTABLE:
-        return this.getSortableHeaderCell(column);
-      case HeaderCellType.FILTERABLE:
-        return this.getFilterableHeaderCell(column);
-      default:
-        return undefined;
-    }
-  };
-
-  getStyle = () => {
```
MayhemYDG: Unused. 🙁

packages/react-data-grid/src/HeaderRow.js
```diff
@@ -1,172 +0,0 @@
...
-      overflowX: 'hidden',
-      overflowY: 'hidden'
-    };
-
-    const { width, height, style, onScroll } = this.props;
-    const cells = this.getCells();
-    return (
-      <div
-        width={width}
-        height={height}
```
MayhemYDG: width and height props set directly on a div do nothing, and break types. 🙁

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -0,0 +1,160 @@
...
+      } else {
+        cells.push(cell);
+      }
+    }
+
+    return cells.concat(frozenCells);
+  }
+
+  setScrollLeft(scrollLeft: number): void {
+    this.props.columns.forEach((column?: Column, i?: number) => {
```
MayhemYDG: Immutable's types aren't top notch, that's why we end up with awkward types here.
image
The package version is also 2 years old so that doesn't help.

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -0,0 +1,157 @@
...
+    if (column.sortable && this.props.rowType !== HeaderRowType.FILTER) {
+      return HeaderCellType.SORTABLE;
+    }
+
+    return HeaderCellType.NONE;
+  }
+
+  getFilterableHeaderCell(column: Column) {
+    const FilterRenderer = column.filterRenderer || FilterableHeaderCell;
+    return <FilterRenderer column={column} onChange={this.props.onFilterChange} />;
```
amanmahajan7: Filter renderer needs `getValidFilterValues` prop otherwise it throws an error 

image

Eventually, I think we should remove `getValidFilterValues` prop as the component can fetch valid values on mount. Wdyt?

packages/react-data-grid/src/common/types.ts
```diff
@@ -7,16 +10,26 @@ export interface Column<T = unknown> {
   left: number;
   cellClass?: string;
 
+  editable?: boolean | ((rowData: unknown) => boolean);
   draggable?: boolean;
   filterable?: boolean;
   frozen?: boolean;
+  locked?: boolean;
```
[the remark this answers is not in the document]
MayhemYDG: locked is used only a few times in the codebase, I'll remove it.

packages/react-data-grid/src/HeaderRow.tsx
MayhemYDG: I've added the prop, but I don't think this is something that should be passed as a prop to ReactDataGrid, as it's only used for custom filter renderers anyway.

packages/react-data-grid/src/ColumnUtils.ts
```diff
@@ -36,5 +36,5 @@ export function canEdit(column: Column, rowData: unknown, enableCellSelect?: boo
 }
 
 export function isFrozen(column: Column): boolean {
```
amanmahajan7: We can remove this function in a separate pr

PR #1585, opened 2019-04-17
tsify InteractionMasks

MayhemYDG:


packages/react-data-grid/src/masks/InteractionMasks.tsx
```diff
@@ -0,0 +1,750 @@
...
+      this.selectFirstCell();
+    }
+  }
+
+  componentWillUnmount() {
+    this.unsubscribeSelectCell!();
+    this.unsubscribeSelectStart!();
+    this.unsubscribeSelectUpdate!();
+    this.unsubscribeSelectEnd!();
+    this.unsubscribeDragEnter!();
```
amanmahajan7: We can do something like the following.

```
unsubscribeHandlers: (() => void)[] = [];

componentDidMount() {
   this.unsubscribeHandlers = [
     eventBus.subscribe(EventTypes.SELECT_START, this.onSelectCellRangeStarted),
     ...
   ];
}

componentWillUnmount() {
   this.unsubscribeHandlers.forEach(h => h());
}
```

wdyt?

packages/react-data-grid/src/masks/InteractionMasks.tsx
```diff
@@ -0,0 +1,750 @@
...
+      // Setting data is required to make an element draggable in FF
+      const transferData = JSON.stringify(selectedPosition);
+      try {
+        e.dataTransfer.setData('text/plain', transferData);
+      } catch (ex) {
+        // IE only supports 'text' and 'URL' for the 'type' argument
+        e.dataTransfer.setData('text', transferData);
+      }
+      this.setState({
+        draggedPosition: selectedPosition
```
amanmahajan7: It looks like we can initialize `overRowIdx` as `selectedPosition.rowIdx` as we return null when `overRowIdx === rowIdx`

packages/react-data-grid/src/masks/InteractionMasks.tsx
```diff
@@ -0,0 +1,750 @@
...
+
+  renderCellRangeSelectView() {
+    const { columns, rowHeight } = this.props;
+    return (
+      <>
+        <SelectionRangeMask
+          {...getSelectedRangeDimensions({ selectedRange: this.state.selectedRange, columns, rowHeight })}
+        />
+        <SelectionMask
+          {...this.getSelectedDimensions(this.state.selectedRange.startCell, true)}
```
amanmahajan7: I think we can probably use `this.state.selectedPosition` as `startCell` is same  as `selectedPosition`

https://github.com/adazzle/react-data-grid/blob/9a612b56c32da4dbbbc2761ccda36d0ab2ee61dd/packages/react-data-grid/src/masks/InteractionMasks.tsx#L516

packages/react-data-grid/src/masks/InteractionMasks.tsx
MayhemYDG: Done.

packages/react-data-grid/src/masks/InteractionMasks.tsx
MayhemYDG: Done, hope it works.

packages/react-data-grid/src/RowGroup.tsx
```diff
@@ -42,7 +42,7 @@ const RowGroup = forwardRef<HTMLDivElement, Props>(function RowGroup(props, ref)
   }
 
   function onClick() {
-    props.eventBus.dispatch(EventTypes.SELECT_CELL, { rowIdx: props.idx });
+    props.eventBus.dispatch(EventTypes.SELECT_CELL, { rowIdx: props.idx, idx: 0 });
```
MayhemYDG: I've set it to `idx: 0`, I don't think we have a live demo of this component in our examples, but I rendered it by tweaking the Canvas, and it works okay.
Since this renders just one grid-wide "cell", I don't think it can have any other idx.

packages/react-data-grid/src/masks/InteractionMasks.tsx
MayhemYDG: Done.

PR #1586, opened 2019-04-17
Tsify EditorContainer

amanmahajan7:
I will update tests in a separate PR

### DIFF
--- a/packages/react-data-grid/src/common/editors/EditorContainer.tsx
+++ b/packages/react-data-grid/src/common/editors/EditorContainer.tsx
@@ -34,7 +34,7 @@ export default class EditorContainer extends React.Component<Props, State> {
   changeCommitted = false;
   changeCanceled = false;
 
-  editor = React.createRef<Editor>();
+  private readonly editor = React.createRef<Editor>();
   readonly state: Readonly<State> = { isInvalid: false };
 
   componentDidMount() {
@@ -96,7 +96,7 @@ export default class EditorContainer extends React.Component<Props, State> {
   };
 
   createEditor() {
-    const editorProps: EditorProps = {
+    const editorProps: EditorProps & { ref: React.RefObject<Editor> } = {
       ref: this.editor,
       column: this.props.column,
       value: this.getInitialValue(),
@@ -182,19 +182,13 @@ export default class EditorContainer extends React.Component<Props, State> {
   };
 
   editorHasResults = () => {
-    if (this.getEditor().hasResults) {
-      return this.getEditor().hasResults!();
-    }
-
-    return false;
+    const { hasResults } = this.getEditor();
+    return hasResults ? hasResults() : false;
   };
 
   editorIsSelectOpen = () => {
-    if (this.getEditor().isSelectOpen) {
-      return this.getEditor().isSelectOpen!();
-    }
-
-    return false;
+    const { isSelectOpen } = this.getEditor();
+    return isSelectOpen ? isSelectOpen() : false;
   };
 
   getRowMetaData() {
@@ -247,8 +241,9 @@ export default class EditorContainer extends React.Component<Props, State> {
   };
 
   isNewValueValid = (value: unknown) => {
-    if (this.getEditor().validate) {
-      const isValid = this.getEditor().validate!(value);
+    const { validate } = this.getEditor();
+    if (validate) {
+      const isValid = validate(value);
       this.setState({ isInvalid: !isValid });
       return isValid;
     }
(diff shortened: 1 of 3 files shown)


packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -56,50 +63,48 @@ export default class EditorContainer extends React.Component {
...
 
-  setEditorRef = (editor) => {
-    this.editor = editor;
-  }
-
-  createEditor = () => {
-    const editorProps = {
-      ref: this.setEditorRef,
+  createEditor() {
+    const editorProps: EditorProps = {
```
MayhemYDG: Maybe we should use `PropsWithRef<EditorProps>`.
That way we can remove the `ref` property on `EditorProps`, and it'll be re-usable for other components.

packages/react-data-grid/src/common/types.ts
```diff
@@ -18,11 +20,12 @@ export interface Column<T = unknown> {
   sortable?: boolean;
   sortDescendingFirst?: boolean;
 
-  editor?: unknown;
+  editor?: React.ComponentType<EditorProps>;
   headerRenderer?: React.ReactElement | React.ComponentType<{ column: Column; rowType: HeaderRowType }>;
```
MayhemYDG: `Column` should probably be `Column<T>` here.

packages/react-data-grid/src/common/editors/SimpleTextEditor.tsx
```diff
@@ -1,36 +1,28 @@
...
-  value: string;
-  onBlur: React.FocusEventHandler<HTMLInputElement>;
-  column: Column;
-}
+type Props = Pick<EditorProps<string>, 'value' | 'column' | 'onBlur'>;
 
-export default class SimpleTextEditor extends React.Component<Props> {
-  input: HTMLInputElement | null = null;
+export default class SimpleTextEditor extends React.Component<Props> implements Editor {
+  input = React.createRef<HTMLInputElement>();
```
MayhemYDG: We should probably mark all refs as `private readonly`.

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -1,50 +1,57 @@
...
+}
+
+export default class EditorContainer extends React.Component<Props, State> {
+  static displayName = 'EditorContainer';
 
-  state = { isInvalid: false };
   changeCommitted = false;
   changeCanceled = false;
 
+  editor = React.createRef<Editor>();
```
MayhemYDG: Let's mark it `private readonly`.

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -178,38 +182,38 @@ export default class EditorContainer extends React.Component {
   };
 
   editorHasResults = () => {
-    if (isFunction(this.getEditor().hasResults)) {
-      return this.getEditor().hasResults();
+    if (this.getEditor().hasResults) {
+      return this.getEditor().hasResults!();
```
MayhemYDG: To avoid the type assertion you could do
```ts
const { hasResults } = this.getEditor();
if (hasResults) return hasResults();
```

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -178,38 +182,38 @@ export default class EditorContainer extends React.Component {
...
     }
 
     return false;
   };
 
   editorIsSelectOpen = () => {
-    if (isFunction(this.getEditor().isSelectOpen)) {
-      return this.getEditor().isSelectOpen();
+    if (this.getEditor().isSelectOpen) {
+      return this.getEditor().isSelectOpen!();
```
MayhemYDG: Same as above.

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -244,9 +246,9 @@ export default class EditorContainer extends React.Component {
     this.props.onCommitCancel();
   };
 
-  isNewValueValid = (value) => {
-    if (isFunction(this.getEditor().validate)) {
-      const isValid = this.getEditor().validate(value);
+  isNewValueValid = (value: unknown) => {
+    if (this.getEditor().validate) {
+      const isValid = this.getEditor().validate!(value);
```
MayhemYDG: Same as above.

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -256,42 +258,33 @@ export default class EditorContainer extends React.Component {
 
   isCaretAtBeginningOfInput = () => {
     const inputNode = this.getInputNode();
-    return inputNode.selectionStart === inputNode.selectionEnd
-      && inputNode.selectionStart === 0;
+    if (inputNode instanceof HTMLInputElement) {
+      return inputNode.selectionStart === inputNode.selectionEnd
+        && inputNode.selectionStart === 0;
+    }
+    return false;
```
MayhemYDG: Can be simplified
```ts
return inputNode instanceof HTMLInputElement
  && inputNode.selectionEnd === 0
  && inputNode.selectionStart === 0;
```
I don't think we even need the `selectionStart` comparison, as it shouldn't ever be greater than `selectionEnd`.

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -256,42 +258,33 @@ export default class EditorContainer extends React.Component {
...
   };
 
   isCaretAtEndOfInput = () => {
     const inputNode = this.getInputNode();
-    return inputNode.selectionStart === inputNode.value.length;
+    if (inputNode instanceof HTMLInputElement) {
+      return inputNode.selectionStart === inputNode.value.length;
+    }
+    return false;
```
MayhemYDG: Similar simplification can be done as my previous comment.

packages/react-data-grid/src/common/types.ts
```diff
@@ -20,8 +20,8 @@ export interface Column<T = unknown> {
   sortable?: boolean;
   sortDescendingFirst?: boolean;
 
-  editor?: React.ComponentType<EditorProps>;
-  headerRenderer?: React.ReactElement | React.ComponentType<{ column: Column; rowType: HeaderRowType }>;
+  editor?: unknown;
```
amanmahajan7: I will change it once we have better types for editors

PR #1589, opened 2019-04-18
tsify Cell

amanmahajan7:


packages/react-data-grid/src/Cell.tsx
```diff
@@ -1,131 +1,102 @@
...
-
-  cell = React.createRef();
-
-  state = {
-    isCellValueChanging: false,
-    isLockChanging: false
+    value: ''
   };
 
-  componentWillReceiveProps(nextProps) {
```
amanmahajan7: `componentWillReceiveProps` is not required as we can check the change in `componentDidUpdate`

packages/react-data-grid/src/Cell.tsx
```diff
@@ -159,8 +130,7 @@ export default class Cell extends React.PureComponent {
         'rdg-last--frozen': lastFrozenColumnIndex === idx,
         'row-selected': isRowSelected,
         'has-tooltip': !!tooltip,
-        'rdg-child-cell': expandableOptions && expandableOptions.subRowDetails && expandableOptions.treeDepth > 0,
-        'last-column': column.isLastColumn
```
amanmahajan7: `isLastColumn` is never set

packages/react-data-grid/src/Cell.tsx
```diff
@@ -233,16 +169,38 @@ export default class Cell extends React.PureComponent {
...
+          };
+        } else {
+          allEvents[event] = (e: Event) => {
+            columnEvents[event](e, eventInfo);
+          };
+        }
+      }
     }
 
-    return this.createEventDTO(gridEvents, columnEvents, onColumnEvent);
```
amanmahajan7: I think the whole purpose of this code is to add an event it does not exist or call the old event along with new if it exists. This can be really simplified

packages/react-data-grid/src/Row.js
```diff
@@ -142,7 +142,14 @@ export default class Row extends React.Component {
...
     const subRowDetails = this.props.subRowDetails;
     if (subRowDetails) {
-      return { canExpand: subRowDetails && subRowDetails.field === columnKey && ((subRowDetails.children && subRowDetails.children.length > 0) || subRowDetails.group === true), field: subRowDetails.field, expanded: subRowDetails && subRowDetails.expanded, children: subRowDetails && subRowDetails.children, treeDepth: subRowDetails ? subRowDetails.treeDepth : 0, subRowDetails };
+      return {
+        canExpand: subRowDetails && subRowDetails.field === columnKey && ((subRowDetails.children && subRowDetails.children.length > 0) || subRowDetails.group === true),
+        field: subRowDetails.field,
+        expanded: subRowDetails && subRowDetails.expanded,
+        hildren: subRowDetails && subRowDetails.children,
+        treeDepth: subRowDetails ? subRowDetails.treeDepth : 0,
+        subRowDetails
```
MayhemYDG: This is wrapped in `if (subRowDetails) {`, so all these `subRowDetails &&` are redundant, as well as the `subRowDetails ?`

```js
getExpandableOptions = (columnKey) => {
  const { subRowDetails } = this.props;
  if (!subRowDetails) return {};

  const { field, expanded, children, treeDepth } = subRowDetails;

  return {
    canExpand: subRowDetails.field === columnKey && ((subRowDetails.children && subRowDetails.children.length > 0) || subRowDetails.group === true),
    field,
    expanded,
    children,
    treeDepth,
    subRowDetails
  };
};
```

You've also typo'd `children` as `hildren` 🤦‍♂️

packages/react-data-grid/src/common/types.ts
```diff
@@ -10,9 +10,14 @@ export interface Column<T = unknown> {
   //FIXME: width and left should be optional
   width: number;
   left: number;
+  hidden?: boolean;
   cellClass?: string;
+  events?: {
+    [key: string]: (e: Event, info: ColumnEventInfo) => void;
```
MayhemYDG: Should type it as `undefined | ((e: Event, info: ColumnEventInfo) => void);` to be safe

packages/react-data-grid/src/common/types.ts
```diff
@@ -10,9 +10,14 @@ export interface Column<T = unknown> {
   //FIXME: width and left should be optional
   width: number;
   left: number;
+  hidden?: boolean;
```
MayhemYDG: Let's group the optional booleans together, wdyt?

packages/react-data-grid/src/common/types.ts
```diff
@@ -81,14 +89,65 @@ export interface Editor {
...
+  rowData: RowData;
   height: number;
   onCommit(args?: { key?: string }): void;
   onCommitCancel(): void;
   onBlur(): void;
   onOverrideKeyDown(e: KeyboardEvent): void;
 }
+
+interface SubRowDetails {
+  anExpand: boolean;
```
MayhemYDG: `canExpand`?

packages/react-data-grid/src/Row.js
amanmahajan7: 🤦‍♂️ 
- Lack of unit tests
- Need to move this to TS ASAP

packages/react-data-grid/src/common/types.ts
amanmahajan7: We probably can, let's do it a separate PR

packages/react-data-grid/src/Cell.tsx
```diff
@@ -135,13 +106,13 @@ export default class Cell extends React.PureComponent {
   }
 
   getRowData = (props = this.props) => {
-    return props.rowData.toJSON ? props.rowData.toJSON() : props.rowData;
+    return typeof props.rowData.toJSON === 'function' ? props.rowData.toJSON() : props.rowData;
```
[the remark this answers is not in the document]
amanmahajan7: I am actually not sure how accurate this is. In some places we use `rowData[x]` directly. It is probably best to revisit this

https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/Cell.js#L208

https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/masks/InteractionMasks.tsx#L640
https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/ReactDataGrid.js#L439

packages/react-data-grid/src/Cell.tsx
```diff
@@ -1,131 +1,102 @@
...
 
   componentDidMount() {
     this.checkScroll();
   }
 
-  componentDidUpdate() {
-    if (this.state.isLockChanging && !isFrozen(this.props.column)) {
+  componentDidUpdate(prevProps: Props) {
+    const { column } = this.props;
+    if (isFrozen(prevProps.column) !== isFrozen(column) && !isFrozen(column)) {
```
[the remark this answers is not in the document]
amanmahajan7: We actually do update the columns probably more than we need. I will leave it as for now and add a ticket to rethink about the props

https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/ReactDataGrid.js#L261

packages/react-data-grid/src/Cell.tsx
```diff
@@ -1,131 +1,102 @@
...
+    value: ''
   };
 
-  componentWillReceiveProps(nextProps) {
-    this.setState({
-      isCellValueChanging: this.props.isCellValueChanging(this.props.value, nextProps.value),
-      isLockChanging: isFrozen(this.props.column) !== isFrozen(nextProps.column)
-    });
-  }
+  cell = React.createRef<HTMLDivElement>();
```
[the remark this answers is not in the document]
amanmahajan7: Done

packages/react-data-grid/src/Cell.tsx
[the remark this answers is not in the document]
amanmahajan7: Done

packages/react-data-grid/src/Row.js
amanmahajan7: Fixed

packages/react-data-grid/src/common/types.ts
amanmahajan7: Fixed

PR #1591, opened 2019-04-19
tsify row

amanmahajan7:


packages/react-data-grid/src/Row.tsx
```diff
@@ -0,0 +1,151 @@
...
+  static displayName = 'Row';
+
+  static defaultProps = {
+    cellRenderer: Cell,
+    isSelected: false,
+    height: 35
+  };
+
+  private readonly row = React.createRef<HTMLDivElement>();
+  private readonly cells: { [key: string]: CellRenderer | null } = {};
```
MayhemYDG: Maps are a more appropriate data structure for this kind of usage than dynamic objects.
```ts
private readonly cells = new Map<string, CellRenderer>();
```
This should perform much better, and won't lead to js deoptimization.

packages/react-data-grid/src/Row.tsx
```diff
@@ -0,0 +1,151 @@
...
+      scrollLeft,
+      lastFrozenColumnIndex
+    };
+
+    return <Renderer key={`${key}-${idx}`} {...cellProps} />;
+  }
+
+  getCells() {
+    const { colOverscanStartIdx, colOverscanEndIdx, columns } = this.props;
+    const frozenColumns = columns.filter((c?: Column) => isFrozen(c!)) as Column[];
```
MayhemYDG: >`as Column[]`

This isn't correct, it can be an immutable list, right?

packages/react-data-grid/src/Row.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const { current } = this.row;
+    return current ? current.offsetTop : 0;
+  }
+
+  getRowHeight(): number {
+    return this.props.height;
+  }
+
+  getCellValue(key: string) {
+    let val;
```
MayhemYDG: Useless variable, we can just return immediately.

packages/react-data-grid/src/Row.tsx
```diff
@@ -0,0 +1,151 @@
...
+    if (key === 'select-row') {
+      return this.props.isSelected;
+    }
+    if (typeof this.props.row.get === 'function') {
```
MayhemYDG: I'm guessing `row` can be an immutable map?
```js
if (Immutable.Map.isMap(this.props.row)) {
```

packages/react-data-grid/src/Row.tsx
amanmahajan7: Done

packages/react-data-grid/src/Row.tsx
amanmahajan7: Done

packages/react-data-grid/src/Row.tsx
amanmahajan7: I think we can completely get rid of immuatable js and as it is not integrated properly. We are using `rows[column.key]` in a few places. I will revisit this

packages/react-data-grid/src/Row.tsx
amanmahajan7: That is correct but cannot make it work without `as Column[]`

PR #1597, opened 2019-04-23
Add UMD builds

MayhemYDG:
I've also tweaked the cells map in the Row component, as promised.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,4 @@
-
+'use strict';
 
 // For a detailed explanation regarding each configuration property, visit:
 // https://jestjs.io/docs/en/configuration.html
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,6 +5,13 @@
 
 module.exports = {
   preset: 'ts-jest/presets/js-with-ts',
+  globals: {
+    'ts-jest': {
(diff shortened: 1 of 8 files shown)


packages/react-data-grid/src/Row.tsx
```diff
@@ -47,7 +47,11 @@ export default class Row extends React.Component<RowRendererProps> implements Ro
 
     const cellProps: CellRendererProps & { ref: (cell: CellRenderer | null) => void } = {
       ref: (cell) => {
```
qili26: Do we have any preference? Your code vs this
```
(cell) => cell ? this.cells.set(key, cell) : this.cells.delete(key);
```

MayhemYDG: Both are good IMO.

jest.config.js
```diff
@@ -1,4 +1,4 @@
-'use strict';
+
```
qili26: Any reason removing the `use strict`? I believe if we remove this, we don't need the ln#1 and 2 anymore.

MayhemYDG: Oops, fixed it now. I need to adjust the eslint config so it doesn't autofix it away.

PR #1599, opened 2019-04-23
Fix draggable + resizable HeaderCell

MayhemYDG:


packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -41,8 +42,7 @@ export default class HeaderCell extends React.Component<Props> {
   };
 
   getWidthFromMouseEvent(x: number): number {
-    const node = ReactDOM.findDOMNode(this) as Element;
-    const left = node.getBoundingClientRect().left;
+    const { left } = this.cell.current!.getBoundingClientRect();
```
amanmahajan7: Let's check for null to be on the safer side

PR #1601, opened 2019-04-25
Fix missing key for HeaderCell in HeaderRow

MayhemYDG:


packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -103,11 +107,13 @@ export default class HeaderRow extends React.Component<Props> {
...
     for (let i = 0, len = getSize(columns); i < len; i++) {
       const column = getColumn(columns, i);
+      const key = this.getHeaderCellKey(column);
       const renderer = column.key === 'select-row' && rowType === HeaderRowType.FILTER ? <div /> : this.getHeaderRenderer(column);
 
       const cell = (
         <HeaderCell
-          ref={node => this.cells[i] = node}
+          key={key}
+          ref={node => node ? this.cells.set(key, node) : this.cells.delete(key)}
```
qili26: Not sure if this is a problem, but don't we need to check if the key value pair already exists in the map before setting it? The result behavior is same, just wondering if React directly pass the node (same obj each time) or create new reference and pass down.

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -130,9 +136,10 @@ export default class HeaderRow extends React.Component<Props> {
   }
 
   setScrollLeft(scrollLeft: number): void {
-    this.props.columns.forEach((column?: Column, i?: number) => {
-      const cell = this.cells[i!];
-      if (!cell) return;
+    this.props.columns.forEach((column?: Column) => {
+      const key = this.getHeaderCellKey(column!);
+      if (!this.cells.has(key)) return;
```
qili26: In Javascript I would prefer something like this

```
const cell = this.cells.get(key);
if (!cell) return;
```

Theoretically, finding the key procedure in the the map will just happen once in my code snippet.

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -130,9 +136,10 @@ export default class HeaderRow extends React.Component<Props> {
   }
 
   setScrollLeft(scrollLeft: number): void {
-    this.props.columns.forEach((column?: Column, i?: number) => {
-      const cell = this.cells[i!];
-      if (!cell) return;
+    this.props.columns.forEach((column?: Column) => {
```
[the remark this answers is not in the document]
MayhemYDG: Yes, the Immutablejs types are not very good.

packages/react-data-grid/src/HeaderRow.tsx
MayhemYDG: No? It doesn't matter either way.

packages/react-data-grid/src/HeaderRow.tsx
MayhemYDG: Sure, but if you're familiar with js engine optimization you'd see that:
- `.has()` returns a boolean, which is faster in an if condition, as the engine doesn't have to cast the object or undefined value to a boolean.
- `cell` is thus monomorphic, so it's easier for the engine to optimize.

https://ponyfoo.com/articles/javascript-performance-pitfalls-v8

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -44,6 +44,10 @@ export default class HeaderRow extends React.Component<Props> {
     );
   }
 
+  getHeaderCellKey(column: Column): string {
+    return `headercell-${column.key}`;
```
amanmahajan7: I think we can removed `header-cell` and just use `column.key`.  

https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -130,9 +136,10 @@ export default class HeaderRow extends React.Component<Props> {
   }
 
   setScrollLeft(scrollLeft: number): void {
-    this.props.columns.forEach((column?: Column, i?: number) => {
-      const cell = this.cells[i!];
-      if (!cell) return;
+    this.props.columns.forEach((column?: Column) => {
+      const key = this.getHeaderCellKey(column!);
+      if (!this.cells.has(key)) return;
+      const cell = this.cells.get(key)!;
```
amanmahajan7: I wish TS could infer that we have already checked that keys exists

PR #1602, opened 2019-04-25
tsify more tests

MayhemYDG:


packages/react-data-grid/src/__tests__/Cell.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+  subRowDetails: {
+    canExpand: false,
+    field: 'field',
+    expanded: false,
+    children: [],
+    treeDepth: 0,
+    siblingIndex: 0,
+    numberSiblings: 0
+  }
+};
```
MayhemYDG: What a weird interface...

packages/react-data-grid/src/__tests__/Row.spec.js
```diff
@@ -1,166 +0,0 @@
-import React from 'react';
-import { shallow } from 'enzyme';
-import TestUtils from 'react-dom/test-utils';
```
MayhemYDG: I just nuked the couple tests that used `TestUtils.renderIntoDocument` out of laziness, and they didn't look very pertinent, let me know if you want them back.

packages/react-data-grid/src/__tests__/Cell.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+    expect(formatterInstance.prop('value')).toEqual('Wicklow');
+  });
+
+  it('should render children when those are passed', () => {
+    const wrapper = renderComponent({ children: <div>Child</div> });
+    expect(wrapper.children().text()).toBe('Child');
+  });
+
+  describe('Rendering Cell component', () => {
+    function shallowRenderComponent(props: Props) {
```
qili26: consider reusing this https://github.com/adazzle/react-data-grid/pull/1602/files#diff-193c6bb6a70f1703324d0ea5dff20df5R51 ?

packages/react-data-grid/src/__tests__/Cell.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+      rowData: helpers.rowGetter(11),
+      expandableOptions,
+      isScrolling: false,
+      scrollLeft: 0
+    };
+
+    it('passes classname property', () => {
+      const wrapper = shallowRenderComponent(requiredProperties);
+      const cellDiv = wrapper.find('div').at(0);
+      expect(cellDiv.hasClass('react-grid-Cell'));
```
qili26: for the test at ln#98, ln#103 and ln#108, I believe they can all be caught by 1 snapshot testing.

packages/react-data-grid/src/common/editors/__tests__/EditorContainer.spec.tsx
```diff
@@ -55,18 +64,8 @@ describe('Editor Container Tests', () => {
...
-        }
-        if (typeof document.selection !== 'undefined') {
-          input.focus();
-          return document.selection.createRange().text === input.value;
-        }
-      };
-      const editorNode = wrapper.instance().getInputNode();
-
-      expect(isTextSelected(editorNode)).toBeDefined();
+      const input = wrapper.instance().getInputNode() as HTMLInputElement;
```
qili26: I think we can add the return type to `getInputNode` directly, create an interface and let the editor implements it.

e.g.
```
interface IEditorBase<T> {
  getInputNode: () => HTMLInputElement
}

class EditorContainer extends React.Component implements IEditorBase
```
Then we can save this `as` cast. `IEditorBase` can be further extended by `IEditor` and add `getValue` for example

packages/react-data-grid/src/common/editors/__tests__/EditorContainer.spec.tsx
```diff
@@ -106,60 +102,61 @@ describe('Editor Container Tests', () => {
...
       wrapper.find('#button1').simulate('click');
       wrapper.find('#button2').simulate('click');
 
-      expect(props.onCommit.mock.calls.length).toEqual(0);
+      expect(props.onCommit).not.toHaveBeenCalled();
     });
 
     it('should call onCommitCancel when editor cancels editing', () => {
+      const { wrapper, props } = innerSetup();
       const editor = wrapper.find(TestEditor);
```
qili26: I remember `find` has a generic type like 
```
wrapper.find<TestEditorProps (or just TestEditor, I forgot it exactly)>(TestEditor)
```
then we can avoid the cast next line

packages/react-data-grid/src/__tests__/Cell.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+import Cell, { Props } from '../Cell';
+import helpers from '../helpers/test/GridPropHelpers';
+import CellAction from '../CellAction';
+import { SimpleCellFormatter } from '../formatters';
+import { Column, CellMetaData } from '../common/types';
+
+const testCellMetaData: CellMetaData = {
+  rowKey: 'row',
+  onCellClick() {},
```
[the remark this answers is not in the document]
MayhemYDG: We probably should, but I'm only tsifying the tests in this PR.

packages/react-data-grid/src/__tests__/Cell.spec.tsx
MayhemYDG: The other function is using `testProps`, so I think it's safe to keep it that way to avoid any possible future conflicts.

packages/react-data-grid/src/__tests__/Cell.spec.tsx
MayhemYDG: I guess so, @amanmahajan7 do we want to add snapshot testing to RDG?

packages/react-data-grid/src/common/editors/__tests__/EditorContainer.spec.tsx
MayhemYDG: https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/common/types.ts#L85
We do have an interface for this already, but `getInputNode` might not necessarily return an input element.
I think it's fine to leave that kind of loose type-casting in tests, as it's not always easy to make TS happy, especially with the enzyme wrappers.

packages/react-data-grid/src/common/editors/__tests__/EditorContainer.spec.tsx
MayhemYDG: Fixed.

packages/react-data-grid/src/common/editors/__tests__/EditorContainer.spec.tsx
qili26: @amanmahajan7 Could you confirm this? I remember it does always return the input (or maybe add select?) element ref so that we can set focus on that.

packages/react-data-grid/src/__tests__/Cell.spec.tsx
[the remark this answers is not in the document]
qili26: sure.

packages/react-data-grid/src/common/editors/__tests__/EditorContainer.spec.tsx
qili26: agree that we should get rid of this method, it's confusing.

PR #1606, opened 2019-04-25
tsify ColumnMetrics

MayhemYDG:
I've removed the Immutable.js tests, as we'll remove support for it eventually.


packages/react-data-grid/src/ColumnMetrics.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  if (keys.size > prevColumnsMap.size) return false;
+
+  // We shouldn't need Array.from(), but TS complains
+  for (const key of Array.from(keys)) {
+    if (!prevColumnsMap.has(key) || !nextColumnsMap.has(key)) return false;
+    const prevColumn = prevColumnsMap.get(key)!;
+    const nextColumn = nextColumnsMap.get(key)!;
+    if (isSameColumn(prevColumn, nextColumn)) return false;
```
amanmahajan7: should it not be `!isSameColumn(prevColumn, nextColumn)`?

MayhemYDG: You're right, fixed it now.

MayhemYDG:
@amanmahajan7 did double-check, and resizing/dragging works as usual.

PR #1607, opened 2019-04-25
Fix: re-focus the mask when it's updated, otherwise it will lose focus

qili26:
## Description

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When cell is being selected, clicking on the same cell will make the mask div lost focus.

**What is the new behavior?**
Re-focus the mask whenever it's updated while in boundary.

```
[ ] Yes
[x] No
```

**Other information**:


qili26:
let me try that action.

qili26:
@MayhemYDG 
I re-run in my local and it's working after performing certain actions (clicking inside grid / outside grid, tabbing), could you be more specific for your example?

MayhemYDG:
@qili26 let's say for example we update the `columns` prop from outside the grid, this should trigger `componentDidUpdate`, which will then call `this.focus()`. I think.
@amanmahajan7 wdyt?

qili26:
I believe `isCellWithinBounds` will handle this, if it's within boundary, it's OK to re-focus, if it's out of boundary, this code will not run.

PR #1614, opened 2019-04-28
tsify Grid

amanmahajan7:


packages/react-data-grid/src/Grid.tsx
```diff
@@ -0,0 +1,136 @@
...
+  scrollLeft?: number;
+  onViewportKeydown(e: React.KeyboardEvent<HTMLDivElement>): void;
+  onViewportKeyup(e: React.KeyboardEvent<HTMLDivElement>): void;
+  onColumnResize(index: number, width: number): void;
+  onSort(columnKey: string, direction: DEFINE_SORT): void;
+  getValidFilterValues?(): unknown;
+  onHeaderDrop?(): void;
+}
+
+export default class Grid extends React.Component<Props> {
```
MayhemYDG: Looks like it's safe and easy to turn this one into a functional component.
We can replace the refs with `useRef`, and `_scrollLeft` with... `useRef`.
https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

packages/react-data-grid/src/Grid.tsx
```diff
@@ -0,0 +1,136 @@
...
+  static displayName = 'Grid';
+
+  private header = React.createRef<Header>();
+  private viewport = React.createRef<Viewport>();
```
MayhemYDG: If you want to keep this a class on the other hand, let's change these two to `readonly`.

packages/react-data-grid/src/Grid.tsx
amanmahajan7: Let's revisit this after we fix the scroll logic

packages/react-data-grid/src/Grid.tsx
amanmahajan7: Done

packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -99,12 +99,12 @@ export default class HeaderCell extends React.Component<Props> {
...
 
-    if (rowType === HeaderRowType.HEADER && column.draggable) {
-      const DraggableHeaderCell = this.props.draggableHeaderCell;
+    const DraggableHeaderCell = this.props.draggableHeaderCell;
+    if (rowType === HeaderRowType.HEADER && column.draggable && DraggableHeaderCell) {
       return (
         <DraggableHeaderCell
           column={column}
-          onHeaderDrop={this.props.onHeaderDrop}
+          onHeaderDrop={this.props.onHeaderDrop!}
```
amanmahajan7: We need a better way to validate `onHeaderDrop` prop. May be we can make `onHeaderProp` optional in `DraggableHeaderCell` but drag/drop does not work without it. I am not a fan of using a wrapper for adding draggable support 

https://github.com/adazzle/react-data-grid/blob/next/examples/scripts/example24-draggable-header.js#L80

packages/react-data-grid/src/Header.tsx
```diff
@@ -15,16 +15,15 @@ interface HeaderRowData extends Pick<HeaderRowProps, 'filterable' | 'onFilterCha
 }
 
 type SharedHeaderRowProps = Pick<HeaderRowProps,
-'draggableHeaderCell'
-| 'getValidFilterValues'
-| 'sortDirection'
-| 'sortColumn'
-| 'onHeaderDrop'
-| 'onSort'
-| 'onScroll'
```
amanmahajan7: Grid never provides `onScroll` prop
https://github.com/adazzle/react-data-grid/pull/1614/files#diff-8c020d9d11e86efdcd2e7b42efd9fa29L139

packages/react-data-grid/src/Header.tsx
```diff
@@ -15,12 +15,12 @@ interface HeaderRowData extends Pick<HeaderRowProps, 'filterable' | 'onFilterCha
...
-  | 'sortDirection'
-  | 'sortColumn'
-  | 'onHeaderDrop'
-  | 'onSort'
+'draggableHeaderCell'
+| 'getValidFilterValues'
+| 'sortDirection'
+| 'sortColumn'
+| 'onHeaderDrop'
+| 'onSort'
```
MayhemYDG: It'll be fixed eventually, gotta give it time.
https://github.com/typescript-eslint/typescript-eslint/issues/455

PR #1615, opened 2019-04-28
tsify Header

MayhemYDG:


packages/react-data-grid/src/Header.js
```diff
@@ -1,203 +0,0 @@
...
-    // cellMetaData: PropTypes.shape(cellMetaDataShape)
-  };
-
-  state = { resizing: null };
-
-  componentWillReceiveProps() {
-    this.setState({ resizing: null });
-  }
-
-  shouldComponentUpdate(nextProps, nextState) {
```
MayhemYDG: I've nuked this, don't think if it's worth keeping.

packages/react-data-grid/src/Header.tsx
```diff
@@ -0,0 +1,185 @@
...
+| 'onSort'
+| 'onScroll'
+>;
+
+export interface Props extends SharedHeaderRowProps {
+  columnMetrics: ColumnMetrics;
+  totalWidth?: number;
+  height: number;
+  headerRows: HeaderRowData[];
+  onColumnResize?(pos: number, width: number): void;
```
amanmahajan7: This can be changed to a required prop
https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/ReactDataGrid.js#L805

packages/react-data-grid/src/Header.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  private readonly row = React.createRef<HeaderRow>();
+  private readonly filterRow = React.createRef<HeaderRow>();
+
+  componentWillReceiveProps(): void {
+    this.setState({ resizing: null });
+  }
+
+  onColumnResize = (column: Column, width: number): void => {
+    const state = this.state.resizing || this.props;
```
amanmahajan7: We can simply it
```
const prevColumnMetrics  = this.props.columnMetrics || state.columnMetrics`;
const columnMetrics = resizeColumn({ ...prevColumnMetrics   }, pos, width);

if (columnMetrics.totalWidth <prevColumnMetrics .totalWidth) { 
  columnMetrics.totalWidth = prevColumnMetrics .totalWidth;
...

packages/react-data-grid/src/Header.tsx
amanmahajan7: I think same for `totalWidth`
https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/ReactDataGrid.js#L802

PR #1619, opened 2019-05-01
RDG: Build declaration files

MayhemYDG:
- add `@types` peer deps
- Removed unused deps
- Export internal types

### DIFF
--- a/packages/react-data-grid/src/common/editors/EditorContainer.tsx
+++ b/packages/react-data-grid/src/common/editors/EditorContainer.tsx
@@ -1,4 +1,4 @@
-import React, { MouseEvent, KeyboardEvent } from 'react';
+import React, { KeyboardEvent } from 'react';
 import classNames from 'classnames';
 import { isElement, isValidElementType } from 'react-is';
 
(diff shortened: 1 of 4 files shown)


packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -252,7 +252,7 @@ export default class EditorContainer extends React.Component<Props, State> {
       && inputNode.selectionStart === inputNode.value.length;
   }
 
-  handleRightClick = (e: MouseEvent<HTMLDivElement>) => {
+  handleRightClick = (e: React.MouseEvent<HTMLDivElement>) => {
```
MayhemYDG: Had to fix this, as otherwise I get this error after tweaking the tsconfig:
```
> tsc

src/common/editors/EditorContainer.tsx:255:3 - error TS4031: Public property 'handleRightClick' of exported class has or is using private name 'MouseEvent'.

255   handleRightClick = (e: MouseEvent<HTMLDivElement>) => {
      ~~~~~~~~~~~~~~~~


Found 1 error.
```

packages/react-data-grid/src/index.ts
```diff
@@ -8,6 +8,7 @@ import * as editors from './common/editors';
 import * as _utils from './common/utils';
 import * as _constants from './common/enums';
 import * as _helpers from './helpers';
+export * from './common/types';
```
amanmahajan7: Let's revisit types and figure out what we want to export

PR #1620, opened 2019-05-01
tsify editors

amanmahajan7:


packages/react-data-grid-addons/src/editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,79 @@
...
+interface Props {
+  value?: string | string[] | number;
+  column: Column;
+  options: (string | Option)[];
+  onBlur: React.FocusEventHandler<HTMLSelectElement>;
+}
+
+interface State {
+  value?: string | string[] | number;
+}
```
richardware: This might be better written like this:

```suggestion
interface State {
  value?: string | string[] | number;
}
```

PR #1621, opened 2019-05-02
Remove docs generation scripts

MayhemYDG:
Let's start over with this, some day.


qili26:
Maybe I should leave this for @amanmahajan7 to review. Just curious, is this for the long generated texts whenever we create a new PR that includes description, whether it's bug feature, whether it includes any break changes stuff?

PR #1622, opened 2019-05-02
Fix column types

amanmahajan7:
I will remove `isFrozen` in a follow up PR

### DIFF
--- a/packages/react-data-grid/src/ColumnMetrics.ts
+++ b/packages/react-data-grid/src/ColumnMetrics.ts
@@ -1,10 +1,12 @@
 export { sameColumn } from './ColumnComparer';
-import { getSize, getColumn, isFrozen, spliceColumn } from './ColumnUtils';
+import { getSize, isFrozen } from './ColumnUtils';
 import getScrollbarSize from './getScrollbarSize';
 import { isColumnsImmutable } from './common/utils';
-import { Column, ColumnList, ColumnMetrics } from './common/types';
+import { Column, CalculatedColumn, ColumnList, ColumnMetrics } from './common/types';
 
-type Metrics = Pick<ColumnMetrics, 'columns' | 'totalWidth' | 'minColumnWidth'>;
+type Metrics = Pick<ColumnMetrics, 'totalWidth' | 'minColumnWidth'> & {
+  columns: ColumnList;
+};
 
 function cloneColumns(columns: ColumnList): Column[] {
   if (Array.isArray(columns)) {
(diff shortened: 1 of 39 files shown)


packages/react-data-grid/src/ColumnMetrics.ts
```diff
@@ -38,17 +40,17 @@ function setDefferedColumnWidths(columns: Column[], unallocatedWidth: number, mi
 
 function setColumnOffsets(columns: Column[]): void {
   let left = 0;
-  for (const column of columns) {
+  for (const column of columns as CalculatedColumn[]) {
```
amanmahajan7: We can probably modify this method and return a list of calculated columns instead. wdyt?

packages/react-data-grid/src/ColumnMetrics.ts
```diff
@@ -63,11 +65,11 @@ export function recalculate(metrics: Metrics) {
 
   const frozenColumns = columns.filter(c => isFrozen(c));
   const nonFrozenColumns = columns.filter(c => !isFrozen(c));
-  columns = frozenColumns.concat(nonFrozenColumns);
-  columns.forEach((c, i) => c.idx = i);
+  const calculatedColumns = frozenColumns.concat(nonFrozenColumns) as CalculatedColumn[];
```
amanmahajan7: Then we won't need type casting

packages/react-data-grid/src/ColumnMetrics.ts
MayhemYDG: Yeah, maybe we can merge some together so we can do
```ts
function x(columns: Column[]): CalculatedColumn[] {
  const cc: CalculatedColumn[] = [];

    const left = ...
    const width = ...
    cc.push({ ...column, left, width });
  }

  return cc;
}
```

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -131,11 +130,11 @@ export default class HeaderRow extends React.Component<HeaderRowProps> {
   }
 
   setScrollLeft(scrollLeft: number): void {
-    this.props.columns.forEach((column?: Column) => {
-      const { key } = column!;
+    this.props.columns.forEach(column => {
```
MayhemYDG: We can use for...of here as well.

packages/react-data-grid/src/Row.tsx
```diff
@@ -66,10 +66,9 @@ export default class Row extends React.Component<RowRendererProps> implements Ro
 
   getCells() {
     const { colOverscanStartIdx, colOverscanEndIdx, columns } = this.props;
-    const frozenColumns = columns.filter((c?: Column) => isFrozen(c!)) as Column[];
-    const nonFrozenColumn = columns.slice(colOverscanStartIdx, colOverscanEndIdx + 1).filter((c?: Column) => !isFrozen(c!)) as Column[];
-    return nonFrozenColumn.concat(frozenColumns)
-      .map((c?: Column) => this.getCell(c!));
+    const frozenColumns = columns.filter(c => isFrozen(c));
```
MayhemYDG:
```suggestion
    const frozenColumns = columns.filter(isFrozen);
```

PR #1623, opened 2019-05-02
Convert ClickOutside and EditorPortal to functional components

MayhemYDG:


packages/react-data-grid/src/common/editors/EditorPortal.tsx
```diff
@@ -1,43 +1,24 @@
-import React from 'react';
+import React, { useState, useLayoutEffect } from 'react';
 import ReactDOM from 'react-dom';
 
 interface Props {
   children: React.ReactNode;
-  target: Node;
+  target: Element;
```
MayhemYDG: This better matches the type `createPortal` accepts
image

packages/react-data-grid/src/common/editors/EditorPortal.tsx
```diff
@@ -1,43 +1,24 @@
...
-}
-
-export default class EditorPortal extends React.Component<Props, State> {
+export default function EditorPortal({ target, children }: Props) {
   // Keep track of when the modal element is added to the DOM
-  readonly state: Readonly<State> = {
-    isMounted: false
-  };
-
-  el = document.createElement('div');
```
MayhemYDG: I don't remember for what we reason we used an extra div here. Let me know if I should add it back.

qili26: I believe that's how the instruction was documented when we first use the portal in a class component.

https://reactjs.org/docs/portals.html#event-bubbling-through-portals
see the modal example
image

amanmahajan7: I do not think we need it

PR #1629, opened 2019-05-03
eslint-plugin-react 7.13

MayhemYDG:
https://github.com/yannickcr/eslint-plugin-react/blob/master/CHANGELOG.md

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -324,7 +324,7 @@ const rules = {
   'react/sort-comp': 0,
   'react/sort-prop-types': 0,
   'react/state-in-constructor': 0,
-  'react/static-property-placement': 0,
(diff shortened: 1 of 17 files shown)


.eslintrc.js
```diff
@@ -323,6 +323,8 @@ const rules = {
   'react/self-closing-comp': 1,
   'react/sort-comp': 0,
   'react/sort-prop-types': 0,
+  'react/state-in-constructor': 0,
+  'react/static-property-placement': 0,
```
qili26: Does this mean we don't want to force the static definition in a class at the top? If so, maybe we should change to 1?

MayhemYDG: Enabled now.

PR #1630, opened 2019-05-03
Add generics to FormatterProps

MayhemYDG:


packages/react-data-grid/src/common/types.ts
```diff
@@ -36,16 +36,16 @@ export interface Column<T = unknown> {
   /** Editor to be rendered when cell of column is being edited. If set, then the column is automatically set to be editable */
   editor?: unknown;
```
amanmahajan7: We need to add editor prop as well

```
editor?: React.ReactElement | React.ComponentType<EditorProps<V, DV>>;
```

packages/react-data-grid/src/common/types.ts
```diff
@@ -36,16 +36,16 @@ export interface Column<T = unknown> {
...
   /** Header renderer for each header cell */
-  headerRenderer?: React.ReactElement | React.ComponentType<{ column: CalculatedColumn<T>; rowType: HeaderRowType }>;
+  headerRenderer?: React.ReactElement | React.ComponentType<{ column: CalculatedColumn<V, T>; rowType: HeaderRowType }>;
```
amanmahajan7: May be export HeaderRendererProps
```
export interface HeaderRendererProps<V = unknown, T= unknown> {
  column: CalculatedColumn<V, T>;
  rowType: HeaderRowType ;
}
```

packages/react-data-grid/src/common/types.ts
```diff
@@ -117,17 +117,17 @@ export interface Editor {
...
   isScrolling: boolean;
-  dependentValues?: unknown;
+  dependentValues?: T;
 }
 
 export interface EditorProps<V = unknown, C = unknown> {
-  column: CalculatedColumn<C>;
+  column: CalculatedColumn<V, C>;
   value: V;
   rowMetaData: unknown;
```
amanmahajan7: `rowMetaData` should not be a required field.
https://github.com/adazzle/react-data-grid/blob/next/packages/react-data-grid/src/common/editors/EditorContainer.tsx#L182

amanmahajan7: Also it should be `C` or `DV` (for clarity?)

PR #1631, opened 2019-05-03
Remove support for react-grid-Row--{even,odd}

MayhemYDG:
Users can style even/odd rows using proper CSS selectors.

### DIFF
--- a/packages/react-data-grid/src/Row.tsx
+++ b/packages/react-data-grid/src/Row.tsx
@@ -120,7 +120,6 @@ export default class Row extends React.Component<RowRendererProps> implements Ro
   render() {
     const className = classNames(
       'react-grid-Row',
-      `react-grid-Row--${this.props.idx % 2 === 0 ? 'even' : 'odd'}`,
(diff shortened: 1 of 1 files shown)


qili26:
I believe it's a breaking change, maybe we should add some description to Change.md

MayhemYDG:
We're doing a lot of breaking changes right now, we'll document it all at some point I guess.

qili26:
I am just thinking it's easy to lose track later, maybe in the title we add `Break`, then it will be easier for us to find which merge contains the break changes later, which are only the tsfying?

MayhemYDG:
I want to clean up all the css and prefix every class names with `rdg`, so maybe we'll just have a single line in the changelog to mention that.

qili26:
That would work.

MayhemYDG:
Actually `nth-child()` selectors wouldn't work since we only render the visible rows only.
We can still remove those classes if we don't care to support them I guess.

MayhemYDG:
This is actually used and `:nth-child()` is not a viable alternative, so I'm closing this.

PR #1632, opened 2019-05-03
Simplify Canvas vertical scroll position

MayhemYDG:
- Fewer dom nodes/updates, less code
- I had tried to use `_scroll.scrollTop` for the padding-top, but after a delay it snaps to the topmost row for some reason.
  - I suspect this it linked to the Viewport.
  - probably also linked to why sometimes there's a blank space above the topmost visible row.


packages/react-data-grid/src/Canvas.tsx
```diff
@@ -317,13 +307,11 @@ export default class Canvas extends React.PureComponent<Props> {
...
-    }
-
-    if (rowsCount - rowOverscanEndIdx > 0) {
-      rows.push(this.renderPlaceholder('bottom', (rowsCount - rowOverscanEndIdx) * rowHeight));
-    }
+    const containerStyle: React.CSSProperties = {
+      width: totalColumnWidth,
+      height: rowsCount * rowHeight,
+      paddingTop: rowOverscanStartIdx * rowHeight
+    };
```
qili26: Maybe we should consider adding a small `paddingBottom`? Currently when the grid has lots of rows and scrolled to the bottom and when the user clicks an editable cell, the right bottom corner will show the small square to indicate draggable status. However, this will push all the rows upward a little bit like `5 ~ 10 px` (not sure the exact amount).

MayhemYDG: Maybe we should just place the icon thing on the top of the row instead?

qili26: Is this icon the square I meant? If so, it would be same. If the grid is scrolled all the way to the top, after clicking on the first row of the cell, the rows will be pushed down a little. Maybe we don't worry about this now. Just want to you be aware that we have this small glitch.

MayhemYDG: Yeah.
What I meant is that we can probably set the square thing at the top of the row, if it's the last row.
Otherwise it'll stay at the bottom.

PR #1633, opened 2019-05-03
Simplify canvas style

MayhemYDG:
It doesn't need to know the exact height and width to position itself correctly.
We should also merge Viewport and Canvas together, I don't see a reason to have these two components on their own.


amanmahajan7:
needs catchup

PR #1635, opened 2019-05-04
Setup react-testing-library and code coverage

MayhemYDG:


packages/react-data-grid/README.md
```diff
@@ -1,17 +1,21 @@
-# react-data-grid
-
-> The core of react-data-grid
```
MayhemYDG: Kind of a useless description. 🤷‍♂

PR #1644, opened 2019-05-09
Use bash instead of script for more colorful logs

MayhemYDG:


MayhemYDG:
Nevermind we're already using bash.
https://github.com/Microsoft/azure-pipelines-tasks/issues/8778

PR #1650, opened 2019-05-10
Cleanup unused css

MayhemYDG:
These still need to be sorted, moved into individual stylesheets, and rewritten probably.

### DIFF
--- a/packages/react-data-grid-addons/style/rdg-toolbar.less
+++ b/packages/react-data-grid-addons/style/rdg-toolbar.less
@@ -47,7 +47,6 @@
   margin-left: 5px;
 }
 .react-grid-Toolbar .tools {
-  display: inline-block;
   float: right;
   margin-top: 0;
   position: relative;
--- a/packages/react-data-grid/style/rdg-cell.less
+++ b/packages/react-data-grid/style/rdg-cell.less
@@ -16,9 +16,6 @@
   background-color: #66afe930;
 }
 
-.moving-element {
-  will-change: transform;
-}
 .react-grid-Cell--frozen {
   /* Should have a higher value than 1 to show in front of cell masks */
   z-index: 2;
@@ -131,10 +128,6 @@
     border-color: transparent transparent black transparent;
 }
 
-.react-grid-Canvas.opaque .react-grid-Cell.cell-tooltip:hover .cell-tooltip-text {
-    visibility: hidden;
-}
-
 .rdg-cell-expand {
     float: right;
     display: table;
@@ -200,14 +193,6 @@
   margin-top:-10px;
   margin-left:-10px;
 
-}
-
-.rdg-empty-child-row:hover .glyphicon-plus-sign {
-  color: green;
-}
-
-.rdg-empty-child-row:hover a {
-  color: green;
 }
 
 .rdg-child-row-btn .glyphicon-remove-sign:hover {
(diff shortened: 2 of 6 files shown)


packages/react-data-grid-addons/style/rdg-toolbar.less
```diff
@@ -47,7 +47,6 @@
   margin-left: 5px;
 }
 .react-grid-Toolbar .tools {
-  display: inline-block;
```
MayhemYDG: Ignored since we use float.

packages/react-data-grid/style/rdg-cell.less
```diff
@@ -131,10 +128,6 @@
     border-color: transparent transparent black transparent;
 }
 
-.react-grid-Canvas.opaque .react-grid-Cell.cell-tooltip:hover .cell-tooltip-text {
```
MayhemYDG: We never set any additional class on `.react-grid-Canvas`, so this whole thing can go.

packages/react-data-grid/style/rdg-checkbox.less
```diff
@@ -1,19 +1,19 @@
-.react-grid-checkbox, .radio-custom {
```
MayhemYDG: I don't know what these "custom" classes are.

packages/react-data-grid/style/rdg-interaction-masks.less
```diff
@@ -70,32 +71,31 @@
...
   color: #999999;
 }
 
-.rdg-editor-container input.editor-main::-webkit-input-placeholder, select.editor-main::-webkit-input-placeholder {
+.rdg-editor-container input.editor-main::placeholder,
+select.editor-main::placeholder {
   color: #999999;
 }
 
-.rdg-editor-container input.editor-main[disabled], select.editor-main[disabled], .rdg-editor-container input.editor-main[readonly], select.editor-main[readonly], fieldset[disabled] .rdg-editor-container input.editor-main, fieldset[disabled] select.editor-main {
```
MayhemYDG: >`fieldset[disabled] .rdg-editor-container input.editor-main`

Don't think that's possible since we use portals, right? so I removed that selector.

packages/react-data-grid/style/rdg-interaction-masks.less
```diff
@@ -70,32 +71,31 @@
...
+.rdg-editor-container input.editor-main[readonly],
+select.editor-main[readonly],
+fieldset[disabled] .rdg-editor-container input.editor-main,
+fieldset[disabled] select.editor-main {
   cursor: not-allowed;
   background-color: #eeeeee;
   opacity: 1;
 }
-
-textarea.rdg-editor-container input.editor-main, textareaselect.editor-main {
```
MayhemYDG: >textareaselect

What even is this?

packages/react-data-grid/style/rdg-checkbox.less
amanmahajan7: It does not seem like they were ever used, I also checked master and no references there also

packages/react-data-grid/style/rdg-interaction-masks.less
amanmahajan7: I am not sure why do we even have a `fieldset` selector and with portals it is not even possible. This selector was possibly copied from `bootstrap.css`

amanmahajan7: Also, I am not sure if we event disable the editors

packages/react-data-grid/style/rdg-interaction-masks.less
amanmahajan7: no idea, probably a typo

PR #1652, opened 2019-05-13
Move the props upstream

MayhemYDG:
When trying to figure out what the props are when using RDG, it's a pain to just see a `Pick<>`ed typed.
I think we should make RDG the single source of props, that internal components will then `Pick<>` on.

Top-to-bottom style.

I think it might also help dealing with generics.


packages/react-data-grid/src/Grid.tsx
```diff
@@ -1,30 +1,56 @@
...
 
+type SharedDataGridState = Pick<DataGridState,
+'columnMetrics'
+| 'sortColumn'
+| 'sortDirection'
+>;
 
-export interface GridProps extends Omit<ViewportProps, 'onScroll'>, SharedHeaderProps {
-  emptyRowsView?: React.ComponentType<{}>;
-  scrollLeft?: number;
```
MayhemYDG: Already found an unused prop.

MayhemYDG:
Bonus: all the tsdoc we'll write in `ReactDataGrid.tsx` will be available downstream:

image

amanmahajan7:
Sounds great to me 👍

packages/react-data-grid/src/ReactDataGrid.tsx
```diff
@@ -134,7 +134,7 @@ export interface DataGridProps {
   RowsContainer?: React.ComponentType<RowsContainerProps>;
   emptyRowsView?: React.ComponentType<{}>;
   onHeaderDrop?(): void;
-  getSubRowDetails?(): SubRowDetails;
+  getSubRowDetails?(row: RowData): SubRowDetails;
```
MayhemYDG: `getSubRowDetails` is called with an argument in Canvas...

packages/react-data-grid/src/RowsContainer.tsx
```diff
@@ -1,8 +1,10 @@
+import React from 'react';
+
 export interface RowsContainerProps {
   id: string;
   children: React.ReactNode;
 }
 
 export default function RowsContainer({ children }: RowsContainerProps) {
-  return children;
+  return <>{children}</>;
```
MayhemYDG: TS complains that components must return a `ReactElement`, and a `ReactNode` is not good enough.

packages/react-data-grid/src/ReactDataGrid.tsx
```diff
@@ -377,7 +392,7 @@ export default class ReactDataGrid extends React.Component<Props, State> {
     const fromRowData = rowGetter(action === UpdateActions.COPY_PASTE ? originRow! : fromRow);
     const fromRowId = fromRowData[rowKey];
     const toRowId = rowGetter(toRow)[rowKey];
-    onGridRowsUpdated({ cellKey, fromRow, toRow, fromRowId, toRowId, rowIds, updated, action, fromRowData });
+    onGridRowsUpdated({ cellKey, fromRow, toRow, fromRowId, toRowId, rowIds, updated: updated as never, action, fromRowData });
```
MayhemYDG: This is bad.

To get the proper type on `onGridRowsUpdated` we need to add a generic on the Props object/RDG component so we can set a map of `{ [column key]: column value type }`, used like this:
```tsx
<ReactDataGrid<{ colKey: string; }> columns={[]} />
```
, but then we'd have to set a generic everywhere.

This'll do for now I think.

packages/react-data-grid/src/ChildRowDeleteButton.tsx
```diff
@@ -1,29 +1,20 @@
...
-import classNames from 'classnames';
 
 export interface ChildRowDeleteButtonProps {
   treeDepth: number;
   cellHeight: number;
-  siblingIndex: number;
-  numberSiblings: number;
   onDeleteSubRow(): void;
   isDeleteSubRowEnabled: boolean;
-  allowAddChildRow?: boolean;
```
MayhemYDG: This prop was never set.
It defaults to true, so only one of those classes was ever used, so I could clean up some css as well.
Then `siblingIndex` and `numberSiblings` were unused, so I could clean those up as well.

We don't use `subRowDetails.{siblingIndex,numberSiblings,canExpand}` internally, but we might externally, so I didn't clean those.

packages/react-data-grid/src/Cell.tsx
```diff
@@ -199,36 +199,40 @@ export default class Cell extends React.PureComponent<Props> implements CellRend
...
-      isScrolling,
-      column,
+  getFormatterProps(): FormatterProps<unknown> {
+    return {
+      rowIdx: this.props.rowIdx,
+      value: this.props.value,
+      isScrolling: this.props.isScrolling,
+      column: this.props.column,
       row: this.getRowData(),
       dependentValues: this.getFormatterDependencies()
```
MayhemYDG: This would log a warning for every single cell if `getRowMetaData` was truthy, even though it's not used with the SimpleCellFormatter.

packages/react-data-grid/src/ReactDataGrid.tsx
amanmahajan7: Do you think we can just set it to `Record<string, unknown>` and let the parent component handle it. It makes sense because we do not know what editor is going to return and typing it makes grid more complicated

PR #1659, opened 2019-05-17
typescript 3.5

MayhemYDG:
https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/


packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -117,7 +117,7 @@ export default class EditorContainer extends React.Component<Props, State> {
 
     return (
       <SimpleTextEditor
-        ref={this.editor as React.RefObject<SimpleTextEditor>}
+        ref={this.editor as unknown as React.RefObject<SimpleTextEditor>}
```
MayhemYDG: image

Even though `input` is private... I wonder if this'll get fixed in the stable release.

PR #1661, opened 2019-05-17
tsify more addons modules

MayhemYDG:


packages/react-data-grid-addons/src/formatters/ImageFormatter.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+interface Props {
+  /** image url, used as background-image */
+  value: string;
+}
+
+export default function ImageFormatter({ value }: Props) {
+  return <div className="react-grid-image" style={{ backgroundImage: `url(${value})` }} />;
```
MayhemYDG: The previous `ImageFormatter` implementation loaded the url in the background before applying the style, I don't know why, doesn't seem worthwhile.

packages/react-data-grid-addons/src/menu/ContextMenu.js
```diff
@@ -1,18 +0,0 @@
...
-import { ContextMenu } from 'react-contextmenu';
-
-export default class ReactDataGridContextMenu extends React.Component {
-  static propTypes = {
-    children: PropTypes.node
-  };
-
-  render() {
-    return (
-      <ContextMenu identifier="reactDataGridContextMenu">
```
MayhemYDG: Bug uncovered: `identifier` is not a valid prop, `id` is though.

packages/react-data-grid-addons/src/cells/headerCells/filters/NumericFilter.tsx
```diff
@@ -0,0 +1,148 @@
...
+  /** Validates the input */
+  function handleKeyPress(event: React.KeyboardEvent<HTMLInputElement>) {
+    const result = /[><,0-9-]/.test(event.key);
+    if (result === false) {
+      event.preventDefault();
+    }
+  }
+
+  function handleChange(event: React.ChangeEvent<HTMLInputElement>) {
+    const value = event.target.value;
```
amanmahajan7: `const { value } = e.target`

packages/react-data-grid-addons/src/cells/headerCells/filters/NumericFilter.tsx
```diff
@@ -0,0 +1,148 @@
...
+    cursor: 'help'
+  };
+
+  const tooltipText = 'Input Methods: Range (x-y), Greater Then (>x), Less Then (<y)';
+
+  return (
+    <div>
+      <div style={columnStyle}>
+        <input
+          key={inputKey}
```
amanmahajan7: missing `type="text"`

packages/react-data-grid-addons/src/cells/headerCells/filters/NumericFilter.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+
+function filterValues(row: RowData, columnFilter: { filterTerm: { [key in string]: Rule } }, columnKey: string) {
+  if (columnFilter.filterTerm == null) {
+    return true;
+  }
+
+  // implement default filter logic
+  const value = parseInt(row[columnKey] as string, 10);
+  for (const ruleKey in columnFilter.filterTerm) {
```
amanmahajan7: `for (const rule of columnFilter.filterTerm)`

packages/react-data-grid-addons/src/toolbars/Toolbar.tsx
```diff
@@ -0,0 +1,40 @@
...
+            {props.filterRowsButtonText}
+          </button>
+        )}
+        {props.children}
+      </div>
+    </div>
+  );
+}
+
+Toolbar.defaultProps = {
```
amanmahajan7: > For stateless function components (SFCs) use ES2015 default initializers for SFCs:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html

packages/react-data-grid-addons/src/cells/headerCells/filters/NumericFilter.tsx
MayhemYDG: It's not an iterable.

packages/react-data-grid-addons/src/cells/headerCells/filters/NumericFilter.tsx
MayhemYDG: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

>If this attribute is not specified, the default type adopted is `text`.

I can add it, but it's not necessary. wdyt?

packages/react-data-grid-addons/src/cells/headerCells/filters/NumericFilter.tsx
amanmahajan7: Ah, I got confused with `filterTerm: Rule[] | null;`

packages/react-data-grid-addons/src/cells/headerCells/filters/NumericFilter.tsx
amanmahajan7: I usually prefer to add it to be explicit but it is fine in this case

PR #1672, opened 2019-05-30
Refactor Cell component

amanmahajan7:
This PR mainly extracts smaller components and does not change any functionality.


packages/react-data-grid/src/Cell/CellAction.tsx
```diff
@@ -10,12 +10,17 @@ export interface CellActionProps extends CellActionButton {
...
+    return classNames('rdg-cell-action', {
+      'rdg-cell-action-last': isFirst
+    });
+  }, [isFirst]);
+
+  const actionButtonClasses = useMemo(() => {
+    return classNames('rdg-cell-action-button', {
+      'rdg-cell-action-button-toggled': isOpen
+    });
+  }, [isOpen]);
```
MayhemYDG: If you _really_ want to optimize these you could just do
```ts
const actionButtonClasses = isOpen
  ? 'rdg-cell-action-button rdg-cell-action-button-toggled'
  : 'rdg-cell-action-button';
```
It might be worth it for simple cases like this one. 🤔 wdyt?

packages/react-data-grid/src/Cell/CellValue.tsx
```diff
@@ -0,0 +1,54 @@
...
+type CellValueProps = Pick<CellContentProps,
+'rowIdx'
+| 'rowData'
+| 'column'
+| 'value'
+| 'isScrolling'
+>;
+
+export default function CellValue({ rowIdx, rowData, column, value, isScrolling }: CellValueProps) {
+  function getRowData() {
```
MayhemYDG: `getRowData` is called within `getFormatterProps` and `getFormatterDependencies`, let's just inline it in `getFormatterProps`, and pass the data to `getFormatterDependencies`, so we don't do the same work twice.

amanmahajan7: Done

packages/react-data-grid/src/Cell/CellAction.tsx
amanmahajan7: I am okay with your suggestion but for now I ended up removing this optimization. I will add it back if needed after we profile the grid. I do remember `classNames` was taking some time and this optimization will be useful for Cell components

PR #1681, opened 2019-06-06
Update testing-library, sonarjs

MayhemYDG:


piehouserat:
hey guys, looks like progress on the next release has paused for a while, is there any kind of ETA on the next major version? It looks like a lot of really nice work has been done (ts conversion, refactoring, styling, etc)

amanmahajan7:
@piehouserat we do not have an ETA at the moment but the alpha version is pretty stable and we have not introduced any major breaking changes. If you prefer then you can start using it now. We will publish a changelog soon

piehouserat:
@amanmahajan7 I'll take the alpha for a spin today to see what benefits it brings us right now and feed back if you guys want. It would be good to see some kind of changelog and/or planned features for v7. Something that would be of use to us would be auto column sizing where if the width of all columns does not overflow the container then the columns should stretch to fill the empty space. Is this something you guys plan to support in v7?

PR #1705, opened 2019-07-08
Update packages

amanmahajan7:


package.json
```diff
@@ -56,13 +56,13 @@
...
-    "react-router-dom": "^5.0.0",
+    "react-router-dom": "^5.0.1",
     "recharts": "^1.6.2",
     "style-loader": "^0.23.1",
     "terser-webpack-plugin": "^1.3.0",
     "ts-jest": "^24.0.1",
-    "typescript": "~3.5.1",
-    "webpack": "^4.32.2",
-    "webpack-dev-server": "^3.4.1"
+    "typescript": "^3.5.2",
```
MayhemYDG: Should keep the `~` range.

amanmahajan7: I think we can keep `^` for now. I will add `~` back if we run into any issues.

PR #1727, opened 2019-07-29
Add row type

amanmahajan7:


packages/react-data-grid/src/Cell/CellValue.tsx
```diff
@@ -25,16 +25,14 @@ export default function CellValue({ rowIdx, rowData, column, value, isScrolling
     }
   }
 
-  function getFormatterProps(): FormatterProps<unknown> {
-    const row = typeof rowData.toJSON === 'function' ? rowData.toJSON() : rowData;
```
amanmahajan7: I do not think this is used anywhere so I removed it

packages/react-data-grid/src/RowGroup.tsx
```diff
@@ -27,10 +28,10 @@ interface Props {
   name: string;
   renderer?: React.ComponentType;
   eventBus: EventBus;
-  renderBaseRow(p: RowRendererProps): React.ReactElement;
+  renderBaseRow(p: RowRendererProps<R>): React.ReactElement;
 }
 
-const RowGroup = forwardRef<HTMLDivElement, Props>(function RowGroup(props, ref) {
+const RowGroup = forwardRef<HTMLDivElement, Props<any>>(function RowGroup(props, ref) {
```
amanmahajan7: I do not think there is a way to specify generic parameters for forwardRef components. I will revisit this

packages/react-data-grid/src/RowUtils.ts
```diff
@@ -1,8 +1,8 @@
 import { RowData } from './common/types';
 
-export function get(row: RowData, property: string) {
-  if (typeof row.get === 'function') {
-    return row.get(property);
+export function get<R extends {}>(row: R, property: keyof R) {
+  if (typeof (row as RowData).get === 'function') {
```
amanmahajan7: I do not think we have a use case where `row.get` is a function. Let's test this and remove this api

packages/react-data-grid/src/common/types.ts
```diff
@@ -34,52 +36,52 @@ export interface Column<V = unknown, DV = unknown> {
   /** Sets the column sort order to be descending instead of ascending the first time the column is sorted */
   sortDescendingFirst?: boolean;
   /** Editor to be rendered when cell of column is being edited. If set, then the column is automatically set to be editable */
-  editor?: React.ReactElement | React.ComponentType<EditorProps<V, DV>>;
+  editor?: React.ReactElement | React.ComponentType<EditorProps<R, V, DV>>;
```
amanmahajan7: I think we may be able to provide better types for editors and formatters. It can probably be `typeof R[key]`

PR #1730, opened 2019-08-06
Add value type

amanmahajan7:
I will fix the following in a future PR
- Fix key type 
- Fix value type in Editors and Formatters. Need to finalize how to handle it internally

### DIFF
--- a/packages/react-data-grid/src/ColumnMetrics.ts
+++ b/packages/react-data-grid/src/ColumnMetrics.ts
@@ -114,8 +114,8 @@ function compareEachColumn<R>(prevColumns: ColumnList<R>, nextColumns: ColumnLis
 
   for (const key of keys) {
     if (!prevColumnsMap.has(key) || !nextColumnsMap.has(key)) return false;
-    const prevColumn = prevColumnsMap.get(key)!;
-    const nextColumn = nextColumnsMap.get(key)!;
+    const prevColumn = prevColumnsMap.get(key) as Column<R>;
+    const nextColumn = nextColumnsMap.get(key) as Column<R>;
     if (!isSameColumn(prevColumn, nextColumn)) return false;
   }
 
--- a/packages/react-data-grid/src/HeaderCell.tsx
+++ b/packages/react-data-grid/src/HeaderCell.tsx
@@ -3,15 +3,15 @@ import classNames from 'classnames';
 import { isElement } from 'react-is';
 import { isFrozen } from './ColumnUtils';
 import { HeaderRowType } from './common/enums';
-import { CalculatedColumn } from './common/types';
+import { CalculatedColumn, HeaderRowProps } from './common/types';
 
-function SimpleCellRenderer<R>({ column, rowType }: { column: CalculatedColumn<R>; rowType: HeaderRowType }) {
+function SimpleCellRenderer<R>({ column, rowType }: HeaderRowProps<R>) {
   const headerText = rowType === HeaderRowType.HEADER ? column.name : '';
   return <div>{headerText}</div>;
 }
 
 interface Props<R> {
-  renderer?: CalculatedColumn<R>['headerRenderer'];
+  renderer?: React.ReactElement | React.ComponentType<HeaderRowProps<R>>;
   column: CalculatedColumn<R>;
   rowType: HeaderRowType;
   height: number;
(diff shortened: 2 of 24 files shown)


packages/react-data-grid/src/ColumnMetrics.ts
```diff
@@ -114,8 +114,8 @@ function compareEachColumn<R>(prevColumns: ColumnList<R>, nextColumns: ColumnLis
 
   for (const key of keys) {
     if (!prevColumnsMap.has(key) || !nextColumnsMap.has(key)) return false;
-    const prevColumn = prevColumnsMap.get(key)!;
-    const nextColumn = nextColumnsMap.get(key)!;
+    const prevColumn = prevColumnsMap.get(key) as Column<R>;
```
amanmahajan7: `!` turns `Column<R>` to `NonNullable<Column<R>>` so had to explicitly cast it

packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -3,15 +3,15 @@ import classNames from 'classnames';
...
 
-function SimpleCellRenderer<R>({ column, rowType }: { column: CalculatedColumn<R>; rowType: HeaderRowType }) {
+function SimpleCellRenderer<R>({ column, rowType }: HeaderRowProps<R>) {
   const headerText = rowType === HeaderRowType.HEADER ? column.name : '';
   return <div>{headerText}</div>;
 }
 
 interface Props<R> {
-  renderer?: CalculatedColumn<R>['headerRenderer'];
+  renderer?: React.ReactElement | React.ComponentType<HeaderRowProps<R>>;
```
amanmahajan7: `CalculatedColumn<R>['headerRenderer']` was failing. Need to revisit it

packages/react-data-grid/src/common/types.ts
```diff
@@ -7,11 +7,11 @@ export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
 export type SelectedRow<TRow> = TRow & { isSelected: boolean };
 
-export interface Column<TRow, TValue = unknown, TDependentValue = unknown> {
+interface ColumnValue<TRow, TDependentValue = any, TField extends keyof TRow = keyof TRow> {
   /** The name of the column. By default it will be displayed in the header cell */
   name: string;
   /** A unique key to distinguish each column */
-  key: keyof TRow;
+  key: TField;
```
amanmahajan7: This makes value strongly typed

image

packages/react-data-grid/src/common/types.ts
```diff
@@ -120,17 +124,17 @@ export interface Editor<TValue = never> extends React.Component {
   readonly disableContainerStyles?: boolean;
 }
 
-export interface FormatterProps<TValue, TDependentValue = unknown, TRow = any> {
+export interface FormatterProps<TValue, TDependentValue = any, TRow = any> {
```
amanmahajan7: I put `any` so we can create generic formatters. `unknown` might work, I will have to check

packages/react-data-grid/src/Row.tsx
```diff
@@ -62,7 +62,7 @@ export default class Row<R extends {}> extends React.Component<RowRendererProps<
       lastFrozenColumnIndex
     };
 
-    return <Renderer key={`${key}-${idx}`} {...cellProps} />;
+    return <Renderer key={`${key as keyof R}-${idx}`} {...cellProps} />; // FIXEME: fix key type
```
MayhemYDG:
```suggestion
    return <Renderer key={`${key as keyof R}-${idx}`} {...cellProps} />; // FIXME: fix key type
```
Might miss this one when searching for `fixme`.

amanmahajan7:
I think this change also highlights some of the issues with RDG, for example

```
value: this.getCellValue(key || String(column.idx));
```

There is no way to type this. I think we can just use `rowData[key]`

```
<SimpleCellFormatter value={value as string} />
```

We should probably change `SimpleCellFormatter` and change `value` from `string | number | boolean` -> `unknown` as we use this formatter for any column

Regarding keys, because we use `keyof R` type for `column.key` instead of a string so we get this type. I am not too sure how to fix it yet.

image. 

Using strongly type value, formatters, and editors did surfaced quite a few type issues in our code base, I think we can try these types and revisit for improvement.

PR #1751, opened 2019-09-13
bug fix: add columns order checking

qili26:
When columns are reordered, and the columns are simple type objects, dragging columns would fail to update to the `Grid`


amanmahajan7:
I think this can be closed as we fixed the column metrics logic

PR #1757, opened 2019-09-27
Ql scroll sync

qili26:
split scroll bar POC


qili26:
@amanmahajan7 check this POC out.

PR #1762, opened 2019-10-02
Re-implement row selection

amanmahajan7:
- [x]  Remove legacy row selection logic
- [x] Fix select all/deselect all 
- [x] Use props to control selected rows
- [x] Support custom cell/header renderer for selection column
- [x] Support shift click
- [x] Support row selection without checkbox
- [x] Update examples
<del>- [ ] Fir row dragging</del> Remove re-ordering example and re-implement if needed


packages/react-data-grid/src/ReactDataGrid.tsx
```diff
@@ -219,194 +207,80 @@ const ReactDataGridBase = forwardRef(function ReactDataGrid<R extends {}>({
...
-          (rowData as SelectedRow<R>).isSelected = true;
-          newSelectedRows.push(rowData as SelectedRow<R>);
-        }
-        setSelectedRows(newSelectedRows);
-        if (onRowSelect) {
-          onRowSelect(newSelectedRows.filter(r => r.isSelected === true));
-        }
+  const gridColumns = useMemo(() => {
+    // TODO: handle row selection without checkbox, i.e using row click
+    if (!rowSelection || !rowSelection.showCheckbox) return columns;
```
amanmahajan7: Need to figure out how to optimize this block
- `rowSelection` is an object so it will always be different.
- it is possible that `rowGetter` and `onSelectedRowsChange` and created on every render so how should we handle it?
https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often

MayhemYDG: - I don't like object props, we should just have `selectedRows?: Set` and `onSelectedRowsChange?(): void` and so on.
- is generating the `gridColumns` a heavy computation at all?

amanmahajan7: 1) Let's remove the object prop 
2) Probably not but changing columns may trigger other unnecessary updates

examples/scripts/example13-all-features.js
```diff
@@ -193,19 +186,24 @@ class Example extends React.Component {
...
+        columns={this.columns}
         rowGetter={this.getRowAt}
         rowsCount={this.getSize()}
         onGridRowsUpdated={this.handleGridRowsUpdated}
         toolbar={<Toolbar onAddRow={this.handleAddRow} />}
-        enableRowSelect
         rowHeight={50}
         minHeight={600}
+        selectedRows={this.state.selectedRows}
+        onSelectedRowsChange={this.onSelectedRowsChange}
```
qili26: This is a new API. We should mark it and add docs later.

PR #1763, opened 2019-10-02
Bind scrollToColumn method

amanmahajan7:


packages/react-data-grid/src/Canvas.tsx
```diff
@@ -126,7 +126,7 @@ export default class Canvas<R> extends React.PureComponent<CanvasProps<R>> {
     );
   }
 
-  scrollToColumn(idx: number) {
+  scrollToColumn = (idx: number) => {
```
amanmahajan7: We add this method to the event bus so it needs to be bound
https://github.com/adazzle/react-data-grid/pull/1763/files#diff-61ebc5f10f9ee6875894862bf0d64cd6R79

PR #1766, opened 2019-10-04
Improve scrolling responsiveness

MayhemYDG:
⚠ This does not exactly improve scroll _performance/rendering_ ⚠

What this does is it decouples scrolling the canvas from the JS thread, so even if a `'scroll'` event handler is still executing, it won't stop the browser (Chrome at least) from scrolling the element.

That makes it extremely responsive, but can lead to rendering artifacts, like cells not being rendered yet if you scroll quickly, or the header row being slightly out of sync, although I don't think you will notice it in practice, unless you try hard.

Since we now use `position: sticky` for frozen cells, they won't be de-synced when scrolling horizontally!

## Promoting the canvas to a new render layer fixes this issue:

image

You can see this by enabling this setting:
image

### Before, when holding the down arrow on the scrollbar:

image

### After, when holding the down arrow on the scrollbar:

image

Repainting only happens when cells get added/removed now.

You can check this by enabling this setting:
image

## I get a very stable ~140fps now when pressing the scrollbar arrows! 🎉

image

## 🚨❗ Although we can now safely reach stable >60fps, this doesn't mean React will render at 60 fps as well!

If you scroll very quickly, you'll see that React can struggle to catch up quickly enough:

image

This is a very minor artifact which you won't really notice in practice I believe, so I think it's very much worth it IMO.

Another caveat is that these scroll handles aren't decoupled from the JS thread for some reason in Chrome, but everything else is afaict: scrollbar arrows, mouse wheel, and so on.
image
They are perfectly decoupled in Edge (the non Chromium version) though.


MayhemYDG:
Another small caveat: the `InteractionMasks` stuff is not perfectly in sync when scrolling horizontally.

packages/react-data-grid/style/rdg-core.less
```diff
@@ -17,4 +17,13 @@
   overflow-x: auto;
   overflow-y: scroll;
   background-color: #fff;
+  // will-change promotes the canvas to a new layer, improving scroll responsiveness
+  will-change: transform;
```
qili26: This seems like a nuclear weapon. 🤔 

https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

MayhemYDG: https://developers.google.com/web/fundamentals/performance/rendering/stick-to-compositor-only-properties-and-manage-layer-count

amanmahajan7: Do we need `will-change` on both `canvas` and `rdg-rows-container`?

MayhemYDG:
@amanmahajan7 
The `will-change` on the rows container reduces repainting (see screenshots above). It looks like we can fix the interaction mask bug while keeping the `will-change` if we move `InteractionMasks` under the rows container div, and it seems to work fine at first glance, though I don't know if that will break anything else.

Cells are not shaky here, I see that you've zoomed out in that tab, maybe reset the zoom level?

amanmahajan7:
I will check `InteractionMasks`, I am not sure if it is safe but I do not see any reason why it won't be. Let's wait before making this change. 

You are right, I am zoomed out. It works fine now 👍

PR #1768, opened 2019-10-04
Remove sortColumn/sortDirection states

amanmahajan7:


packages/react-data-grid/src/ReactDataGrid.tsx
```diff
@@ -424,9 +414,9 @@ const ReactDataGridBase = forwardRef(function ReactDataGrid<R extends {}>({
           selectedRows={selectedRows}
           onSelectedRowsChange={onSelectedRowsChange}
           rowOffsetHeight={rowOffsetHeight}
-          sortColumn={sortColumn}
-          sortDirection={sortDirection}
-          onSort={handleSort}
+          sortColumn={props.sortColumn}
+          sortDirection={props.sortDirection}
+          onSort={props.onGridSort}
```
qili26: Do we need to print error message if the `sortColumn`, `sortDirection`, `onGridSort` are not used together?

amanmahajan7: We should but may be a in a different PR. I did discuss to Nico about the same thing.

PR #1759, opened 2019-10-01
Totals row

amanmahajan7:
- [x] Add pinnedRows to `ReactDataGrid.tsx` and rows sticky container
- [x] Add Special logic for IE / IE Edge
- [x] Add isBottomPinned to `Row.tsx` / `Cell.tsx`

pinned rows demo

PR #1773, opened 2019-10-07
Summary rows

qili26:
- [x] Add `summaryRows` to `ReactDataGrid.tsx` and rows sticky container
- [x] Add Special logic for IE / IE Edge
- [x] Add `isSummaryRow` to `Row.tsx` / `Cell.tsx`
- [ ] unit test


examples/assets/css/main.css
```diff
@@ -567,4 +560,13 @@ code.tab3{
 }
 code.tab4{
 	margin-left:12px;
-}
\ No newline at end of file
+}
+
+.rdg-bottom-pinned-rows-sticky-wrapper .react-grid-checkbox-container {
```
qili26: For example only.

examples/scripts/example32-total-rows.js
```diff
@@ -0,0 +1,167 @@
...
+    key: 'cost',
+    name: 'Resource cost (USD)',
+    width: 200,
+    formatter: ({ value }) => <div>{ value.toLocaleString('en-US') }</div>
+  },
+  {
+    key: 'hours',
+    name: '# of working hours',
+    width: 200,
+    formatter: ({ value }) => <div>{ value.toLocaleString('en-US', { maximumFractionDigits: 0 }) }</div>
```
qili26: Need to display the formatting like `1,000`

examples/scripts/example32-total-rows.js
```diff
@@ -0,0 +1,167 @@
...
+
+  const rows = useMemo(() => {
+    const rows = [];
+
+    for (let i = 1; i < 1000; i++) {
+      rows.push({
+        id: i,
+        task: `Task ${i}`,
+        complete: Math.min(100, Math.round(Math.random() * 110)),
+        priority: ['Critical', 'High', 'Medium', 'Low'][Math.floor(Math.random() * 4)],
```
qili26: `Array[0 ~ 4]`

packages/react-data-grid/src/ReactDataGrid.tsx
```diff
@@ -153,6 +153,12 @@ export interface ReactDataGridProps<R extends {}> {
    * Note that using this parameter will result in an additional render call after scrolling has stopped (when isScrolling changes from true to false).
    */
   enableIsScrolling?: boolean;
+  /**
+   * Rows to be pinned at the bottom of the rows view, the vertical scroll bar will not scroll these rows.
+   * The rows can be used as total rows to display some summary information.
+   * Bottom horizontal scroll bar can move the row left / right. Or a customized row renderer can be used to disabled the scrolling support.
+   */
+  pinnedRows?: R[];
```
qili26: maybe we should use `bottomPinnedRows`?

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -343,8 +355,27 @@ export default function Canvas<R>({
...
-  const paddingBottom = (rowsCount - 1 - rowOverscanEndIdx) * rowHeight;
+  const pinnedRowsHeight = isIEOrEdge && pinnedRows ? pinnedRows.length * rowHeight : 0;
+  const paddingTop = rowOverscanStartIdx > 0 ? rowOverscanStartIdx * rowHeight : 0;
+  const paddingBottom = rowsCount - rowOverscanEndIdx > 0 ? (rowsCount - 1 - rowOverscanEndIdx) * rowHeight + pinnedRowsHeight : pinnedRowsHeight;
+  const { totalColumnWidth: width } = columnMetrics;
+
+  // Set minHeight to show horizontal scrollbar when there are no rows
+  const scrollableRowsWrapperStyle: React.CSSProperties = { width, paddingTop, paddingBottom };
+  const pinnedRowsWrapperStyle: React.CSSProperties = { width };
+  let ieStickyWrapperStyle: React.CSSProperties | undefined;
```
qili26: Any better ideas for IE specific codes?

qili26:
supersedes https://github.com/adazzle/react-data-grid/pull/1759

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -343,8 +355,27 @@ export default function Canvas<R>({
     );
   }
 
-  const paddingTop = rowOverscanStartIdx * rowHeight;
-  const paddingBottom = (rowsCount - 1 - rowOverscanEndIdx) * rowHeight;
+  const pinnedRowsHeight = isIEOrEdge && pinnedRows ? pinnedRows.length * rowHeight : 0;
+  const paddingTop = rowOverscanStartIdx > 0 ? rowOverscanStartIdx * rowHeight : 0;
```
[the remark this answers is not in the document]
qili26: `rowOverscanStartIdx` will be guaranteed to be a positive integer, right?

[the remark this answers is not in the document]
MayhemYDG: Yeah it's 0 minimum.

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -241,7 +272,8 @@ export default function Canvas<R>({
       colOverscanEndIdx,
       lastFrozenColumnIndex: columnMetrics.lastFrozenColumnIndex,
       isScrolling,
-      scrollLeft
+      scrollLeft,
+      isSummaryRow: summaryRows && summaryRows.includes(row)
```
amanmahajan7: This will always be false, correct?

packages/react-data-grid/src/Cell.tsx
```diff
@@ -159,31 +160,96 @@ export default class Cell<R> extends React.Component<CellProps<R>> implements Ce
...
+                  column={column}
+                  value={value}
+                  isScrolling={isScrolling}
+                  isRowSelected={this.props.isRowSelected}
+                  onRowSelectionChange={this.props.onRowSelectionChange}
+                  isSummaryRow={isSummaryRow}
+                />
+              </span>
+            </div>
+            {tooltip && <span className="cell-tooltip-text">{tooltip}</span>}
```
amanmahajan7: I think we should deprecate `tooltip` prop. `Row/Cell` renderers must be used to set it and it makes using grid difficult IMO. We should just set the tooltip in the formatter. We do not have to remove `tooltip` prop in this PR but we can remove it from `summaryRow`. MediaPlan extensively use this prop so we have to keep in on the regular row for now

packages/react-data-grid/style/rdg-row.less
```diff
@@ -73,3 +73,19 @@
...
+
+@supports (position: sticky) or (position: -webkit-sticky) {
+  .rdg-bottom-pinned-rows-sticky-wrapper {
+    position: -webkit-sticky;
+    position: sticky;
+    bottom: 0;
+  }
+}
+
+.rdg-bottom-pinned-rows {
```
amanmahajan7: May be we can usse `rdg-summary-*`  classnames to match the prop name

packages/react-data-grid/src/Cell.tsx
```diff
@@ -159,31 +160,96 @@ export default class Cell<R> extends React.Component<CellProps<R>> implements Ce
...
+    if (isSummaryRow) {
+      return (
+        <div
+          ref={this.cell}
+          className={cellClassName}
+          style={style}
+        >
+          <div className={cellValueClassName}>
+            <div className="react-grid-Cell__container">
+              <span>
```
amanmahajan7: We can remove the span?

packages/react-data-grid/src/Cell.tsx
```diff
@@ -159,31 +160,96 @@ export default class Cell<R> extends React.Component<CellProps<R>> implements Ce
...
+                <CellValue<R>
```
amanmahajan7: We can do something like 

```
const cellValue = (<CellValue<R> ... />);


if (isSummaryRow) {
 ....
   <div className="react-grid-Cell__container">
      {cellValue}
   </div>
```

amanmahajan7: Now I think about it, I think it is better to use `CellValue` for summary rows instead of `CellContent`. This way we do not have to move the `Expand/Collapse` functionality here

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -261,6 +293,16 @@ export default function Canvas<R>({
     return <Row<R> {...rendererProps} />;
   }
 
+  function getRowRef(row: R, idx: number) {
```
amanmahajan7: Summary rows `ref` will never reset from what I can tell. Consider a case
1) 10 rows + 2 summary rows are renderered
2) rowCount is changed to 30. This will override summary row refs

I am quite sure this will happen but we can double check. We may have to use 2 maps 

> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts.

examples/demos/example32-summary-rows.js
```diff
@@ -0,0 +1,167 @@
+import React, { useCallback, useMemo, useState } from 'react';
+import ReactDataGrid, { SelectColumn } from 'react-data-grid';
+import Wrapper from './Wrapper';
+
+function DecimalFormatter({ value, isSummaryRow, maximumFractionDigits }) {
+  const text = value.toLocaleString('en-US', maximumFractionDigits == null ? { maximumFractionDigits: 2 } : { maximumFractionDigits: 0 });
+  return isSummaryRow ? <strong>{text}</strong> : <div>{text}</div>;
```
amanmahajan7: @MayhemYDG thoughts on the this API where we pass `isSummaryRow ` prop to `Row/Cell/Formatters`? Another option is to have a dedicated `column.GroupFormatter`

MayhemYDG: I'm fine with it.

packages/react-data-grid/src/Canvas.tsx
qili26: It will always be `false` only when `summaryRows` doesn't exist. Same story like `isRowSelected`, if it's not enabled, it will just be false and populated to next levels.

packages/react-data-grid/src/Cell.tsx
qili26: I believe we need to update some styles as well. (we should do this in another PR). Currently if we set the tooltip in the formatter, the tooltip will only display when the formatter area is hovered but not the whole cell.

packages/react-data-grid/style/rdg-row.less
qili26: I will change this.

packages/react-data-grid/src/Cell.tsx
qili26: I want to remove the span, but I am not sure if we should do it in this PR? Maybe in another cell refactor PR?

packages/react-data-grid/src/Canvas.tsx
qili26: you are right, I will use another way for the summary rows ref.

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -291,8 +330,27 @@ export default function Canvas<R>({
...
+  let ieStickyWrapperStyle: React.CSSProperties | undefined;
+
+  if (isIEOrEdge) {
+    summaryRowsWrapperStyle.position = 'absolute';
+    summaryRowsWrapperStyle.bottom = 0;
+
+    ieStickyWrapperStyle = {
+      position: 'absolute',
+      top: 0,
+      height: height - 17, // IE scroll bar width
```
amanmahajan7: We should use `getScrollbarSize`

packages/react-data-grid/src/Cell.tsx
amanmahajan7: I was referring to that we should remove `{tooltip && <span className="cell-tooltip-text">{tooltip}</span>}` from this block. We do not need to handle tooltip on the totals row

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -171,35 +177,64 @@ export default function Canvas<R>({
...
       colOverscanEndIdx,
-      lastFrozenColumnIndex: columnMetrics.lastFrozenColumnIndex,
-      scrollLeft
+      lastFrozenColumnIndex,
+      scrollLeft,
+      isSummaryRow: false
+    };
+  }
+
+  function renderSummaryRow(row: R, idx: number) {
```
amanmahajan7: This component is becoming little complex to understand, I will check if we can cleanup somehow

amanmahajan7: https://github.com/adazzle/react-data-grid/pull/1787 should address it

qili26:
After moving the summary rows to the bottom. IE EDGE is rendering like this. Don't know why this happens, but it seems like the `position sticky` is just buggy in EDGE.

image

examples/index.html
```diff
@@ -46,6 +46,7 @@
     .example {
       overflow: auto;
       flex-grow: 1;
+      margin: 0 8px;
```
qili26: added this margin so that I can see the RDG border clearly.

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -275,47 +313,63 @@ export default function Canvas<R>({
...
+        />
+        <RowsContainer id={contextMenu ? contextMenu.props.id : 'rowsContainer'}>
+          <div className="rdg-rows-container" style={canvasRowsContainerStyle}>
+            {getRows()}
+          </div>
+        </RowsContainer>
+      </div>
+      {summaryRows && summaryRows.length && (
+        <div className="rdg-summary">
+          <div ref={summary} style={boundaryStyle}>
```
qili26: This is to mimic canvas horizontal scrolling so that for Summary it can just reuse the existing `Row.tsx`.
1. The container is fixed width. so that when this `div` is set the `scrollLeft`, the rows container (with full width, the next line) could be moving around.
2. For sticky frozen divs, we do nothing. For IE 11 specifically, we need to set the `scrollLeft` to each frozen cells. (This logic is just same as what we had)


https://github.com/w3c/csswg-drafts/issues/865
An insteresting finding is that the reason why previously `Header` sticky doesn't work for EDGE is because we set the `.react-grid-HeaderRow` as `overflow: hidden`, it will break the cell position sticky only in EDGE. However, the `overflow: hidden` like the structure of canvas and summary area would work.

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -21,6 +21,7 @@ type SharedDataGridProps<R, K extends keyof R> = Pick<ReactDataGridProps<R, K>,
 | 'RowsContainer'
 | 'getSubRowDetails'
 | 'selectedRows'
+| 'summaryRows'
```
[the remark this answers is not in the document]
MayhemYDG: Circular imports are fine, type imports will be erased in the resulting JS, and proper es module implementations should support circular imports.
https://twitter.com/awbjs/status/1160933193320525825

packages/react-data-grid/src/SummaryRowRenderer.tsx
```diff
@@ -25,28 +25,25 @@ function RowRenderer<R, K extends keyof R>({
...
-    scrollLeft,
-    isSummaryRow: true
-  };
-
-  return <Row<R> {...rendererProps} />;
+  return (
+    <Row<R>
+      idx={idx}
+      row={rowData}
+      width={columnMetrics.totalColumnWidth + getScrollbarSize()}
```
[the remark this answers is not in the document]
MayhemYDG: Maybe we'll do that when we'll refactor header/summary rows to render them directly in the grid (if we can work around the edge bug) 🤔

PR #1775, opened 2019-10-07
Grid refactor

MayhemYDG:
TODO:
- [ ] Move header rows rendering into `ActualRow` somehow.
  - Right now `HeaderRow` always re-renders, that should take care of it.
- [ ] Only render the visible cells in header rows.
  - Should we just move Cell rendering in `ActualRow` as well? Something like `<Row {...props}>{getCells()}</Row>`? Composition!
- [ ] Reimplement scrolling on browsers that do not support `position: sticky`
  - Just pass `scrollLeft` and `scrollTop` as props on those browsers (just IE11?).
  - For browsers that support `position: sticky` we'll just pass `undefined` or `null` so we do not trip `React.memo` or `SCU`.
- [ ] Fix `InteractionMasks`
  - normal cells: `absolute` positioning
  - fixed cells: `fixed` positioning
  - no need for `scrollLeft`/`scrollTop` anymore


amanmahajan7:
> Just pass scrollLeft and scrollTop as props on those browsers
This will affect scroll performance in IE as `SCU/memo` will return false so each cell will render on scroll event

MayhemYDG:
Yes, that's very true, but the experience in IE is already beyond mediocre, so I think it's fine if we don't optimize for it, and let us optimize everything else.

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -209,144 +217,60 @@ export default function Canvas<R>({
...
-    const row = rows.get(rowIdx);
-    if (row && row.getRowTop) {
-      return row.getRowTop();
-    }
     return rowHeight * rowIdx;
   }
 
-  function getRowHeight(rowIdx: number) {
-    const row = rows.get(rowIdx);
-    if (row && row.getRowHeight) {
```
qili26: Maybe this was for dynamic row height consideration? I am not sure if we can remove this.

packages/react-data-grid/style/rdg-header.less
```diff
@@ -10,12 +10,14 @@
 }
 
 .react-grid-HeaderRow {
-  position: relative;
-  overflow: hidden;
+  position: sticky;
```
qili26: So you plan to stick the header rows like normal rows? How will the canvas vertical scroll bar look like?

MayhemYDG: The scrollbar will cover the header rows, I think that's fine.

qili26: hmm... seriously, I had never seen a grid columns are covered by it's rows canvas vertical scroll bar.

https://github.com/FancyGrid/awesome-grid

MayhemYDG: Why spend more effort on such a benign issue?

qili26: For IE 11 and IE Edge, vertical scrolling will make the headers and summary rows jumping around after setting the `transform: translateY()`.

Just FYI, here is how it looks like for IE11 and IE Edge.
IE 11 (header columns are scrolled to top, with the transform calculated later, it will be scrolled to top and jump back)
IE 11 vertical scrolling

IE Edge (header columns are gone and reappear)
IE Edge vertical scrolling

MayhemYDG: This pr isn't ready for review.

PR #1776, opened 2019-10-08
Move all utilities to the utils folder

amanmahajan7:
- [x] Move all utilities to the utils folder
- [x] Deleted unused utilities and combine files where make sense
- [x] Use `camelCase` for utility file names
- [x] Delete `keyCodes` file


packages/react-data-grid/src/index.ts
```diff
@@ -1,16 +1,13 @@
...
-import * as _utils from './common/utils';
 import * as _helpers from './helpers';
 export * from './Columns';
 export * from './formatters';
 export * from './common/editors';
 export * from './common/enums';
 export * from './common/types';
 export {
-  _utils,
   _helpers
```
amanmahajan7: Need to delete this as well

MayhemYDG: Maybe we can start moving stuff off of the addons package.

PR #1779, opened 2019-10-09
Improve scrolling performance by batching rendered rows

MayhemYDG:
Right now the current strategy is to render and remove rows as soon as they enter or leave the viewport (±overscan), which isn't ideal for the rendering rate, as we have to render new rows, or remove rows very often when scrolling.

`removeChild` is also one of the most taxing operation by the way:
image

What I'm proposing here is to render rows in batches, that way we render more rows at once, but we don't add or remove rows as often, which will reduce the flashing of rows when scrolling, and improve React's render rate by having fewer big updates.


MayhemYDG:
image

The improvement is quite noticeable in Firefox, where it can lag quite a bit without this optimization.

packages/react-data-grid/src/utils/__tests__/viewportUtils.spec.ts
```diff
@@ -116,8 +117,8 @@ describe('viewportUtils', () => {
       expect(getRange({ scrollLeft: 500, columnMetrics })).toEqual({
```
amanmahajan7: We should add a test the number for a different batchSize?

PR #1780, opened 2019-10-09
CSS cleanups, fixes, and reimplementations

MayhemYDG:
Some things to note:
- cells are now centered vertically by setting the line-height === row height.
  - This fixes header cells text not being vertically centered
  - We were previously using `position: absolute; top: 50%; transform: translateY(-50%)` which may hurt text rendering or layout times.
- The generic editors should now be perfectly aligned with the formatters
- Selected rows now have a hover style
- cell actions icons are missing, should we add a dependency on some icons?


packages/react-data-grid/src/Cell/CellContent.tsx
```diff
@@ -69,18 +69,16 @@ export default function CellContent<R>({
...
+          rowData={rowData}
+          column={column}
+          value={value}
+          isScrolling={isScrolling}
+          isRowSelected={isRowSelected}
+          onRowSelectionChange={onRowSelectionChange}
+        />
         {cellControls}
       </div>
       {tooltip && <span className="cell-tooltip-text">{tooltip}</span>}
```
qili26: https://github.com/adazzle/react-data-grid/pull/1773/files/cba04c56c295e32fa3f036cb6cd9dbd5f8d5ec60#r333000852

Thoughts about removing this here? Since the style is touched, if we want to remove, I bet it's better in this PR. Or we keep it for revisiting.

MayhemYDG: Let's do it in a separate PR.

packages/react-data-grid/style/rdg-header.less
```diff
@@ -22,6 +22,13 @@
   width: 10px;
 }
 
+@supports (position: sticky) or (position: -webkit-sticky) {
```
amanmahajan7: Do we need this? Currently we render all the header cells and adjust the scrolleft on scroll

MayhemYDG: We also iterate through all the columns to set/unset the scroll position of each header row cells.

I figure we should optimize this for browsers that support sticky position, no?

packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -132,24 +132,21 @@ export default class HeaderCell<R> extends React.Component<Props<R>> {
   }
 
   render() {
-    const { column, rowType, height } = this.props;
+    const { column, rowType } = this.props;
 
-    const className = classNames('react-grid-HeaderCell', {
+    const className = classNames('react-grid-Cell', {
```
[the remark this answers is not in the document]
MayhemYDG: Sure, there's still a lot of opportunities to clean up the existing stylesheets, and rename classes, but I've mostly focused on replacing bootstrap in this PR.

packages/react-data-grid/style/rdg-cell.less
```diff
@@ -46,22 +41,10 @@
...
 }
 
-.react-grid-Cell:not(.editing) .react-grid-Cell__value {
-  white-space: nowrap;
-  text-overflow: ellipsis;
-  overflow: hidden;
-  height: inherit;
-}
-
 .react-grid-Cell.readonly {
```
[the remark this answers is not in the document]
MayhemYDG: Probably not, I haven't fully cleaned up this file.

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -140,15 +145,11 @@ export default class HeaderRow<R> extends React.Component<HeaderRowProps<R>> {
...
-
     return (
       <div
         ref={this.headerRow}
-        style={{
-          height: rowType === HeaderRowType.FILTER ? 500 : height
-        }}
-        className="react-grid-HeaderRow"
+        className="rdg-header-row"
+        style={{ height: this.props.height }}
```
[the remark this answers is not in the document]
MayhemYDG: Yeah, imo we should just reuse the editor container for such custom filters, otherwise it makes it hard to work with `overflow: hidden`.

packages/react-data-grid/src/common/cells/headerCells/SortableHeaderCell.tsx
```diff
@@ -52,8 +52,8 @@ export default function SortableHeaderCell<R>(props: Props<R>) {
 
   return (
     <div className="rdg-sortable-header-cell" onClick={onClick}>
-      <span className="pull-right">{SORT_TEXT[sortDirection]}</span>
       {content}
+      <span className="rdg-header-sort">{SORT_TEXT[sortDirection]}</span>
```
qili26: Since we changed to MUI in this PR, should we change this as well to be an icon?

packages/react-data-grid-addons/src/draggable/RowDragLayer.js
```diff
@@ -69,9 +69,9 @@ class CustomDragLayer extends Component {
           if (c.formatter) {
             const Formatter = c.formatter;
             const dependentValues = typeof c.getRowMetaData === 'function' ? c.getRowMetaData(item, c) : {};
-            cells.push(<td key={`dragged-cell-${rowIdx}-${c.key}`} className="react-grid-Cell" style={{ padding: '5px' }}><Formatter dependentValues={dependentValues} value={item[c.key]} /></td>);
+            cells.push(<td key={`dragged-cell-${rowIdx}-${c.key}`} className="rdg-cell" style={{ padding: '5px' }}><Formatter dependentValues={dependentValues} value={item[c.key]} /></td>);
```
qili26: `{ padding: 5 }`? Any ideas why this is not moved to css?

packages/react-data-grid/style/rdg-editor.less
```diff
@@ -0,0 +1,26 @@
...
+  position: absolute;
+}
+
+.rdg-text-editor,
+.rdg-select-editor {
+  -moz-appearance: none;
+  -webkit-appearance: none;
+  appearance: none;
+
+  box-sizing: border-box;
```
qili26: do we need to specify this?

packages/react-data-grid/src/Header.tsx
```diff
@@ -124,17 +122,12 @@ export default forwardRef(function Header<R>(props: HeaderProps<R>, ref: React.R
     props.cellMetaData.onCellClick({ rowIdx: -1, idx: -1 });
   }
 
-  const className = classNames('react-grid-Header', {
-    'react-grid-Header--resizing': resizing !== null
```
qili26: no more this?

packages/react-data-grid/style/rdg-header.less
```diff
@@ -38,97 +22,93 @@
...
-.react-grid-HeaderCell .Select-menu-outer .Select-menu .Select-option.is-focused {
+.rdg-cell .Select-menu-outer .Select-menu .Select-option.is-focused {
     color: #555555;
 }
 
 .rdg-sortable-header-cell {
   cursor: pointer;
 }
+
+.rdg-header-sort {
```
qili26: maybe using `flex` and put 2 `div`s inside rather than just `float: right`?

packages/react-data-grid/src/common/cells/headerCells/SortableHeaderCell.tsx
MayhemYDG: Let's leave it like this for now.

packages/react-data-grid-addons/src/draggable/RowDragLayer.js
MayhemYDG: This PR doesn't aim to fix every styling issues in the repo, it's just trying to address issues that arose after removing the bootstrap styles. There are too many things to fix or cleanup otherwise.

packages/react-data-grid/style/rdg-editor.less
MayhemYDG: Yes, it's too big otherwise.

packages/react-data-grid/src/Header.tsx
MayhemYDG: Doesn't look like we need it no.

packages/react-data-grid/style/rdg-header.less
MayhemYDG: It just works though?

qili26: My concern is that when the columns width is `100px` while the header text width is like `200px`, will the sort icon being pushed to next line or keep in the view. I believe using `float: right` will push it to next line. I am trying to run this PR in my local but having some packages installing issue.

PR #1781, opened 2019-10-10
Editor 2.0

amanmahajan7:
- [x] Remove ref dependency
- [x] Use state to manage changed value
- [x] <del>Provide a hook to handle cursor position for cell navigation</del> <del>Added a new `editorRef` prop</del>Each editor should prevent event propagation to cancel the default navigation
- [ ] <del>Support validation</del> Will add in a separate PR
- [ ] <del>Do we need `hasResults`, `isSelectOpen` methods?</del> It is needed for the legacy API. We should mark these methods as unsafe 
- [x] Improve types


packages/react-data-grid/src/common/editors/SimpleTextEditorNew.tsx
```diff
@@ -0,0 +1,39 @@
...
+  onBlur
+}: Props) {
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+      inputRef.current.select();
+    }
+  }, [inputRef]);
```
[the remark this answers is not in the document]
amanmahajan7: ESLint plugin automatically added it

amanmahajan7:
@MayhemYDG I have committed my local changes in case you want to work on it while I am away

MayhemYDG:
I probably won't have the time for it, so I'll leave it up to you.

qili26:
@amanmahajan7 I can give it a try but might need some time as a side prj.

amanmahajan7:
@qili26 I did not realize you are working on a separate branch :( I ended up making some changes (some are similar to yours). The main difference is how we handle textbox navigation. I ended up providing a ref which editors can set on the input element

packages/react-data-grid/src/common/editors/TextEditor.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { forwardRef } from 'react';
+
+import { EditorProps } from '../types';
+
+type SimpleTextEditorProps = Pick<EditorProps<string>, 'value' | 'onChange' | 'onCommit'>;
+
+export default forwardRef<HTMLInputElement, SimpleTextEditorProps>(function TextEditor({
```
qili26: I took a quick look at your structure. My main concern is that for the editors provided by others, they must use `forwardRef` or a class component, i.e. a ref is still needed for RDG API. For mine, it might be similar, the user must hook the `useDataGridEditor` to their editor. and if their editor was not a class component, they might have to use render props pattern to delegate the keyboard event to their component.

amanmahajan7: I added an `inputRef` prop to make it explicit. This way it can be used with the class component as well. Take a look and let me know what you think. I will check your changes in the meantime.

packages/react-data-grid/src/common/types.ts
```diff
@@ -117,27 +119,28 @@ export interface Editor<TValue = never> extends React.Component {
...
-export interface EditorProps<TValue, TDependentValue = unknown, TRow = any> {
-  column: CalculatedColumn<TRow, TDependentValue>;
+export interface EditorProps<TValue, TDependentValue = unknown, TRow = any, TField extends keyof TRow = any> {
+  column: CalculatedColumn<TRow, TDependentValue, TField>;
   value: TValue;
   rowMetaData?: TDependentValue;
   rowData: TRow;
   height: number;
-  onCommit(args?: { key?: string }): void;
+  inputRef: React.RefObject<HTMLInputElement>;
```
qili26: I believe it's better to set `HTMLInputElement` and `HTMLSelectElement`

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -1,274 +1,246 @@
...
+  const { enableNewEditor, getRowMetaData, editor } = column;
+
+  const getInputNode = useCallback(() => {
+    return enableNewEditor
+      ? inputRef.current
+      : editorRef.current && editorRef.current.getInputNode();
+  }, [enableNewEditor]);
+
+  useEffect(() => {
+    const inputNode = getInputNode();
```
qili26: My only concern is that if we put this `useEffect` in the editor container, the RDG might delegate too much for the `CustomEditor`. We should leave it to the consumer project to decide if they want to focus it by entering the editor or not. Just to be flexible. That's the reason why I added that hook.

We could consider passing the `firstEditorKeyPress` down to the `CustomEditor` as well to give the `CustomEditor` more flexibility. As currently for the line#73, it has a bug that if you input the value like `1` and `2` and `3`. the result would be just `23`. as the input `1` will open the editor and make the input selecting all, and the second input `2` will replace the `1` and set the cursor to the end which results in having `23` in the input.

amanmahajan7: > My only concern is that if we put this useEffect in the editor container, the RDG might delegate too much for the CustomEditor. We should leave it to the consumer project to decide if they want to focus it by entering the editor or not. Just to be flexible

This is totally optional with the new editor api. If `editorRef(previously inputRef)` is set and it is an `input/select` element then we add additional behavior which is 
- prevent navigation based on the caret position
- focus/select element


> We could consider passing the firstEditorKeyPress down to the CustomEditor as well to give the CustomEditor more flexibility if we don't use the above hook that I mentioned. As currently for the line#73, it has a bug that if you input the value like 1 and 2 and 3. the result would be just 23. as the input 1 will open the editor and make the input selecting all, and the second input 2 will replace the 1 and set the cursor to the end which results in having 23 in the input.

I am not sure if we should open editor or every key. Activating a complex editor like a date-picker does not make much sense. May be we can somehow allow users to define the keys that should open the editor. I am not too sure at this point and we need some investigation. I like how slick grid editors work

https://mleibman.github.io/SlickGrid/examples/example3-editing.html

I think the new API is easy to work with. We only introduced two new props
- `onChange` - Call it when the value is changed. This way `getInputNode` is not required
- `editorRef` - set it if you want the grid to handle keyboard navigation for input elements and selection for `input/select` elements

This PR does not modify how editors are opened and this is still up for discussion

Thanks for playing with the idea of using hooks for navigation but it seems like it may be adding additional complexity. @MayhemYDG what do you think about the two approaches. Other ideas?

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -1,274 +1,247 @@
...
 
-  componentDidMount() {
-    const inputNode = this.getInputNode();
     if (inputNode instanceof HTMLElement) {
       inputNode.focus();
     }
     if (inputNode instanceof HTMLInputElement) {
```
qili26: Just FYI, I believe there is a bug here for a long time. Please see my fix.

https://github.com/adazzle/react-data-grid/pull/1813/files#diff-bbe7676c3301843d12329fb8ff07e0f3R29

amanmahajan7: @qili26 wdyt about the last commit? The issue is that we enable the editor on any key press. I think users should have some control over when to enable the editor and whether to select the value or not. I ended up removing `editorRef`, I think each editor should handle `keyDown` event and prevent default navigation if needed.

qili26: I think removing the `editorRef` falls back to my implementation, which provides couple hooks for others to composite. Using context was just for passing the internal props directly to the hook so that when people are using it, they could just use like `useRDGEditor` (or like `useRDGKeys()`, etc..).

packages/react-data-grid/src/common/editors/TextEditor.tsx
```diff
@@ -0,0 +1,34 @@
...
+}: TextEditorProps) {
+  const inputRef = useRef<HTMLInputElement>(null);
+  useEffect(() => {
+    inputRef.current?.focus();
+    inputRef.current?.select();
+  }, []);
+
+  function onKeyDown(e: React.KeyboardEvent<HTMLInputElement>) {
+    if (e.key === 'ArrowLeft' || e.key === 'ArrowRight') {
+      e.stopPropagation();
```
nstepien: I think we should just disable keyboard shortcuts while we have an editor open in rdg itself, otherwise we'll need a variation of this for every editor.

packages/react-data-grid/src/common/editors/EditorContainer.tsx
```diff
@@ -1,274 +1,230 @@
...
 }
 
-interface State {
-  isInvalid: boolean;
-}
-
-export default class EditorContainer<R, K extends keyof R> extends React.Component<Props<R, K>, State> {
-  static displayName = 'EditorContainer';
+function getInitialValue<R, CK extends keyof R>({ key, value }: { key: string | null; value: R[CK] }): R[CK] | string {
+  // TODO: Do we need to activate editor when Delete, Backspace or other keys are pressed?
```
nstepien: I think in most cases pressing delete/backspace on a cell should wipe the value, update the row data, then open the editor.
There will be cases where a cell's value cannot be deleted, so I think we should add an optional `onDelete`/`onClear`/`onReset` handler, or something like that.

```js
const column = {
  ...column,
  onReset({ commit, ...data }) {
    commit(DEFAULT_VALUE);
    // commit(0);
    // commit('');
    // commit(null);
    return false; // optional, returning false would not open the editor
  }
};

// and/or...
const column = {
  ...,
  defaultValue: DEFAULT_VALUE // or 0 / '' / null / etc...
};
```

I guess the handler/default value would have to be mandatory for editable columns.

packages/react-data-grid/src/common/editors/EditorContainer.tsx
nstepien:
```js
const columns = {
  ...column,
  onCellInput({ commit, ...data }) {
    // let users handle it here, per column
    commit(value);
    return true; // to open the editor
  }
}
```

I think this is what we should do. In most cases users will only use a text input column, so we can provide a simple default handler for that. `onCellInput: defaultCellInputHandler`.
If `onCellInput` is missing, then nothing will happen.

This handler could also handle backspace/delete, as well as pasting data.

explicit behavior > implicit assumptions, even if it means writing more code.

>it has a bug that if you input the value like 1 and 2 and 3. the result would be just 23. as the input 1 will open the editor and make the input selecting all, and the second input 2 will replace the 1 and set the cursor to the end which results in having 23 in the input.

Sounds like what we want is a `openReason: 'click' | 'input' | anything else?`. Then you can adjust the initial state of your editor as desired.

PR #1782, opened 2019-10-10
Remove enableIsScrolling

MayhemYDG:
- Triggering re-renders all all rows and cells every time you start and stop scrolling is not very nice.
- Formatters not using `isScrolling` properly to avoid "derendering" is not very nice.
- Users can use a custom hook based on `requestIdleCallback` to keep smooth scrolling.


qili26:
hmm... IE and EDGE are broken.

PR #1785, opened 2019-10-10
Delete Grid component

MayhemYDG:
`Grid` is an unnecessary component layer, so I've moved its stuff into the root component.


packages/react-data-grid/src/Canvas.tsx
```diff
@@ -6,38 +6,40 @@ import RowGroup from './RowGroup';
 import InteractionMasks from './masks/InteractionMasks';
 import { getColumnScrollPosition, isPositionStickySupported, getScrollbarSize } from './utils';
 import { EventTypes } from './common/enums';
-import { CalculatedColumn, Position, ScrollPosition, SubRowDetails, RowRenderer, RowRendererProps, RowData } from './common/types';
-import { GridProps } from './Grid';
+import { CalculatedColumn, Position, ScrollPosition, SubRowDetails, RowRenderer, RowRendererProps, RowData, ColumnMetrics, CellMetaData, InteractionMasksMetaData } from './common/types';
+import { ReactDataGridProps } from './ReactDataGrid';
+import EventBus from './EventBus';
```
qili26: Just a general discussion regarding this `EventBus`, looks like it can be replaced by `Context` now? After several layers being removed, we directly render the `Canvas` in `RDG.tsx` in this PR, it seems we start to not need the `eventBus` anymore?

MayhemYDG: How would you use a context to trigger `scrollToColumn`?

qili26: For `scrollToColumn` case, can't we move it one level up? Then we can move canvas ref in `scrollToColumn` in RDG level like this. 
```javascript
  function scrollToColumn(idx: number, columns: CalculatedColumn<R>[]) {
    const { current } = canvas.current.ref;
    ...
  }
```

This `evenBus` `subscribe` and `unsubscribe` makes me feel like we are going back and forth to avoid passing functions as props. With the layers being removed, maybe the `eventBus` can be removed gradually, or did I miss something?

PR #1787, opened 2019-10-11
Add RowRenderer component

MayhemYDG:
- All row components will now be `memo`'d for free!
- Named it `ActualRowRenderer` because the `RowRenderer` name is already being used right now, I'm open to suggestions. `RowMaterializer`? `RowConjurer`? `RowRendererRenderer`? 🤔 
- Cleans up `Canvas` quite a bit.


packages/react-data-grid/src/Canvas.tsx
```diff
@@ -168,59 +151,34 @@ export default function Canvas<R>({
...
-
-    if (rowRenderer) {
-      return renderCustomRowRenderer(rendererProps);
+      const rowData = rowGetter(idx);
+      rowElements.push(
+        <RowRenderer<R>
+          key={idx}
+          idx={idx}
+          rowData={rowData}
+          rowRefs={rowRefs}
```
qili26: Instead of passing the `rowRefs`, should we pass a ref setting function? It seems weird for each individual row to have access to all other row refs.

packages/react-data-grid/src/common/types.ts
```diff
@@ -166,7 +166,7 @@ export interface RowsContainerProps {
   children: React.ReactElement;
 }
 
-export interface RowRendererProps<TRow> {
+export interface IRowRendererProps<TRow> {
```
qili26: `IRowRendererProps<TRow>` vs `IRowProps<TRow>`, wdyt?

Then for this line:
https://github.com/adazzle/react-data-grid/pull/1787/files#diff-c77bf2dcbfe5c1705d249cd2d09835aeR8
```typescript
export default class Row<R> extends React.Component<RowProps<R>> implements IRowRenderer
```

packages/react-data-grid/src/Canvas.tsx
MayhemYDG: If I did this we'd pass a new ref function on each render, which might trip up `memo`, and i'm having trouble typing the ref parameter for some reason.

packages/react-data-grid/src/common/types.ts
MayhemYDG: Mmh, I'm not sure, it's shared across multiple components I think, not just the `Row` component.

packages/react-data-grid/src/Canvas.tsx
qili26: In `Canvas.tsx`
```typescript
const setRowRef = useCallback((rowRef: (IRowRenderer & React.Component<IRowRendererProps<R>>) | null, idx: number) => {
    if (rowRef) {
      rowRefs.set(idx, rowRef);
    } else {
      rowRefs.delete(idx);
    }
  }, [rowRefs]);

<RowRenderer<R>
  key={idx}
  idx={idx}
  rowData={rowData}
  setRowRef={setRowRef}
  ...
/>
```

In `RowRenderer.tsx`
```typescript
interface RowRendererProps<R> extends SharedCanvasProps<R> {
  ...
  setRowRef(rowRef: (IRowRenderer & React.Component<IRowRendererProps<R>>) | null, idx: number): void;
}

const rendererProps: RendererProps<R> = {
    ref(row) {
      setRowRef(row, idx);
    },
   ...
};
```

This should work isn't it?

packages/react-data-grid/src/common/types.ts
qili26: Then I guess we should leave this naming for now and revisit later.

packages/react-data-grid/src/Row.tsx
```diff
@@ -18,13 +17,6 @@ export default class Row<R> extends React.Component<RowRendererProps<R>> impleme
   private readonly row = React.createRef<HTMLDivElement>();
```
qili26: I just found maybe we can remove this as well.

packages/react-data-grid/src/RowRenderer.tsx
```diff
@@ -0,0 +1,143 @@
...
+| 'selectedRows'
+>;
+
+interface RowRendererProps<R> extends SharedCanvasProps<R> {
+  idx: number;
+  rowData: R;
+  colOverscanStartIdx: number;
+  colOverscanEndIdx: number;
+  scrollLeft: number;
+  setRowRef(row: Row<R> | null, idx: number): void;
```
qili26: 👍 this is more concise.

PR #1791, opened 2019-10-14
horizontal scroll sync - fix: 1 interaction mask is moving during scroll dragging. 2 columns not in sync

qili26:


qili26:
With a few refactor to `Canvas.tsx` this approach is more doable. Currently for the existing behaviors when only columns were affected, it might seem fine. However when we enable the [summary rows](https://github.com/adazzle/react-data-grid/pull/1773), it will look awful for non Chrome browsers. 

scroll not sync jumping - summary rows

qili26:
Due to IE 11 weird bug, this implementation is halted.

PR #1793, opened 2019-10-16
Remove setScrollLeft on Row and Cell

MayhemYDG:
- It was only used in IE11
- We now only pass the `scrollLeft` prop in IE11, as it's not used in other browsers.
  - Vertical scrolling isn't impacted in any way.
- memo/SCU/PureComponent simplification:
  - we don't depend on `shallowequal` anymore
  - React's internal props comparison might be more optimized than `shallowequal`
- Fewer refs, we'll be able to move `Cell` to a stateless functional component.
- Performance isn't that much worse in IE11?

### DIFF
--- a/packages/react-data-grid/src/Row.tsx
+++ b/packages/react-data-grid/src/Row.tsx
@@ -3,9 +3,9 @@ import classNames from 'classnames';
 
 import Cell from './Cell';
 import { isFrozen } from './utils/columnUtils';
-import { IRowRenderer, IRowRendererProps, CellRenderer, CellRendererProps, CalculatedColumn } from './common/types';
+import { IRowRendererProps, CellRendererProps, CalculatedColumn } from './common/types';
 
-export default class Row<R> extends React.Component<IRowRendererProps<R>> implements IRowRenderer {
+export default class Row<R> extends React.Component<IRowRendererProps<R>> {
   static displayName = 'Row';
 
   static defaultProps = {
@@ -13,8 +13,6 @@ export default class Row<R> extends React.Component<IRowRendererProps<R>> implem
     isSelected: false,
     height: 35
   };
-
-  private readonly cells = new Map<keyof R, CellRenderer>();
 
   handleDragEnter = (e: React.DragEvent<HTMLDivElement>) => {
     // Prevent default to allow drop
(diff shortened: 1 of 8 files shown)


packages/react-data-grid/src/masks/InteractionMasks.tsx
```diff
@@ -692,12 +664,7 @@ export default class InteractionMasks<R, K extends keyof R> extends React.Compon
         onKeyDown={this.onKeyDown}
         onFocus={this.onFocus}
       >
-        {copiedPosition && (
-          <CopyMask
-            {...this.getSelectedDimensions(copiedPosition)}
-            ref={this.copyMask}
-          />
-        )}
+        {copiedPosition && <CopyMask {...this.getSelectedDimensions(copiedPosition)} />}
```
qili26: what is the `CopyMask`? Like `ctrl c` / `ctrl v` background?

packages/react-data-grid/src/Row.tsx
```diff
@@ -34,34 +33,44 @@ export default class Row<R> extends React.Component<IRowRendererProps<R>> implem
...
+      row,
+      scrollLeft
+    } = this.props;
+    const Renderer = this.props.cellRenderer!;
+    const canSticky = isPositionStickySupported();
 
-    return <Renderer key={`${key as keyof R}-${idx}`} {...cellProps} />; // FIXME: fix key type
-  }
+    // FIXME: do we need this, considering columnsMetrics should have these columns sorted already?
+    const frozenColumns = columns.slice(0, this.props.lastFrozenColumnIndex + 1);
```
qili26: I remember I tried in my local that to put `frozenColumns` first and then `nonFrozenColumns` later. The row cells displaying was screwed up. Not sure why that happened as I didn't take time to investigate.

qili26:
Just want to add one more comment that the difference between the existing code and the new code is just the SCU on `CellContent` and children component. The performance would be improved after we reach the Cell refactoring.

The bottleneck in IE 11 is after horizontal scrolling there were bunch of `appendChild()`, `removeChild()`, `insertBefore()` happening for react `mount` `unmount`. IE 11 just cannot handle those large amount of actions....

packages/react-data-grid/src/masks/InteractionMasks.tsx
MayhemYDG: image
A mask that appears when you copy a cell.

PR #1796, opened 2019-10-17
Squeeze more performance out of Row.getCells

MayhemYDG:
- `Row.getCells()` is now less wasteful:
  - Before: `columns.slice` * 2 + `.filter()` + `.concat()` + `.map()` = we create 5 arrays, not super efficient
  - After: we create 1 array, and we only loop through the columns once, up to `colOverscanEndIdx`, that's better, and should put less stress on the gc.
- We now render frozen cells before the non-frozen cells.

### DIFF
--- a/packages/react-data-grid/src/Row.tsx
+++ b/packages/react-data-grid/src/Row.tsx
@@ -3,7 +3,6 @@ import classNames from 'classnames';
 
 import Cell from './Cell';
 import { isFrozen } from './utils/columnUtils';
-import { isPositionStickySupported } from './utils';
 import { IRowRendererProps } from './common/types';
 
 export default class Row<R> extends React.Component<IRowRendererProps<R>> {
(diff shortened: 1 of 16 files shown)


packages/react-data-grid/src/Row.tsx
```diff
@@ -3,16 +3,13 @@ import classNames from 'classnames';
...
-import { isPositionStickySupported } from './utils';
 import { IRowRendererProps } from './common/types';
 
 export default class Row<R> extends React.Component<IRowRendererProps<R>> {
   static displayName = 'Row';
 
   static defaultProps = {
-    cellRenderer: Cell,
-    isSelected: false,
-    height: 35
```
MayhemYDG: `isSelected` is not a used prop, and `height` is already default-ed in the root component, so it can't be undefined.

packages/react-data-grid/src/utils/viewportUtils.ts
```diff
@@ -77,9 +77,9 @@ export function getHorizontalRangeToRender<R>({
   viewportWidth = viewportWidth > 0 ? viewportWidth + firstVisibleColumnHiddenWidth : totalColumnWidth;
   const availableWidth = viewportWidth - totalFrozenColumnWidth;
   const nonFrozenRenderedColumnCount = getColumnCountForWidth(columns, availableWidth, colVisibleStartIdx);
-  const colVisibleEndIdx = Math.min(columns.length, colVisibleStartIdx + nonFrozenRenderedColumnCount);
+  const colVisibleEndIdx = Math.min(columns.length - 1, colVisibleStartIdx + nonFrozenRenderedColumnCount);
   const colOverscanStartIdx = Math.max(0, colVisibleStartIdx - 1);
-  const colOverscanEndIdx = Math.min(columns.length, colVisibleEndIdx + 1);
+  const colOverscanEndIdx = Math.min(columns.length - 1, colVisibleEndIdx + 1);
```
MayhemYDG: I caught a bug here, the max column index could be too big and crash the grid.

PR #1797, opened 2019-10-18
FC-ize Cell, refactor CellContent

MayhemYDG:
- Cell is now a functional component with memo instead of a `PureComponent`
- I've merged `CellValue` into `CellContent`, as we don't need the overhead of 2 components when 1 is enough
- I've got ideas to refactor `Cell` and `CellContent` further and improve performance 👀


packages/react-data-grid/src/Cell/CellContent.tsx
```diff
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { createElement, cloneElement } from 'react';
```
qili26: I am wondering if this layer is needed as well. However for this particular PR, I would prefer less removal, so removing `CellValue` should be enough.

packages/react-data-grid/src/Cell.tsx
```diff
@@ -79,46 +81,45 @@ export default class Cell<R> extends React.PureComponent<CellProps<R>> {
...
+      onDoubleClick: handleCellDoubleClick,
+      onContextMenu: handleCellContextMenu,
+      onDragOver: handleDragOver
     };
 
     if (!columnEvents) {
       return allEvents;
     }
 
+    const eventInfo: ColumnEventInfo<R> = {
```
qili26: 👍

qili26:
One more question is like we don't cache functions a lot, so each render will create new instance. Is it because `Cell` is relative end component so that we won't benefit a lot?

PR #1798, opened 2019-10-18
ReactDataGrid -> DataGrid, remove class root wrapper

MayhemYDG:
- I've renamed `ReactDataGrid` (the component and file) to `DataGrid`. This makes more sense, we don't import `react-select` as `ReactSelect` for example, or some button components as `ReactButton`.
- I've removed the root component class wrapper. Since we figured out a way to work around the forwardRef types issue when using generics, the class isn't needed anymore.


packages/react-data-grid-addons/src/draggable-header/__tests__/DraggableContainer.spec.js
```diff
@@ -2,11 +2,11 @@ import React from 'react';
...
 
 import DraggableContainer from '../DraggableContainer';
-import ReactDataGrid from 'react-data-grid';
+import DataGrid from 'react-data-grid';
 
 describe('<DraggableContainer />', () => {
   it('should render grid wrapper to be used as Drag and Drop context with grid as a child component inside', () => {
     const wrapper = shallow(<DraggableContainer />);
-    expect(wrapper.find(ReactDataGrid));
+    expect(wrapper.find(DataGrid));
```
qili26: Don't know what this test is for....

packages/react-data-grid-addons/src/menu/ContextMenu.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ContextMenu } from 'react-contextmenu';
 
-export default function ReactDataGridContextMenu({ children }: { children: React.ReactNode }) {
-  return <ContextMenu id="reactDataGridContextMenu">{children}</ContextMenu>;
+export default function DataGridContextMenu({ children }: { children: React.ReactNode }) {
+  return <ContextMenu id="dataGridContextMenu">{children}</ContextMenu>;
```
qili26: This seems a break change, maybe for `id` we keep it?

MayhemYDG: Reverted.

packages/react-data-grid-addons/src/draggable-header/__tests__/DraggableContainer.spec.js
MayhemYDG: It doesn't even have a `toBeDefined()` or anything, so I've removed this test file and another useless one.
The tests are in a dire state right now, and we'll need to overhaul them, but not in this PR.

qili26: Agree. This reminds me a jest eslint rule that to always require assertion.

PR #1800, opened 2019-10-19
Add support for custom cell content renderers

MayhemYDG:
- Main info in here: https://gist.github.com/MayhemYDG/090298c3c2d94324cb332c33d82fdcfb
- Custom cell content rendering is an opt-in feature, so it shouldn't break anything.
- Removed handling the `value` in `Row`/`Cell`, it's now handled within cell content renderers, which makes types a bit easier to deal with.
- Removed `cellControls` prop from `Cell`, which doesn't seem to be an actually used feature
- Removed `tooltip` prop from `Cell`, custom cell content renderers should just implement this themselves. We can add it back if needed.
- I want to remove the `children` prop from `Cell` as well, I think it's only used for custom cell renderers that wrap `Cell`. WDYT?

### DIFF
--- a/packages/react-data-grid/src/Cell.tsx
+++ b/packages/react-data-grid/src/Cell.tsx
@@ -1,9 +1,7 @@
 import React, { memo } from 'react';
 import classNames from 'classnames';
 
-import CellActions from './Cell/CellActions';
 import CellContent from './Cell/CellContent';
-import CellExpand from './Cell/CellExpander';
 import { CellRendererProps, ColumnEventInfo } from './common/types';
 import { isFrozen } from './utils/columnUtils';
 
@@ -11,12 +9,9 @@ export interface CellProps<R> extends CellRendererProps<R> {
   children?: React.ReactNode;
   // TODO: Check if these props are required or not. These are most likely set by custom cell renderer
   className?: string;
-  tooltip?: string | null;
-  cellControls?: unknown;
 }
 
 function Cell<R>({
-  cellControls,
   cellMetaData,
   children,
   className,
@@ -30,7 +25,6 @@ function Cell<R>({
   rowData,
   rowIdx,
   scrollLeft,
-  tooltip,
   value = ''
 }: CellProps<R>) {
   function handleCellClick() {
@@ -58,12 +52,6 @@ function Cell<R>({
     cellMetaData.onCellDoubleClick({ idx, rowIdx });
   }
 
-  function handleCellExpand() {
-    if (cellMetaData.onCellExpand) {
-      cellMetaData.onCellExpand({ rowIdx, idx, rowData, expandArgs: expandableOptions });
-    }
-  }
-
   function handleDragOver(e: React.DragEvent<HTMLDivElement>) {
     e.preventDefault();
   }
@@ -89,7 +77,6 @@ function Cell<R>({
       className, {
         'rdg-cell-frozen': colIsFrozen,
         'rdg-cell-frozen-last': colIsFrozen && column.idx === lastFrozenColumnIndex,
-        'has-tooltip': !!tooltip,
         'rdg-child-cell': expandableOptions && expandableOptions.subRowDetails && expandableOptions.treeDepth > 0
       }
     );
@@ -137,6 +124,7 @@ function Cell<R>({
     return allEvents;
   }
 
+  // FIXME: is this property used?
   if (column.hidden) {
     return null;
   }
@@ -151,6 +139,7 @@ function Cell<R>({
           idx={idx}
           rowIdx={rowIdx}
           column={column}
+          cellMetaData={cellMetaData}
           rowData={rowData}
           value={value}
           expandableOptions={expandableOptions}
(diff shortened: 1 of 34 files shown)


packages/react-data-grid/src/Cell.tsx
```diff
@@ -137,70 +124,28 @@ function Cell<R>({
     return allEvents;
   }
 
+  // FIXME: is this property used?
```
qili26: I am not using it but not sure if others are using. I believe this is not a easily removed field.

MayhemYDG: image
It looks like that if a column is hidden, I think we should remove this feature, wdyt?

qili26: What a weird API. Yeah, I think so, we should just remove it. I assumed it wrong that the columns will just be hidden but later realized that the cell is already rendered, and we only get an empty cell.

For this case it's same to putting `formatter` to return null

MayhemYDG: Removed it now.

packages/react-data-grid/src/Cell.tsx
```diff
@@ -1,22 +1,16 @@
...
-import CellActions from './Cell/CellActions';
-import CellContent from './Cell/CellContent';
-import CellExpand from './Cell/CellExpander';
 import { CellRendererProps, ColumnEventInfo } from './common/types';
 import { isFrozen } from './utils/columnUtils';
 
 export interface CellProps<R> extends CellRendererProps<R> {
-  children?: React.ReactNode;
   // TODO: Check if these props are required or not. These are most likely set by custom cell renderer
+  children?: React.ReactNode;
```
qili26: I believe we don't have the usage in canary code base internally from beginning(?) or at least for a long time. To have the `children` in `Cell`, we need to render it in `Row.tsx` like `<Cell><Something /></Cell>` in `getCells` function, However, [this line](https://github.com/adazzle/react-data-grid/blob/canary/packages/react-data-grid/src/Row.tsx#L59) directly renders the `Cell` (`Renderer`) and passes no children inside.

One other possibility is that since we directly export `Cell` component, potentially people could do `<Cell><TheirCode /></Cell>` in their code base. I believe it's no harm to keep it for now and we might need to decide to keep or kill it with `column.formatter` together, I believe we don't need to rush in this PR.

packages/react-data-grid/src/Cell.tsx
```diff
@@ -137,70 +98,43 @@ function Cell<R>({
...
-      tooltip={tooltip}
-      expandableOptions={expandableOptions}
-      onDeleteSubRow={cellMetaData.onDeleteSubRow}
-      cellControls={cellControls}
-      isRowSelected={isRowSelected}
-      onRowSelectionChange={onRowSelectionChange}
-      isSummaryRow={isSummaryRow}
-    />
+  const colIsFrozen = isFrozen(column);
+  className = classNames(
```
qili26: 👍

packages/react-data-grid/src/Cell/CellContent.tsx
```diff
@@ -66,9 +46,25 @@ export default function CellContent<R>({
     };
   }
 
+  function getFormattedValue(formatter: typeof column['formatter']) {
```
qili26: Should we create a type called `Formatter<TRow, TDependentValue = unknown, TField extends keyof TRow = keyof TRow>` and assign to both column and here?

packages/react-data-grid/src/__tests__/utils/index.ts
```diff
@@ -1,18 +1,20 @@
+import { valueCellContentRenderer } from '../../Cell/cellContentRenderers';
 import { CalculatedColumn } from '../../common/types';
 
-function createColumn(index: number): CalculatedColumn<{ [key: string]: unknown }> {
+function createColumn(index: number): CalculatedColumn<{ [key: string]: React.ReactNode }> {
```
qili26: Is this correct? [This line](https://github.com/adazzle/react-data-grid/blob/canary/packages/react-data-grid/src/common/types.ts#L50) stands for the RowData type, doesn't it? Then probably we should not return the `ReactNode`?

packages/react-data-grid/src/Cell/cellContentRenderers.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import CellContent from './CellContent';
+import { CellContentRendererProps } from '../common/types';
+
+export function legacyCellContentRenderer<R>(props: CellContentRendererProps<R>) {
+  return React.createElement<CellContentRendererProps<R>>(CellContent, props);
```
qili26: Any reason not to write it like this `<CellContent {...props} />`?

packages/react-data-grid/src/Cell/CellContent.tsx
MayhemYDG: We could, I'm just lazy and we're deprecating formatters anyway 🤷‍♂

packages/react-data-grid/src/__tests__/utils/index.ts
MayhemYDG: `valueCellContentRenderer` needs to return a property on `RowData`, and cell content formatters, need to return a `React.ReactNode`, so that's why I made the change here.
In tests we generally just use string or number values, I'm just making the type broad so it's easier to deal with.

packages/react-data-grid/src/Cell/cellContentRenderers.tsx
MayhemYDG: I recently found out that TS will transpile such code like this:
image

which is less efficient than manually using `React.createElement`:
image

Babel handles it better: https://babeljs.io/repl#?code_lz=GYVwdgxgLglg9mABAGwKYHMCGECeBhVZZPBKVMKAJXIBNUAnBgCgAd64WBnASkQG8AUIkSMoIekgA8BIiQrko_AHQq2HTgF9EAegB8AbgEagA

MayhemYDG:
>Currently in this PR, if I want to use `cellContentRenderer`, I will lose the ability to have the cell actions for example. Are you planning to add the support in the coming up PRs?

For now you'll have to stick to the legacy renderer + formatter. We'll add more renderers as we need them.

packages/react-data-grid/src/__tests__/utils/index.ts
qili26: I believe `React.ReactNode` is too broad which is similar to `any` but missing some basic types and adding some unnecessary types, e.g. a field in the row data could have an object or an array of objects which is not defined in `React.ReactNode`. In addition in terms of considering what is valid data type, I believe it's inappropriate as well because generally speaking, a `Row` data object should not contain `fragment` for example. (Though with that being said, people could still potentially save an element in the row data, but generally speaking they won't save into database) My point was that for RDG code, it actually doesn't care about what does the `RowData` looks like, so I think `CalculatedColumn<{ [key: string]: any }>` would be more appropriate in this case if `CalculatedColumn<{ [key: string]: unknown }>` will cause test type issue.

PR #1801, opened 2019-10-21
Resize the all the cells when resizing a column

MayhemYDG:
Since we've optimized rendering quite a bit, I think it's now safe to re-render the whole grid when resizing columns. 👍


MayhemYDG:
Yes unfortunately you can't set a cursor in css on pseudo element on those browsers 😭

qili26:
As it actually breaks IE / EDGE behavior, maybe we need to bring it back some day in the future. Otherwise the user really doesn't know the column could be resized.

MayhemYDG:
Yeah, I'll look into it someday.

PR #1802, opened 2019-10-22
Fix columns metrics columns generating logic, improve performance

qili26:
`column.left` was calculated based on obsoleted columns order when `isFrozne` column is applied, We should sort the columns first and then calculate it. This PR is only for fixing, no performance updates, as I believe after moving the code around we could save some more loops.

### DIFF
--- a/packages/react-data-grid/src/utils/columnUtils.ts
+++ b/packages/react-data-grid/src/utils/columnUtils.ts
@@ -8,14 +8,12 @@ type Metrics<R> = Pick<ColumnMetrics<R>, 'viewportWidth' | 'minColumnWidth' | 'c
 export function getColumnMetrics<R>(metrics: Metrics<R>): ColumnMetrics<R> {
   let left = 0;
   let totalWidth = 0;
-  let allocatedWidths = 0;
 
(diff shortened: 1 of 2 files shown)


packages/react-data-grid/src/utils/columnUtils.ts
```diff
@@ -8,21 +8,41 @@ type Metrics<R> = Pick<ColumnMetrics<R>, 'viewportWidth' | 'minColumnWidth' | 'c
...
+    metrics.minColumnWidth
+  );
 
-  setRemainingWidths(columns, unallocatedWidth, metrics.minColumnWidth);
-
-  let calculatedColumns: CalculatedColumn<R>[] = columns.map(column => {
+  const calculatedColumns: CalculatedColumn<R>[] = columns.map((column, idx) => {
     // Every column should have a valid width as this stage
-    const width = column.width as number;
+    const width = column.width || unallocatedColumnWidth;
```
qili26: This would be crazy that maybe no one would intentionally do that, but what if the user set `column.width` to `0`? Should we check based on `undefined` like above rather than `undefined null and 0`?

PR #1803, opened 2019-10-22
Render more cells in all features example

qili26:
1. make side bar visible / hidden. toggle by clicking on example title
2. auto adjust grid height for all features example

image

### DIFF
--- a/examples/demos/example13-all-features.js
+++ b/examples/demos/example13-all-features.js
@@ -208,8 +208,8 @@ export default class extends React.Component {
                 rowsCount={this.getSize()}
                 onGridRowsUpdated={this.handleGridRowsUpdated}
                 rowHeight={30}
-                minWidth={width - 2}
-                minHeight={height - 16}
+                minWidth={width}
+                minHeight={height}
                 selectedRows={this.state.selectedRows}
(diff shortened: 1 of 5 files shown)


examples/demos/example13-all-features.js
```diff
@@ -196,18 +197,25 @@ export default class extends React.Component {
...
+              <DataGrid
+                ref={node => this.grid = node}
+                enableCellSelect
+                columns={this.columns}
+                rowGetter={this.getRowAt}
+                rowsCount={this.getSize()}
+                onGridRowsUpdated={this.handleGridRowsUpdated}
+                rowHeight={30}
+                minWidth={width - 2}
+                minHeight={height - 16}
```
MayhemYDG: Do we need the `- 2` and `- 16`?

qili26: `width - 2 ` is needed, otherwise you will see the scroll bar is constantly added and removed, height doesn't need to subtract that much, but 16 is a personal preference. I believe for height we at least need to `- 1` for border size.

qili26: Let me double check, maybe it's due to the height with border.

PR #1806, opened 2019-10-24
Fix root div width with specified minWidth prop

qili26:
Subtract 2px due to border left and right from minWidth.


packages/react-data-grid/style/rdg-core.less
```diff
@@ -6,13 +6,14 @@
...
   border: 1px solid @borderColor;
+  box-sizing: border-box;
   .user-select();
   font-size: 14px;
 
   *,
   *::before,
   *::after {
-    box-sizing: border-box;
+    box-sizing: inherit;
```
qili26: 👍

PR #1811, opened 2019-10-30
Fix header cell resize cursor in IE11/Edge

MayhemYDG:


packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -148,6 +147,7 @@ export default class HeaderCell<R> extends React.Component<HeaderCellProps<R>> {
         onTouchStart={column.resizable ? this.onTouchStart : undefined}
       >
         {this.getCell()}
+        {column.resizable && <div className="rdg-header-cell-resizer" />}
```
qili26: How about using that `isIE` thing (maybe we removed in the totalRows PR). so we still keep the pseudo css code for firefox and chrome but check like `isIE && colum.resizable && <div className="rdg-header-cell-resizer" />`.

MayhemYDG: Would have to add back the isIE check.

qili26: Do you think that is reasonable? For performance consideration, I don't think chrome will benefit any as chrome is already fast enough. My only concern is just after IE / Edge support is dropped, we have to change everything back to pseudo code style. Adding `isIE` will change slightly less places...

MayhemYDG: Changing back to a pseudo element will be a very small task, I wouldn't worry about it.

PR #1813, opened 2019-10-31
Editor 2.0

qili26:
This PR is a new branch of https://github.com/adazzle/react-data-grid/pull/1781 to avoid commit pollution.

image

image


qili26:
@amanmahajan7 @MayhemYDG 
My POC code was to mainly remove the ref support so I created the following API for selecting the input value when the editor is open.

1. editor hook.
2. render props.
3. hoc.

So that for the existing editors that's already implemented, it can apply the latest changes by different ways.

pedroffd:
With this changes I notice that you've made some features to deal with hotkeys such as tab/enter/arrow left, there's a way to edit the key behavior with those changes?

qili26:
@pedroffd 
That's what I intended to do by providing a separate hook so that the consumer devs could compose with necessary overriding. However, it's still a thought under discussion.

Issue #1816, opened 2019-11-07
[canary] need a way to set SelectColumn width, to be always the same size.


jwallet:
- Have you already searched for similar issues? Please help us out and double-check first!

- Also, please don't be that person who deletes this template. It's here for a reason.

- Thanks!

---

### Which version of React JS are you using?

✅ Officially supported ✅
- [ ] v15.4.x

⚠️ Not officially supported, expect warnings ⚠️
- [ ] v15.5.x
- [ ] v15.6.x

☣️ Not officially supported, expect warnings and errors ☣️
- [x] v16.x.x

---

### Which browser are you using?

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

⚠️ Not officially supported, but "should work" ⚠️
- [x] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [x] 🐛 Bug Report
- [ ] 💡 Feature Request

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---
In `columns` array if we add `SelectColumn` to add a checkbox to the row : 
- we can't specify a width to SelectColumn to use
- SelectColumn is affected by `minColumnWidth`

```
// does nothing, it's still 60px
this.columns = [
     {...SelectColumn, width: 40},
     // { ... }
];
```
adding `minColumnWidth={200}` to <ReactDataGrid />` will give a width of 200px to the checkbox, which makes no sense

```

amanmahajan7:
This should be fixed in v7 canary 16

PR #1818, opened 2019-11-08
[canary] new prop for column 'minWidth' and SelectColumn always use `width` prop

jwallet:
> * Column new prop: minWidth
> 
> * Grid prop minColumnWidth is ignored if column has its prop minWidth specified, then it uses minWidth/width highest value
> 
> * SelectColumn can now have a width specified to stay fixed when minColumnWidth is specified
> 
> * updated all-features-13 which was already using SelectColumn to demo the added feature

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

- #1816 Can't specify a width to SelectColumn
- Can't specifiy a minWidth to individual column
- minColumnWidth affects SelectColumn and breaks the UI

**What is the new behavior?**
- SelectColumn width can be overwritten
- Column have a new prop : minWidth

```
[ ] Yes
[x] No
```

**Other information**:
image

close issue #1816

### DIFF
--- a/examples/demos/example13-all-features.js
+++ b/examples/demos/example13-all-features.js
@@ -22,6 +22,7 @@ export default class extends React.Component {
         key: 'id',
         name: 'ID',
         width: 80,
+        minWidth: 60,
         resizable: true,
         frozen: true
       },
@@ -29,6 +30,7 @@ export default class extends React.Component {
         key: 'avatar',
         name: 'Avatar',
         width: 60,
+        minWidth: 60,
         formatter: ImageFormatter,
         resizable: true,
         headerRenderer: <ImageFormatter value={faker.image.cats()} />
@@ -38,6 +40,7 @@ export default class extends React.Component {
         name: 'Title',
         editor: <DropDownEditor options={titles} />,
         width: 200,
+        minWidth: 50,
         resizable: true,
         events: {
           onClick: (ev, { idx, rowIdx }) => {
@@ -65,28 +68,28 @@ export default class extends React.Component {
         key: 'email',
         name: 'Email',
         editable: true,
-        width: 200,
+        minWidth: 250,
         resizable: true
       },
       {
         key: 'street',
         name: 'Street',
         editable: true,
-        width: 200,
+        minWidth: 200,
         resizable: true
       },
       {
         key: 'zipCode',
         name: 'ZipCode',
         editable: true,
-        width: 200,
+        minWidth: 80,
         resizable: true
       },
       {
         key: 'date',
         name: 'Date',
         editable: true,
-        width: 200,
+        minWidth: 100,
         resizable: true
       },
       {
@@ -125,7 +128,7 @@ export default class extends React.Component {
     };
   }
 
-  createRows = numberOfRows => {
+  createRows = (numberOfRows) => {
     const rows = [];
     for (let i = 0; i < numberOfRows; i++) {
       rows[i] = this.createFakeRowObjectData(i);
@@ -133,7 +136,7 @@ export default class extends React.Component {
     return rows;
   };
 
-  createFakeRowObjectData = index => {
+  createFakeRowObjectData = (index) => {
     return {
       id: `id_${index}`,
       avatar: faker.image.avatar(),
@@ -177,7 +180,7 @@ export default class extends React.Component {
     this.setState({ rows });
   };
 
-  getRowAt = index => {
+  getRowAt = (index) => {
     if (index < 0 || index > this.getSize()) {
       return undefined;
     }
@@ -189,7 +192,7 @@ export default class extends React.Component {
     return this.state.rows.length;
   };
 
-  onSelectedRowsChange = selectedRows => {
+  onSelectedRowsChange = (selectedRows) => {
     this.setState({ selectedRows });
   };
 
--- a/packages/react-data-grid/src/DataGrid.tsx
+++ b/packages/react-data-grid/src/DataGrid.tsx
@@ -13,7 +13,7 @@ import { isValidElementType } from 'react-is';
 import Header, { HeaderHandle } from './Header';
 import Canvas from './Canvas';
 import { legacyCellContentRenderer } from './Cell/cellContentRenderers';
-import { getColumnMetrics, getWidth } from './utils/columnUtils';
+import { getColumnMetrics } from './utils/columnUtils';
 import EventBus from './EventBus';
 import { CellNavigationMode, EventTypes, UpdateActions, HeaderRowType, DEFINE_SORT } from './common/enums';
 import {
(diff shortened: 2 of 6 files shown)


packages/react-data-grid/src/utils/columnUtils.ts
```diff
@@ -64,22 +65,77 @@ export function getColumnMetrics<R>(metrics: Metrics<R>): ColumnMetrics<R> {
...
+
 function getSpecifiedWidth<R>(
   column: Column<R>,
   columnWidths: Map<keyof R, number>,
   viewportWidth: number,
   minColumnWidth: number
 ): number | void {
+  // get minColumnWidth from column prop or grid prop
+  const minColumnWidthSpecified: number = getMinColumnWidthSpecified(
+    column,
```
qili26: We will only use `column.minWidth` for getting the `minWidth`, so it's better to be specific like 
```suggestion
  const minWidth = getMinWidth(column.minWidth, defaultMinColumnWidth);
```
I also reduced the noise words in the `function` and `variable` to be more concise.

In addition, in the `getMinWidth`, I believe we need to refine the definition for which field to use, as I believe `column.minWidth` and `grid.minColumnWidth` should be precedence relationship rather than comparison relationship, which means when `column.minWidth` is available while it's greater than `0`, we should just return it for the `minWidth` we will be using for the rest. Hence, `getMinWidth` will be like the following
```javascript
function getMinWidth(columnMinWidth, defaultMinColumnWidth): number {
  if (columnMinWidth != null && columnMinWidth > 0) {
    return columnMinWidth;
  }

  return defaultMinColumnWidth;
}
```
And we can further refine it to be just an one line statement with tertiary operator.
```javascript
function getSpecifiedWidth<R>(
  column: Column<R>,
  columnWidths: Map<keyof R, number>,
  viewportWidth: number,
	defaultMinColumnWidth: number
): number | void {
const minWidth = column.minWidth  != null && column.minWidth > 0 ? column.minWidth : defaultMinColumnWidth;

...
```

examples/demos/example13-all-features.js
```diff
@@ -125,15 +125,15 @@ export default class extends React.Component {
     };
   }
 
-  createRows = (numberOfRows) => {
```
jwallet: sorry for my eslint config, will disable it after code review, to set those back

packages/react-data-grid/src/utils/columnUtils.ts
jwallet: sounds good

jwallet: - on load `width` value will be the column width if above `minWidth` value (if `minWidth` undefined, ... above `minColumnWidth` value) otherwise it will be `minWidth || minColumnWidth`
- on resize the width will go down until one of the `minWidth` or `minColumnWidth` is reached
- checkbox SelectColumn can have its `width` re-defined, if by any reason the `width` is undefined, it will fallback on `minWidth`

packages/react-data-grid/src/utils/columnUtils.ts
```diff
@@ -64,22 +65,45 @@ export function getColumnMetrics<R>(metrics: Metrics<R>): ColumnMetrics<R> {
   };
 }
 
+// get column width, fallback on defaultWidth when width is undefined
+function getWidth(
+  viewportWidth: number,
+  minWidth?: number | string,
+  defaultWidth?: number
+): number | void {
+  switch (typeof minWidth) {
```
qili26: https://github.com/adazzle/react-data-grid/pull/1818/files#diff-71bddbe841a1657d0cf905e3011e1fd7R14

`minWidth` is defined as an optional number, so this case will never run unless the dev cheats the RDG like put `minWidth = '100' as any`, however that's not a case we want to cover inside RDG which is why we migrated to the `typescript`, so basically this function is not needed, and we just need to move the code to that `getSpecifiedWidth`.

MayhemYDG:
Wouldn't removing the `Math.max()` call here be enough?
https://github.com/adazzle/react-data-grid/blob/canary/packages/react-data-grid/src/utils/columnUtils.ts#L79
Do we have a need for column-specific min width otherwise?

qili26:
@MayhemYDG It still make sense to be specific.

For example when you have 5 columns while you don’t want 2 specific columns width header text being cropped to display ellipsis accordingly. There is no way to set like this by existing API.

packages/react-data-grid/src/DataGrid.tsx
```diff
@@ -247,7 +247,8 @@ function DataGrid<R, K extends keyof R>({
 
   function handleColumnResize(column: CalculatedColumn<R>, width: number) {
     const newColumnWidths = new Map(columnWidths);
-    width = Math.max(width, minColumnWidth);
+    const minWidth = typeof column.minWidth === 'number' ? column.minWidth : minColumnWidth;
+    width = Math.max(width, minWidth);
```
MayhemYDG: We can simplify a tiny bit:
```ts
width = Math.max(width, column.minWidth || minColumnWidth);
```

packages/react-data-grid/src/utils/columnUtils.ts
```diff
@@ -64,22 +65,45 @@ export function getColumnMetrics<R>(metrics: Metrics<R>): ColumnMetrics<R> {
...
+  defaultMinColumnWidth: number
 ): number | void {
+  // get column min width from grid or column prop, fallback on defaultMinColumnWidth if undefined
+  const minWidth = typeof column.minWidth === 'number' ? column.minWidth : defaultMinColumnWidth;
+  const width = getWidth(viewportWidth, column.width);
+
+  // SelectColumn must always use width prop when defined
+  if (column.key === SelectColumn.key) {
+    return width || minWidth;
+  }
```
MayhemYDG: We shouldn't treat `SelectColumn` as a special case.

We only need to update this function that way:
```ts
  if (typeof column.width === 'number') {
    return column.width;
  }
  if (typeof column.width === 'string' && /^\d+%$/.test(column.width)) {
    return Math.max(Math.floor(viewportWidth * parseInt(column.width, 10) / 100), column.minWidth || minColumnWidth);
  }
  if (typeof column.minWidth === 'number' && column.minWidth > minColumnWidth) {
    return column.minWidth;
  }
```

The new `if` at the end should allow us to properly handle `unallocatedWidth` in `getColumnMetrics`.
For example:
- you have 3 columns with undefined widths, one column has a `minWidth` of `300`
- the container is 500px wide
- `minColumnWidth` is set to `80`
- you'll get 1 column with `300px` width, and the 2 other columns will be `100px` wide.

nstepien:
This has been addressed in c08c29de2781cd5820e83844a8ca6de850508b51

PR #1820, opened 2019-11-08
Fix dragging issue

5joseph9:
When dragging upwards the `fromRow` and `toRow` are set incorrectly.

- [x] Assign correct values to `fromRow` and `toRow` when dragging upwards

### DIFF
--- a/packages/react-data-grid/src/masks/InteractionMasks.tsx
+++ b/packages/react-data-grid/src/masks/InteractionMasks.tsx
@@ -600,10 +600,8 @@ export default class InteractionMasks<R> extends React.Component<InteractionMask
     const column = columns[draggedPosition.idx];
     const value = getSelectedCellValue({ selectedPosition: draggedPosition, columns, rowGetter });
     const cellKey = column.key;
-    const fromRow = rowIdx;
-    const toRow = overRowIdx;
 
-    onGridRowsUpdated(cellKey, fromRow, toRow, { [cellKey]: value }, UpdateActions.CELL_DRAG);
+    onGridRowsUpdated(cellKey, rowIdx, overRowIdx, { [cellKey]: value }, UpdateActions.CELL_DRAG);
 
     this.setState({
       draggedPosition: null
--- a/packages/react-data-grid/src/masks/InteractionMasks.tsx
+++ b/packages/react-data-grid/src/masks/InteractionMasks.tsx
@@ -600,8 +600,8 @@ export default class InteractionMasks<R> extends React.Component<InteractionMask
     const column = columns[draggedPosition.idx];
     const value = getSelectedCellValue({ selectedPosition: draggedPosition, columns, rowGetter });
     const cellKey = column.key;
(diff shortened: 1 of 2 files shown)


packages/react-data-grid/src/masks/InteractionMasks.tsx
```diff
@@ -600,8 +600,8 @@ export default class InteractionMasks<R> extends React.Component<InteractionMask
     const column = columns[draggedPosition.idx];
     const value = getSelectedCellValue({ selectedPosition: draggedPosition, columns, rowGetter });
     const cellKey = column.key;
-    const fromRow = rowIdx < overRowIdx ? rowIdx : overRowIdx;
-    const toRow = rowIdx > overRowIdx ? rowIdx : overRowIdx;
+    const fromRow = rowIdx;
```
qili26: I feel that we should check the dragging orientation and then set the value respectively.

@MayhemYDG  wdth?

MayhemYDG: We set the correct `overRowIdx` value in `handleDragEnter`, so we just have to use it correctly, which is what this PR does.
Not sure what you're trying to suggest.

packages/react-data-grid/src/masks/InteractionMasks.tsx
```diff
@@ -600,8 +600,8 @@ export default class InteractionMasks<R> extends React.Component<InteractionMask
...
+    const toRow = overRowIdx;
 
     onGridRowsUpdated(cellKey, fromRow, toRow, { [cellKey]: value }, UpdateActions.CELL_DRAG);
```
MayhemYDG: @5joseph9 since we're just renaming variables, we could simply use them in the `onGridRowsUpdated` call,

```suggestion
    onGridRowsUpdated(cellKey, rowIdx, overRowIdx, { [cellKey]: value }, UpdateActions.CELL_DRAG);
```

packages/react-data-grid/src/masks/__tests__/InteractionMasks.spec.tsx
```diff
@@ -738,15 +738,17 @@ describe('<InteractionMasks/>', () => {
   });
 
   describe('Drag functionality', () => {
-    const setupDrag = () => {
-      const selectedPosition = { idx: 1, rowIdx: 2 };
+    const setupDrag = (overrideProps?: any) => {
```
MayhemYDG: ESLint is failing here.

packages/react-data-grid/src/masks/InteractionMasks.tsx
qili26: I think your comment replied me. I thought we need to get the correct id from correct dragging orientation, but as long as this is set properly, I am good.

MayhemYDG: That comment is a reply to you indeed.

MayhemYDG:
@5joseph9 do you need me to release a new version asap?

5joseph9:
@MayhemYDG no it is okay you can create the new version with the next release

PR #1822, opened 2019-11-14
Set correct rowIds when copying cell with upward dragging

5joseph9:


packages/react-data-grid/src/ReactDataGrid.tsx
```diff
@@ -390,8 +390,14 @@ export default class ReactDataGrid<R extends {}> extends React.Component<DataGri
     }
 
     const rowIds = [];
-    for (let i = fromRow; i <= toRow; i++) {
-      rowIds.push(rowGetter(i)[rowKey]);
+    if (fromRow >= toRow) {
+      for (let i = fromRow; i <= toRow; i++) {
+        rowIds.push(rowGetter(i)[rowKey]);
```
qili26: Since the action is same, in this case I would prefer DRY which only set the `start` and `end` index based on the `fromRow` and `toRow`, and the benefit would be we don't need to think about whether it's `<=` with `i++` or `>=` with `i--` which is easy to read and maintain.

For example.
```
const start = Math.min(fromRow, toRow);
const end = Math.max(fromRow, toRow);

for (let i = start; i <= end; i++) {
  rowIds.push(rowGetter(i)[rowKey]);
}
```

PR #1824, opened 2019-11-15
Add scroll reach bottom api for lazy loading usage

qili26:
scrollReachBottom


packages/react-data-grid/style/rdg-core.less
```diff
@@ -29,4 +29,5 @@
 .rdg-grid {
   // min-height is here to show the horizontal scrollbar when there are no rows
   min-height: 1px;
+  margin-bottom: 8px;
```
qili26: Add small space between the bottom row and the bottom border, as the editable small square will increase the height size a little bit, so it's better to leave the margin in advance to avoid some randomly bug that a cell cannot be edited.

image

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -110,6 +110,12 @@ export default function Canvas<R, K extends keyof R>({
     }
   }, [rowHeight, scrollToRowIndex]);
 
+  useEffect(() => {
+    if (scrollTop + clientHeight - rowHeight * rowsCount === 8) {
```
qili26: `8` is corresponding to this line.

https://github.com/adazzle/react-data-grid/pull/1824/files#diff-ee619d375026eab3db5bbae675917325R32

packages/react-data-grid/src/common/types.ts
```diff
@@ -254,6 +254,10 @@ export interface ScrollPosition {
   scrollTop: number;
 }
 
+export interface ScrollOption extends ScrollPosition {
+  reachedBottom?: boolean;
```
qili26: We might want to revisit the API name, but I believe this should be a reasonable API.

qili26: Or should I add a stand alone API like `onReachedBottom` to be explicitly?

MayhemYDG:
Why not just surface the scroll event object, then users can just check for themselves:
```ts
event.target.clientHeight + event.target.scrollTop === event.target.scrollHeight
```

qili26:
blocked by https://github.com/adazzle/react-data-grid/pull/1852

qili26:
Superseded by this https://github.com/adazzle/react-data-grid/pull/1854

Issue #1474, opened 2019-02-24
"onGridRowsUpdated" callback is being passed the 'toRow' argument into both the 'toRow' AND 'fromRow' parameters.


SmartArtsStudio:
- Have you already searched for similar issues? Please help us out and double-check first!
   Yes
- Also, please don't be that person who deletes this template. It's here for a reason.
  OK
- Thanks!

---

### Which version of React JS are you using? 
    
✅ Officially supported ✅
- [ ] v15.4.x

⚠️ Not officially supported, expect warnings ⚠️
- [ ] v15.5.x
- [ ] v15.6.x

☣️ Not officially supported, expect warnings and errors ☣️
- [x] v16.x.x     But the issue is inside react-data-grid      

---

### Which browser are you using?
   Chrome

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

⚠️ Not officially supported, but "should work" ⚠️
- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [x] 🐛 Bug Report

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---

### Issue Details

```
Please include:
- What the current behavior is

onGridRowsUpdated callback according to the source code is suppose to provide 
fromRow and toRow parameters  when invoked. However the toRow value is being 
passed into both parameters.

- What the desired behvaior is

onGridRowsUpdated  fromRow and toRow parameters to be passed their respective values.

- (If Bug) Steps to reproduce the issue

```javascript
    // With cellRangeSelection turned on pass the following callback intoappropriate  prop on a
    // <ReactDataGrid onGridRowsUpdated=onGridRowsUpdated >

        onGridRowsUpdated({ cellKey, fromRow, toRow, updated, action }) {
            console.log( 'fromRow: ', fromRow, 'toRow: ', toRow);
        };

// Here is the handlePaste method from InteractionMasks.js implementation I have highlighted the
// toRow duplicated arguments passed into the onGridRowsUpdated invocation.

  handlePaste = () => {
    const { columns, onCellCopyPaste, onGridRowsUpdated } = this.props;
    const { selectedPosition, copiedPosition } = this.state;
    const { rowIdx: toRow } = selectedPosition;

    if (copiedPosition == null) {
      return;
    }

[...]

Issue #1406, opened 2018-11-28
Turning Off Copy and Paste


ZackKnopp:
- Have you already searched for similar issues? Please help us out and double-check first!

- Also, please don't be that person who deletes this template. It's here for a reason.

- Thanks!

---

### Which version of React JS are you using?

✅ Officially supported ✅
- [ ] v15.4.x

⚠️ Not officially supported, expect warnings ⚠️
- [ ] v15.5.x
- [ ] v15.6.x

☣️ Not officially supported, expect warnings and errors ☣️
- [x] v16.x.x

---

### Which browser are you using?

✅ Officially supported ✅
- [ ] IE 9 / IE 10 / IE 11
- [ ] Edge
- [x] Chrome

⚠️ Not officially supported, but "should work" ⚠️
- [ ] Firefox
- [ ] Safari

---

### I'm submitting a ...

- [ ] 🐛 Bug Report
- [x] 💡 Feature Request

> 👋 Need general support? Not sure about how to use React itself, or how to get started with the Grid?
> Please do not submit support request here. Instead see

> https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md

---

### Issue Details

```
Please include:
- What the current behavior is
- What the desired behvaior is
- (If Bug) Steps to reproduce the issue
- (If Feature) The motivation / use case for the feature

We especially love screenshots / videos of problems, and remember
The Best Issue Is A Pull Request™
```

Hi All,

I'm using the new cellRangeSelection functionality to add cell range copy and pasting, however the default one-cell-only copy and pasting is getting in the way of copy and pasting a range from an external source. **Is there any way to turn off the default copy and paste functionality**? I'm using v5.0.4.

Thanks,

Zack

ZackKnopp:
Nevermind, I think I got it. I can handle it in `OnGridRowsUpdated()` by using the 'action' argument and doing something special if it is 'COPY_PASTE'.

I'll upload a gist when I'm done the copy-paste selection code so other people can use it.

Issue #822, opened 2017-06-07
Enable Cell Editing, and Disable Cell Drag


lord-xeon:
> **HAVE YOU ALREADY SEARCHED FOR SIMILAR ISSUES? PLEASE HELP US OUT AND DOUBLE-CHECK FIRST!**

> **ALSO, PLEASE DON'T BE THAT PERSON WHO DELETES THIS TEMPLATE. IT'S HERE FOR A REASON.**

> **THANKS!**

**WHICH VERSION OF REACT ARE YOU USING?**
```
Officially Supported:
[ ] v0.14.x
```

```
Community Supported:
[x] v15.0.x
```

**WHICH BROWSER ARE YOU USING?**

```
Officially Supported:
[ ] IE 9 / IE 10 / IE 11
[ ] Edge
[x] Chrome
```

```
Should work:
[ ] Firefox
[ ] Safari
```

--------

**I'm submitting a ...**  (check one with "x")
```
[ ] bug report
[x] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

**Current behavior**
When you pass in `enableCellSelect={true}` and a `onGridRowsUpdated={props.onRowUpdate}`, you can enter editing mode for cells.  By default this enables the click and drag up/down to copy the given value into the selected cells (provided your `onGridRowsUpdated` function does that.

**Expected/desired behavior**

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem.

**What is the expected behavior?**
I would like a way to still enable cell editing and selection, but disallow the click and drag functionality.

**What is the motivation / use case for changing the behavior?**
Users are not expected/allowed to change that many rows with the same data in the project I'm working on.

Perhaps there is a property that does this already, but I can't seem to find it in the (little) documentation that exists.

audiolion:
Agreed, it looks like in the code if the cell is editable, it adds the drag properties without a way to turn it off

lord-xeon:
I was able to fake it by hiding the drag-handle via CSS.  But I think a programatic way on the grid itself  would be best.

`
.drag-handle {
	display: none;
}
`

dbraynard:
I would like to see a way to disable the `onDoubleClick={this.onDragHandleDoubleClick}` somehow. I like the drag, but I don't like how double clicking can overwrite all cells below that one.

[...]

Issue #669, opened 2017-02-19
Show filter text boxes and set default filtering values in filter text boxes when initializing


sahil37:
**WHICH VERSION OF REACT ARE YOU USING?**
```
Officially Supported:
[ ] v0.14.x
```

```
Community Supported:
[ x] v15.0.x
```

**WHICH BROWSER ARE YOU USING?**

```
Officially Supported:
[ ] IE 9 / IE 10 / IE 11
[ ] Edge
[ x] Chrome
```

```
Should work:
[ ] Firefox
[ ] Safari
```

--------

**I'm submitting a ...**  (check one with "x")
```
[ ] bug report
[x ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/adazzle/react-data-grid/blob/master/CONTRIBUTING.md
```

**Current behavior**

There is no way at the moment to show the filter boxes for each column by default on initialization without the user having the click on the filter rows button. Additionally there is no way to set default filter values for the filter boxes for each column.

**Expected/desired behavior**

A way to show the filter boxes for each column by default on initialization as well as populate these filter boxes for each column with initial values.

**What is the motivation / use case for changing the behavior?**

I am saving the filter values entered by the user on the back-end and I want to use them to repopulate the filter text boxes when the component is initialized again.

dsego:
My workaround to show filters:

```js
    componentDidMount() {
        this.grid.onToggleFilter();
    }
    render() {
         return (
         <ReactDataGrid
                ref={(grid) => { this.grid = grid; }}

```

tranquilize:
```javascript
componentDidMount() {
  this.refs.dataGrid.setState({canFilter: true}, () => this.hideFilterButton())
}

hideFilterButton = () => $('div.react-grid-Toolbar .tools button.btn').hide();
``` 
works for me

mathieuripert:
+1 for having the ability to set default values for filters at initialization.

Prashant89Chaudhary:
+1 for having the ability to set default values for filters at initialization or from any funcation

Prashant89Chaudhary:
Any update on this.

bostondevin:
+1

KevinMongiello:
Here's my workaround.  In this case, I know the order of my columns never change, so I used the element selector in dev tools to find the input i wanted to set default to and voila.
```
componentDidMount() {
		this.grid.onToggleFilter();
		setTimeout(function () {
[...]

PR #1827, opened 2019-11-21
Cleanup props

amanmahajan7:
- [x] Remove `interactionMasksMetaData`
- [ ] <del>Remove `cellMetaData`</del> I will address it in a separate PR
- [x] Remove `onCellCopyPaste` prop and use `onGridRowsUpdated` instead. 
- [x] Remove `getValidFilterValues`. This is only used with `AutoCompleteFilter` and is not generic. It is fairly easy to populate options without this prop
- [x] Replace `onCellSelected` and `onCellDeSelected` with `onSelectedCellChange` prop. I cannot think of a use case where `onCellDeSelected` is needed. Moreover we can always use a local state to keep track of the previously selected cell
- [x] Replace `cellRangeSelection.onStart/onUpdate/onEnd` props with a single `onCellRangeSelectionChange`
- [x] Add `filters/onFiltersChange` props. Currently there is no easy way to set the initial value of the filters https://github.com/adazzle/react-data-grid/issues/669 and it seems reasonable to move the filter [update logic](https://github.com/adazzle/react-data-grid/pull/1827/files#diff-196b1adeb970fd60f04bbc18bf20ec14R53) inside the grid 
- [x] Add `enableCellCopyPaste` prop to toggle copy/paste https://github.com/adazzle/react-data-grid/issues/1406
- [x] Add `enableCellDragAndDrop` prop to toggle drag fill https://github.com/adazzle/react-data-grid/issues/822
- [x] Cleanup `onDragHandleDoubleClick` params
- [x] Revisit `onGridRowsUpdated` params. Do we need `fromRowId`, `toRowId`, and `fromRowData`? These can be calculated from `fromRow` and `toRow` fields. Also fix params for `COPY_PASTE` action https://github.com/adazzle/react-data-grid/issues/1474


amanmahajan7:
@qili26 @MayhemYDG I wanted to make the above changes for sometime and finally had the opportunity to work on them. Please take a look when you have sometime.

examples/demos/example14-all-features-immutable.js
```diff
@@ -189,6 +189,8 @@ export default class extends React.Component {
           onGridRowsUpdated={this.handleGridRowsUpdated}
           rowHeight={50}
           minHeight={600}
+          enableCellCopyPaste
```
qili26: Should we remove this whole example as we don't support immutable any more?

examples/demos/example21-cell-selection-events.js
```diff
@@ -30,20 +30,15 @@ export default class extends React.Component {
     return this._rows[index];
   };
 
-  onCellSelected = ({ rowIdx, idx }) => {
+  onSelectedCellChange = ({ rowIdx, idx }) => {
```
qili26: So there will never be a deselection api, right? What I am trying to say is to not change the example behavior in this case, so perhaps we can add `selectedRowIdx` and `selectedColIdx` in this example `state`, and the next time when the `onSelectedCellChange` is called, the `this.state.selectedRowIdx` and `this.state.selectedColIdx` are the deselected selection and the new selection is `rowIdx, idx`.

packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
```diff
@@ -66,21 +66,22 @@ export default class AutoCompleteFilter extends React.Component {
...
   render() {
+    const { value, multiSelection } = this.props;
+
     return (
       <Select
         autosize={false}
         name={`filter-${this.props.column.key}`}
         options={this.state.options}
-        placeholder={this.state.placeholder}
+        placeholder="Search"
```
qili26: To API standpoint, maybe we should do `placeholder = {this.props.placeHolder != null ? this.props.placeholder : 'Search'}`

packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
```diff
@@ -66,21 +66,22 @@ export default class AutoCompleteFilter extends React.Component {
...
         onChange={this.handleChange}
         escapeClearsValue
-        multi={this.props.multiSelection !== undefined && this.props.multiSelection !== null ? this.props.multiSelection : true}
-        value={this.state.filters}
+        multi={multiSelection !== undefined && multiSelection !== null ? multiSelection : true}
```
qili26:
```suggestion
        multi={multiSelection != null ? multiSelection : true}
```

packages/react-data-grid/src/HeaderRow.tsx
```diff
@@ -50,11 +54,19 @@ export default class HeaderRow<R, K extends keyof R> extends React.Component<Hea
 
   getFilterableHeaderCell(column: CalculatedColumn<R>) {
     const FilterRenderer = column.filterRenderer || FilterableHeaderCell;
+    const { filters, onFiltersChange } = this.props;
+
+    const onChange = (value: unknown) => {
```
qili26: Consider moving this outside and bind this so that the instance would be same in each render? I don't think we should convert this class component to functional component in this PR.

examples/demos/example14-all-features-immutable.js
amanmahajan7: I think we should revisit demos and remove obsolete/duplicate examples. Probably suited for a separate PR

packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
amanmahajan7: I think we should look into deprecating the `addons` package because of the following reasons. 
- It is not very well maintained and does not address all the requirements from the community
- Many of the addons are dependent on third party libraries like `react-select` or `react-dnd` and it is extra work to keep them upto date.
- We should continue improving the API so it is easy to customize it but do not provide any custom implementation. I think eventually we should move all the editors/filters etc to the demos folder but do not publish them. 

wdyt?

packages/react-data-grid/src/HeaderRow.tsx
amanmahajan7: I am going to cleanup header/filter rows in a future PR and convert them to functions

examples/demos/example21-cell-selection-events.js
amanmahajan7: I think it is extra work and probably we can skip it :)

qili26: I believe it's better to keep it so that people won't question where is the `onDeselectedCellChange`. Or at least put a comment about how to use it like `onDeselectedCellChange`

packages/react-data-grid-addons/src/cells/headerCells/filters/AutoCompleteFilter.js
qili26: I'm in favor of removing the `addons` packages as it doesn't provide any value other than examples in my perspective. It's not powerful enough for day to day use that I barely use it except the dragging which we cannot avoid.

examples/demos/example21-cell-selection-events.js
amanmahajan7: I have updated the example

packages/react-data-grid/src/DataGrid.tsx
```diff
@@ -169,22 +161,27 @@ export interface DataGridHandle {
 function DataGrid<R, K extends keyof R>({
   rowKey = 'id' as K,
   rowHeight = 35,
+  headerRowHeight,
```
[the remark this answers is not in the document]
qili26: I would prefer to keep the `headerRowHeight` and revisit once we refactor the `HeaderRow`, I would be flexible to make header height adjustable other than the normal row height.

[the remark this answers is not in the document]
MayhemYDG: Problem is I'm not sure we actually properly support this prop right now.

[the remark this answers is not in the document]
qili26: I think it is working now, as I am using it without any issues so far. So at the moment we could just leave it and revisit.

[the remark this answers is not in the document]
amanmahajan7: I think we can leave it for not but let's revisit it

packages/react-data-grid/src/DataGrid.tsx
```diff
@@ -135,13 +130,10 @@ export interface DataGridProps<R, K extends keyof R> {
...
-  /** Deprecated: Function called when grid is updated via a copy/paste. Use onGridRowsUpdated instead*/
-  onCellCopyPaste?(event: CellCopyPasteEvent<R>): void;
-  /** Function called whenever a cell is selected */
-  onCellSelected?(position: Position): void;
-  /** Function called whenever a cell is deselected */
-  onCellDeSelected?(position: Position): void;
+  /** Function called whenever selected cell is changed */
+  onSelectedCellChange?(position: Position): void;
+  /** Function called whenever selected cell range is changed */
+  onSelectedCellRangeChange?(selectedRange: SelectedRange): void;
```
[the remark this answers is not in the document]
amanmahajan7: It will require some refactoring but I think we can probably use the same event handler. We will have to address a few things
- Currently we attach the mousdown and mouseenter events only when `onCellSelectionChange` is specified. 
- CellRange selection does not work properly when there are frozen columns.

I think it is best to address it separately as the logic can be cleanup

qili26:
@MayhemYDG 
For next release, I believe we need 2 versions, `v7.alpha` to `v7.0.0` and `canary` to `v8`

MayhemYDG:
Nah we'll stay on v7.alpha and v7.canary, we have a lot of breaking changes to make still, and we don't want to end up with `v19` by the end of next year.

amanmahajan7:
@MayhemYDG is there a reason for not merging the canary branch to next? My understanding was that we will not stop work on the Canary branch and fix any integration issues on the next branch.

MayhemYDG:
The `next` branch is the "stable" branch, sort of. We don't merge `canary` onto `next` yet in case we need an emergency fix on the `next` branch. We could add a `pre-canary` or `dev` branch if you want.

amanmahajan7:
Sounds good! If we end up with many open PRs then we can consider creating a `pre-canary` branch

PR #1833, opened 2019-12-02
[canary] Horizontal scroll calculation after removing columns breaks the dataGrid

jwallet:
## Description

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Having 30 columns and the horizontal scroll bar at the far right, then removing some (keep 10 of those) the method getHorizontalRangeToRender tries to move the horizontal scrollbar to the proper column index but fails because there is always a remaining scroll value after substracting the width of each column. it fails trying to get a invalid column index of the column array. columns[10] index = 10. 

**What is the new behavior?**
moved the columnIndex incrementer at the end of the while loop so the condition in the while loop will actually be validated.

```
  while (remainingScroll >= 0 && columnIndex < columns.length) {
    const { width = 0 } = columns[columnIndex];
    remainingScroll -= width;
    if (remainingScroll >= 0) {
      hiddenColumnsWidth += width;
    }
    columnIndex++; <------
  }
```

```
[ ] Yes
[x] No
```

**Other information**:

### DIFF
--- a/packages/react-data-grid/src/utils/viewportUtils.ts
+++ b/packages/react-data-grid/src/utils/viewportUtils.ts
@@ -59,7 +59,12 @@ export function getHorizontalRangeToRender<R>({
   const { columns, totalColumnWidth, lastFrozenColumnIndex } = columnMetrics;
   let { viewportWidth } = columnMetrics;
 
-  let remainingScroll = scrollLeft;
+  // Get valid left scroll position.
+  // When we're scrolled all the way to the right and we remove columns, the scrollLeft value will be outdated,
+  // and greater than what's possible, so we need to handle that case here.
+  let remainingScroll = scrollLeft + viewportWidth > totalColumnWidth
+    ? totalColumnWidth - viewportWidth
+    : scrollLeft;
   let columnIndex = Math.max(lastFrozenColumnIndex, 0);
   let hiddenColumnsWidth = 0;
   while (remainingScroll >= 0 && columnIndex < columns.length) {
--- a/packages/react-data-grid/src/utils/viewportUtils.ts
+++ b/packages/react-data-grid/src/utils/viewportUtils.ts
@@ -65,15 +65,16 @@ export function getHorizontalRangeToRender<R>({
   let remainingScroll = scrollLeft + viewportWidth > totalColumnWidth
     ? totalColumnWidth - viewportWidth
     : scrollLeft;
-  let columnIndex = Math.max(lastFrozenColumnIndex, 0);
+
+  let columnIndex = lastFrozenColumnIndex;
   let hiddenColumnsWidth = 0;
-  while (remainingScroll >= 0 && columnIndex < columns.length) {
+  while (remainingScroll >= 0 && columnIndex < columns.length - 1) {
+    columnIndex++;
     const { width = 0 } = columns[columnIndex];
     remainingScroll -= width;
     if (remainingScroll >= 0) {
       hiddenColumnsWidth += width;
     }
-    columnIndex++;
   }
   const colVisibleStartIdx = Math.max(columnIndex, 0);
 
@@ -82,8 +83,6 @@ export function getHorizontalRangeToRender<R>({
   if (viewportWidth > 0) {
     const firstVisibleColumnHiddenWidth = scrollLeft - hiddenColumnsWidth;
     viewportWidth += firstVisibleColumnHiddenWidth;
-  } else if (remainingScroll > totalColumnWidth) {
-    viewportWidth = scrollLeft - totalColumnWidth;
   } else {
     viewportWidth = totalColumnWidth;
   }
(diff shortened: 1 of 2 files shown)


packages/react-data-grid/src/utils/viewportUtils.ts
```diff
@@ -63,12 +63,12 @@ export function getHorizontalRangeToRender<R>({
   let columnIndex = lastFrozenColumnIndex;
   let hiddenColumnsWidth = 0;
   while (remainingScroll >= 0 && columnIndex < columns.length) {
```
qili26: I believe we only need to do this.
```suggestion
  while (remainingScroll >= 0 && columnIndex <= columns.length - 2) {
```

jwallet: why skipping index 0

qili26: hmm... I think there is another bug, if select column is enabled, i.e. the checkbox column, then we should skip the 0, I believe we need to check exactly like the start condition. However, the end condition should be correct.

wdyt? @MayhemYDG

packages/react-data-grid/src/utils/viewportUtils.ts
```diff
@@ -60,21 +60,29 @@ export function getHorizontalRangeToRender<R>({
   let { viewportWidth } = columnMetrics;
 
   let remainingScroll = scrollLeft;
```
MayhemYDG:
```suggestion
  // Get valid left scroll position.
  // When we're scrolled all the way to the right and we remove columns, the scrollLeft value will be outdated,
  // and greater than what's possible, so we need to handle that case here.
  let remainingScroll = scrollLeft + viewportWidth > totalColumnWidth
    ? totalColumnWidth - viewportWidth
    : scrollLeft;
```

I've tested this locally, and I think this is the only change we need, you can revert the other changes and apply this, unless I've missed something.

packages/react-data-grid/src/utils/viewportUtils.ts
MayhemYDG: @qili26 No? The select column isn't special in this context.

qili26: nvm, I believe I mixed his change and the original code when leaving the comment. If there is no frozen columns, then the `lastFrozenColumnIndex` would just return `-1`, right? So his change for this `Math.max(lastFrozenColumnIndex, 0)` is not needed.

For the loop break condition, shouldn't we do `columnIndex <= columns.length - 2`? The current code will exceed the array boundary and get an `undefined` column, which will break `column.width` when we get the width.

let's say `columns` has 10 elements, and `columnIndex` is 9, `columnIndex++` will result in 10 and `columns[10]` is `undefined` and then we do `remainingScroll -= undefined.width;`. Am I missing anything here?

MayhemYDG: >If there is no frozen columns, then the lastFrozenColumnIndex would just return -1, right? So his change for this Math.max(lastFrozenColumnIndex, 0) is not needed.

Correct.

>For the loop break condition, shouldn't we do `columnIndex <= columns.length - 2`?

Possibly, I'm not 100% certain. Addressing the "impossible" scroll left value fixes this edge case,  and maybe more, so I think it's enough.

>Am I missing anything here?

As long as we use valid dimensions/scroll positions, then `remainingScroll >= 0` will break us out of the loop before we try to get an out of bounds column.

packages/react-data-grid/src/utils/viewportUtils.ts
```diff
@@ -60,21 +60,29 @@ export function getHorizontalRangeToRender<R>({
...
-  let columnIndex = lastFrozenColumnIndex;
+  let columnIndex = Math.max(lastFrozenColumnIndex, 0);
   let hiddenColumnsWidth = 0;
   while (remainingScroll >= 0 && columnIndex < columns.length) {
-    columnIndex++;
-    const column = columns[columnIndex];
```
jwallet: @MayhemYDG even with your change above, this will fail because the end validation inside while() does nothing, because the columnIndex is incremented after validating the end condition
`columnIndex < columns.length` : `0 < 1`
`columnIndex++` : `1`
`columns[columnIndex` : `columns[1] // invalid`

packages/react-data-grid/src/utils/viewportUtils.ts
jwallet: > Possibly, I'm not 100% certain. Addressing the "impossible" scroll left value fixes this edge case, and maybe more, so I think it's enough.
so `columnIndex < columns.length` will always return true then, why keeping it

MayhemYDG: >so `columnIndex < columns.length` will always return true then, why keeping it

You're right, let's remove it.

This method is kind of confusing, I think I'll refactor it after this PR.

MayhemYDG:
Some tests are still failing, but I'll go ahead and merge this, and do a follow up PR.

PR #1835, opened 2019-12-04
Update dependencies and eslint rules

MayhemYDG:


packages/react-data-grid-addons/package.json
```diff
@@ -27,7 +27,7 @@
   "dependencies": {
     "lodash": "^4.17.15",
     "prop-types": "^15.7.2",
-    "react-contextmenu": "^2.11.0",
+    "react-contextmenu": "^2.13.0",
     "react-data-grid": "^7.0.0-canary.6",
     "react-dnd": "^2.6.0",
```
qili26: Do we want to upgrade this as well?

MayhemYDG: Eventually yes, if we still have a need for it.

amanmahajan7:
FYI https://twitter.com/brian_d_vaughn/status/1191792689563357184

PR #1836, opened 2019-12-05
Remove support for getRowMetaData, dependentValues, rowMetaData

MayhemYDG:
To be merged into `canary` once `canary` stabilizes and we merge it onto `next`.

### DIFF
--- a/packages/react-data-grid-addons/src/draggable/RowDragLayer.js
+++ b/packages/react-data-grid-addons/src/draggable/RowDragLayer.js
@@ -68,8 +68,7 @@ class CustomDragLayer extends Component {
         if (item.hasOwnProperty(c.key)) {
           if (c.formatter) {
             const Formatter = c.formatter;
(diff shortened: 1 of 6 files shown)


MayhemYDG:
@amanmahajan7 IMO, seeing the glacial pace at which we can adopt the ongoing rdg work, we may as well merge this now, or we'll have to wait for another 6-12 months.

amanmahajan7:
@qili26  are you okay with merging this PR?

qili26:
Yes, I think we could merge this. However, I think we should add something to the migration plan later, like for a shot time, you can customize like 
```
const cellRenderer = () => {
  // not from column anymore, if the dev prefer, they can accommodate the logic here.
  const metaData = getRowMetaData(); 
  return <CellContent value={metaData.value} />;
}
```

MayhemYDG:
Yes devs can either wrap their formatters on a per-column basis, if they're using the legacy cell content renderer, or edit their custom cell content renderers, shouldn't be too much of an issue I think.

PR #1837, opened 2019-12-05
Refactor header components

amanmahajan7:
- [x] Remove `cellMetaData`
- [x] Render visible header cells.
- [x] Remove `setScrollLeft` method and use `scrollLeft` prop to freeze columns in IE
- [x] Simplify `HeaderRow` component
  - [x] Add `FilterRow` component and move all the filter logic to it
  - [x] Use component composition to handle sorting and resizing
- [x] Remove `scrollToRowIndex` prop and add a new `scrollToRow` method. `scrollToRowIndex` only works if new value is different
- [x] Fix unit tests


rozeskjm:
Got bored did you? 😉

packages/react-data-grid/src/Cell.tsx
```diff
@@ -23,31 +24,49 @@ function Cell<R>({
...
+  onRowClick,
+  onRowDoubleClick,
+  onDeleteSubRow,
+  onCellExpand,
+  getCellActions,
+  enableCellRangeSelection
 }: CellProps<R>) {
+  const position: Position = { idx, rowIdx };
+
+  function selectCell(openEditor?: boolean) {
```
amanmahajan7: I added ``eventBus` prop as it easier to pass it down instead of all the event handlers. It also works better with memo as we do not have to use `useCallback` everywhere

packages/react-data-grid/src/DataGrid.tsx
```diff
@@ -92,8 +91,6 @@ export interface DataGridProps<R, K extends keyof R> {
   rowsCount: number;
   /** The minimum height of the grid in pixels */
   minHeight?: number;
-  /** When set, grid will scroll to this row index */
-  scrollToRowIndex?: number;
```
amanmahajan7: This is the only breaking change in this PR

packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -1,168 +1,118 @@
...
-    }
-  }
+  let cell = getCell();
 
-  removeScroll() {
-    const node = this.cell.current;
-    if (node) {
-      node.style.transform = 'none';
-    }
+  if (column.sortable) {
```
amanmahajan7: I think composition makes this code cleaner and easier to maintain and can add another `if (draggable)` block in the future. It will be nice to completely drop `react-dnd` dependency

packages/react-data-grid/src/SummaryRowRenderer.tsx
```diff
@@ -4,28 +4,28 @@ import Row from './Row';
...
   columnMetrics,
   idx,
   rowData,
   rowHeight,
-  scrollLeft
+  scrollLeft,
+  eventBus
 }: SummaryRowRendererProps<R, K>) {
   return (
     <Row<R>
```
amanmahajan7: Do we need this component? It does not have any logic and just forward the props

packages/react-data-grid/src/common/types.ts
```diff
@@ -168,11 +161,13 @@ export type CellContentRendererProps<TRow> = Pick<CellRendererProps<TRow>,
...
 | 'rowData'
 | 'column'
-| 'cellMetaData'
 | 'expandableOptions'
 | 'isRowSelected'
 | 'onRowSelectionChange'
 | 'isSummaryRow'
+| 'onDeleteSubRow'
+| 'onCellExpand'
+| 'getCellActions'
```
amanmahajan7: I am not sure why we added these props but SubRows and CellActions can be handled via a custom formatter

packages/react-data-grid/src/masks/InteractionMasks.tsx
```diff
@@ -594,11 +587,29 @@ export default class InteractionMasks<R, K extends keyof R> extends React.Compon
...
+      cellKey,
+      fromRow: selectedPosition.rowIdx,
+      toRow: rowsCount - 1,
+      updated: { [cellKey]: value } as never,
+      action: UpdateActions.COLUMN_FILL
+    });
   };
 
-  onCommit = (args: CommitEvent<R>): void => {
-    this.props.onCommit(args);
```
amanmahajan7: `onCommit` and `onDragHandleDoubleClick`  call `onGridRowsUpdated` so I just moved the logic here

packages/react-data-grid/src/DataGrid.tsx
```diff
@@ -205,6 +194,14 @@ function DataGrid<R, K extends keyof R>({
     });
   }, [columnWidths, columns, defaultCellContentRenderer, minColumnWidth, viewportWidth]);
 
+  const horizontalRange = useMemo(() => {
```
MayhemYDG:
```tsx
  const [horizontalRange, viewportColumns] = useMemo((): [HorizontalRangeToRender | null, CalculatedColumn<R>[]] => {
    if (!columnMetrics) return [null, []];

    const horizontalRange = getHorizontalRangeToRender({
      columnMetrics,
      scrollLeft
    });

    const viewportColumns = getViewportColumns(
      columnMetrics.columns,
      horizontalRange.colOverscanStartIdx,
      horizontalRange.colOverscanEndIdx
    );

    return [horizontalRange, viewportColumns];
  }, [columnMetrics, scrollLeft]);
```

WDYT?

packages/react-data-grid/src/DataGrid.tsx
```diff
@@ -394,25 +276,43 @@ function DataGrid<R, K extends keyof R>({
...
+              rowsCount={rowsCount}
+              rowGetter={rowGetter}
+              height={headerRowHeight}
+              width={columnMetrics.totalColumnWidth + getScrollbarSize()}
+              columns={viewportColumns}
+              onColumnResize={handleColumnResize}
+              lastFrozenColumnIndex={columnMetrics.lastFrozenColumnIndex}
+              draggableHeaderCell={props.draggableHeaderCell}
+              onHeaderDrop={props.onHeaderDrop}
+              allRowsSelected={selectedRows !== undefined && selectedRows.size === rowsCount}
```
MayhemYDG:
```suggestion
              allRowsSelected={selectedRows?.size === rowsCount}
```

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -135,39 +115,49 @@ export default function Canvas<R, K extends keyof R>({
 
   function onHitBottomCanvas({ rowIdx }: Position) {
     const { current } = canvas;
-    if (current) {
-      // We do not need to check for the index being in range, as the scrollTop setter will adequately clamp the value.
-      current.scrollTop = (rowIdx + 1) * rowHeight - clientHeight;
-    }
+    if (!current) return;
+    // We do not need to check for the index being in range, as the scrollTop setter will adequately clamp the value.
+    current.scrollTop = (rowIdx + 1) * rowHeight - clientHeight;
```
MayhemYDG:
```suggestion
    canvas.current?.scrollTop = (rowIdx + 1) * rowHeight - clientHeight;
```

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -135,39 +115,49 @@ export default function Canvas<R, K extends keyof R>({
...
+    const newScrollLeft = getColumnScrollPosition(columnMetrics.columns, idx, scrollLeft, clientWidth);
     if (newScrollLeft !== 0) {
       current.scrollLeft = scrollLeft + newScrollLeft;
     }
   }
 
+  function scrollToRow(rowIdx: number) {
+    const { current } = canvas;
+    if (!current) return;
+    current.scrollTop = rowIdx * rowHeight;
```
MayhemYDG:
```suggestion
    canvas.current?.scrollTop = rowIdx * rowHeight;
```

packages/react-data-grid/src/Canvas.tsx
```diff
@@ -176,6 +166,13 @@ export default function Canvas<R, K extends keyof R>({
     }
   }, [rowRefs]);
 
+  useImperativeHandle(ref, () => ({
```
MayhemYDG: That's going to update the ref on every render 🤔 
Hope no one uses a complicated ref function prop.

packages/react-data-grid/src/FilterRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+          width: column.width,
+          left: column.left
+        };
+
+        if (colIsFrozen && typeof scrollLeft === 'number') {
+          style.transform = `translateX(${scrollLeft}px)`;
+        }
+
+        return (
+          <div
```
MayhemYDG:
```suggestion
          <div
            key={key}
```

packages/react-data-grid/src/FilterRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+  columns,
+  lastFrozenColumnIndex,
+  scrollLeft,
+  filters,
+  onFiltersChange
+}: FilterRowProps<R, K>) {
+  function onChange(key: keyof R, value: unknown) {
+    const newFilters: Filters<R> = { ...filters };
+    newFilters[key] = value;
+    onFiltersChange && onFiltersChange(newFilters);
```
MayhemYDG:
```suggestion
    onFiltersChange?.(newFilters);
```

packages/react-data-grid/src/FilterRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+        if (colIsFrozen && typeof scrollLeft === 'number') {
+          style.transform = `translateX(${scrollLeft}px)`;
+        }
+
+        return (
+          <div
+            style={style}
+            className={className}
+          >
+            {key !== 'select-row' && column.filterable && createElement(renderer, {
```
MayhemYDG: Let's add a `filterRenderer` to `SelectColumn` that just renders null, we shouldn't have special cases for columns.

packages/react-data-grid/src/FilterRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+              column,
+              value: filters ? filters[column.key] : undefined,
```
MayhemYDG:
```suggestion
              value: filters?.[key],
```

packages/react-data-grid/src/Row.tsx
```diff
@@ -42,38 +42,39 @@ export default class Row<R> extends React.Component<IRowRendererProps<R>> {
...
-          scrollLeft={colIsFrozen && typeof scrollLeft === 'number' ? scrollLeft : undefined}
-          isRowSelected={isRowSelected}
-          onRowSelectionChange={onRowSelectionChange}
-          isSummaryRow={isSummaryRow}
-        />
-      );
-    }
-
-    return cellElements;
+    return getViewportColumns(columns, colOverscanStartIdx, colOverscanEndIdx)
```
MayhemYDG: Don't we pre-filter the visible columns now? We should only pass the visible columns so we don't have to call `getViewportColumns` in each Row component, wdyt?

packages/react-data-grid/src/DataGrid.tsx
```diff
@@ -177,21 +169,18 @@ function DataGrid<R, K extends keyof R>({
...
-  onSelectedCellRangeChange,
   selectedRows,
   onSelectedRowsChange,
-  filters,
-  onFiltersChange,
   ...props
 }: DataGridProps<R, K>, ref: React.Ref<DataGridHandle>) {
   const [columnWidths, setColumnWidths] = useState(() => new Map<keyof R, number>());
-  const [eventBus] = useState(() => new EventBus());
+  const [scrollLeft, setScrollLeft] = useState(0);
```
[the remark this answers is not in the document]
amanmahajan7: I profiled it and It has not affected the performance. The only extra work is rendering the header cells on scroll which is minimal as we are only rendering the visible cells now. If that becomes an issue then we can memoized  the header cells as well. I am not sure if there is another way to share the horizontal range

packages/react-data-grid/src/Canvas.tsx
amanmahajan7: Is this a valid syntax?

> The left-hand side of an assignment expression may not be an optional property access

MayhemYDG: I guess not. 😕

packages/react-data-grid/src/utils/viewportUtils.ts
```diff
@@ -1,4 +1,5 @@
 import { CalculatedColumn, ColumnMetrics } from '../common/types';
+import { isFrozen } from './columnUtils';
```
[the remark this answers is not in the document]
amanmahajan7: I agree, let's tackle this in a follow up PR

packages/react-data-grid/src/HeaderCell.tsx
qili26: I think it's still worth using the `react-dnd` as it provides some features that we don't want to rework. I have a PR to upgrade it by using v10, will assign you for review later.

packages/react-data-grid/src/Row.tsx
amanmahajan7: I thought about it but decided to keep this PR simple. `viewportColumns` is recomputed when `scrollLeft` is changed so it may break memoization. Let's revisit it in the future

packages/react-data-grid/src/RowGroup.tsx
```diff
@@ -52,9 +52,9 @@ export default forwardRef<HTMLDivElement, Props<any>>(function RowGroup(props, r
       <Renderer {...props} ref={ref} onRowExpandClick={onRowExpandClick} onRowExpandToggle={onRowExpandToggle} />
     </div>
   );
-});
+}) as <R>(props: Props<R>) => JSX.Element;
```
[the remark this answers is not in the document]
amanmahajan7: We do pass the ref but I am not too sure if it is needed or not. We need to check if `renderBaseRow` is needed anymore or not. IIRC `renderBaseRow` was added because there was no easy way to forward the ref earlier

MayhemYDG:
Looks like I accidentally force-pushed this branch, not sure how. I've fixed it back to how it was.

packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -1,168 +1,118 @@
...
+  const style: React.CSSProperties = {
+    width: column.width,
+    left: column.left
+  };
 
-    const className = classNames('rdg-cell', {
-      'rdg-cell-frozen': colIsFrozen,
-      'rdg-cell-frozen-last': colIsFrozen && column.idx === this.props.lastFrozenColumnIndex
-    }, this.props.className, column.cellClass);
+  if (typeof scrollLeft === 'number') {
```
MayhemYDG: Do we not need to check if the column is frozen here as well?
We seem to do it in `FilterRow.tsx` and `Row.tsx`.

packages/react-data-grid/src/RowGroup.tsx
```diff
@@ -1,15 +1,15 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import React, { forwardRef } from 'react';
 import { EventTypes } from './common/enums';
-import { CellMetaData, IRowRendererProps, CalculatedColumn } from './common/types';
+import { IRowRendererProps, CalculatedColumn, RowExpandToggleEvent, Omit, CellRendererProps } from './common/types';
 import EventBus from './EventBus';
 
 interface Props<R> {
+  ref?: React.Ref<HTMLDivElement>;
```
MayhemYDG: `ref` Refs shouldn't be on the `Props` type. otherwise you end up with `props.ref` being possible type-wise, when it's not in practise.

packages/react-data-grid/src/common/cells/headerCells/ResizableHeaderCell.tsx
```diff
@@ -0,0 +1,100 @@
...
+  column: CalculatedColumn<R>;
+  onResize(column: CalculatedColumn<R>, width: number): void;
+}
+
+export default function ResizableHeaderCell<R>({
+  children,
+  column,
+  ...props
+}: ResizableHeaderCellProps<R>) {
+  const cellRef = useRef<HTMLDivElement>(null);
```
MayhemYDG: We can get rid of this ref, and just pass `event.currentTarget` around instead.

packages/react-data-grid/src/common/cells/headerCells/ResizableHeaderCell.tsx
```diff
@@ -0,0 +1,100 @@
+import React, { useRef, cloneElement } from 'react';
+import { CalculatedColumn } from '../../../common/types';
+
+export interface ResizableHeaderCellProps<R> {
+  children: React.ReactElement<React.ComponentProps<'div'>>;
```
MayhemYDG: Small simplification:
```suggestion
  children: React.ReactHTMLElement<HTMLDivElement>;
```
`React.ReactComponentElement<'div'>` is also available, but it won't let us set the ref with it for some reason.

packages/react-data-grid/src/HeaderCell.tsx
amanmahajan7: We need to check it as we set `scrollLeft` on frozen columns only. `FilterRow` does not use `HeaderCell` component or the `Row` component, it directly renders the filter component. We can probably do some clean up and extract this common logic somewhere

packages/react-data-grid/src/Cell.tsx
```diff
@@ -23,31 +24,49 @@ function Cell<R>({
...
+    eventBus.dispatch(EventTypes.SELECT_CELL, position, openEditor);
+  }
+
   function handleCellClick() {
-    cellMetaData.onCellClick({ idx, rowIdx });
+    selectCell();
+    onRowClick?.(rowIdx, rowData, column);
```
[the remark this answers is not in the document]
amanmahajan7: We probably can but lets revisit it. I was thinking of using `useReducer` and pass down `dispatch` instead of `eventBus`, not sure if it is possible but it uses more of less the same pattern.

packages/react-data-grid/src/common/cells/headerCells/ResizableHeaderCell.tsx
amanmahajan7: Hmm, seems to be failing

> src/HeaderCell.tsx:98:9 - error TS2741: Property 'ref' is missing in type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)>' but required in type 'ReactHTMLElement<HTMLDivElement>'.

packages/react-data-grid/src/RowGroup.tsx
amanmahajan7: I made a small change, let me know if there is a better way

MayhemYDG: Yup looks good.

packages/react-data-grid/src/common/cells/headerCells/ResizableHeaderCell.tsx
MayhemYDG: Ah, nevermind then.

MayhemYDG:
@amanmahajan7 Feel free to merge it.

PR #1841, opened 2019-12-10
react dnd v10 support

qili26:
To replace the existing react-dnd v2 by v10 with some existing drag / drop clean up.


packages/react-data-grid/src/HeaderCell.tsx
```diff
@@ -146,23 +147,16 @@ export default class HeaderCell<R> extends React.Component<HeaderCellProps<R>> {
         onMouseDown={column.resizable ? this.onMouseDown : undefined}
         onTouchStart={column.resizable ? this.onTouchStart : undefined}
       >
-        {this.getCell()}
+        {rowType === HeaderRowType.HEADER && column.draggable && DraggableHeaderCell ? (
```
qili26: Move the `dragTarget` to cell inner level so that we don't see the issue in this link.

https://github.com/react-dnd/react-dnd/issues/1608

qili26:
Changed files records were not correct, probably it is due to the inappropriate process taken from `rebase + pull`. I just cherry picked another PR to supersede this.

PR #1843, opened 2019-12-12
Drop lerna, start ignoring rdg-addons

MayhemYDG:
I think we can stop working on rdg-addons now. We can keep using the previously published versions until we completely integrate its features into the main package.

- Started using some ES modules for our node scripts, so make sure you're on the latest Node.js version
- Deleted the `.github/` files. The repo's not in a state where we can provide proper support, so I feel like the templates were not useful as they were, and are not as relevant as before. We'll add them back once we have an actual stable version we can support.
- `packages/react-data-grid-addons/` is still there, but it'll now be ignored by the tools like eslint and we won't build it anymore.
- `npm start` works without having to build the lib all the time.
  - TS-only now, JS is not allowed anymore.

---

- [x] Merge the READMEs
- [x] Fix publishing guide in CONTRIBUTING.md
- [ ] Should we finish porting the examples to TS?


amanmahajan7:
@MayhemYDG I think we can create `pre-canary` branch as there are many PRs that will be ready soon. It will probably be better to start merging things so there are less conflicts

MayhemYDG:
Sure, you can go ahead, or I'll do it tomorrow morning.

qili26:
Do we want to finish this PR first and then do the removal?

https://github.com/adazzle/react-data-grid/pull/1841

MayhemYDG:
I'm fine doing it either way.

MayhemYDG:
@qili26 @amanmahajan7 since I'm off until January maybe we could merge this now, and we'll check the dnd PR later?

amanmahajan7:
Is this PR ready? Looks like we comment some code including examples

MayhemYDG:
It's ready for review. I've made it so running the examples will only allow ts (from src or examples) from now on, so I haven't TSified all the examples yet, as we can do it later as we need it. Also to avoid git conflicts.

amanmahajan7:
I will check

PR #1845, opened 2019-12-15
Remove getCellActions prop

amanmahajan7:
A separate `getCellActions` prop is not required and cell actions feature can be easily implemented using a custom formatter.

Next we can remove props to implement subrows also


examples/demos/example27-cell-actions.js
```diff
@@ -52,7 +53,42 @@ const columns = [
   {
     key: 'county',
     name: 'County',
-    width: 200
+    width: 200,
+    cellClass: 'rdg-cell-action',
+    formatter({ row, value }) {
```
MayhemYDG: If you want we can use the new `cellContentRenderer` API here.
```js
    cellContentRenderer({ rowData }) {
      const value = rowData.county;
      if (rowData.id === 'id_0') {
```

amanmahajan7: We probably do not need `cellContentRenderer` once subrows functionality is removed. We can also delete `CellContent` component and use `formatter` as `cellContentRenderer`. We can then replace `defaultCellContentRenderer` with `defaultFormatter`. Wdyt? 

We can merge this PR after the lerna removal PR

MayhemYDG: I agree.

IMO RDG shouldn't access the value from the row object, so we can have renderers like
```js
({ rowData }) => rowData.nested.value;
({ rowData }) => rowData.x + rowData.y;
````
for example. And it might simplify types as well, maybe.

qili26:
Is this `getCellActions()` only used for customized html actions passed into RDG cell? I had never used this before, so want to double check.

amanmahajan7:
That is correct. It basically provides a way to add actions to a cell but the same functionality can be achieved with a custom formatter. Check the cell actions example. Also with this change we can use any component for the actions. .

ValYouW:
@amanmahajan7 columns definition is usually done "outside" of the component, same goes for the formatter, which makes it difficult to do anything with an action fired from within the formatter - no way to "communicate" with component.
If the params to the formatter included some callback function that when called triggers another callback on the grid it would have been easier. Something like `onRowChange` that is passes to `Editor2`

PR #1849, opened 2019-12-16
Compute viewportColumns only once in DataGrid

MayhemYDG:


qili26:
@amanmahajan7 IMHO Lerna PR seems way too big now that we might want to merge it asap.

MayhemYDG:
I've split computing `viewportColumns` into a separate `useMemo`, so we don't end up with a new array on each `scrollLeft` change, that should prevent unnecessary memoized component re-renders.

qili26:
@MayhemYDG Do we really need to separate it? Whenever `columnMetrics` changes, they both changes, the the new added `useMemo` is using the result of the previous memo result, so they will always change together, won't they?

MayhemYDG:
>Whenever `columnMetrics` changes, they both changes

Yes, that's how it should be. `columnMetrics` shouldn't change often though, check its `useMemo` dependencies.

>the new added `useMemo` is using the result of the previous memo result, so they will always change together, won't they?

Yes and no, `horizontalRange` will always change on `scrollLeft` change for example, but `horizontalRange.colOverscanEndIdx` and `horizontalRange.colOverscanStartIdx` on the other hand won't always change on `scrollLeft` change. Again, check the dependencies used for `useMemo`.

PR #1850, opened 2019-12-16
update vscode-eslint settings

MayhemYDG:
Make sure to update your eslint vscode extension. It'll finally fix all the issues at once 👏 
https://github.com/microsoft/vscode-eslint/blob/bc6f9f98d0b35d875875c9729b8ec864a87d266e/README.md#release-notes

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,18 +1,7 @@
 {
+  "editor.codeActionsOnSave": {
+    "source.fixAll": true
+  },
   "eslint.enable": true,
-  "eslint.autoFixOnSave": true,
-  "eslint.validate": [{
-    "language": "javascript",
-    "autoFix": true
-  }, {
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -1,18 +1,7 @@
 {
+  "editor.codeActionsOnSave": {
+    "source.fixAll": true
+  },
   "eslint.enable": true,
-  "eslint.autoFixOnSave": true,
-  "eslint.validate": [{
```
qili26: don’t we need to do this?

“eslint.validate": ["javascript", ..., “typescript”]

MayhemYDG: `eslint.validate` is deprecated and `eslint.probe` should be used instead. `eslint.probe` defaults to supporting js and ts so we don't need to manually configure it.

PR #1851, opened 2019-12-20
react-dnd v10 upgrade

qili26:
1. upgrade react-dnd to v10
2. move some draggable header related code from addons package to example.
3. remove the draggable container and move into the DataGrid src.


examples/demos/example21-grouping.js
```diff
@@ -1,42 +1,46 @@
+import faker from 'faker';
```
qili26: I tried to convert it to tsx but it will eventually touch tons of files like grouping related js files, so maybe we should do the conversion later in another PR.

examples/demos/example23-immutable-data-grouping.js
```diff
@@ -1,13 +1,11 @@
+import faker from 'faker';
```
qili26: Do we want to just delete this file?

src/DataGrid.tsx
```diff
@@ -298,8 +304,9 @@ function DataGrid<R, K extends keyof R>({
               columns={viewportColumns}
               onColumnResize={handleColumnResize}
               lastFrozenColumnIndex={columnMetrics.lastFrozenColumnIndex}
-              draggableHeaderCell={props.draggableHeaderCell}
-              onHeaderDrop={props.onHeaderDrop}
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              draggableHeaderCell={draggableHeaderCell || (onHeaderDrop && DraggableHeaderCell as React.ComponentType<any>)}
```
qili26: Not sure how to fix this type at the moment.

qili26:
No tests were added at the moment as not sure how to test the Drag feature yet.

src/DataGrid.tsx
```diff
@@ -1,37 +1,47 @@
...
 } from 'react';
+import { DndProvider } from 'react-dnd';
+import HTML5Backend from 'react-dnd-html5-backend';
 import { isValidElementType } from 'react-is';
 
-import HeaderRow from './HeaderRow';
-import FilterRow from './FilterRow';
 import Canvas, { CanvasHandle as DataGridHandle } from './Canvas';
 import { legacyCellContentRenderer } from './Cell/cellContentRenderers';
-import { getColumnMetrics, getHorizontalRangeToRender, isPositionStickySupported, getViewportColumns, getScrollbarSize, HorizontalRangeToRender } from './utils';
```
qili26: import order was changed by automated auto order `cmd`.

style/rdg-cell.less
```diff
@@ -149,3 +149,21 @@
...
+  cursor: grabbing;
+}
+
+.rdg-draggable-header-cell {
+  cursor: move;
+  display: inline-block;
+  height: inherit;
+  background-color: inherit;
+  width: 100%;
+  margin-left: -8px;
```
qili26: This will affect the dragging preview. Without this, the preview text will have no left space in the preview container. Maybe we can revisit the `header` css together later.

src/HeaderCell.tsx
```diff
@@ -35,7 +35,7 @@ export default function HeaderCell<R, K extends keyof R>({
   scrollLeft,
   ...props
 }: HeaderCellProps<R, K>) {
-  function getCell() {
+  function getCell(): React.ReactText | JSX.Element {
```
nstepien: We don't need to enforce a return type here.

src/HeaderCell.tsx
```diff
@@ -50,7 +50,15 @@ export default function HeaderCell<R, K extends keyof R>({
...
-  let cell = getCell();
+  const DraggableHeaderCell = props.draggableHeaderCell;
+  let cell = column.draggable && DraggableHeaderCell ? (
+    <DraggableHeaderCell
+      column={column}
+      onHeaderDrop={props.onHeaderDrop!}
+    >
+      {getCell()}
+    </DraggableHeaderCell>
+  ) : getCell();
```
nstepien: Why are we moving the draggable header wrapping here?

examples/demos/GroupedColumnsPanel.js
```diff
@@ -0,0 +1,87 @@
+import React, { Component } from 'react';
```
nstepien: Do we plan on using these js files?

examples/demos/example23-immutable-data-grouping.js
nstepien: If they don't show something unique compared to other demos, we can remove them.

nstepien: IMO let's not edit js files without tsifying them. You can revert individual files like this:
```sh
git fetch
git checkout origin/pre-canary -- examples/demos/example23-immutable-data-grouping.js
```

examples/index.tsx
```diff
@@ -40,6 +40,9 @@ ReactDOM.render((
       <ul className="nav">
         <li><NavLink to="/">All Features Example</NavLink></li>
```
nstepien: We missed this in the previous PR, but we should add the `exact` prop here so we don't always highlight this link:
```suggestion
        <li><NavLink to="/" exact>All Features Example</NavLink></li>
```

package.json
```diff
@@ -71,7 +71,9 @@
     "markdown": "^0.5.0",
     "react": "^16.12.0",
     "react-contextmenu": "^2.13.0",
-    "react-dnd-test-backend": "^2.6.0",
+    "react-dnd": "^10.0.2",
+    "react-dnd-html5-backend": "^10.0.2",
```
nstepien: Looks like `react-dnd` and `react-dnd-html5-backend` are `dependencies` now, not `devDependencies`.

src/index.ts
```diff
@@ -1,11 +1,11 @@
...
-export * from './formatters';
+export * from './Cell/cellContentRenderers';
+export * from './Columns';
 export * from './common/editors';
 export * from './common/enums';
 export * from './common/types';
+export { DataGridHandle, DataGridProps, default } from './DataGrid';
+export * from './formatters';
+export { default as HeaderCell } from './HeaderCell';
+export { default as Row } from './Row';
```
nstepien: Why rearrange the exports? 😐

src/DataGrid.tsx
nstepien: I don't like this auto order thing. 🤷‍♂

examples/demos/example23-immutable-data-grouping.js
qili26: Those immutable / group files were changed before your addon removal PR, so the changes were kept here as well. I thought it's no harm to keep them as they were working. I will try to clean them then.

src/DataGrid.tsx
qili26: hmm... I really don't like to spend time to manually manage the imports. Though with that being said, if you guys both agree to manually handle this, I can stop using that.

One of the benefit of using this can be merging files. For example, when imports lines has conflicts, we can just accept both and use the short cut key to remove any redundant imports (this is part of the re-ordering) without checking them one by one.

src/HeaderCell.tsx
qili26: It's the preview thing I mentioned before. It's a bug only in Chrome. Without this change, the dragging preview will show the whole column headers rather than the dragged one.

examples/demos/GroupedColumnsPanel.js
qili26: Like I mentioned in the previous comment, this was used in the Group example but we cannot tsfy them at the moment. However, it's working from my old PR changes.

examples/index.tsx
qili26: will create a new PR to add this.

package.json
qili26: Yep, I missed that after the addon removal.

src/index.ts
qili26: same from the auto order cmd, it just sort the import export alphabetically.

amanmahajan7:
I have not reviewed the code but I was wondering if is it possible to use a custom `headerRenderer` and use a `DraggableHeaderCell`? This way we can avoid adding adding `react-dnd` dependency. 

```
headerRenderer(props) {
  <DraggableHeaderCell
      column={props.column}
      onHeaderDrop={onHeaderDrop}
    >
      {column.name}
    </DraggableHeaderCell>
}
```

qili26:
@amanmahajan7 
This would be the first iteration for upgrading the `react-dnd` to be v10. Next is to change to use `react-dnd` hook. Then we can think if we could completely remove the dependency, but I feel in a short term we cannot.

PR #1852, opened 2019-12-23
tsfy all features example

qili26:
After pre-canary, only 1 base example was tsfied. This is to just tsfy the all features example so that when we have new codes pushed, we have a demo page for examination.

1. Tsfy all features exmaple.
1.1 Move DropdownEditor to the demo folder with tsfying.
1.2 Move Toolbar to the demo folder
1.3 Move ImageFormatter to the demo folder.
2. Remove immutable examples.
3. Redirect to all features example by default.


tsconfig.json
```diff
@@ -20,6 +20,7 @@
     "types": ["jest"]
   },
   "files": [
-    "src/index.ts"
+    "src/index.ts",
+    "examples/index.tsx"
```
nstepien: Running `npm run build` will also build the examples with this change, we don't want to do that.

examples/index.tsx
```diff
@@ -73,44 +67,44 @@ ReactDOM.render((
...
+        {/* <Route path="/examples/draggable-header" component={DraggableHeader} /> */}
+        {/* <Route path="/examples/tree-view" component={TreeView} /> */}
+        {/* <Route path="/examples/tree-view-no-add-delete" component={TreeViewNoAddDelete} /> */}
+        {/* <Route path="/examples/cell-actions" component={CellActions} /> */}
+        {/* <Route path="/examples/scroll-to-row-index" component={ScrollToRowIndex} /> */}
+        {/* <Route path="/examples/descendingFirstSortable" component={DescendingFirstSortable} /> */}
+        {/* <Route path="/examples/selection-range-events" component={SelectionRangeEvents} /> */}
+        {/* <Route path="/examples/isScrolling" component={IsScrolling} /> */}
+        {/* <Route path="/examples/summary-rows" component={SummaryRows} /> */}
+        <Redirect from="/" exact to="/examples/all-features" />
```
nstepien:
```suggestion
        <Route>
          <AllFeatures />
        </Route>
```
We could just do that, wdyt?

examples/index.tsx
```diff
@@ -73,44 +67,44 @@ ReactDOM.render((
...
+        {/* <Route path="/examples/resizable-cols" component={ResizableCols} /> */}
+        {/* <Route path="/examples/frozen-cols" component={FrozenCols} /> */}
+        {/* <Route path="/examples/editable" component={Editable} /> */}
+        {/* <Route path="/examples/custom-formatters" component={CustomFormatters} /> */}
+        {/* <Route path="/examples/built-in-editors" component={BuiltInEditors} /> */}
+        {/* <Route path="/examples/sortable-cols" component={SortableCols} /> */}
+        {/* <Route path="/examples/filterable-grid" component={FilterableGrid} /> */}
+        {/* <Route path="/examples/one-million-rows" component={OneMillionRows} /> */}
+        {/* <Route path="/examples/customRowRenderer" component={CustomRowRenderer} /> */}
+        <Route path="/examples/all-features" component={AllFeatures} />
```
nstepien:
```suggestion
```
See comment below

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+      ))}
+    </select>
+  );
+}
+
+const DropDownEditor = forwardRef(
+  DropDownEditorWithRef as React.RefForwardingComponent<DropDownEditorHandle, DropDownEditorProps<{ [key: string]: unknown }>>
+) as <R>(props: DropDownEditorProps<R> & { ref?: React.Ref<DropDownEditorHandle> }) => JSX.Element;
+
+(DropDownEditor as React.ComponentType).propTypes = {
```
nstepien: `PropTypes` is redundant and a maintenance burden when using ts, let's remove them.

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+  column: Column<TRow>;
+  options: Array<Option | string>;
+  value?: string;
+  onBlur(): void;
+}
+
+type DropDownEditorHandle = Pick<Editor<{[key: string]: string | undefined}>, 'getInputNode' | 'getValue'>;
+
+function DropDownEditorWithRef<TRow>({ column, value, onBlur, options }: DropDownEditorProps<TRow>, ref: React.Ref<DropDownEditorHandle>): JSX.Element {
+  const selectRef = React.createRef<HTMLSelectElement>();
```
nstepien: You're creating a new ref on each render.
```suggestion
  const selectRef = useRef<HTMLSelectElement>(null);
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+      const option = selectRef.current.querySelector('option');
+      if (option) {
+        const optionsHeight = option.clientHeight * options.length + 8;
+        setOptionsHeight(optionsHeight > 200 ? 200 : optionsHeight);
+      }
+    }
+  }, [options.length, selectRef]);
+
+  useImperativeHandle<DropDownEditorHandle, DropDownEditorHandle>(ref, () => ({
```
nstepien: It's fine to omit the generics completely here.

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (selectRef && selectRef.current) {
+      selectRef.current.size = options.length;
+
+      const option = selectRef.current.querySelector('option');
+      if (option) {
+        const optionsHeight = option.clientHeight * options.length + 8;
+        setOptionsHeight(optionsHeight > 200 ? 200 : optionsHeight);
+      }
+    }
+  }, [options.length, selectRef]);
```
nstepien:
```suggestion
  }, [options.length]);
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [optionsHeight, setOptionsHeight] = useState<number | undefined>();
+
+  const onFocus = useCallback(() => {
+    if (selectRef && selectRef.current) {
+      selectRef.current.size = options.length;
+
+      const option = selectRef.current.querySelector('option');
+      if (option) {
+        const optionsHeight = option.clientHeight * options.length + 8;
+        setOptionsHeight(optionsHeight > 200 ? 200 : optionsHeight);
```
nstepien:
```suggestion
        setOptionsHeight(Math.min(200, optionsHeight));
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type DropDownEditorHandle = Pick<Editor<{[key: string]: string | undefined}>, 'getInputNode' | 'getValue'>;
+
+function DropDownEditorWithRef<TRow>({ column, value, onBlur, options }: DropDownEditorProps<TRow>, ref: React.Ref<DropDownEditorHandle>): JSX.Element {
+  const selectRef = React.createRef<HTMLSelectElement>();
+  const [optionsHeight, setOptionsHeight] = useState<number | undefined>();
+
+  const onFocus = useCallback(() => {
+    if (selectRef && selectRef.current) {
```
nstepien:
```suggestion
    if (selectRef.current === null) return;
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+  onBlur(): void;
+}
+
+type DropDownEditorHandle = Pick<Editor<{[key: string]: string | undefined}>, 'getInputNode' | 'getValue'>;
+
+function DropDownEditorWithRef<TRow>({ column, value, onBlur, options }: DropDownEditorProps<TRow>, ref: React.Ref<DropDownEditorHandle>): JSX.Element {
+  const selectRef = React.createRef<HTMLSelectElement>();
+  const [optionsHeight, setOptionsHeight] = useState<number | undefined>();
+
+  const onFocus = useCallback(() => {
```
nstepien: We don't really need `useCallback` here IMO.

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+      if (option) {
+        const optionsHeight = option.clientHeight * options.length + 8;
+        setOptionsHeight(optionsHeight > 200 ? 200 : optionsHeight);
+      }
+    }
+  }, [options.length, selectRef]);
+
+  useImperativeHandle<DropDownEditorHandle, DropDownEditorHandle>(ref, () => ({
+    getInputNode() {
+      return selectRef?.current;
```
nstepien:
```suggestion
      return selectRef.current;
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+    },
+    getValue() {
+      return {
+        [column.key]: selectRef?.current?.value
```
nstepien:
```suggestion
        [column.key]: selectRef.current?.value
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+type DropDownEditorHandle = Pick<Editor<{[key: string]: string | undefined}>, 'getInputNode' | 'getValue'>;
+
+function DropDownEditorWithRef<TRow>({ column, value, onBlur, options }: DropDownEditorProps<TRow>, ref: React.Ref<DropDownEditorHandle>): JSX.Element {
+  const selectRef = React.createRef<HTMLSelectElement>();
+  const [optionsHeight, setOptionsHeight] = useState<number | undefined>();
+
+  const onFocus = useCallback(() => {
+    if (selectRef && selectRef.current) {
+      selectRef.current.size = options.length;
```
nstepien: What does this do?

examples/demos/components/Toolbar/Toolbar.tsx
```diff
@@ -7,16 +8,15 @@ interface Props {
...
   filterRowsButtonText?: string;
-  children: React.ReactNode;
 }
 
-export default function Toolbar(props: Props) {
-  function onAddRow() {
-    props.onAddRow!({ newRowIndex: props.numberOfRows });
-  }
+export default function Toolbar(props: PropsWithChildren<Props>) {
+  const onAddRow = useCallback(() => {
```
nstepien: Why use `useCallback`?

examples/demos/example13-all-features.tsx
```diff
@@ -0,0 +1,228 @@
...
+    rows[i] = createFakeRowObjectData(i);
+  }
+
+  return rows;
+}
+
+export default function AllFeaturesExample(): JSX.Element {
+  const [rows, setRows] = useState(() => createRows(2000));
+  const [selectedRows, setSelectedRows] = useState(() => new Set<string>());
+  const gridRef = React.createRef<DataGridHandle>();
```
nstepien:
```suggestion
  const gridRef = useRef<DataGridHandle>(null);
```

examples/demos/example13-all-features.tsx
```diff
@@ -0,0 +1,228 @@
...
+    },
+    {
+      key: 'title',
+      name: 'Title',
+      editor: <DropDownEditor<Row> options={titles} column={{} as Column<Row>} onBlur={() => {}} />,
+      width: 200,
+      resizable: true,
+      cellContentRenderer(props) {
+        return (
+          <div onClick={() => gridRef?.current?.openCellEditor(props.rowIdx, props.idx)}>
```
nstepien:
```suggestion
          <div onClick={() => gridRef.current?.openCellEditor(props.rowIdx, props.idx)}>
```

examples/demos/example13-all-features.tsx
```diff
@@ -0,0 +1,228 @@
...
+      resizable: true
+    },
+    {
+      key: 'sentence',
+      name: 'Sentence',
+      editable: true,
+      width: 200,
+      resizable: true
+    }
+  ], [gridRef]);
```
nstepien:
```suggestion
  ], []);
```

tsconfig.json
qili26: The reason why I added this was for the examples files, the ts checking seemed not working in vs code file editor. I must miss something here.

examples/index.tsx
qili26: Is this for the default `<Route />`?

examples/demos/components/Editors/DropDownEditor.tsx
qili26: I think it still makes sense. For those who don't use typescript, they will see the console error. In fact, even though we have typescript, it still makes sense to keep the backward capability.

examples/demos/components/Editors/DropDownEditor.tsx
qili26: shoot, forgot change this after converting from a class component

examples/index.tsx
nstepien: Yeah.

examples/demos/components/Editors/DropDownEditor.tsx
qili26: by default, when you open this `<select />` editor, it opens, but acts like not opened. Because the `<select />` tag won't open the `options` inside, so it makes the users feeling it's broken.

examples/demos/components/Editors/DropDownEditor.tsx
nstepien: It's really not worth it IMO. @amanmahajan7 wdyt?

examples/index.tsx
qili26: nvm, I got it.

examples/demos/components/Editors/DropDownEditor.tsx
amanmahajan7: I think we should stick to TS. We are not going to publish this component so we can keep it simple

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }));
+
+  return (
+    <select
+      ref={selectRef}
+      className="rdg-select-editor"
+      defaultValue={value}
+      onBlur={onBlur}
+      onFocus={onFocus}
+      style={{ height: optionsHeight && optionsHeight > 0 ? optionsHeight : 'initial', overflowY: 'auto' }}
```
nstepien:
```suggestion
      size={options.length}
      style={{ maxHeight: 200, height: 'auto', overflowY: 'auto' }}
```

With this we can just remove the `onFocus` handler, and the `optionsHeight` state.

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  options: Array<Option | string>;
+  value?: string;
+  onBlur(): void;
+}
+
+type DropDownEditorHandle = Pick<Editor<{[key: string]: string | undefined}>, 'getInputNode' | 'getValue'>;
+
+function DropDownEditorWithRef<TRow>({ column, value, onBlur, options }: DropDownEditorProps<TRow>, ref: React.Ref<DropDownEditorHandle>): JSX.Element {
+  const selectRef = useRef<HTMLSelectElement>(null);
+  const [optionsHeight, setOptionsHeight] = useState<number | undefined>();
```
nstepien:
```suggestion
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { forwardRef, useImperativeHandle, useRef, useState } from 'react';
```
nstepien:
```suggestion
import React, { forwardRef, useImperativeHandle, useRef } from 'react';
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }));
+
+  return (
+    <select
+      ref={selectRef}
+      className="rdg-select-editor"
+      defaultValue={value}
+      onBlur={onBlur}
+      onFocus={onFocus}
```
nstepien:
```suggestion
```

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  onBlur(): void;
+}
+
+type DropDownEditorHandle = Pick<Editor<{[key: string]: string | undefined}>, 'getInputNode' | 'getValue'>;
+
+function DropDownEditorWithRef<TRow>({ column, value, onBlur, options }: DropDownEditorProps<TRow>, ref: React.Ref<DropDownEditorHandle>): JSX.Element {
+  const selectRef = useRef<HTMLSelectElement>(null);
+  const [optionsHeight, setOptionsHeight] = useState<number | undefined>();
+
+  function onFocus() {
```
nstepien: We can completely remove this function.

examples/demos/components/Editors/DropDownEditor.tsx
qili26: right, make sense.

PR #1853, opened 2019-12-23
Remove props to implement subrows

amanmahajan7:
- [x] Remove SubRow props
   - `getSubRowDetails`
   - `onCellExpand`
   - `onDeleteSubRow`
   - `onAddSubRow` - this prop was not used
- [x] Update `TreeView` example and implement subrows without the above props. SubRows implementation felt too complicated and I ended up using a simpler model
- [x] Combine `TreeView` and `TreeViewNoDelete` examples
- [x] Tsify `CellActions` example

Next step is to replace `defaultCellContentRenderer` with `defaultFormatter`


examples/demos/example27-cell-actions.tsx
```diff
@@ -147,38 +146,19 @@ const columns = [
...
-        />
-      </Wrapper>
-    );
-  }
+  return (
+    <Wrapper title="Cell Actions Example">
+      <DataGrid<Row, 'id'>
+        enableCellSelect
+        columns={columns}
+        rowGetter={i => rows[i]}
```
nstepien: It doesn't matter too much for demos, but we should avoid passing new functions on each render to DataGrid.

examples/demos/example25-tree-view.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+function toggleSubRow(rows: Row[], id: string): Row[] {
+  const rowIndex = rows.findIndex(r => r.id === id);
+  const row = rows[rowIndex];
+  const { children } = row;
+  if (!children) return rows;
+
+  const newRows = [...rows];
+  if (!row.isExpanded) {
+    newRows[rowIndex] = { ...newRows[rowIndex], isExpanded: true };
```
nstepien:
```suggestion
    newRows[rowIndex] = { ...row, isExpanded: true };
```
We could also move this line out of the if/else, and do
```suggestion
    newRows[rowIndex] = { ...row, isExpanded: !row.isExpanded };
```

examples/demos/example25-tree-view.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const subRows = children.map(c => ({ ...c }));
```
nstepien: Don't think we really need to clone the subrow objects.

examples/demos/example25-tree-view.tsx
```diff
@@ -0,0 +1,170 @@
...
+    newRows.splice(rowIndex + 1, 0, ...subRows);
+  } else {
+    newRows[rowIndex] = { ...newRows[rowIndex], isExpanded: false };
```
nstepien: See above.

examples/demos/example25-tree-view.tsx
amanmahajan7: Done

examples/demos/example27-cell-actions.tsx
amanmahajan7: I am thinking of using `eventBus` more aggressively and avoid passing callbacks down. I have a few ideas that we can discuss next week

PR #1854, opened 2020-01-02
Extend onScroll API

qili26:
1. Add event in the onScroll api for endless scrolling.

lazyloading


style/rdg-core.less
```diff
@@ -29,4 +29,5 @@
 .rdg-grid {
   // min-height is here to show the horizontal scrollbar when there are no rows
   min-height: 1px;
+  margin-bottom: 8px;
```
nstepien: I wonder if this could break anything.

qili26: No, this doesn't break anything, and in fact this fixes a jumping issue that if the last row cells are editable, and when you click on it, the the drag square will push the rows above a little bit.

qili26:
They all called lazy loading, but this one is data focused. Maybe endless scrolling would be a better term. The real use case is like you only want to pull 100 records initially from an API and pull another 100 records from another API call when the scroll bar reaches the bottom.

nstepien:
You could just trigger loading in `rowGetter` when the row number requested is close or at the `rows.length`.

qili26:
Aren't we going to deprecate the `rowGetter` and passing all the rows into the grid?

ShantanuNair:
Why close this??

PR #1855, opened 2020-01-02
Remove CellContentRenderer

amanmahajan7:
A separate `cellContentRenderer` prop is no longer required as we have simplified the `Cell` component and removed the `CellActions` and `SubRows` logic

Next we can remove the value prop and simplify types


examples/demos/example13-all-features.tsx
```diff
@@ -87,10 +87,10 @@ export default function AllFeaturesExample(): JSX.Element {
       editor: <DropDownEditor<Row> options={titles} column={{} as Column<Row>} onBlur={() => {}} />,
       width: 200,
       resizable: true,
-      cellContentRenderer(props) {
+      formatter(props) {
         return (
-          <div onClick={() => gridRef.current?.openCellEditor(props.rowIdx, props.idx)}>
-            <CellContent {...props} />
+          <div onClick={() => gridRef.current?.openCellEditor(props.rowIdx, props.column.idx)}>
+            <ValueFormatter {...props} />
```
qili26: Maybe this is better?
```typescript
{ props.row[props.column.key] }
```

examples/demos/example13-all-features.tsx
```diff
@@ -87,10 +87,10 @@ export default function AllFeaturesExample(): JSX.Element {
       editor: <DropDownEditor<Row> options={titles} column={{} as Column<Row>} onBlur={() => {}} />,
       width: 200,
       resizable: true,
-      cellContentRenderer(props) {
+      formatter(props) {
```
qili26: Generally speaking, render props is still faster than using a component for basic use case, because if a function only returns a number or string, it avoided unnecessary React related expense. I am wondering if we should do the opposite like removing the `formatter` but keep the `cellContentRenderer`?

In addition to the `formatter`, I believe we should start thinking removing the element support, i.e. at least we should not pass an element as a `formatter` because this could potentially result in not re-render problem.

BTW, not sure if this this working, but I am guessing if we want to put a hook inside this formatter function, it might not work.

src/common/types.ts
```diff
@@ -51,7 +50,6 @@ export type CalculatedColumn<TRow, TField extends keyof TRow = keyof TRow> =
     idx: number;
     width: number;
     left: number;
-    cellContentRenderer: CellContentRenderer<TRow>;
```
nstepien: I'd rather we keep using this type for `formatter`. Making the `formatter` a function makes things so much simpler/better IMO. That way we can just do
```js
const column = {
  key: '...',
  formatter(rowData) {
    return intlFormatter.format(rowData.value);
  }
};
```

Removing the requirement for formatters to be a React component means we can reuse utils/libs directly as formatters, and we don't have to unnecessarily go through another React component instantiation/rendering.

https://gist.github.com/nstepien/090298c3c2d94324cb332c33d82fdcfb

examples/demos/example13-all-features.tsx
nstepien: If you want to use hooks, then you'll just have to separate the component from the formatter function:
```jsx
  formatter(props) {
    return <Formatter {...props} />;
  }
```

src/common/types.ts
amanmahajan7: I did some quick profiling and did not notice any performance difference when using a component instead of a function. After the initial render, memoization should anyway prevent unnecessary cell rendering.

Here are a few reasons I prefer a component
- There is nothing that prevent us from calling component as function. For example `formatter: ImageFormatter` would work but it is technically not correct. It is fairly common to have `JSX` in a formatter and we will have to use `formatter(props) {  return <Formatter {...props} />;  }` whenever the formatter needs `JSX`. 
- Calling a component as a function sometimes works but it can lead to confusing [errors](https://kentcdodds.com/blog/dont-call-a-react-function-component), one example is hooks. We can use `formatter(props) {  return <Formatter {...props} />;  }` but I wonder if it necessary? I think it would be simple to just use `formatter: Formatter` instead. We will have to add additional documentation explaining the differences which is extra work IMO. 
- We won't catch this bug with types
- `formatter(rowData) { return intlFormatter.format(rowData.value) }` is a valid component. It does not work because of the type [issue](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20544). A workaround is `formatter(rowData) { return <>intlFormatter.format(rowData.value)</> }`

examples/demos/example13-all-features.tsx
amanmahajan7: >Generally speaking, render props is still faster than using a component for basic use case, because if a function only returns a number or string, it avoided unnecessary React related expense. I am wondering if we should do the opposite like removing the formatter but keep the cellContentRenderer?
https://github.com/adazzle/react-data-grid/pull/1855#discussion_r362855946

> In addition to the formatter, I believe we should start thinking removing the element support
I agree we should remove element support. I think we do not 

>at least we should not pass an element as a formatter because this could potentially result in not re-render problem.
I do not think we have to worry about re-rendering formatters anymore. A formatter should only be re-rendered if any of the Cell component props change which should not happen often

>BTW, not sure if this this working, but I am guessing if we want to put a hook inside this formatter function, it might not work.
As formatter is a component so hooks should work

src/common/types.ts
nstepien: >We will have to add additional documentation explaining the differences which is extra work IMO.

The documentation is already written for us: https://reactjs.org/docs/render-props.html

>We won't catch this bug with types

I haven't tested this, but I'd expect React to fail immediately if you use hooks outside a separate component, so tests would catch this bug instead.

qili26: @amanmahajan7 

1. `formatter: ImageFormatter` is a little bit restricted as in `ImageFormatter`, it cannot access any outside scope data and it can only rely on `redux` or `context`, which actually makes things more difficult, so for this usage, I would actually want to gradually deprecate it. When we pass an element directly, it will result in the issue that I mentioned in the previous comment in some cases.
1.1 Let's say if `ImageFormatter` is a class component, what we do inside the data grid is to check whether it's a valid component and then render it, It's same if the user provides the formatter function and just do `formatter(props) { return <ImageFormatter/>; }`
1.2 If the `ImageFormatter` is a function component without hook, it can just be used as a function.
1.3 If the `ImageFormatter` is a function component with hook. it falls to 1.1

> We won't catch this bug with types
Could you be more specific? I think it's the user's decision to decide what's being rendered in the cell, By default we only need to support `row[column.key]`, but they can even only display some hard coded text based on their business requirement.

examples/demos/example13-all-features.tsx
qili26: @amanmahajan7 
My hook comment above is for if we write code like this

```typescript
function MyGrid() {
const someDataInMyGrid;
const columns = [
  {
    key: xx,
    ...
    formatter(props) {
       const xxx = useState(someDataInMyGrid);
       return someElement;
    }
  }
];

...
}
```

examples/demos/example31-isScrolling.js
```diff
@@ -53,7 +53,7 @@ const createColumns = (numberCols) => [...Array(numberCols).keys()].map(i => {
     name: `col${i}`
   };
   if (i === 3) {
-    column.cellContentRenderer = () => <ExpensiveFormatter />;
+    column.formatter = () => <ExpensiveFormatter />;
```
nstepien:
```suggestion
    column.formatter = ExpensiveFormatter;
```

src/formatters/SimpleCellFormatter.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 export interface SimpleCellFormatterProps {
-  value: string | number | boolean;
+  value?: string | number | boolean;
 }
 
 export function SimpleCellFormatter({ value }: SimpleCellFormatterProps) {
```
nstepien: Should change div -> span.
Blocks prevent text ellipsis from working.

image
image

src/index.ts
```diff
@@ -3,8 +3,6 @@ export { default as Cell } from './Cell';
 export { default as Row } from './Row';
 export { default as HeaderCell } from './HeaderCell';
 export * from './Columns';
-export * from './Cell/cellContentRenderers';
-export { default as CellContent } from './Cell/CellContent';
```
nstepien: https://github.com/adazzle/react-data-grid/pull/1831
@qili26 Why did we need this export for again?

src/Cell.tsx
```diff
@@ -133,21 +134,34 @@ function Cell<R>({
...
+      isSummaryRow
+    };
+    const { formatter } = column;
+
+    if (isValidElementType(formatter)) {
+      children = createElement<typeof formatterProps>(formatter, formatterProps);
+    }
+
+    if (isElement(formatter)) {
+      children = cloneElement(formatter, formatterProps);
```
nstepien: Let's remove the `React.ReactElement` type from `column.formatter` now.

examples/demos/example13-all-features.tsx
amanmahajan7: This would work because formatter is currently a component. It is same as writing

```
function Formatter() {
   const xxx = useState(someDataInMyGrid);
   return someElement;
}

const columns = [
  {
    key: xx,
    ...
    formatter: Formatter
  }
];
```

This however won't work with render props

src/common/types.ts
amanmahajan7: Render prop is mainly for sharing common logic but we are using it to improve performance but I did not notice any difference in my testing . With component we can prevent any edge cases so I am not sure what is the advantage of using render props in this case. 

> I haven't tested this, but I'd expect React to fail immediately if you use hooks outside a separate component, so tests would catch this bug instead.

This would be a run time error which can totally be avoided with component

> We won't catch this bug with types
I meant the following would be valid with render props and there would not be any type error

```
formatter(props) {
  const xxx = useState(someDataInMyGrid);
  return someElement;
}
```

src/index.ts
qili26: @nstepien  @amanmahajan7  Just DM you guys.

src/Cell.tsx
amanmahajan7: Done

src/formatters/SimpleCellFormatter.tsx
amanmahajan7: Done

amanmahajan7:
Let's remove `value` in the next PR

examples/demos/example13-all-features.tsx
amanmahajan7: Done

src/Cell.tsx
```diff
@@ -133,21 +134,30 @@ function Cell<R>({
     style.transform = `translateX(${scrollLeft}px)`;
   }
 
+  if (!children) {
+    const { formatter } = column;
+
+    if (isValidElementType(formatter)) {
```
nstepien: IMO we don't need to bother with this check.
```tsx
if (!children) {
  children = createElement<typeof formatterProps>(column.formatter, {
    value: rowData[column.key],
    column,
    rowIdx,
    row: rowData,
    isRowSelected,
    onRowSelectionChange,
    isSummaryRow
  });
}
```

We should also remove support for the `children` prop on `Cell`, maybe in the future.

amanmahajan7: Do you think it is better to pass down `defaultFormatter` instead setting the formatter on the calculated column? We have two options 
- Make formatter a required prop in `CalculatedColumn`
- or use `const formatter = column.formatter || props.defaultFormatter`

nstepien: I think I'd rather make the `formatter` be a required type on `CalculatedColumn`, instead of prop drilling.

src/common/types.ts
```diff
@@ -50,6 +50,7 @@ export type CalculatedColumn<TRow, TField extends keyof TRow = keyof TRow> =
     idx: number;
     width: number;
     left: number;
+    formatter: React.ComponentType<FormatterProps<any, TRow>>;
```
qili26: Maybe this?
```typescript
{ [key: string]: any }
```
`any` includes primary type and array, which are apparently not correct to be a `Row`. However I am not sure the above type is accurate either. Have you discussed with @nstepien about this before when you guys first worked on migrating to typescript?

src/Cell.tsx
```diff
@@ -133,21 +133,27 @@ function Cell<R>({
     style.transform = `translateX(${scrollLeft}px)`;
   }
 
+  if (!children) {
+    const { formatter } = column;
```
nstepien: Don't need to destructure this anymore.

PR #1856, opened 2020-01-03
Use eventBus to select rows

amanmahajan7:
As`handleRowSelectionChange` is dependent on `selectedRows`, the whole grid is re-rendered when a row is selected.

**Before**
image

**After**

image


src/EventBus.ts
```diff
@@ -1,7 +1,15 @@
 import { Position } from './common/types';
 
+export interface SelectRowEvent<R> {
```
nstepien: Should probably move this in our `types` file.

src/Cell.tsx
```diff
@@ -141,7 +146,7 @@ function Cell<R>({
       rowIdx,
       row: rowData,
       isRowSelected,
-      onRowSelectionChange,
+      onRowSelectionChange: isSummaryRow ? noop : onRowSelectionChange,
```
nstepien: fwiw our checkbox column doesn't display any checkbox for summary rows, so maybe we don't need this check here.
And maybe users might want to display set up a custom column/row selection behavior.

src/Canvas.tsx
```diff
@@ -155,6 +158,34 @@ function Canvas<R, K extends keyof R>({
     selectCell({ rowIdx, idx }, true);
   }
 
+  useEffect(() => {
```
nstepien: Any particular reason to move this from DataGrid to Canvas?

amanmahajan7: Need access to `eventBus` which is initialized in the Canvas component

src/Cell.tsx
amanmahajan7: Makes sense, I have removed it

PR #1857, opened 2020-01-04
Move Cell events in Row (POC)

qili26:
1. Move the following events from `Cell` to `Row`
```
  onClick,
  onDoubleClick
  onContextMenu
  onDragOver
  onMouseDown
```
2. put data-idx in Cell div and use element.closest() to get the `.rdg-cell` to access the `idx`,
https://developer.mozilla.org/en-US/docs/Web/API/Element/closest
3. Then we would know the position in Row level without binding the event in each Cell.

By doing so, for editable cells we could roughly save the following numbers of events binding (All features max width example, 1080P screen) `9 * 27 * 5 - 27 * 5 = 1080`

If this is the right way for a better performance, we could even move the common events 1 level above to `Canvas` by assiging `data-rowIdx` and `data-idx` to the cell. Then eventually for those common 5 events, we only need to bind to the Canvas once.

Events in Row
image

Events in Cell
image

### DIFF
--- a/src/Cell.tsx
+++ b/src/Cell.tsx
@@ -24,48 +24,48 @@ function Cell<R>({
   rowIdx,
   scrollLeft,
   eventBus,
-  onRowClick,
-  onRowDoubleClick,
+  // onRowClick,
+  // onRowDoubleClick,
   enableCellRangeSelection
 }: CellProps<R>) {
   const position: Position = { idx, rowIdx };
 
-  function selectCell(openEditor?: boolean) {
-    eventBus.dispatch(EventTypes.SELECT_CELL, position, openEditor);
-  }
+  // function selectCell(openEditor?: boolean) {
+  //   eventBus.dispatch(EventTypes.SELECT_CELL, position, openEditor);
+  // }
 
-  function handleCellClick() {
-    selectCell();
-    onRowClick?.(rowIdx, rowData, column);
-  }
+  // function handleCellClick() {
+  //   selectCell();
+  //   onRowClick?.(rowIdx, rowData, column);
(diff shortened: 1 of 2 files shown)


qili26:
@nstepien @amanmahajan7 
Do you guys think this is worth a shot?

amanmahajan7:
React uses event delegation internally and does not attach events to individual nodes. 
https://github.com/facebook/react/issues/13635

qili26:
That thread asked for memory saving, however, our case is for rendering speed consideration (maybe as well as the code maintenance complexity).
1. The events are still there in the Cell though not directly bond to the dom.
2. When we lift those handlers to Canvas level, eventually we can remove the `eventbus` and Canvas will just be the only parent component for `Interactionmask` and `Cell` / `Row` communication.
3. I had tried to remove all events in Cell and in Row component, it's definitely faster than the Cell with events. Maybe 20ms ~ 40ms difference doesn't seem persuasive enough?

nstepien:
>Maybe 20ms ~ 40ms difference doesn't seem persuasive enough?

Did you test this in production mode? How much of an impact does it have when you scroll vertically/horizontally?

Would custom row renderers (if we still support this feature) have to re-implement the handlers when re-using the Cell component?

Also fwiw you can probably get the correct row idx/cell idx by checking the click's coordinates relative to the grid, instead of using dom attributes.

amanmahajan7:
From what I understand, React does not attach events to individual nodes so whether we add an event handler on the row or on the cell should not make a difference. Notice how cell does not have a click handler. `noop` is added to fix a safari bug

image

PR #1863, opened 2020-01-07
Stricter rowRenderer/headerRenderer types

nstepien:
Would be nice if we could get rid of `react-is`.

### DIFF
--- a/src/HeaderCell.tsx
+++ b/src/HeaderCell.tsx
@@ -1,8 +1,7 @@
-import React, { createElement, cloneElement } from 'react';
+import React, { createElement } from 'react';
 import classNames from 'classnames';
-import { isElement } from 'react-is';
 
-import { CalculatedColumn, HeaderRendererProps } from './common/types';
+import { CalculatedColumn } from './common/types';
 import { DEFINE_SORT } from './common/enums';
 import { HeaderRowProps } from './HeaderRow';
 import SortableHeaderCell from './common/cells/headerCells/SortableHeaderCell';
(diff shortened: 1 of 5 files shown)


src/HeaderCell.tsx
```diff
@@ -19,7 +18,6 @@ type SharedHeaderRowProps<R, K extends keyof R> = Pick<HeaderRowProps<R, K>,
 
 export interface HeaderCellProps<R, K extends keyof R> extends SharedHeaderRowProps<R, K> {
   column: CalculatedColumn<R>;
-  renderer?: React.ReactElement | React.ComponentType<HeaderRendererProps<R>>;
```
nstepien: This was actually unused.

src/HeaderCell.tsx
```diff
@@ -93,7 +82,7 @@ export default function HeaderCell<R, K extends keyof R>({
         column={column}
         onResize={props.onResize}
       >
-        {cell}
+        {cell as React.ReactElement<React.ComponentProps<'div'>>}
```
amanmahajan7: May be returning `<>column.name</>` is better if TS is not recognizing the string

src/common/types.ts
```diff
@@ -167,6 +167,10 @@ export interface IRowRendererProps<TRow> {
   onRowExpandToggle?(event: RowExpandToggleEvent): void;
 }
 
+export interface CustomRowRendererProps<TRow> extends IRowRendererProps<TRow> {
+  renderBaseRow: React.FunctionComponent<IRowRendererProps<TRow>>;
+}
+
```
amanmahajan7: We should only export the public interfaces instead of `export * from './common/types';`. We can check which ones we need in a separate PR

src/HeaderCell.tsx
nstepien: I think the issue is that we're assigning various types to `cell`, and TS doesn't understand that it's a div element at this moment, so I think the type assertion is fair here.

PR #1865, opened 2020-01-08
Kill row refs

nstepien:


src/DataGrid.tsx
```diff
@@ -23,7 +23,7 @@ import {
   RowExpandToggleEvent,
   RowGetter,
   SelectedRange,
-  CustomRowRendererProps,
+  IRowRendererProps,
```
amanmahajan7: Should we rename it to `RowRendererProps` instead as we not using the `I` prefix for any other interface?

src/RowRenderer.tsx
```diff
@@ -63,41 +57,26 @@ function RowRenderer<R, K extends keyof R>({
...
-      return renderGroupRow();
+      return (
+        <RowGroup<R>
+          {...rendererProps}
+          {...__metaData!}
+          name={(rowData as RowData).name!}
+          eventBus={eventBus}
+          renderer={rowGroupRenderer}
+          onRowExpandToggle={props.onRowExpandToggle}
+          renderBaseRow={(p: IRowRendererProps<R>) => <Row {...p} />}
```
amanmahajan7: I think we should remove `renderBaseRow`. IIRC it was added to forward the ref correctly to prevent breaking `setScrollRef` logic.

amanmahajan7: We can handle it it a separate PR if you prefer

src/DataGrid.tsx
nstepien: It conflicts with `RowRenderer`'s `RowRendererProps` unfortunately.

PR #1866, opened 2020-01-08
Enforce column.editor: ComponentType

nstepien:


src/common/editors/EditorContainer.tsx
```diff
@@ -85,25 +84,20 @@ export default class EditorContainer<R, K extends keyof R> extends React.Compone
...
-      return React.cloneElement(CustomEditor, editorProps);
-    }
-    if (isValidElementType(CustomEditor)) {
-      return <CustomEditor {...editorProps} />;
+    if (this.props.column.editor) {
+      return (
+        <this.props.column.editor
+          ref={this.editor}
+          column={this.props.column}
+          value={this.getInitialValue() as R[keyof R & string] & R[keyof R & number] & R[keyof R & symbol]}
```
nstepien: 🤷‍♂

examples/demos/example13-all-features.tsx
```diff
@@ -84,7 +84,7 @@ export default function AllFeaturesExample(): JSX.Element {
     {
       key: 'title',
       name: 'Title',
-      editor: <DropDownEditor<Row> options={titles} column={{} as Column<Row>} onCommit={() => { }} />,
+      editor: ({ column }) => <DropDownEditor<Row> options={titles} column={column} onCommit={() => {}} />,
```
amanmahajan7: I think this will break because we are not forwarding the ref

examples/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -8,26 +8,21 @@ interface Option {
...
 
   useImperativeHandle(ref, () => ({
     getInputNode() {
       return selectRef.current;
     },
     getValue() {
-      return {
-        [column.key]: selectRef.current?.value
-      };
+      return selectRef.current!.value;
```
amanmahajan7: Would this not fail?
https://github.com/adazzle/react-data-grid/blob/131a9d3f13c39527c19adb8210823bd5d057b53c/examples/demos/example13-all-features.tsx#L186

nstepien: Mmh, might be a type issue somewhere, as it wasn't allowing me to return an object.

nstepien: Nevermind, I've fixed it now.

PR #1867, opened 2020-01-08
Remove FormatterProps.value

nstepien:
I'd like to see if it makes sense to decouple `row` from `column.key`, as we might not always have a perfect 1:1 mapping between the data we want to display, and the actual data.

For example:

| a  | b | sum (a + b) |
| - | - | - |
| 2  | 3  | 5 |

Removing `value` from formatters is a first step, the second one is in editors. then maybe we can make `column.key` a simple `string` type.
WDYT?


amanmahajan7:
I think it makes sense, we do not need value as it can be derive from row and column. I will also simply the types

PR #1871, opened 2020-01-09
Compute editor position using selection coordinates and grid position

nstepien:


src/masks/InteractionMasks.tsx
```diff
@@ -139,24 +132,14 @@ export default function InteractionMasks<R, K extends keyof R>({
...
-
-    const { left: portalTargetLeft, top: portalTargetTop } = editorPortalTarget.getBoundingClientRect();
-    const { scrollLeft, scrollTop } = editorPortalTarget;
+  function getEditorPosition() {
+    if (!canvas.current) return null;
+    const { left, top } = canvas.current.getBoundingClientRect();
     return {
-      left: selectionMaskLeft - portalTargetLeft + scrollLeft,
-      top: selectionMaskTop - portalTargetTop + scrollTop
+      left: columns[selectedPosition.idx].left - scrollLeft + left,
```
amanmahajan7: Should we remove `editorPortalTarget` prop?

src/masks/InteractionMasks.tsx
```diff
@@ -85,6 +86,7 @@ export default function InteractionMasks<R, K extends keyof R>({
   enableCellDragAndDrop,
   editorPortalTarget,
   cellNavigationMode,
+  canvas,
```
amanmahajan7: May be we can rename it to `canvasRef` for clarity

src/masks/InteractionMasks.tsx
amanmahajan7: nm, it is needed for the portals

nstepien: I guess we could, do we have a need for it anymore?

amanmahajan7: I am not too sure actually 🤔

amanmahajan7:
We can probably add a new function 

```
function getCellPosition({ idx, rowIdx }: Position) {
    if (!canvas.current) return null;
    const { left, top } = canvas.current.getBoundingClientRect();
    return {
      left: columns[idx].left - scrollLeft + left,
      top: rowIdx * rowHeight - scrollTop + top
    };
  }
```

And pass this to interaction mask. And may be use it to set `selectionMasks` dimensions also. Wdyt?

nstepien:
We only need to get the canvas left/top when positioning elements outside the viewport though.

PR #1881, opened 2020-01-13
Reset invalid positions

amanmahajan7:
This can happen when row or columns are removed


amanmahajan7:
We will have to run some tests. It looks like react will re-run the component with updated state immediately **after** exiting the first render so it might break.

https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

May be we can use a key to reset `InteractionMasks`  instead 🤔

amanmahajan7:
This change does not bypass the render so we need to check if a position is within bound or not. We were already doing that for `selectedPosition` and I added a check for `copied` and `dragged` position as well

PR #1884, opened 2020-01-13
Code cleanup

amanmahajan7:
- Remove `DragHandle`. It is just a wrapper and we can use the `div` directly
- Remove `SelectionMasks` and `CopyMasks` components and use `CellMask` directly
- Remove `getSelectedCellValue` 
- Fix `act` warnings


amanmahajan7:
@nstepien wdyt?

PR #1886, opened 2020-01-14
Rearrange props

nstepien:
- Rearranged the main props interface so it makes a bit more sense
- `minHeight` -> `height`
- `minWidth` -> `width`
- `RowsContainer` -> `rowsContainer`
- `onGridSort` -> `onSort`
- `onGridRowsUpdated` -> `onRowsUpdate`
- Fixed header/filter row line heights

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -31,34 +31,20 @@ import {
 export { DataGridHandle };
 
 export interface DataGridProps<R, K extends keyof R> {
+  /**
+   * Grid and data Props
+   */
   /** An array of objects representing each column on the grid */
   columns: readonly Column<R>[];
-  /** The minimum width of the grid in pixels */
-  minWidth?: number;
-  /** The height of the header row in pixels */
-  headerRowHeight?: number;
-  /** The height of the header filter row in pixels */
-  headerFiltersHeight?: number;
-  /** Toggles whether filters row is displayed or not */
-  enableHeaderFilters?: boolean;
-  /** Minimum column width in pixels */
-  minColumnWidth?: number;
-  /** Function called whenever row is clicked */
-  onRowClick?(rowIdx: number, rowData: R, column: CalculatedColumn<R>): void;
-  /** Function called whenever row is double clicked */
-  onRowDoubleClick?(rowIdx: number, rowData: R, column: CalculatedColumn<R>): void;
-
-  filters?: Filters<R>;
-  onFiltersChange?(filters: Filters<R>): void;
-  /** Function called whenever keyboard key is released */
-  onGridKeyUp?(event: React.KeyboardEvent<HTMLDivElement>): void;
-  /** Function called whenever keyboard key is pressed down */
-  onGridKeyDown?(event: React.KeyboardEvent<HTMLDivElement>): void;
-
-  selectedRows?: ReadonlySet<R[K]>;
-  /** Function called whenever row selection is changed */
-  onSelectedRowsChange?(selectedRows: Set<R[K]>): void;
-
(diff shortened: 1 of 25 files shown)


src/DataGrid.tsx
```diff
@@ -31,34 +31,20 @@ import {
...
-
+  /** A function called for each rendered row that should return a plain key/value pair object */
+  rows: readonly R[];
+  /**
+   * Rows to be pinned at the bottom of the rows view for summary, the vertical scroll bar will not scroll these rows.
+   * Bottom horizontal scroll bar can move the row left / right. Or a customized row renderer can be used to disabled the scrolling support.
+   */
+  summaryRows?: readonly R[];
+  /** The primary key property of each row */
+  rowKey?: K;
```
amanmahajan7: I think we should make it a required prop. We can do it in a separate PR

src/DataGrid.tsx
```diff
@@ -68,58 +54,94 @@ export interface DataGridProps<R, K extends keyof R> {
    * 4. Update all cells under a given cell by double clicking the cell's fill handle.
    */
   onGridRowsUpdated?<E extends GridRowsUpdatedEvent<R>>(event: E): void;
```
amanmahajan7: May be `onRowsUpdated` would make more sense. Wdyt?

src/DataGrid.tsx
```diff
@@ -68,58 +54,94 @@ export interface DataGridProps<R, K extends keyof R> {
...
+  /** Set of selected row keys */
+  selectedRows?: ReadonlySet<R[K]>;
+  /** Function called whenever row selection is changed */
+  onSelectedRowsChange?(selectedRows: Set<R[K]>): void;
   /** The key of the column which is currently being sorted */
   sortColumn?: keyof R;
   /** The direction to sort the sortColumn*/
   sortDirection?: DEFINE_SORT;
   /** Function called whenever grid is sorted*/
   onGridSort?(columnKey: keyof R, direction: DEFINE_SORT): void;
```
amanmahajan7: Same here, `onSort`?

src/DataGrid.tsx
qili26: For this one, I would say maybe keep it as it is, since the event is actually `GridRowsUpdatedEvent`, so the names are consistent.

src/DataGrid.tsx
qili26: For this one, I would agree with you, we can ignore the `Grid` word like `onSort` only.

src/HeaderRow.tsx
```diff
@@ -50,7 +50,7 @@ export default function HeaderRow<R, K extends keyof R>({
   return (
     <div
       className="rdg-header-row"
-      style={{ width, height }}
+      style={{ width, height, lineHeight: `${height}px` }}
```
amanmahajan7: Needs `pre-canary` catchup?

nstepien: No?

src/DataGrid.tsx
amanmahajan7: Nm, for some reason I thought we fixed the line height in a separate PR

src/HeaderRow.tsx
amanmahajan7: Nvm, for some reason I thought we fixed the line height in a separate PR

src/DataGrid.tsx
amanmahajan7: I think we can also rename `GridRowsUpdatedEvent` to RowsUpdateEvent`

izanf:
A good proposal to width is create a flexible width like a table columns width.

PR #1889, opened 2020-01-14
Changelog

amanmahajan7:
Added some details on what has changed in different branches. This is mainly for reference for now but let's add future breaking changes here as well.


docs/CHANGELOG.md
```diff
@@ -0,0 +1,147 @@
...
+- Rename `minHeight` to `height`
+- Rename `minWidth` to `width`
+- Rename `RowsContainer` to `rowsContainer`
+- Rename `onGridSort` to `onSort`
+- Rename `onGridRowsUpdated` to `onRowsUpdate`
+
+## next -> canary
+- Column resize now resizes all the cells which means re-rendering the whole grid when resizing columns.
+- Add `column.cellContentRenderer`. This is an opt-in feature, so it should not break anything. More info in [here](https://gist.github.com/nstepien/090298c3c2d94324cb332c33d82fdcfb)
+- `ReactDataGrid` component is renamed to `DataGrid`
```
qili26: Do we really need to mention this? I mean this seems something internal as they can still import like 
```
import ReactDataGrid from 'react-data-grid';
```

amanmahajan7: We don't need it as it is a default import. I just added it to emphasize that we should use `import DataGrid` instead of `import ReactDataGrid`. We can remove it from the final release notes

docs/CHANGELOG.md
```diff
@@ -0,0 +1,147 @@
...
+- Remove row reordering support. This was a buggy feature so it has been removed. 
+- Remove toolbar prop. Check [here](https://github.com/adazzle/react-data-grid/pull/1769) on how to migrate.
+- Remove internal `sortColumn` and `sortDirection` states. Sorting is now only controlled by `sortColumn` and `sortDirection` props. Check [here](https://github.com/adazzle/react-data-grid/pull/1768) on how to migrate.
+- Improve scrolling performance
+- Improve frozen columns performance
+- Re-implement row selection. Previous row selection API has been replaced by the new `selectedRows` and `onSelectedRowsChange` props. A new `SelectColumn` component is now needed to add checkbox column. Check [here](https://github.com/adazzle/react-data-grid/pull/1762) on how to migrate.
+- Add `renderBatchSize` prop [1779](https://github.com/adazzle/react-data-grid/pull/1779)
+- Remove bootstrap dependency and re-implement grid css [1780](https://github.com/adazzle/react-data-grid/pull/1780)
+
+## master -> next
```
nstepien:
```suggestion
## master -> alpha
```
`next` is the branch name, while `alpha` is the npm tag, should probably use it instead. And rename the next branch maybe.

docs/CHANGELOG.md
```diff
@@ -0,0 +1,147 @@
...
+- Remove `formatter.isScrolling` prop. `isScrolling` prop was introduced to improve performance of expensive formatters. This is no longer required as the internals of the grid were rewritten to improve performance
+- Add summary rows [1773](https://github.com/adazzle/react-data-grid/pull/1773)
+- Remove row reordering support. This was a buggy feature so it has been removed. 
+- Remove toolbar prop. Check [here](https://github.com/adazzle/react-data-grid/pull/1769) on how to migrate.
+- Remove internal `sortColumn` and `sortDirection` states. Sorting is now only controlled by `sortColumn` and `sortDirection` props. Check [here](https://github.com/adazzle/react-data-grid/pull/1768) on how to migrate.
+- Improve scrolling performance
+- Improve frozen columns performance
+- Re-implement row selection. Previous row selection API has been replaced by the new `selectedRows` and `onSelectedRowsChange` props. A new `SelectColumn` component is now needed to add checkbox column. Check [here](https://github.com/adazzle/react-data-grid/pull/1762) on how to migrate.
```
nstepien:
```suggestion
- Re-implement row selection. Previous row selection API has been replaced by the new `selectedRows` and `onSelectedRowsChange` props. A new `SelectColumn` is now available to import and add a row selection column. Check [here](https://github.com/adazzle/react-data-grid/pull/1762) on how to migrate.
```

PR #1891, opened 2020-01-15
Remove EventTypes enum

amanmahajan7:


qili26:
Any reason for not using predefined strings?

nstepien:
@qili26 We already do that: https://github.com/adazzle/react-data-grid/blob/pre-canary/src/EventBus.ts#L12

qili26:
Got it, then it totally makes sense. Thanks!

PR #1895, opened 2020-01-16
FC-ize EditorContainer

amanmahajan7:
I copied the changes from https://github.com/adazzle/react-data-grid/pull/1781. I will add a separate `EditorContainer` for the new API


src/editors/EditorContainer.tsx
```diff
@@ -19,220 +19,155 @@ export interface Props<R, K extends keyof R> extends SharedInteractionMasksProps
...
+  firstEditorKeyPress: key
+}: EditorContainerProps<R, K>) {
+  const editorRef = useRef<Editor>(null);
+  const [isInvalid, setIsInvalid] = useState(false);
+  const prevScrollLeft = useRef(scrollLeft);
+  const prevScrollTop = useRef(scrollTop);
+
+  const getInputNode = useCallback(() => editorRef.current?.getInputNode(), []);
+
+  useLayoutEffect(() => {
```
amanmahajan7: Need `useLayoutEffect` to mimic the existing selection behavior in Chrome

src/editors/EditorContainer.tsx
```diff
@@ -19,220 +19,155 @@ export interface Props<R, K extends keyof R> extends SharedInteractionMasksProps
...
+  left,
+  top,
+  onCommit,
+  onCommitCancel,
+  scrollLeft,
+  scrollTop,
+  firstEditorKeyPress: key
+}: EditorContainerProps<R, K>) {
+  const editorRef = useRef<Editor>(null);
+  const [isInvalid, setIsInvalid] = useState(false);
```
nstepien:
```suggestion
  const [isValid, setValid] = useState(true);
```
We don't need to do this now as it's kind of a deprecated component, but I prefer to avoid negative variable names, they can make it a little bit harder to read the code.

src/editors/EditorContainer.tsx
```diff
@@ -19,220 +19,155 @@ export interface Props<R, K extends keyof R> extends SharedInteractionMasksProps
...
 
-  onPressArrowRight = (e: KeyboardEvent) => {
-    // prevent event propogation. this disables right cell navigation
-    if (!this.isCaretAtEndOfInput()) {
+  function onKeyDown(e: KeyboardEvent<HTMLElement>) {
+    if (preventDefaultNavigation(e.key)) {
       e.stopPropagation();
-    } else {
-      this.commit();
+      return;
```
nstepien: Could just use `else if`

PR #1896, opened 2020-01-16
Fix SelectCellFormatter export

nstepien:


packages/react-data-grid/src/formatters/SelectCellFormatter.tsx
```diff
@@ -5,7 +5,7 @@ export interface SelectCellFormatterProps {
   onChange(value: boolean, isShiftClick: boolean): void;
 }
 
-export default function SelectCellFormatter({ value, onChange }: SelectCellFormatterProps) {
+export function SelectCellFormatter({ value, onChange }: SelectCellFormatterProps) {
```
qili26: Quick question, why not change in the `formatters/index.tsx` like `export { default as SelectCellFormatter } from './SelectCellFormatter`

nstepien: Just consistency with other formatters.

PR #1897, opened 2020-01-16
Add Column.cellClass function support

nstepien:
Not sure if passing `row` is enough.
Not sure if it makes sense for header/filter cells to have custom classes as well, so I removed them.
WDYT?

### DIFF
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -12,7 +12,7 @@ interface ColumnValue<TRow, TField extends keyof TRow = keyof TRow> {
   key: TField;
   /** Column width. If not specified, it will be determined automatically based on grid width and specified widths of other columns*/
   width?: number | string;
-  cellClass?: string;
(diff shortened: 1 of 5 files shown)


amanmahajan7:
I think it makes sense as it is a common requirement to set different classes on difference rows. May be we can also add `column.headerCssClass`. It can just be a string. Or we can wait until there is a requirement

src/common/types.ts
```diff
@@ -12,7 +12,7 @@ interface ColumnValue<TRow, TField extends keyof TRow = keyof TRow> {
   key: TField;
   /** Column width. If not specified, it will be determined automatically based on grid width and specified widths of other columns*/
   width?: number | string;
-  cellClass?: string;
+  cellClass?: string | ((rowData: TRow) => string);
```
qili26: Maybe `rowData?: TRow` for the header cells?

PR #1898, opened 2020-01-16
rowData -> row

nstepien:


src/Row.tsx
```diff
@@ -47,7 +47,7 @@ export default function Row<R>({
           rowIdx={idx}
```
amanmahajan7: May be we should rename `idx` -> `rowIdx` in the next PR

PR #1899, opened 2020-01-16
rm onRowClick

nstepien:
Small gain in simplicity and flexibility.

### DIFF
--- a/src/Cell.tsx
+++ b/src/Cell.tsx
@@ -21,8 +21,6 @@ function Cell<R>({
   rowIdx,
   scrollLeft,
   eventBus,
-  onRowClick,
-  onRowDoubleClick,
   enableCellRangeSelection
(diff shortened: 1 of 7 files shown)


src/Cell.tsx
```diff
@@ -56,7 +53,6 @@ function Cell<R>({
   function handleCellDoubleClick(e: React.MouseEvent<HTMLDivElement>) {
     e.stopPropagation();
```
nstepien: Why do we do this? It'll prevent the double click event to be triggered on the Row `div`.

amanmahajan7: I am not sure
https://github.com/adazzle/react-data-grid/commit/1c273ed8b2a8b8eae01496b4da51deec496cc821#diff-d22f3ef92a32a6dcd0ca1ea0fe869cebR83

PR #1900, opened 2020-01-16
rm onRowDoubleClick prop

nstepien:


src/Cell.tsx
```diff
@@ -56,7 +55,6 @@ function Cell<R>({
   function handleCellDoubleClick(e: React.MouseEvent<HTMLDivElement>) {
     e.stopPropagation();
```
amanmahajan7: Do we want to keep `e.stopPropagation()`?

nstepien: https://github.com/adazzle/react-data-grid/pull/1123
It originates from this PR, but I can't think of a reason to keep this, I'll remove it.

PR #1906, opened 2020-01-17
Remove rowsContainer

amanmahajan7:
We can use `rowRenderer` to add `contextMenu`

### DIFF
--- a/src/Canvas.tsx
+++ b/src/Canvas.tsx
@@ -209,19 +209,6 @@ function Canvas<R, K extends keyof R>({
     return rowElements;
   }
 
-  const grid = (
-    <div
-      className="rdg-grid"
-      style={{
-        width: columnMetrics.totalColumnWidth,
-        paddingTop: rowOverscanStartIdx * rowHeight,
-        paddingBottom: (rows.length - 1 - rowOverscanEndIdx) * rowHeight
-      }}
-    >
-      {getViewportRows()}
-    </div>
-  );
-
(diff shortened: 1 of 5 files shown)


src/Canvas.tsx
```diff
@@ -212,18 +210,16 @@ function Canvas<R, K extends keyof R>({
...
-        style={{
-          width: columnMetrics.totalColumnWidth,
-          paddingTop: rowOverscanStartIdx * rowHeight,
-          paddingBottom: (rows.length - 1 - rowOverscanEndIdx) * rowHeight
-        }}
-      >
-        {getViewportRows()}
-      </div>
-    </RowsContainer>
+    <div
```
nstepien: We don't need to keep this in a separate variable now.

stories/demos/ContextMenu.tsx
```diff
@@ -4,7 +4,7 @@ import faker from 'faker';
 import { AutoSizer } from 'react-virtualized';
 import { ContextMenu, MenuItem, SubMenu, ContextMenuTrigger } from 'react-contextmenu';
 
-import DataGrid, { Column, Position, RowsContainerProps } from '../../src';
+import DataGrid, { Column, Row, RowRendererProps } from '../../src';
 import './react-contextmenu.less';
 
 interface Row {
```
nstepien: We have `Row` and `Row` now 🤔

amanmahajan7: Wow, not sure how it worked :)

PR #1907, opened 2020-01-17
rm enableCellSelect

nstepien:


src/utils/columnUtils.test.ts
```diff
@@ -123,319 +123,37 @@ describe('getColumnScrollPosition', () => {
...
-      });
-    });
+  it('should return correct booleans', () => {
+    const row: Row = { id: 1 };
+    const editor = () => null;
+
+    expect(canEdit(column, row)).toBe(false);
+    expect(canEdit({ ...column, editable: false }, row)).toBe(false);
+    expect(canEdit({ ...column, editable: true }, row)).toBe(true);
+    expect(canEdit({ ...column, editor }, row)).toBe(true);
```
amanmahajan7: duplicate line

src/utils/columnUtils.ts
```diff
@@ -85,11 +85,11 @@ function getSpecifiedWidth<R>(
...
 // Logic extented to allow for functions to be passed down in column.editable
 // this allows us to decide whether we can be editing from a cell level
-export function canEdit<R>(column: CalculatedColumn<R>, row: R, enableCellSelect?: boolean): boolean {
+export function canEdit<R>(column: CalculatedColumn<R>, row: R): boolean {
   if (typeof column.editable === 'function') {
-    return enableCellSelect === true && column.editable(row);
+    return column.editable(row);
   }
-  return enableCellSelect === true && (!!column.editor || !!column.editable);
+  return Boolean(column.editor || column.editable);
```
amanmahajan7: I think this creates some confusion. We allow editing if `column.editor` exists and `column.editable` is false and vice versa. May be we do not need to check `editor` as we default to `TextEditor`. Let's revisit in the future

PR #1909, opened 2020-01-17
rm onRowClick

nstepien:
Small gain in simplicity and flexibility.


jwallet:
what exactly can replace that, I found it useful to have an event that can tell me which column the click event came from. just having an event.target to determine which dom element was clicked is a bit fragile. would it be possible to have these data includes as a currying function `onClick = event => (rowIdx, row, column) => { };` or has added parameters to the function.

CHANGELOG.md
```diff
@@ -29,13 +29,20 @@
...
     - ⚠️ `onCellCopyPaste`
-    - ⚠️ `onGridKeyDown`
-    - ⚠️ `onGridKeyUp`
-    - ⚠️ `onRowDoubleClick`
     - ⚠️ `rowsContainer`
+    - ⚠️ `onRowDoubleClick`, `onRowClick`, `onGridKeyDown`, `onGridKeyUp`
+      - Use standard `div` props on the `Row` component itself instead, via `rowRenderer`:
+      ```jsx
+      function RowRenderer(props) {
+        return <Row onClick={(event) => {}} {...props} />;
```
jwallet, line 37: if the new way is to use `onClick` we should pass the old props to the clickEvent as well or as a currying function. so we can identify quickly which cell has been clicked on. Otherwise we have to do some DOM querySelector.

For instance, I want to navigate when pressing ctrl+click on the row, so I look for event.ctrlKey === true sent by the user and a click performed anywhere but on column.key === 'select-row'. currently there was a way with onRowClick to do with and using `window.event` to capture the current event.

nstepien:
>I found it useful to have an event that can tell me which column the click event came from

Yes we're not sure if we're going to merge this PR or not. I think renaming it to `onCellClick` would be appropriate at least, since we don't trigger `onRowClick` when clicking on a row itself.

PR #1913, opened 2020-01-17
Rename idx to rowIdx for clarity

amanmahajan7:


src/Canvas.tsx
```diff
@@ -182,12 +182,12 @@ function Canvas<R, K extends keyof R>({
...
     const rowElements = [];
-    for (let idx = rowOverscanStartIdx; idx <= rowOverscanEndIdx; idx++) {
-      const row = rows[idx];
+    for (let rowIdx = rowOverscanStartIdx; rowIdx <= rowOverscanEndIdx; rowIdx++) {
+      const row = rows[rowIdx];
       rowElements.push(
         <RowRenderer<R, K>
-          key={idx}
-          idx={idx}
+          key={rowIdx}
```
```suggestion
          key={row[rowKey]}
```
Shouldn't we do this instead? 🤔

amanmahajan7: I remember thinking about it but do not remember why I did not change this. Let's revisit in the next PR

amanmahajan7: We show also make `rowKey` a required prop before we make this change

PR #1916, opened 2020-01-21
Editor 2

nstepien:
https://www.youtube.com/watch?v=SSbBvKaM6sk


src/editors/EditorContainer2.tsx
```diff
@@ -0,0 +1,64 @@
...
+      fromRow: rowIdx,
+      toRow: rowIdx,
+      updated: rowUpdate as never, // TODO: fix type, or data structure
+      action: UpdateActions.CELL_UPDATE
+    });
+  }
+
+  function getContainerStyle(): React.CSSProperties | undefined {
+    if (!canvasRef.current) return;
+    const { top, left } = canvasRef.current.getBoundingClientRect();
```
amanmahajan7: Let's fix this bug in canary in case someone use it? Should we also not check editorPortal
https://github.com/adazzle/react-data-grid/pull/1871/files#diff-63a962509ad28f9bf4dbc6c8a1c50569L153

PR #1918, opened 2020-01-22
Add Header Filters story

amanmahajan7:
I have also removed unused CSS


stories/demos/HeaderFilters.tsx
```diff
@@ -0,0 +1,191 @@
...
+  return rows;
+}
+
+export default function HeaderFilters() {
+  const [rows] = useState(createRows);
+  const [filters, setFilters] = useState<Filters<Row>>({
+    priority: 'Critical',
+    issueType: ''
+  });
+  const [filteredRows, setFilteredRows] = useState(() => filterRows(filters));
```
nstepien: For derived states we should use `useMemo` instead of a state. It's so much safer IMO.
Example: https://github.com/adazzle/react-data-grid/blob/canary/stories/demos/CommonFeatures.tsx#L204

stories/demos/HeaderFilters.tsx
```diff
@@ -0,0 +1,191 @@
...
+  function onFiltersChange(filters: Filters<Row>) {
+    setFilters(filters);
+    setFilteredRows(filterRows(filters));
+  }
+
+  return (
+    <>
+      <Toolbar
+        enableFilter
+        numberOfRows={rows.length}
```
nstepien: Probably not the toolbar's responsibility to pick the `newRowIndex` 🤔

stories/demos/HeaderFilters.tsx
```diff
@@ -0,0 +1,191 @@
...
+        onToggleFilter={onToggleFilter}
+      />
+      <AutoSizer>
+        {({ height, width }) => (
+          <DataGrid<Row, 'id'>
```
nstepien: FWIW if you type everything correctly, then TS will correctly infer the generics.

stories/demos/HeaderFilters.tsx
amanmahajan7: I completely agree 👍

stories/demos/HeaderFilters.tsx
amanmahajan7: Yeah, this is an old component. Not sure if we should keep it

stories/demos/components/Filters/NumericFilter.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const value = parseInt(row[columnKey] as unknown as string, 10);
+  for (const ruleKey in columnFilter.filterTerm) {
+    const rule = columnFilter.filterTerm[ruleKey];
+
+    switch (rule.type) {
+      case RuleType.Number:
+        if (rule.value === value) {
+          return true;
+        }
+        break;
```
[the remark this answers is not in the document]
amanmahajan7: It is inside a loop so we cannot use it

stories/demos/HeaderFilters.tsx
amanmahajan7: I removed it and added custom buttons

PR #1921, opened 2020-01-23
Refactor summary row type

qili26:
1. add SummaryRowRenderer, SummaryRow, SummaryCell.
2. refine SummaryRow type.


stories/demos/CommonFeatures.tsx
```diff
@@ -17,6 +17,17 @@ function CurrencyFormatter({ value }: { value: number }) {
...
+interface SummaryRow {
+  id: string;
+  totalCount: number;
+  availableCount: {
+    yes: number;
+    no: number;
+  };
+}
+
+type SummaryRowFormatterProps = SummaryFormatterProps<SummaryRow, Row>;
```
qili26: Eventually it ends up like this. However, I still don't think we need to put the `SummaryRow` generic type in the `DataGrid` definition. Unlike regular `Row` which is associated with `Column` definition, the SummaryRow is completely siloed.

src/SummaryRowRenderer.tsx
```diff
@@ -1,6 +1,6 @@
 import { memo } from 'react';
```
qili26: Do we really want to keep this layer for `SummaryRow`?

src/SummaryRow.tsx
```diff
@@ -0,0 +1,39 @@
...
+  scrollLeft,
+  viewportColumns,
+  width
+}: SummaryRendererProps<R>) {
+  return (
+    <div
+      className={classNames(
+        'rdg-row',
+        `rdg-row-${rowIdx % 2 === 0 ? 'even' : 'odd'}`,
+        extraClasses
```
amanmahajan7: Is `extraClasses` being used?

src/common/types.ts
```diff
@@ -14,8 +14,11 @@ interface ColumnValue<TRow, TField extends keyof TRow = keyof TRow> {
   width?: number | string;
   cellClass?: string | ((row: TRow) => string);
   headerCellClass?: string;
+  summaryCellClass?: string | ((row: any) => string);
```
amanmahajan7: We are adding similar props for each section (header, filter, body, and footer)
- height, headerRowHeight, headerFiltersHeight
- cellClass, headerCellClass, summaryCellClass

I wonder if there is a better way to organize props so we do have to repeat them 🤔

src/SummaryRow.tsx
```diff
@@ -0,0 +1,39 @@
...
+  rowIdx,
+  lastFrozenColumnIndex,
+  row,
+  scrollLeft,
+  viewportColumns,
+  width
+}: SummaryRendererProps<R>) {
+  return (
+    <div
+      className={classNames(
```
amanmahajan7: May be we can add `rdg-summary-row` class as well so it is easy to customize styles. wdyt?

stories/demos/CommonFeatures.tsx
amanmahajan7: May be `DataGrid<R, K, SR = unknown>` is not a bad idea. Let's go with the current approach and we can revisit

src/SummaryCell.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  const style: React.CSSProperties = { width, left };
+
+  if (scrollLeft !== undefined) {
+    style.transform = `translateX(${scrollLeft}px)`;
+  }
+
+  return (
+    <div className={className} style={style}>
+      {SummaryFormatter && <SummaryFormatter rowIdx={rowIdx} column={column} row={row} />}
```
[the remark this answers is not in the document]
qili26: Any benefit for using `createElement` in this case? They are same, aren't they? https://reactjs.org/docs/react-api.html#createelement

src/SummaryRowRenderer.tsx
```diff
@@ -1,6 +1,6 @@
...
 
-import { RowRendererProps } from './common/types';
-import Row from './Row';
+import { SummaryRendererProps } from './common/types';
+import SummaryRow from './SummaryRow';
 
-export default memo(Row) as <R>(props: RowRendererProps<R>) => JSX.Element;
+export default memo(SummaryRow) as <R>(props: SummaryRendererProps<R>) => JSX.Element;
```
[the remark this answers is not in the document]
qili26: yeah, will do. I just kept the structure initially.

src/Canvas.tsx
```diff
@@ -218,11 +218,8 @@ function Canvas<R, K extends keyof R>({
           width={columnMetrics.totalColumnWidth + getScrollbarSize()}
           height={rowHeight}
           viewportColumns={viewportColumns}
-          isRowSelected={false}
           lastFrozenColumnIndex={columnMetrics.lastFrozenColumnIndex}
           scrollLeft={nonStickyScrollLeft}
-          isSummaryRow
```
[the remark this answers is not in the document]
qili26: nice catch, I forgot this instance. I will double check again.

src/SummaryRow.tsx
qili26: hm... I thought it was used for passing the extra class directly to the row, but looks like even for `Row` component it's a dead props as well. Do we want to align this props together in a new PR?

stories/demos/CommonFeatures.tsx
qili26: `DataGrid` level really doesn't need to know what does the `summaryRow` looks like as we directly pass the `summaryRow` to the formatter. Maybe we can see if we want to expend the Column type like `ColumnValue<TRow, , TField extends keyof TRow = keyof TRow, TSummaryRow - unknown>`. then we can do this like 

```
const columns: Column: readonly Column<Row, keyof Row, SummaryRow>[] = [
  { 
     ... ,
    summaryFormatter({ row }) {  }
  }
];
```

However, comparing this approach vs not even letting `Column` know the SR types, IMHO, having a separate types like `type SummaryRowFormatterProps = SummaryFormatterProps<SummaryRow, Row>;` is still better.

src/SummaryCell.tsx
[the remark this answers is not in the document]
amanmahajan7: There is no difference, I find `createElement` cleaner in this case 

```
const { summaryFormatter: SummaryFormatter } = column;
{SummaryFormatter && <SummaryFormatter rowIdx={rowIdx} column={column} row={row} />}
```

vs 

```
createElement(column.summaryFormatter, { rowIdx, row, column });
```

But I am fine with either

src/SummaryRow.tsx
amanmahajan7: I think we can remove it. `extraClasses` is set using `rowRenderer` but for summary rows we do not use the `rowRenderer`. This is the only place we call it

https://github.com/adazzle/react-data-grid/pull/1921/files#diff-6ee01b9be15a1d3a39e71ca4ea0d1a91R214

src/common/types.ts
nstepien: Shouldn't we change this `any`?

src/common/types.ts
```diff
@@ -14,8 +14,11 @@ interface ColumnValue<TRow, TField extends keyof TRow = keyof TRow> {
...
   /** Formatter to be used to render the cell content */
   formatter?: React.ComponentType<FormatterProps<TRow>>;
+  /** Formatter to be used to render the summary cell content */
+  summaryFormatter?: React.ComponentType<SummaryFormatterProps<any, TRow>>;
```
nstepien: Why `any`?

stories/demos/CommonFeatures.tsx
```diff
@@ -201,6 +233,20 @@ export default function CommonFeatures() {
   const [[sortColumn, sortDirection], setSort] = useState<[keyof Row, DEFINE_SORT]>(['id', DEFINE_SORT.NONE]);
   const [selectedRows, setSelectedRows] = useState(() => new Set<number>());
 
+  const summaryRows = useMemo(() => {
+    const summaryRows: SummaryRow[] = [{ id: 'total_0', totalCount: 0, availableCount: { yes: 0, no: 0 } }];
```
nstepien:
```suggestion
    const summaryRow: SummaryRow = { id: 'total_0', totalCount: rows.length, availableCount: { yes: 0, no: 0 } };
```

stories/demos/CommonFeatures.tsx
```diff
@@ -201,6 +233,20 @@ export default function CommonFeatures() {
...
+
+    for (const row of rows) {
+      if (row.available) {
+        summaryRows[0].availableCount.yes++;
```
nstepien:
```suggestion
        summaryRow.availableCount.yes++;
```

stories/demos/CommonFeatures.tsx
```diff
@@ -201,6 +233,20 @@ export default function CommonFeatures() {
...
+      } else {
+        summaryRows[0].availableCount.no++;
```
nstepien:
```suggestion
        summaryRow.availableCount.no++;
```

stories/demos/CommonFeatures.tsx
```diff
@@ -201,6 +233,20 @@ export default function CommonFeatures() {
...
+      }
+    }
+
+    summaryRows[0].totalCount = rows.length;
```
nstepien:
```suggestion
```

stories/demos/CommonFeatures.tsx
```diff
@@ -201,6 +233,20 @@ export default function CommonFeatures() {
...
+    return summaryRows;
```
nstepien:
```suggestion
    return [summaryRow];
```

stories/demos/CommonFeatures.tsx
```diff
@@ -201,6 +233,20 @@ export default function CommonFeatures() {
...
+  }, [rows]);
```
nstepien:
```suggestion
  }, [rows]);
  
```

src/Canvas.tsx
```diff
@@ -108,7 +108,7 @@ function Canvas<R, K extends keyof R>({
       const isCellAtLeftBoundary = left < scrollLeft + width + getFrozenColumnsWidth();
       const isCellAtRightBoundary = left + width > clientWidth + scrollLeft;
       if (isCellAtLeftBoundary || isCellAtRightBoundary) {
-        const newScrollLeft = getColumnScrollPosition(columns, idx, scrollLeft, clientWidth);
+        const newScrollLeft = getColumnScrollPosition(columns as readonly CalculatedColumn<R, never>[], idx, scrollLeft, clientWidth);
```
nstepien: Please avoid type assertions, we can fix `getColumnScrollPosition` by adding `SR` generic.
```suggestion
        const newScrollLeft = getColumnScrollPosition(columns, idx, scrollLeft, clientWidth);
```

src/Canvas.tsx
```diff
@@ -197,8 +197,8 @@ function Canvas<R, K extends keyof R>({
           key={key}
           rowIdx={rowIdx}
           row={row}
-          columnMetrics={columnMetrics}
-          viewportColumns={viewportColumns}
+          columnMetrics={columnMetrics as ColumnMetrics<R, never>}
+          viewportColumns={viewportColumns as readonly CalculatedColumn<R, never>[]}
```
nstepien: Here as well we should avoid type assertions as they may mask broken types in the future.

src/Columns.tsx
```diff
@@ -4,7 +4,7 @@ import { Column } from './common/types';
 
 // TODO: fix type
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const SelectColumn: Column<any> = {
+export const SelectColumn: Column<any, never> = {
```
nstepien:
```suggestion
export const SelectColumn: Column<any> = {
```
We default it to `never` so we can skip it.

src/DataGrid.tsx
```diff
@@ -94,13 +94,13 @@ export interface DataGridProps<R, K extends keyof R> {
   rowGroupRenderer?: React.ComponentType;
   emptyRowsView?: React.ComponentType<{}>;
   /** Component used to render a draggable header cell */
-  draggableHeaderCell?: React.ComponentType<{ column: CalculatedColumn<R>; onHeaderDrop(): void }>;
+  draggableHeaderCell?: React.ComponentType<{ column: CalculatedColumn<R, never>; onHeaderDrop(): void }>;
```
nstepien:
```suggestion
  draggableHeaderCell?: React.ComponentType<{ column: CalculatedColumn<R, SR>; onHeaderDrop(): void }>;
```
Why use `never` here?

src/DataGrid.tsx
```diff
@@ -94,13 +94,13 @@ export interface DataGridProps<R, K extends keyof R> {
...
 
   /**
    * Event props
    */
   /** Function called whenever a row is clicked */
-  onRowClick?(rowIdx: number, row: R, column: CalculatedColumn<R>): void;
+  onRowClick?(rowIdx: number, row: R, column: CalculatedColumn<R, never>): void;
```
nstepien:
```suggestion
  onRowClick?(rowIdx: number, row: R, column: CalculatedColumn<R, SR>): void;
```

src/DataGrid.tsx
```diff
@@ -172,7 +172,7 @@ function DataGrid<R, K extends keyof R>({
     if (viewportWidth <= 0) return null;
 
     return getColumnMetrics<R>({
-      columns,
+      columns: columns as unknown as readonly Column<R, never>[],
```
nstepien:
```suggestion
      columns,
```

src/DataGrid.tsx
```diff
@@ -172,7 +172,7 @@ function DataGrid<R, K extends keyof R>({
     if (viewportWidth <= 0) return null;
 
     return getColumnMetrics<R>({
```
nstepien:
```suggestion
    return getColumnMetrics<R, SR>({
```

src/DataGrid.tsx
```diff
@@ -191,7 +191,7 @@ function DataGrid<R, K extends keyof R>({
     });
   }, [columnMetrics, scrollLeft]);
 
-  const viewportColumns: readonly CalculatedColumn<R>[] = useMemo(() => {
+  const viewportColumns: readonly CalculatedColumn<R, never>[] = useMemo(() => {
```
nstepien:
```suggestion
  const viewportColumns: readonly CalculatedColumn<R, SR>[] = useMemo(() => {
```

src/DataGrid.tsx
```diff
@@ -216,7 +216,7 @@ function DataGrid<R, K extends keyof R>({
     };
   }, [width]);
 
-  function handleColumnResize(column: CalculatedColumn<R>, width: number) {
+  function handleColumnResize(column: CalculatedColumn<R, never>, width: number) {
```
nstepien:
```suggestion
  function handleColumnResize(column: CalculatedColumn<R, SR>, width: number) {
```

src/DataGrid.tsx
qili26: For regular `Row` related definitions, what's the point for passing the `SR` into them? Same question for the `Column` definition being used in calculation.

nstepien: >For regular Row related definitions, what's the point for passing the SR into them?

Because it's the correct thing to do.
We pass these columns back to user code, so their types much match the types the user code has, even if it's outside the summary row rendering context..

qili26: It's still not convincing to me. For the regular `Row`, in the `Row` context, it doesn't have the `Summary Row` definition at all which will always point to `never`. That's why I thought it make sense to default `SummaryRow` to `never`.

http://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults

My understanding for the default type is that, the user doesn't need to constantly put the default type whey they use it. In our case, most of time when the grid doesn't have summary row being defined or passed, it is just `never`, so the user doesn't need to explicitly write like 
```
return <DataGrid<Row, 'id', never> xxx />;
```
and the following would just work
```
return <DataGrid<Row, 'id'> xxx />;
```

nstepien: Imagine this:
```
<DataGrid
  onSomeEvent={({ column }) => myEventHandler(column)}
/>
```
Here your `myEventHandler` will use the `Column<Row, SummaryRow>` type, so `Column<Row, never>` will be incompatible.

qili26: In your example, I think it makes sense to define `myEventHandler` as `myEventHandler(c: Column<Row, never>) {}`, and the call back functions should follow RDG type definition. However, I will try to remove all `= never` in another commit.

src/common/types.ts
```diff
@@ -5,7 +5,7 @@ import EventBus from '../EventBus';
 
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
-export interface Column<TRow> {
+export interface Column<TRow, TSummaryRow> {
```
nstepien:
```suggestion
export interface Column<TRow, TSummaryRow = never> {
```
That way we can remove a bunch of `never` in `Column<..., never>`

src/utils/index.ts
```diff
@@ -4,7 +4,7 @@ export * from './viewportUtils';
 export * from './keyboardUtils';
 export * from './selectedCellUtils';
 
-export function assertIsValidKey<R>(key: unknown): asserts key is keyof R {
+export function assertIsValidKey<R, SR>(key: unknown): asserts key is keyof R {
```
nstepien:
```suggestion
export function assertIsValidKey<R>(key: unknown): asserts key is keyof R {
```
The generic is unused here.

src/DataGrid.tsx
```diff
@@ -171,8 +171,8 @@ function DataGrid<R, K extends keyof R>({
   const columnMetrics = useMemo(() => {
     if (viewportWidth <= 0) return null;
 
-    return getColumnMetrics<R>({
-      columns,
+    return getColumnMetrics<R, SR>({
+      columns: columns as unknown as readonly Column<R, SR>[],
```
nstepien:
```suggestion
      columns,
```

src/Canvas.tsx
```diff
@@ -108,7 +108,7 @@ function Canvas<R, K extends keyof R>({
       const isCellAtLeftBoundary = left < scrollLeft + width + getFrozenColumnsWidth();
       const isCellAtRightBoundary = left + width > clientWidth + scrollLeft;
       if (isCellAtLeftBoundary || isCellAtRightBoundary) {
-        const newScrollLeft = getColumnScrollPosition(columns, idx, scrollLeft, clientWidth);
+        const newScrollLeft = getColumnScrollPosition(columns as readonly CalculatedColumn<R, SR>[], idx, scrollLeft, clientWidth);
```
```suggestion
        const newScrollLeft = getColumnScrollPosition(columns, idx, scrollLeft, clientWidth);
```

src/Canvas.tsx
```diff
@@ -193,12 +193,12 @@ function Canvas<R, K extends keyof R>({
...
 
       rowElements.push(
-        <RowRenderer<R>
+        <RowRenderer<R, SR>
           key={key}
           rowIdx={rowIdx}
           row={row}
-          columnMetrics={columnMetrics}
-          viewportColumns={viewportColumns}
+          columnMetrics={columnMetrics as ColumnMetrics<R, SR>}
```
```suggestion
          columnMetrics={columnMetrics}
```

src/Canvas.tsx
```diff
@@ -193,12 +193,12 @@ function Canvas<R, K extends keyof R>({
...
+          viewportColumns={viewportColumns as readonly CalculatedColumn<R, SR>[]}
```
```suggestion
          viewportColumns={viewportColumns}
```

src/Canvas.tsx
```diff
@@ -243,10 +240,10 @@ function Canvas<R, K extends keyof R>({
         ref={canvasRef}
         onScroll={handleScroll}
       >
-        <InteractionMasks<R>
+        <InteractionMasks<R, SR>
           rows={rows}
           rowHeight={rowHeight}
-          columns={columns}
+          columns={columns as readonly CalculatedColumn<R, SR>[]}
```
```suggestion
          columns={columns}
```

stories/demos/CommonFeatures.tsx
```diff
@@ -163,8 +178,21 @@ const columns: readonly Column<Row>[] = [
...
+    summaryCellClass: 'common-feature-summary-available',
     formatter(props) {
       return <>{props.row.available ? '✔️' : '❌'}</>;
+    },
+    summaryFormatter({ column, row }) {
+      return (
+        <>{column.key === 'available' && (
+          <>
+            <div>{`✔️: ${row.availableCount.yes}`}</div>
+            <div>{`❌: ${row.availableCount.no}`}</div>
```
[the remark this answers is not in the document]
qili26: sure, will make the change.

stories/demos/AllFeatures.tsx
```diff
@@ -62,7 +62,7 @@ export default function AllFeatures() {
   const gridRef = useRef<DataGridHandle>(null);
 
   const columns = useMemo((): Column<Row>[] => [
-    SelectColumn,
+    getSelectColumn<Row>(),
```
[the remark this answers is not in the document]
qili26: Typescript will emit error for this case, saying `Column<Row,never>` and `Column<Row, any>` are different. (or cannot assign `any` to `never`, forgot the exact term). In fact, I believe `never` is not in the `any` set.

src/utils/columnUtils.ts
```diff
@@ -1,21 +1,21 @@
 import { Column, CalculatedColumn, ColumnMetrics, FormatterProps, Omit } from '../common/types';
 import { getScrollbarSize } from './domUtils';
 
-interface Metrics<R> {
-  columns: readonly Column<R>[];
+interface Metrics<R, SR = never> {
```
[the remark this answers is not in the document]
qili26: I think it make sense, will make it more strict to only allow the exported types to have the default `= never`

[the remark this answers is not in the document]
qili26: Actually, this might end up with adding tons of `never` back in test files like `getColumnMetrics<Row, never>`

src/FilterRow.tsx
```diff
@@ -4,28 +4,28 @@ import classNames from 'classnames';
 import { CalculatedColumn, Filters } from './common/types';
 import { DataGridProps } from './DataGrid';
 
-type SharedDataGridProps<R> = Pick<DataGridProps<R, never>,
+type SharedDataGridProps<R, SR = never> = Pick<DataGridProps<R, never, SR>,
```
amanmahajan7: I agree with Nico's comment. We do not need ` = never` here. Moreover we already have `SR` on line 20

nstepien:
I've removed `{SummaryCellProps,SummaryFormatterProps}.rowIdx`, I don't think we have a need for it. We can always add it back if we actually do.

stories/demos/components/Editors/DropDownEditor.tsx
```diff
@@ -8,13 +8,13 @@ interface Option {
   text: string;
 }
 
-interface DropDownEditorProps<TRow> extends EditorProps<string, TRow> {
+interface DropDownEditorProps<TRow, TSummaryRow> extends EditorProps<string, TRow, TSummaryRow> {
```
amanmahajan7: Do we need this change? It should work without specifying `SR`

stories/demos/components/Filters/NumericFilter.tsx
```diff
@@ -12,14 +12,14 @@ type Rule =
   | { type: RuleType.Range; begin: number; end: number }
   | { type: RuleType.GreaterThan | RuleType.LessThan | RuleType.Number; value: number };
 
-interface ChangeEvent<R> {
+interface ChangeEvent<R, SR> {
```
amanmahajan7: Same here. We will probably have to change `TSummaryRow = any` in `FilterRendererProps`

src/common/types.ts
```diff
@@ -5,7 +5,7 @@ import EventBus from '../EventBus';
 
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
-export interface Column<TRow> {
+export interface Column<TRow, TSummaryRow = unknown> {
```
amanmahajan7: May be `any` type is better here?

amanmahajan7: NM, let;s keep `unknown` here

src/Cell.tsx
```diff
@@ -89,26 +88,25 @@ function Cell<R>({
...
+      onMouseEnter={!enableCellRangeSelection ? onMouseEnter : wrapEvent(handleCellMouseEnter, onMouseEnter)}
       {...props}
     >
       {children}
     </div>
   );
 }
 
-export default memo(Cell) as <R>(props: CellRendererProps<R>) => JSX.Element;
+export default memo(Cell) as <R, SR = unknown>(props: CellRendererProps<R, SR>) => JSX.Element;
```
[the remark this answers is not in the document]
nstepien: We do export it publicly. We can make it stricter and see how it goes though.

src/SummaryRow.tsx
```diff
@@ -0,0 +1,52 @@
+import classNames from 'classnames';
+import React, { memo } from 'react';
+
+import SummaryCell from './SummaryCell';
+import { RowRendererProps } from './common/types';
+
+type SharedRowRendererProps<TRow, TSummaryRow> = Pick<RowRendererProps<TRow, TSummaryRow>,
```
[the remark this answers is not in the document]
nstepien: How so?

src/Cell.tsx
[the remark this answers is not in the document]
amanmahajan7: ah, then we can keep it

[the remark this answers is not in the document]
amanmahajan7: Do we want to do the same for `Row`?

src/SummaryRow.tsx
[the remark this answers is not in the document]
amanmahajan7: Just the name `R vs TRow` and `SR vs TSummaryRow` but we can can keep what we have

amanmahajan7:
We also need to update change log

src/Cell.tsx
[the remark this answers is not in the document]
nstepien: I'll do it. TBH if you just do pass the props through, ts should infer the generic correctly.

src/SummaryRow.tsx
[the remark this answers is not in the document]
qili26: For `TRow` vs `R`, I remember we had some inconsistencies when I went through the whole code base, maybe we should handle this separately together.

PR #1928, opened 2020-01-28
Square SelectColumn

nstepien:
https://www.youtube.com/watch?v=LB5YkmjalDg

### DIFF
--- a/src/utils/columnUtils.ts
+++ b/src/utils/columnUtils.ts
@@ -75,8 +75,7 @@ function getSpecifiedWidth<R>(
     return columnWidths.get(column.key);
   }
   if (typeof column.width === 'number') {
-    // TODO: allow width to be less than minWidth?
-    return Math.max(column.width, minColumnWidth);
+    return column.width;
   }
   if (typeof column.width === 'string' && /^\d+%$/.test(column.width)) {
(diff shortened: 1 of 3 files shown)


src/utils/columnUtils.ts
```diff
@@ -75,8 +75,7 @@ function getSpecifiedWidth<R>(
     return columnWidths.get(column.key);
   }
   if (typeof column.width === 'number') {
-    // TODO: allow width to be less than minWidth?
-    return Math.max(column.width, minColumnWidth);
```
amanmahajan7: What happens when we resize a column that has a width less than `minColumnWidth`? I will probably jump to min width
https://github.com/adazzle/react-data-grid/blob/fb8229d5c12ee459a89f7439af59fd0b5663ed8e/src/DataGrid.tsx#L224

nstepien: I think that's an edge case we can live with.

amanmahajan7: @qili26 fixed this issue a few times so I think we will have to fix it

https://github.com/adazzle/react-data-grid/pulls?utf8=%E2%9C%93&q=is%3Apr+resize+is%3Aclosed+author%3Aqili26+

PR #1929, opened 2020-01-29
Require rowKey prop for selection

nstepien:


src/test/index.tsx
```diff
@@ -20,6 +20,7 @@ export function getProps(extraProps?: Partial<DataGridProps<RowType, 'id'>>): Da
       name: 'Column 2'
     }],
     rows,
+    rowKey: 'id',
```
nstepien: Do we need this still?

src/Canvas.tsx
```diff
@@ -184,20 +185,29 @@ function Canvas<R, K extends keyof R>({
     const rowElements = [];
     for (let rowIdx = rowOverscanStartIdx; rowIdx <= rowOverscanEndIdx; rowIdx++) {
       const row = rows[rowIdx];
+      let key: string | number = rowIdx;
+      let isRowSelected = false;
+      if (rowKey !== undefined) {
+        const rowKeyValue = row[rowKey];
```
```suggestion
        const rowId = row[rowKey];
```

src/editors/EditorContainer.tsx
```diff
@@ -127,18 +127,16 @@ export default function EditorContainer<R, K extends keyof R>({
...
-          column={column}
-          value={getInitialValue() as R[keyof R & string] & R[keyof R & number] & R[keyof R & symbol]}
-          row={row}
-          height={height}
-          onCommit={commit}
-          onCommitCancel={onCommitCancel}
-          onOverrideKeyDown={onKeyDown}
-        />
-      );
+      return createElement(column.editor as ComponentType<EditorProps<string | R[(keyof R & string) | (keyof R & number) | (keyof R & symbol)], R>>, {
```
nstepien: We didn't need to change this?

src/utils/index.ts
```diff
@@ -3,3 +3,9 @@ export * from './columnUtils';
 export * from './viewportUtils';
 export * from './keyboardUtils';
 export * from './selectedCellUtils';
+
+export function assertIsValidKey<R>(key: unknown): asserts key is keyof R {
+  if (key === undefined) {
```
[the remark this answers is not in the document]
amanmahajan7: We need to check that the `rowKey` is defined. At this point let's not restrict the key type. we can revisit if needed in the future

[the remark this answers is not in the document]
qili26: We are checking it here, aren't we? That's why I thought they would be consistent.

https://github.com/adazzle/react-data-grid/pull/1929/files#diff-6ee01b9be15a1d3a39e71ca4ea0d1a91R193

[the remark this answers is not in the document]
amanmahajan7: We are checking the value here `row[rowKey]` and this function checks `rowKey`

src/editors/EditorContainer.tsx
amanmahajan7: Reverted

PR #1934, opened 2020-01-31
column.key: string

nstepien:


src/formatters/ValueFormatter.tsx
```diff
@@ -2,5 +2,5 @@ import React from 'react';
 import { FormatterProps } from '../common/types';
 
 export function ValueFormatter<R>(props: FormatterProps<R>) {
-  return <>{props.row[props.column.key]}</>;
+  return <>{props.row[props.column.key as keyof R]}</>;
```
nstepien: IMO we should remove the default defaultFormatter, and delete `ValueFormatter`. Columns should just do
```tsx
formatter(props) { return <>{props.row.specificValue}</>; }
```

amanmahajan7: we can consider it

src/utils/columnUtils.ts
```diff
@@ -15,7 +15,7 @@ export function getColumnMetrics<R>(metrics: Metrics<R>): ColumnMetrics<R> {
   let allocatedWidths = 0;
   let unassignedColumnsCount = 0;
   let lastFrozenColumnIndex = -1;
-  const columns: Array<Column<R> & { width: number | void }> = [];
+  const columns: Array<Column<R> & { width: number | undefined }> = [];
```
amanmahajan7: `void` was failing for some reason

qili26: I think `void` means no data type, `undefined` is a valid `nothing exists` type. They are still slightly different.

src/common/types.ts
```diff
@@ -37,13 +37,12 @@ export interface Column<TRow> {
...
 }
 
-export type CalculatedColumn<TRow> =
-  Column<TRow> & {
-    idx: number;
-    width: number;
-    left: number;
-    formatter: React.ComponentType<FormatterProps<TRow>>;
-  };
+export interface CalculatedColumn<TRow> extends Column<TRow> {
```
amanmahajan7: Do we need to omit `Column<Row>.formatter`?

nstepien: Doesn't look like it, it's overriding both width and formatter accordingly.

PR #1941, opened 2020-02-03
Add unsafe_onCelllInput hook

amanmahajan7:
- [x] Add `unsafe_onCellInput` prop to control editor activation. We can rename it once the API is finalized
- [x]  Editor and Selection masks component are mutually exclusive so we only render one of them. Using a separate `isEditorEnabled` makes things little complicated as we have to check this flag in various places. I added a new `status` field to `selectedPosition` state.


amanmahajan7:
I will fix the tests after the review

src/editors/EditorContainer.tsx
```diff
@@ -119,8 +106,6 @@ export default function EditorContainer<R>({
       e.stopPropagation();
     } else if (['Enter', 'Tab', 'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'].includes(e.key)) {
       commit();
-    } else if (e.key === 'Escape') {
```
amanmahajan7: `InteractionsMasks` handles `keydown` as well and it seems like we can remove this event from `EditorContainer` completely in the future. `onCommitCancel` is handled by `InteractionMasks`

src/masks/InteractionMasks.tsx
```diff
@@ -140,54 +125,83 @@ export default function InteractionMasks<R>({
...
-    }
+  function canOpenEditor(event: React.KeyboardEvent<HTMLDivElement>): boolean {
+    if (selectedPosition.status === 'EDIT' || !isCellEditable(selectedPosition)) return false;
+
+    const column = columns[selectedPosition.idx];
+    const row = rows[selectedPosition.rowIdx];
+    const isEditable = column.unsafe_onCellInput?.(event, row);
+
+    // Let the user override the activation logic
+    if (typeof isEditable === 'boolean') return isEditable;
```
amanmahajan7: I am not 100% sure about this but I think it makes sense to allow users to override activation logic completely

src/masks/InteractionMasks.tsx
```diff
@@ -86,34 +72,33 @@ export default function InteractionMasks<R>({
   onRowsUpdate,
   scrollToCell
 }: InteractionMasksProps<R>) {
-  const [selectedPosition, setSelectedPosition] = useState<Position>(() => {
+  const [selectedPosition, setSelectedPosition] = useState<SelectCellState | EditCellState>(() => {
     if (enableCellAutoFocus && document.activeElement === document.body && columns.length > 0 && rows.length > 0) {
-      return { idx: 0, rowIdx: 0 };
+      return { idx: 0, rowIdx: 0, status: 'SELECT' };
```
amanmahajan7: Using a `status` simplifies state logic little bit

stories/demos/AllFeatures.tsx
```diff
@@ -85,11 +85,7 @@ export default function AllFeatures() {
       width: 200,
       resizable: true,
       formatter(props) {
-        return (
-          <div onClick={() => gridRef.current?.openCellEditor(props.rowIdx, props.column.idx)}>
-            {props.row.title}
-          </div>
```
amanmahajan7: This stopped working and I think the following happens
- Click on the cell.
- Event is handled by the formatted first and it calls `openCellEditor` which opens the editor
- Event is handled by the cell, it fires `SELECT_CELL` event which selects the cell and hides the editor.

I think previously there was some timing issue

src/masks/InteractionMasks.tsx
```diff
@@ -86,34 +76,33 @@ export default function InteractionMasks<R>({
...
   useEffect(() => {
     return eventBus.subscribe('SELECT_CELL', selectCell);
   });
 
+  const closeEditor = useCallback(() => {
+    setSelectedPosition(({ idx, rowIdx }) => ({ idx, rowIdx, status: 'SELECT' }));
+  }, []);
+
+  // close editor when scrolling
+  useEffect(() => closeEditor, [scrollTop, scrollLeft, closeEditor]);
```
nstepien: This should be moved in the editor container.
- every time we scroll, we re-render InteractionMasks, and the effect cleanup function (`closeEditor`) will be called
- `closeEditor` then updates selectedPosition with a new state. Since it's an new object, it'll trigger a re-render

That doesn't sound too good. 🤔

src/masks/InteractionMasks.tsx
```diff
@@ -140,53 +129,85 @@ export default function InteractionMasks<R>({
...
+        nextPosition = { idx, rowIdx: rowIdx + 1 };
+        break;
+      case 'ArrowLeft':
+        nextPosition = { idx: idx - 1, rowIdx };
+        break;
+      case 'ArrowRight':
+        nextPosition = { idx: idx + 1, rowIdx };
+        break;
+      case 'Tab':
+        nextPosition = shiftKey ? { idx: idx - 1, rowIdx } : { idx: idx + 1, rowIdx };
```
nstepien:
```suggestion
        nextPosition = { idx: idx + (shiftKey ?  -1 : 1), rowIdx };
```

nstepien: Also shouldn't this handle cellNavigationMode?

src/masks/InteractionMasks.tsx
```diff
@@ -140,53 +129,85 @@ export default function InteractionMasks<R>({
...
   }
 
-  function openEditor(event: React.KeyboardEvent<HTMLDivElement>): void {
-    if (!isEditorEnabled && isCellEditable(selectedPosition)) {
-      setFirstEditorKeyPress(event.key);
-      setIsEditorEnabled(true);
+  function onKeyDown(event: React.KeyboardEvent<HTMLDivElement>): void {
+    const column = columns[selectedPosition.idx];
+    const row = rows[selectedPosition.rowIdx];
+    const isActivedByUser = (column.unsafe_onCellInput ?? legacyCellInput)(event, row) === true;
```
nstepien: `Actived` isn't a word.
```suggestion
    const isActivatedByUser = (column.unsafe_onCellInput ?? legacyCellInput)(event, row) === true;
```
`Triggered` is also good.

stories/demos/AllFeatures.tsx
nstepien: `event.stopPropagation()` might fix that.

stories/demos/CommonFeatures.tsx
```diff
@@ -136,6 +136,10 @@ const columns: readonly Column<Row>[] = [
     sortable: true,
     formatter(props) {
       return <CurrencyFormatter value={props.row.budget} />;
+    },
+    editable: true,
+    unsafe_onCellInput({ key }) {
+      return /^\d$/.test(key);
```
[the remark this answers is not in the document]
amanmahajan7: probably, I will look into it

src/masks/InteractionMasks.tsx
amanmahajan7: Moved it back to EditorContainer

src/masks/InteractionMasks.tsx
amanmahajan7: It is handled by the mode param

https://github.com/adazzle/react-data-grid/pull/1941/files#diff-63a962509ad28f9bf4dbc6c8a1c50569R233

stories/demos/CommonFeatures.tsx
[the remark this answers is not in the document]
amanmahajan7: I ended up removing this for now. I will add back when we add `editor2`

src/masks/InteractionMasks.tsx
```diff
@@ -140,94 +122,104 @@ export default function InteractionMasks<R>({
...
+      default:
+        if (canOpenEditor && isActivatedByUser) {
+          setSelectedPosition(({ idx, rowIdx }) => ({ idx, rowIdx, status: 'EDIT', key }));
+        }
+        break;
     }
   }
 
   function onPressTab(e: React.KeyboardEvent<HTMLDivElement>): void {
-    // When there are no rows in the grid, we need to allow the browser to handle tab presses
```
amanmahajan7: I do not think this is possible. We show selection mask if the cell is within row/column bound and if there are no rows then `selectionMask` is not rendered. We set the `tabIndex` on the selection mask so `onKeyDown` will never be trigged

amanmahajan7:
@qili26 okay with merging this?

stories/demos/AllFeatures.tsx
```diff
@@ -188,6 +184,12 @@ export default function AllFeatures() {
 
   const handleAddRow = useCallback(({ newRowIndex }: { newRowIndex: number }): void => setRows([...rows, createFakeRowObjectData(newRowIndex)]), [rows]);
 
+  const handleRowClick = useCallback((rowIdx: number, row: Row, column: CalculatedColumn<Row>) => {
+    if (column.key === 'title') {
+      gridRef.current?.openCellEditor(rowIdx, column.idx);
```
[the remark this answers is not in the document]
amanmahajan7: I am not sure yet. We can probably just use `grid.selectCell`

PR #1947, opened 2020-02-10
Add github actions CI

nstepien:
Docs:
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows
- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-nodejs-with-github-actions
- https://github.com/actions/setup-node
- https://github.com/actions/cache


nstepien:
Could also speed up typechecking with the `incremental` option but it doesn't work with `noEmit` enabled.
https://github.com/microsoft/TypeScript/issues/30661

PR #1953, opened 2020-02-13
Editor2

amanmahajan7:
- [ ] Finalize `editor2` api
- [ ] Finalize `onCellInput` behavior
- [ ] How to close and commit editor value? Old `Editor` and `EditorContainer` use a few ways
  - `Text` and `Drodown` editors call `commit` on `blur` event
  - `EditorContainer` calls `commit` on [`componentWillUnmount`](https://github.com/adazzle/react-data-grid/blob/alpha/packages/react-data-grid/src/common/editors/EditorContainer.tsx#L56)
  - `ClickOutside` component


amanmahajan7:
Replaced by https://github.com/adazzle/react-data-grid/pull/2102

PR #1971, opened 2020-03-02
Update quick-start with import css example

samselfridge:


nstepien:
We do not publish a stylesheet anymore

PR #1982, opened 2020-03-24
Drop support for IE11

nstepien:


style/variables.less
```diff
@@ -12,13 +10,4 @@
...
   }
-
-  // IE
-  &:-ms-input-placeholder {
-    color: @color;
-  }
-  // Edge
-  &::-ms-input-placeholder {
-    color: @color;
-  }
```
nstepien: Since canary won't be used for a while still, I figure we may as well remove those now.

src/DataGrid.tsx
```diff
@@ -265,15 +264,13 @@ function DataGrid<R, K extends keyof R, SR>({
...
               onSort={props.onSort}
-              scrollLeft={nonStickyScrollLeft}
             />
             {enableFilters && (
               <FilterRow<R, SR>
                 height={headerFiltersHeight}
                 width={columnMetrics.totalColumnWidth + getScrollbarSize()}
                 lastFrozenColumnIndex={columnMetrics.lastFrozenColumnIndex}
                 columns={viewportColumns}
-                scrollLeft={nonStickyScrollLeft}
```
amanmahajan7, line 276: Do you think we can remove `scrollLeft` from `InteractionMasks` also?

src/masks/InteractionMasks.tsx
```diff
@@ -114,7 +114,7 @@ export default function InteractionMasks<R, SR>({
   function getEditorPosition() {
     if (!canvasRef.current) return null;
     const { left, top } = canvasRef.current.getBoundingClientRect();
-    const { scrollTop: docTop, scrollLeft: docLeft } = document.scrollingElement || document.documentElement;
+    const { scrollTop: docTop, scrollLeft: docLeft } = document.scrollingElement!;
```
amanmahajan7, line 117: Do we need `!`? Can it be null?

style/variables.less
```diff
@@ -2,8 +2,6 @@
 
 .user-select(@value: none) {
   -webkit-user-select: @value;
-  -moz-user-select: @value;
```
amanmahajan7, line 5: Do we not need `moz-user-select`?

src/masks/InteractionMasks.tsx
nstepien: image

Unlikely at best. 🤔

src/DataGrid.tsx
nstepien: I believe we need it to position the editor, and close the editor when scrolling.
We can look into it further in the future.

style/variables.less
nstepien: It's been unprefixed since Firefox 69.

src/masks/InteractionMasks.tsx
amanmahajan7: Although we are not supporting IE, I think it is okay to add back `document.scrollingElement || document.documentElement;` so rdg is somewhat usable in IE. It is not a big change anyway

src/masks/InteractionMasks.tsx
```diff
@@ -276,12 +276,7 @@ export default function InteractionMasks<R, SR>({
     e.dataTransfer.effectAllowed = 'copy';
     // Setting data is required to make an element draggable in FF
     const transferData = JSON.stringify(selectedPosition);
-    try {
-      e.dataTransfer.setData('text/plain', transferData);
-    } catch (ex) {
-      // IE only supports 'text' and 'URL' for the 'type' argument
-      e.dataTransfer.setData('text', transferData);
-    }
+    e.dataTransfer.setData('text/plain', transferData);
```
amanmahajan7, line 279: Same here, an easy fix for IE

qili26:
@amanmahajan7 
I agree with you. Stopping supporting shouldn't make IE 11 critical malfunction (e.g. result in a blank page). I think we shouldn't assume the value is always there. Other than that, this is really a thrilling change!

nstepien:
image
We don't render columns that would be hidden behind frozen columns, so it'd be buggy anyway in IE. Do we really want to support IE11?

qili26:
Just wondering if we should block the frozen feature for IE to not show the incorrect UI? By saying block, I meant when we see the flag is on, we still don't render that feature at all.

amanmahajan7:
We do not want to do extra work to support IE but if we can prevent the page from crashing that would be nice. We would mention in the docs that IE is not supported and IE related bugs will not be addressed. I think it is fine if frozen columns do not work.

PR #1986, opened 2020-03-26
Commit changes when EditorContainer is unmounted

amanmahajan7:
This matches the legacy behavior.
https://github.com/adazzle/react-data-grid/blob/alpha/packages/react-data-grid/src/common/editors/EditorContainer.tsx#L56

`ClickOutside` does not commit changes when another cell is selected. I think it may be due to fact that component is unmounted before `ClickOutside` has a chance to commit changes. You can remove `onBlur` event handler in `SimpleTextEditor` to test this fix
https://github.com/adazzle/react-data-grid/blob/canary/src/editors/SimpleTextEditor.tsx#L25

Most of the textbox editors do not handle onBlur event so I think we should add the commit logic back

### DIFF
--- a/src/editors/EditorContainer.tsx
+++ b/src/editors/EditorContainer.tsx
@@ -139,6 +139,8 @@ export default function EditorContainer<R, SR>({
       e.stopPropagation();
     } else if (['Enter', 'Tab', 'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'].includes(e.key)) {
       commit();
+    } else if (e.key === 'Escape') {
+      commitCancel();
     }
   }
 
--- a/src/editors/EditorContainer.tsx
+++ b/src/editors/EditorContainer.tsx
@@ -40,6 +40,7 @@ export default function EditorContainer<R, SR>({
   const [isValid, setValid] = useState(true);
   const prevScrollLeft = useRef(scrollLeft);
   const prevScrollTop = useRef(scrollTop);
+  const isUnmounting = useRef(false);
 
   const getInputNode = useCallback(() => editorRef.current?.getInputNode(), []);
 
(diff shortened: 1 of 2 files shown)


src/editors/EditorContainer.tsx
```diff
@@ -114,6 +139,8 @@ export default function EditorContainer<R, SR>({
       e.stopPropagation();
     } else if (['Enter', 'Tab', 'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'].includes(e.key)) {
       commit();
+    } else if (e.key === 'Escape') {
+      commitCancel();
```
amanmahajan7: Need to call cancel changes otherwise changes are committed on `Escape`
https://github.com/adazzle/react-data-grid/blob/canary/src/masks/InteractionMasks.tsx#L180 this line closes the editor so we need to set `changeCanceled` flag

src/editors/EditorContainer.tsx
```diff
@@ -50,7 +60,21 @@ export default function EditorContainer<R, SR>({
   }, [getInputNode]);
 
   // close editor when scrolling
-  useEffect(() => onCommitCancel, [scrollTop, scrollLeft, onCommitCancel]);
+  useEffect(() => {
+    if (scrollLeft !== prevScrollLeft.current || scrollTop !== prevScrollTop.current) {
+      commitCancel();
+    }
+  }, [commitCancel, scrollLeft, scrollTop]);
```
nstepien: Why not keep it like that?
```suggestion
  useEffect(() => commitCancel, [scrollTop, scrollLeft, commitCancel]);
```

src/editors/EditorContainer.tsx
```diff
@@ -50,7 +60,21 @@ export default function EditorContainer<R, SR>({
...
+
+  // commit changes when editor is closed
+  useEffect(() => {
+    return () => {
+      if (!changeCommitted.current && !changeCanceled.current) {
+        commit();
+      }
+    };
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
nstepien: Looks dangerous.
I wonder if we do something stupid like this:
```ts
  useEffect(() => () => {
    markAsActuallyUnmounting.current = true;
  }, []);

  useEffect(() => () => {
    if (markAsActuallyUnmounting.current && !changeCommitted.current && !changeCanceled.current) {
      commit();
    }
  });
```

src/editors/EditorContainer.tsx
amanmahajan7: It fires when the component is unmounted. We want to call `commitCancel` only when the scroll position is changed

src/editors/EditorContainer.tsx
amanmahajan7: I think it is relatively safe to omit `commit` from the dependency list here but I try your suggession

PR #1988, opened 2020-03-26
Cleanup onRowsUpdate type

amanmahajan7:
I am not sure if there is an easy way to support types for `onRowsUpdate` handler. We might as well cleanup the internal types

### DIFF
--- a/src/masks/InteractionMasks.tsx
+++ b/src/masks/InteractionMasks.tsx
@@ -241,7 +241,7 @@ export default function InteractionMasks<R, SR>({
       cellKey,
       fromRow,
       toRow,
-      updated: { [cellKey]: value } as never,
+      updated: { [cellKey]: value },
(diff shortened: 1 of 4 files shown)


src/masks/InteractionMasks.tsx
```diff
@@ -241,7 +241,7 @@ export default function InteractionMasks<R, SR>({
       cellKey,
       fromRow,
       toRow,
-      updated: { [cellKey]: value } as never,
+      updated: { [cellKey]: value },
```
qili26: Since the update is not tied to the row key, I'm wondering if we should make this loosen?

amanmahajan7: Not sure I follow? Can you give an example?

qili26: Ignore this, I thought it was something else.

PR #1989, opened 2020-03-26
Add column.{min,max}Width

nstepien:


src/Columns.tsx
```diff
@@ -8,6 +8,7 @@ export const SelectColumn: Column<any, any> = {
   key: 'select-row',
   name: '',
   width: 35,
+  maxWidth: 35,
```
nstepien: `width` doesn't override `minColumnWidth` anymore, so we specify `maxWidth`.
`width` is just the default width.
We still specify `width` here so we properly calculate `unallocatedColumnWidth`.

PR #1993, opened 2020-04-01
gitignore on subfolders of /node_modules /dist /lib package-lock.json

jwallet:
when switching between branches from 7.0.0, to alpha to canary and get thousands of files added to my unstagged files because the folder hierarchy changed


nstepien:
We've only got the `main` branch now, so this isn't needed anymore. Thanks

PR #1994, opened 2020-04-02
EmptyRowsView in canvas to keep horizontal scroll position when filtering

jwallet:
Currently:
EmptyRowsView is outside the viewport,
It does not render the proper width, h-scroll loosing position when having overflow headers
** it also breaks the headers position **

current

Wanted Behavior:
EmptyRowsView inside the viewport,
This way we can share to this component the calculated height and width of the viewport so it can adjusted itself to keep track of the headers (when many) and so we don't lose current position of the horizontal scroll when filtering results

expected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,10 @@
 /coverage
-/dist
-/lib
-/node_modules
+**/dist
+**/lib
+**/node_modules
 /storybook-static
 /.eslintcache
-/package-lock.json
+**/package-lock.json
 
 npm-debug.log
 **.orig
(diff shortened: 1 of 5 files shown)


jwallet:
@nstepien if you can look at this as well

if you have any other way to implement this and have the same behavior I'm all open to change it.

amanmahajan7:
Closing this PR as the internals of canary have been changed so this PR is not longer needed

https://github.com/adazzle/react-data-grid/blob/canary/CHANGELOG.md
https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--common-features
https://github.com/adazzle/react-data-grid/tree/canary/stories/demos

PR #1995, opened 2020-04-06
Delete react-data-grid-addons

nstepien:


muXxer:
@nstepien 
Why was that removed btw?
I can't find a reason in the changelogs, and I really liked the old examples with the addons.
Is there a new way I'm missing? :)

PR #2000, opened 2020-04-16
Remove material-ui packages, fixes #1999

### LINKED ISSUE #1999: Material-ui is a heavy dependency for a few icons
Is `@material-ui/core` and `@material-ui/icons` really needed as dependencies? 

I was sort of surprised when checking in a `package-lock.json` file that so many packages had been installed when I added react-data-grid. After looking in to it, it seams that only icons are used from material-ui, which makes it seem kind of a heavy dependency.

I know a dependency does not equal me deploying a larger bundle, but I still want my dependencies to be manageable.

nstepien:
image
image

### DIFF
--- a/stories/demos/components/Formatters/ChildRowDeleteButton.tsx
+++ b/stories/demos/components/Formatters/ChildRowDeleteButton.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { RemoveCircle } from '@material-ui/icons';
 
 export interface ChildRowDeleteButtonProps {
   onDeleteSubRow: () => void;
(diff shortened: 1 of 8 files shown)


stories/demos/components/Formatters/ChildRowDeleteButton.tsx
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import { RemoveCircle } from '@material-ui/icons';
```
oliviertassinari, line 2: This would have removed the bundle size overhead in dev mode (tree shaking is ineffective in dev mode).

```jsx
import RemoveCircle from '@material-ui/icons/RemoveCircle';
```

PR #2006, opened 2020-04-20
Remove openCellEditor, Cell.props.children, onSelectedCellRangeChange

nstepien:
- removed `...props` parameter spread in
  - `DataGrid`
  - `HeaderCell`,
  - `HeaderRow`
  - `RowRenderer`
  - `ResizableHeaderCell`
- removed `openCellEditor` in the ref handle
- removed support for `children` prop in `Cell`
- removed `onSelectedCellRangeChange` prop, and cell range selection related code
  - We can revisit this when we refactor/remove InteractionMasks
  - Does it need to be opt-in? what features does it provide?


ShantanuNair:
> removed `onSelectedCellRangeChange` prop, and cell range selection related code
> * We can revisit this when we refactor/remove InteractionMasks
> * Does it need to be opt-in? what features does it provide?

onSelectedCellRangeChange was useful for handling copy/paste functionality.

PR #2007, opened 2020-04-20
Add columns reordering example

amanmahajan7:


CHANGELOG.md
```diff
@@ -36,6 +36,9 @@
     - ⚠️ `onGridKeyDown`
     - ⚠️ `onGridKeyUp`
     - ⚠️ `onRowDoubleClick`
+    - ⚠️ `onHeaderDrop`
+    - ⚠️ `draggableHeaderCell`
+      - Check [#2006](https://github.com/adazzle/react-data-grid/pull/2006) on how to migrate
```
nstepien, line 41:
```suggestion
      - Check [#2007](https://github.com/adazzle/react-data-grid/pull/2007) on how to migrate
```

stories/demos/ColumnsReordering.tsx
```diff
@@ -0,0 +1,129 @@
...
+      return <DraggableHeaderRenderer {...props} onColumnsReorder={handleColumnsReorder} />;
+    }
+
+    function handleColumnsReorder(sourceKey: string, targetKey: string) {
+      const sourceColumn = columns.find(c => c.key === sourceKey)!;
+      const targetColumn = columns.find(c => c.key === targetKey)!;
+
+      const reorderedColumns = columns.map(c => {
+        if (c === sourceColumn) return targetColumn;
+        if (c === targetColumn) return sourceColumn;
```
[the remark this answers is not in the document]
amanmahajan7: Makes sense, I will update it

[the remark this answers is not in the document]
amanmahajan7: Copied old logic
https://codesandbox.io/s/z6l3xrjlo4?from-embed=&file=/src/index.js:865-978

stories/demos/ColumnsReordering.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }, []);
+
+  const draggableColumns = useMemo(() => {
+    function HeaderRenderer(props: HeaderRendererProps<Row>) {
+      return <DraggableHeaderRenderer {...props} onColumnsReorder={handleColumnsReorder} />;
+    }
+
+    function handleColumnsReorder(sourceKey: string, targetKey: string) {
+      const sourceColumnIndex = columns.findIndex(c => c.key === sourceKey)!;
+      const targetColumnIndex = columns.findIndex(c => c.key === targetKey)!;
```
nstepien, line 81:
```suggestion
      const sourceColumnIndex = columns.findIndex(c => c.key === sourceKey);
      const targetColumnIndex = columns.findIndex(c => c.key === targetKey);
```

stories/demos/components/HeaderRenderers/DraggableHeaderRenderer.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
```
qili26, line 1: I'm just not sure about this file. What do you guys think if we provide this in the DataGrid and export this as a default `DraggableHeaderRenderer`?

stories/demos/components/HeaderRenderers/DraggableHeaderRenderer.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useDrag, useDrop, DragObjectWithType } from 'react-dnd';
+
+import { HeaderRendererProps } from '../../../../src';
+
```
qili26, line 5: And we don't need this empty line.

stories/demos/components/HeaderRenderers/DraggableHeaderRenderer.tsx
amanmahajan7: I would prefer to keep the dependencies to a minimum and provide a flexible API so users can write their own implementations. Composition is always more maintainable than adding extra props

nstepien: Yes we have to maintain the feature + dependency on other libs.
That way users can use whatever implementation they want to header dragging.

PR #2008, opened 2020-04-20
Fix for Browserify without babelify

bradleat:
Since you are building to the esnext standard, some builders (browserify) are unable to build your code into a bundle without help or special configuration. 

This simple change allows more people to run your package without problems.

Also, we use the `--experimental-modules ` flag so more versions of node can run your `buildStylesheets.mjs` script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "scripts": {
     "start": "start-storybook --quiet -p 6006",
     "build": "tsc",
-    "postbuild": "node tools/buildStylesheets.mjs",
+    "postbuild": "node --experimental-modules tools/buildStylesheets.mjs",
     "test": "jest",
     "test:watch": "jest --watch",
     "eslint": "eslint --ext mjs,ts,tsx --max-warnings 0 -f codeframe --cache --color src stories jest tools",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,7 +6,7 @@
     "inlineSources": true,
     "jsx": "react",
     "lib": ["es2020", "dom", "dom.iterable"],
-    "module": "esnext",
+    "module": "commonjs",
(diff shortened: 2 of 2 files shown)


bradleat:
@qili26 can you let me know who can review this

right now your npm package doesn't work for some because of the build rules you are using. this set of build rules opens it up to more targets

bradleat:
@amanmahajan7

@nstepien

nstepien:
It looks like you should be able to configure babelify to transform esm into cjs.
https://github.com/browserify/browserify/issues/1186#issuecomment-572119852

bradleat:
@nstepien 
I've tried babelify doesn't work because it breaks other well known packages. What's the problem with exporting the commonjs style modules until the esnext ecosystem is more mature?

nstepien:
https://github.com/babel/babelify#why-arent-files-in-node_modules-being-transformed
Looks like this config should work for you.
```js
browserify().transform("babelify", {
  global: true,
  ignore: /\/node_modules\/(?!react-data-grid\/lib\/)/
});
```

Modern bundlers like rollup/webpack/etc have great support for es modules. We can reconsider publishing additional cjs modules if there's really a need for it.

bradleat:
Seeing as browserify is the second most downloaded bundler according to npm, I imagine there is a need for it. As for my needs: I've published a package with these changes.

Browserify will have better support for mjs soon, until then I don't understand the benefit you are getting for chasing the bleeding edge before the standard is actually set.

nstepien:
@bradleat 
We publish both esm and cjs modules now, targeting the [latest browsers and maintained Node.js versions](https://github.com/adazzle/react-data-grid/blob/main/browserslist), can you confirm if this bundling with browserify is still an issue?

PR #2012, opened 2020-04-21
Remove legacy grouping props

amanmahajan7:
Cell selection on the grouped row is off on the grouped row but that is how [legacy ](https://w741nqnxv7.csb.app/)worked so I am not changing it. Let's revisit when we add the new grouping API


src/RowRenderer.tsx
```diff
@@ -1,40 +1,26 @@
...
 
-interface IRowRendererProps<R, SR> extends SharedDataGridProps<R, SR> {
-  viewportColumns: readonly CalculatedColumn<R, SR>[];
-  lastFrozenColumnIndex: number;
-  rowIdx: number;
-  row: R;
-  eventBus: EventBus;
-  isRowSelected: boolean;
-}
+type IRowRendererProps<R, SR> = RowRendererProps<R, SR> & SharedDataGridProps<R, SR>;
```
qili26, line 12: Do we still need this? Or we should use it at ln#23

src/RowRenderer.tsx
```diff
@@ -1,40 +1,26 @@
...
   row,
-  rowGroupRenderer,
   rowRenderer,
   isRowSelected,
-  onRowClick,
-  onRowExpandToggle
-}: IRowRendererProps<R, SR>) {
-  const { __metaData } = row as RowData;
+  onRowClick
+}: RowRendererProps<R, SR> & SharedDataGridProps<R, SR>) {
```
amanmahajan7, line 23: I wonder if we should use memo on the `Row` component and delete `RowRenderer`. Thoughts?

qili26: We still want to have the ability to customize the rowrenderer, right? I believe that's the reason why we kept this layer.

stories/demos/LegacyGrouping.tsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  return gridRows;
+}
+
+export default function LegacyGrouping() {
+  const [rows] = useState(createRows);
+  const [groupByColumns] = useState(['priority', 'issueType']);
+  const [expandedGroups, setExpandedGroups] = useState(new Set(['Low', 'Low_Epic']));
```
qili26, line 87: Don't we need this `() => new Set(['Low', 'Low_Epic'])` in `useState`?

stories/demos/components/RowRenderers/GroupRowRenderer.less
```diff
@@ -0,0 +1,15 @@
...
+    color: #777;
+  }
+}
+
+.rdg-row-default-group {
+  width: var(--row-width);
+  height: var(--row-height);
+  line-height: var(--row-height);
+  padding: 0 8px;
+  border-bottom: 1px solid #ddd;
```
qili26, line 14: `border-bottom` seems like a common variable, I'm wondering if we want to create a css variable for this.

package.json
```diff
@@ -74,6 +75,7 @@
     "jest-environment-jsdom-sixteen": "^1.0.3",
     "less": "^3.11.1",
     "less-loader": "^5.0.0",
+    "lodash.groupby": "^4.6.0",
```
nstepien, line 78: Use the `lodash` package directly instead. It's more likely to already be installed by other deps, and we get less duplicate code (`lodash/groupBy` + `lodash.groupby`).
If you're worried about bundle size, then `import ... from 'lodash/...'` should be used instead.

src/RowRenderer.tsx
nstepien, line 23:
```suggestion
}: IRowRendererProps<R, SR>) {
```

nstepien: Let's just delete this module, memoize `Row`, and use the Row/rowRenderer directly.
Custom row renderers can memo themselves.

stories/demos/LegacyGrouping.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [columnKey, ...remainingColumnKeys] = columnKeys;
+  const groupedRows = groupBy(rows, columnKey);
+  const groupedKeys = Object.keys(groupedRows);
+  for (const key of groupedKeys) {
+    const groupKey = parentKey ? `${parentKey}_${key}` : key;
+    const isExpanded = expandedGroups.has(groupKey);
+    const rowGroupHeader = {
+      groupKey,
+      name: key,
+      __metaData: {
```
nstepien, line 67: I think it'd be better to refactor this to a more realistic data structure, maybe something like
```ts
interface RowGroup {
  type: 'group';
  // ...
}

interface RowData {
  type: 'data';
  // ...
}

type Row = RowGroup | RowData;
```

Not worth our time right now though.

stories/demos/LegacyGrouping.tsx
```diff
@@ -0,0 +1,111 @@
...
+      gridRows.push(...groupByColumn(groupedRows[key], remainingColumnKeys, expandedGroups, treeDepth + 1, key));
+    }
+  }
+
+  return gridRows;
+}
+
+export default function LegacyGrouping() {
+  const [rows] = useState(createRows);
+  const [groupByColumns] = useState(['priority', 'issueType']);
```
nstepien, line 86: Let's move this outside the component.
```suggestion
  const groupByColumns = ['priority', 'issueType'];
```

stories/demos/LegacyGrouping.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [expandedGroups, setExpandedGroups] = useState(new Set(['Low', 'Low_Epic']));
+
+  const gridRows = useMemo(() => {
+    return groupByColumn(rows, groupByColumns, expandedGroups);
+  }, [rows, groupByColumns, expandedGroups]);
```
nstepien, line 91:
```suggestion
  }, [rows, expandedGroups]);
```

stories/demos/LegacyGrouping.tsx
amanmahajan7: Simplified types a bit

stories/demos/LegacyGrouping.tsx
amanmahajan7: It is not required but I have made the change

PR #2014, opened 2020-04-21
Optimize rendering with css-contain

nstepien:
- https://developers.google.com/web/updates/2016/06/css-containment
- https://developer.mozilla.org/en-US/docs/Web/CSS/contain
- https://drafts.csswg.org/css-contain/

`contain` allows the browser to enable further optimizations.

I've also enabled `flex` layout for all rows, as I found out by inspecting the layers that they were taller than the actual elements, and may reduce rendering performance. Setting `vertical-top: top` on the cells can also fix it, but I figure we should just adopt a box-y layout, rather than a line-y layout for this, it just makes more sense.

## Before
image

## After
image

Much cleaner! We can get rid of the "Overlaps other composited content." reason by removing the interaction masks and the drop shadow on last frozen cells.

---

Speaking of drop shadows, we won't don't bleed over the previous row anymore! I've set the shadow color to `red` so it's more visible.

## Before
image

## After
image

---

If you scroll very quickly, you'll also notice that Chrome defers the painting of parts of the grid.
We basically render faster than Chrome can paint.
As I understand it, Chrome favors scrolling rate over painting rate. You should notice that scrolling is a bit smoother now as well.

image


style/core.less
```diff
@@ -19,3 +18,11 @@
     box-sizing: inherit;
   }
 }
+
+// set stacking context in safari
```
nstepien, line 22: https://bugs.webkit.org/show_bug.cgi?id=172026

style/cell.less
```diff
@@ -1,5 +1,5 @@
 .rdg-cell {
-  display: inline-block;
+  contain: strict;
```
qili26, line 2: We don't need to do the checking like this? https://github.com/adazzle/react-data-grid/pull/2014/files#diff-95d4fa4e583fb3fa7c9d33f79b431c52R23

nstepien: I've removed `display: inline-block` as it's redundant with `display: flex` on the rows.
It shouldn't break anything if `contain; strict` isn't supported here.

qili26: Cool. I'm still not sure what does the `contain: strict` really do here. I'm still reading the doc that you linked. Based on your screenshots, this looks nice.

style/row.less
```diff
@@ -1,4 +1,6 @@
 .rdg-row {
+  contain: strict;
```
amanmahajan7, line 2: Do we need to set contain at the row level? Does it result in better performance? Or is it mainly to fix shadows? Should change the shadow css instead?

nstepien: It should result in better performance as the browser has more opportunity to optimize rendering. We can safely enable it afaict, so I think it's worth keeping.

PR #2015, opened 2020-04-22
Boost layout performance

nstepien:
Since we know all the widths/heights/positions of rows/cells, we can absolutely position the rows, and replace **two** _padding_ divs with very dynamic `height`s, with **one** div with a more static `height`.
This improves layout perf from ~5ms to ~3.1ms on my machine. Might also improve more than layout rendering, not sure.

## Before:
image

## After:

image


nstepien:
I've also fixed summary rows positioning so they're always at the bottom of the grid, even if we don't have enough rows to fill the available space.

## Before:
image
## After:
image

src/Row.tsx
```diff
@@ -47,6 +48,7 @@ function Row<R, SR = unknown>({
       onDragEnter={wrapEvent(handleDragEnter, onDragEnter)}
       onDragOver={wrapEvent(handleDragOver, onDragOver)}
       onDrop={wrapEvent(preventDefault, onDrop)}
+      style={{ top }}
```
qili26, line 51: Just 1 concern here. Let's say if the user has their own `RowRenderer` and they forgot to add this `top` to their wrapper. The whole layout would be messed up, right?

nstepien: Yuyp. If you pass `top` back to `Row`, or to your wrapper, you'll be fine though.

qili26: To be honest, I don't know if this worth it given by 5ms to 3ms improvement. As it makes the `RowRenderer` error prone. And I bet people will create issues for this case when they use `CustomRowRenderer`.

nstepien: Aren't custom row renderers already error-prone?

amanmahajan7:
@nstepien can you catchup?

nstepien:
@amanmahajan7 Done.

PR #2016, opened 2020-04-22
Remove onCheckCellIsEditable prop

nstepien:
I've also made it so `editable: false` disables the editor, whether we have an editor or not.


src/utils/columnUtils.ts
```diff
@@ -109,7 +109,7 @@ export function canEdit<R, SR>(column: CalculatedColumn<R, SR>, row: R): boolean
   if (typeof column.editable === 'function') {
     return column.editable(row);
   }
-  return Boolean(column.editor || column.editable);
+  return column.editable || column.editor != null && column.editable !== false;
```
nstepien, line 112: editor | editable | result
-------|----------|-------
no | false | false
no | true | true => `!editor && editable`
yes | false | false
yes | true | true => `editor && editable`
no | nullish | false
yes | nullish | true => `editor && editable == null`

```
(!editor && editable || editor && editable)
  => editable
```
So you end up with
```js
editable || editor != null && editable != null
// or
editable || editor != null && editable !== false
```

qili26: To be honest, I'm not sure about the `nullish` cases. I think to reduce the complexity for understanding it would be nice to make sure the editor doesn't open when the `editable` is not set to any value. It means less flexibility, but helpful for consumer dev debugging.

nstepien: Adding `editable: true` for all your columns with editors is cumbersome and unintuitive IMO.

qili26: It is cumbersome but more precise.

src/utils/columnUtils.ts
```diff
@@ -164,7 +164,7 @@ export function canEdit<R, SR>(column: Column<R, SR>, row: R): boolean {
   if (typeof column.editable === 'function') {
     return column.editable(row);
   }
-  return Boolean(column.editor ?? column.editor2 ?? column.editable);
+  return column.editable || (column.editor != null || column.editor2 != null) && column.editable !== false;
```
amanmahajan7, line 167: A bit confusing with many or and and conditions.

```
if (typeof column.editable === 'function') return column.editable(row);
if (column.editable === false) return false;
return column.editor != null || column.editor2 != null;
```

Wdyt?

src/utils/columnUtils.ts
```diff
@@ -164,7 +164,7 @@ export function canEdit<R, SR>(column: Column<R, SR>, row: R): boolean {
   if (typeof column.editable === 'function') {
```
amanmahajan7, line 164: I wonder if we should move `canEdit` to `selectedCellUtils.ts`. It is only used once

src/utils/columnUtils.ts
nstepien: This won't enable the default editor if `editable` is `true` and `editor` is nullish.
Any confusion could be addressed by having unit tests for this function.

amanmahajan7: Are you thinking of using a default editor?

nstepien: https://github.com/adazzle/react-data-grid/blob/canary/src/editors/SimpleTextEditor.tsx
https://github.com/adazzle/react-data-grid/blob/canary/src/editors/EditorContainer.tsx#L161
We default to the `SimpleTextEditor` 🤔

amanmahajan7: But only for the old editor. How would we handle it for the new editor once the old editor is removed?

src/utils/columnUtils.ts
nstepien: Done.

amanmahajan7 (approved):

PR #2018, opened 2020-04-24
Replace classnames with clsx

nstepien:
- https://github.com/lukeed/clsx
- `classnames` is an umd module,, `clsx` provides an es module, types are bundled as well.
- Should be faster than `classnames`
  https://github.com/lukeed/clsx/tree/master/bench
- `classnames` uses a temporary array + `join(' ')`, while `clsx` just does string concatenation, which is well optimized by js engines.
- `@material-ui` packages use `clsx`
- There's also a babel plugin to optimize both :eyes:
  https://github.com/merceyz/babel-plugin-optimize-clsx/tree/master/benchmark

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "postpublish": "git push --follow-tags origin HEAD"
   },
   "dependencies": {
-    "classnames": "^2.2.6"
+    "clsx": "^1.1.0"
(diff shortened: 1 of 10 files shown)


oliviertassinari:
👌

PR #2061, opened 2020-06-23
Remove InteractionMask

amanmahajan7:
Scrolling profile

**Before**

image

**After**

image

It is more or less the same.


src/DataGrid.tsx
```diff
@@ -441,29 +630,8 @@ function DataGrid<R, K extends keyof R, SR>({
...
-              scrollToCell={scrollToCell}
-              editorPortalTarget={editorPortalTarget}
-              onCheckCellIsEditable={onCheckCellIsEditable}
-              onRowsUpdate={handleRowsUpdate}
-              onSelectedCellChange={onSelectedCellChange}
-            />
-          )}
           <div style={{ height: Math.max(rows.length * rowHeight, clientHeight) }} />
-          {getViewportRows()}
+          {viewportWidth > 0 && getViewportRows()}
```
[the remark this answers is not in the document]
amanmahajan7: it was causing the same issue where the grid is initially scrolled

src/DataGrid.tsx
```diff
@@ -297,9 +309,107 @@ function DataGrid<R, K extends keyof R, SR>({
       onSelectedRowsChange(newSelectedRows);
     };
 
-    return eventBus.subscribe('SELECT_ROW', handleRowSelectionChange);
+    return eventBus.subscribe('ROW_SELECT', handleRowSelectionChange);
   }, [eventBus, onSelectedRowsChange, rows, rowKey, selectedRows]);
 
+  useEffect(() => {
+    return eventBus.subscribe('CELL_SELECT', selectCell);
+  });
```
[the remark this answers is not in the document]
amanmahajan7: Removed most of the effect. I will check the remaining but I think it is fine as are used for dragging and not very expensive

src/DataGrid.tsx
```diff
@@ -129,8 +145,6 @@ export interface DataGridProps<R, K extends keyof R, SR = unknown> {
    */
   /** Toggles whether filters row is displayed or not */
   enableFilters?: boolean;
-  /** Toggles whether cells should be autofocused */
-  enableCellAutoFocus?: boolean;
```
qili26, line 133: What was this for at first? Making (0,0) cell selected upon mounting the grid?

src/DataGrid.tsx
```diff
@@ -393,16 +651,34 @@ function DataGrid<R, K extends keyof R, SR>({
...
+        onKeyDown={handleKeyDown}
+      >
+        {rowElements}
+        {getEditorContainer()}
+      </div>
+    );
+  }
+
+  // Reset the positions if the current values are no longer valid. This can happen if a column or row is removed
+  if (selectedPosition.idx > columns.length || selectedPosition.rowIdx > rows.length) {
```
qili26, line 678: As `idx` is 0 based, shouldn't we use `>=`?

src/DataGrid.tsx
```diff
@@ -393,16 +651,34 @@ function DataGrid<R, K extends keyof R, SR>({
...
+    setSelectedPosition({ idx: -1, rowIdx: -1, mode: 'SELECT' });
```
qili26, line 679: I'd say we create this `{ idx: -1, rowIdx: -1, mode: 'SELECT' }` to a separate standalone variable to be reused.

https://github.com/adazzle/react-data-grid/pull/2061/files#diff-60532bd8286dc6c925b8ea6ee3e2573bR241

src/DataGrid.tsx
```diff
@@ -206,6 +215,9 @@ function DataGrid<R, K extends keyof R, SR>({
   const [scrollTop, setScrollTop] = useState(0);
   const [scrollLeft, setScrollLeft] = useState(0);
   const [columnWidths, setColumnWidths] = useState<ReadonlyMap<string, number>>(() => new Map());
+  const [selectedPosition, setSelectedPosition] = useState<SelectCellState | EditCellState>({ idx: -1, rowIdx: -1, mode: 'SELECT' });
```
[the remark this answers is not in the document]
qili26: I think this makes sense to stand for `defaultSelectionState`. In addition, we can skip empty check for this usage.

https://github.com/adazzle/react-data-grid/pull/2061/files#diff-60532bd8286dc6c925b8ea6ee3e2573bR436

src/DataGrid.tsx
```diff
@@ -297,9 +332,36 @@ function DataGrid<R, K extends keyof R, SR>({
...
-    return eventBus.subscribe('SELECT_ROW', handleRowSelectionChange);
+    return eventBus.subscribe('ROW_SELECT', handleRowSelectionChange);
   }, [eventBus, onSelectedRowsChange, rows, rowKey, selectedRows]);
 
+  useEffect(() => {
+    return eventBus.subscribe('CELL_SELECT', selectCell);
+  });
+
+  useEffect(() => {
+    const handleDragEnter = (overRowIdx: number) => {
```
qili26, line 343: Use regular function rather than arrow for consistency?

src/DataGrid.tsx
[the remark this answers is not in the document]
qili26: How about moving these 3 together in 1 useEffect?

```js
useEffect(() => {
  const fns = [
    eventBus.subscribe(),
    eventBus.subscribe(),
    eventBus.subscribe()
  ];
  return () => fns.forEach(fn => fn());
});
```

https://github.com/adazzle/react-data-grid/pull/2061/files#diff-60532bd8286dc6c925b8ea6ee3e2573bR357
https://github.com/adazzle/react-data-grid/pull/2061/files#diff-60532bd8286dc6c925b8ea6ee3e2573bR361

src/DataGrid.tsx
```diff
@@ -332,13 +418,114 @@ function DataGrid<R, K extends keyof R, SR>({
...
+  function handleCommit({ cellKey, rowIdx, updated }: CommitEvent) {
+    onRowsUpdate?.({
+      cellKey,
+      fromRow: rowIdx,
+      toRow: rowIdx,
+      updated,
+      action: UpdateActions.CELL_UPDATE
+    });
+
+    setSelectedPosition(({ idx, rowIdx }) => ({ idx, rowIdx, mode: 'SELECT' }));
```
qili26, line 430: Why not just using spread operator, I thought that would be more concise and we can avoid same variable name though they are in different scope?

```javascript
    setSelectedPosition(position => ({ ...position, mode: 'SELECT' }));
```

src/DataGrid.tsx
```diff
@@ -332,13 +418,114 @@ function DataGrid<R, K extends keyof R, SR>({
...
+      fromRow: rowIdx,
+      toRow: draggedOverRowIdx,
+      updated: { [cellKey]: value } as unknown as never,
+      action: UpdateActions.CELL_DRAG
+    });
+
+    setDraggedOverRowIdx(undefined);
+  }
+
+  function handleDragHandleDoubleClick(): void {
```
qili26, line 452: maybe clear this a little bit `: void` for consistency

src/DataGrid.tsx
```diff
@@ -332,13 +418,114 @@ function DataGrid<R, K extends keyof R, SR>({
...
+  }
+
+  function handleCopy() {
+    const { idx, rowIdx } = selectedPosition;
+    const value = rows[rowIdx][columns[idx].key as keyof R];
+    setCopiedPosition({ idx, rowIdx, value });
+  }
+
+  function handlePaste() {
+    if (copiedPosition === null || !isCellEditable(selectedPosition)) {
```
qili26, line 473: Maybe also check if the `copiedPosition` is same as the `selectCellPosition` to skip dispatching the update event?

src/DataGrid.tsx
amanmahajan7: good catch

src/DataGrid.tsx
```diff
@@ -11,20 +11,26 @@ import React, {
 } from 'react';
 
 import EventBus from './EventBus';
-import InteractionMasks from './masks/InteractionMasks';
 import HeaderRow from './HeaderRow';
 import FilterRow from './FilterRow';
 import Row from './Row';
 import SummaryRow from './SummaryRow';
 import { ValueFormatter } from './formatters';
+import { legacyCellInput } from './editors/CellInputHandlers';
```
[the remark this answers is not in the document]
amanmahajan7: Done

src/DataGrid.tsx
amanmahajan7: `selectedPosition` also has a `key` field that we do not want to copy

src/DataGrid.tsx
amanmahajan7: Correct. I think we should not steal the focus and let the user select the first cell using `grid.selectCell`

src/DataGrid.tsx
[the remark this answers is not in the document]
amanmahajan7: Moved all the effects to handle dragging

src/Cell.tsx
```diff
@@ -72,6 +111,15 @@ function Cell<R, SR>({
         isRowSelected={isRowSelected}
         onRowSelectionChange={onRowSelectionChange}
       />
+      {enableCellDragAndDrop && isSelected && (
```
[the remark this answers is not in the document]
amanmahajan7: Fixed

src/DataGrid.tsx
```diff
@@ -670,6 +673,7 @@ function DataGrid<R, K extends keyof R, SR>({
           rowClass={rowClass}
           top={rowIdx * rowHeight + totalHeaderHeight}
           setDraggedOverRowIdx={isDragging ? setDraggedOverRowIdx : undefined}
+          dragHandle={selectedPosition.rowIdx === rowIdx ? getDragHandle() : undefined}
```
qili26, line 676: This looks weird combining this condition with the checks inside the `getDragHandle` function at ln#635. I think it will be better for maintenance to either add a checking outside like this

```js
dragHandle={displayDragHandle() ? getDragHandle() : undefined} // remove the checks from getDragHandle
```

or completely move all conditions inside `getDragHandle` like this

```js
function getDragHandle() {
    if (selectedPosition.rowIdx !== rowIdx || !enableCellDragAndDrop || !isCellEditable(selectedPosition) || selectedPosition.mode === 'EDIT') return null;
    ...
```

amanmahajan7:
@nstepien is it worth adding a new `Viewport` component? Currently we render header and summary rows on each cell navigation. Header and SummaryRow are memoized so I am not sure if there will be any performance gain. It however does extract Viewport logic in a separate component which may be easier to maintain. Thoughts?

nstepien:
Rather than creating separate components, I'd rather we split logic into custom hooks, for example we can have all the columns handling in one hook.

amanmahajan7:
That sounds good to me, may be we can try in the next PR. It does feel the DataGrid component is getting a bit complex

src/DataGrid.tsx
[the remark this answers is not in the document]
amanmahajan7: I think it is okay to use a new `{ idx: -1, rowIdx: -1 }` for now. Let's revisit once we add hooks and cleanup code

src/DataGrid.tsx
```diff
@@ -297,9 +309,107 @@ function DataGrid<R, K extends keyof R, SR>({
...
+        return;
+      }
+
+      const { rowIdx: toRow } = position;
+
+      const cellKey = columns[position.idx].key;
+      const { rowIdx: fromRow, idx, value } = copiedPosition;
+      const fromCellKey = columns[idx].key;
+
+    onRowsUpdate?.({
```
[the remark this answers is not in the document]
amanmahajan7: Fixed

src/Cell.tsx
```diff
@@ -1,67 +1,106 @@
...
+      cellRef.current?.focus();
+    }
+  }, [isSelected]);
+
   function selectCell(openEditor?: boolean) {
-    eventBus.dispatch('SELECT_CELL', { idx: column.idx, rowIdx }, openEditor);
+    eventBus.dispatch('CELL_SELECT', { idx: column.idx, rowIdx }, openEditor);
   }
 
-  function handleCellClick() {
```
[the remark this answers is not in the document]
amanmahajan7: It is the Cell component and `HandleCellClick` seemed unnecessary

src/DataGrid.tsx
amanmahajan7: I think it is fine. We need to add dragHandle only on the row that has the selected position. I will have to change the signature otherwise `getDragHandle(rowIdx)`. I am happy to revisit after code cleanup

nstepien: I agree with Qi. We can also do `return` instead of `return null` in `getDragHandle`.

src/DataGrid.tsx
```diff
@@ -441,29 +712,14 @@ function DataGrid<R, K extends keyof R, SR>({
...
-            />
-          )}
           <div style={{ height: Math.max(rows.length * rowHeight, clientHeight) }} />
-          {getViewportRows()}
+          <div
+            className={clsx({ 'rdg-viewport-dragging': isDragging })}
+            onKeyDown={handleKeyDown}
+          >
+            {viewportWidth > 0 && getViewportRows()}
+            {getEditorContainer()}
```
nstepien, line 721: Maybe we should move this outside the div, so we can simplify `handleKeyDown`?

src/Cell.tsx
```diff
@@ -1,38 +1,50 @@
...
   onDoubleClick,
   onContextMenu,
-  onDragOver,
   ...props
 }: CellRendererProps<R, SR>, ref: React.Ref<HTMLDivElement>) {
+  const cellRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (isSelected) {
+      cellRef.current?.focus();
```
nstepien, line 30: This causes issues when scrolling, as we re-focus the cell when it reappears in the viewport, and the browser will try scrolling to it.

src/Cell.tsx
```diff
@@ -45,24 +57,27 @@ function Cell<R, SR>({
     'rdg-cell',
     {
       'rdg-cell-frozen': column.frozen,
-      'rdg-cell-frozen-last': column.idx === lastFrozenColumnIndex
+      'rdg-cell-frozen-last': column.idx === lastFrozenColumnIndex,
+      'rdg-cell-selected': isSelected,
+      'rdg-cell-copied': isCopied,
+      'rdg-cell-dragged-over': isDraggedOver
```
nstepien, line 63: I wonder if we should set all these classes and event handlers at the row level, that way we can pass fewer props down to the Cell. Though that might require some `useCallback` 🤔

src/DataGrid.tsx
```diff
@@ -441,29 +712,14 @@ function DataGrid<R, K extends keyof R, SR>({
...
-              onRowsUpdate={handleRowsUpdate}
-              onSelectedCellChange={onSelectedCellChange}
-            />
-          )}
           <div style={{ height: Math.max(rows.length * rowHeight, clientHeight) }} />
-          {getViewportRows()}
+          <div
+            className={clsx({ 'rdg-viewport-dragging': isDragging })}
+            onKeyDown={handleKeyDown}
+          >
```
nstepien, line 719: If you want we can move the className and onKeyDown props to the top div, remove this div, then add `onKeyDown={e => e.stopPropagation()}` in header/filter/summary rows. WDYT?

src/Cell.tsx
amanmahajan7: Running the effect only on update now. This fixes the issue

src/DataGrid.tsx
```diff
@@ -300,6 +318,14 @@ function DataGrid<R, K extends keyof R, SR>({
     return eventBus.subscribe('SELECT_ROW', handleRowSelectionChange);
   }, [eventBus, onSelectedRowsChange, rows, rowKey, selectedRows]);
 
+  useEffect(() => {
+    return eventBus.subscribe('SELECT_CELL', selectCell);
+  });
+
+  useEffect(() => {
+    latestDraggedOverRowIdx.current = draggedOverRowIdx;
```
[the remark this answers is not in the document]
amanmahajan7: We are using `setDraggedOverRowIdx={isDragging ? setDraggedOverRowIdx : undefined}`. When the drag starts we rerender all the rows but on subsequent render we only render the row that fires `onMouseEnter`. If we add a setter then all the rows will be rerendered during dragging, We can use `useCallback` to prevent that. What do you prefer?

src/DataGrid.tsx
amanmahajan7: Moved `onKeyDown` to the cell component

src/Cell.tsx
amanmahajan7: Fixed by rendering an row that contains the selected cell

src/DataGrid.tsx
amanmahajan7: No longer needed

src/Cell.tsx
amanmahajan7: We probably can. Let's revisit in the next PR

src/DataGrid.tsx
```diff
@@ -371,6 +548,88 @@ function DataGrid<R, K extends keyof R, SR>({
...
+
+  function getEditorContainer() {
+    if (selectedPosition.mode === 'SELECT') return null;
+
+    const column = columns[selectedPosition.idx];
+    const row = rows[selectedPosition.rowIdx];
+    let editorLeft = 0;
+    let editorTop = 0;
+
+    if (gridRef.current !== null) {
```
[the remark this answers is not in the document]
amanmahajan7: No longer used

src/DataGrid.tsx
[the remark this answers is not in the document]
nstepien: I'd prefer `useCallback` over `useEffect` for this. I feel like `useEffect` could cause rare timing issues.

src/DataGrid.tsx
```diff
@@ -306,15 +337,39 @@ function DataGrid<R, K extends keyof R, SR>({
...
 
   /**
    * event handlers
    */
-  function onGridScroll(event: React.UIEvent<HTMLDivElement>) {
+  function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
+    if (enableCellCopyPaste && isCtrlKeyHeldDown(event)) {
+      // event.key may be uppercase `C` or `V`
+      const lowerCaseKey = event.key.toLowerCase();
+      if (lowerCaseKey === 'c') return handleCopy();
```
qili26, line 350: Not a big fan for this way of writing code. Intuitively I thought `handleCopy` would return something, but it's `void`, so my preference would just to split this into multiple lines.

```typescript
if (lowerCaseKey === 'c') {
  handleCopy();
  return;
}
```

src/DataGrid.tsx
```diff
@@ -368,43 +550,130 @@ function DataGrid<R, K extends keyof R, SR>({
...
+
+    for (let rowIdx = rowOverscanStartIdx; rowIdx <= rowOverscanEndIdx; rowIdx++) {
+      rowElements.push(getRow(rowIdx));
     }
 
     return rowElements;
   }
 
+  // Reset the positions if the current values are no longer valid. This can happen if a column or row is removed
+  if (selectedPosition.idx >= columns.length || selectedPosition.rowIdx >= rows.length) {
```
qili26, line 668: Is it valid for resetting state in during render? Shouldn't we move these lines in an effect?

src/DataGrid.tsx
```diff
@@ -368,43 +550,130 @@ function DataGrid<R, K extends keyof R, SR>({
...
+    const { rowIdx } = selectedPosition;
+
+    const isDraggedOver = rowIdx < draggedOverRowIdx
+      ? rowIdx < currentRowIdx && currentRowIdx <= draggedOverRowIdx
+      : rowIdx > currentRowIdx && currentRowIdx >= draggedOverRowIdx;
+
+    return isDraggedOver ? selectedPosition.idx : undefined;
+  }
+
+  function getSelectedCellProps(rowIdx: number): SelectedCellProps | undefined {
```
qili26, line 590: This function is called on each render, and then passed the return result into the `Row`. I am wondering if we shall add memorization to the output to prevent unnecessary Row re-render.

src/DataGrid.tsx
```diff
@@ -368,43 +550,130 @@ function DataGrid<R, K extends keyof R, SR>({
     }
   }
 
-  function getViewportRows() {
-    const rowElements = [];
+  function navigate(key: string, shiftKey: boolean) {
+    let nextPosition = getNextPosition(key, shiftKey, selectedPosition);
+    let mode = cellNavigationMode;
+    if (key === 'Tab') {
```
qili26, line 556: When `Tab` is being held, you will see the following that the tab focus element is not moving to the next possible cell but only focus on the select `input`. Not a big deal, but we might need to revisit later.

tab gap

src/common/types.ts
```diff
@@ -103,13 +95,38 @@ export interface HeaderRendererProps<TRow, TSummaryRow = unknown> {
   onAllRowsSelectionChange: (checked: boolean) => void;
 }
 
+export interface SharedEditorContainerProps {
+  editorPortalTarget: Element;
```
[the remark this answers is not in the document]
nstepien: `HTMLDivElement` extends/is a superset of `Element` so it should be fine.

src/DataGrid.tsx
nstepien: AFAIK it's valid if it's the same component's state. If you're updating a parent component's state on the other hand React will give you a warning.

src/Cell.tsx
```diff
@@ -1,77 +1,126 @@
...
   onClick,
   onDoubleClick,
   onContextMenu,
-  onDragOver,
   ...props
 }: CellRendererProps<R, SR>, ref: React.Ref<HTMLDivElement>) {
+  const cellRef = useRef<HTMLDivElement>(null);
+  const isSelected = selectedCellProps !== undefined;
+  const isEditing = selectedCellProps?.mode === 'EDIT';
+  const setFocus = isSelected && !isEditing;
```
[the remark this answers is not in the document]
amanmahajan7: Changed

src/common/types.ts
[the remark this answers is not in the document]
amanmahajan7:
```
export function createPortal(children: ReactNode, container: Element, key?: null | string): ReactPortal;
```
This is the definition

src/DataGrid.tsx
```diff
@@ -203,6 +214,22 @@ function DataGrid<R, K extends keyof R, SR>({
   const [scrollTop, setScrollTop] = useState(0);
   const [scrollLeft, setScrollLeft] = useState(0);
   const [columnWidths, setColumnWidths] = useState<ReadonlyMap<string, number>>(() => new Map());
+  const [selectedPosition, setSelectedPosition] = useState<SelectCellState | EditCellState>({ idx: -1, rowIdx: -1, mode: 'SELECT' });
+  const [copiedPosition, setCopiedPosition] = useState<Position & { value: unknown } | null>(null);
+  const [isDragging, setDragging] = useState(false);
```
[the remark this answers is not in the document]
amanmahajan7: If we use context then each cell needs to read the selected/copied/dragged position to set the styles. That means rerendering every cell when context value changes. I think it will be slower

src/DataGrid.tsx
amanmahajan7: That is correct. Check https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

amanmahajan7: Also check https://github.com/facebook/react/pull/17099

src/DataGrid.tsx
amanmahajan7: I copied the old code. I am okay with either style

src/DataGrid.tsx
amanmahajan7: Fixed

src/DataGrid.tsx
amanmahajan7: It is only passed to the row containing the selected cell so we only rerender 2 rows max. The performance is good and it is tricky to memoize this function. I think it is fine for now

src/Cell.tsx
```diff
@@ -1,77 +1,126 @@
...
+      <>
+        <column.formatter
+          column={column}
+          rowIdx={rowIdx}
+          row={row}
+          isRowSelected={isRowSelected}
+          onRowSelectionChange={onRowSelectionChange}
+        />
+        {selectedCellProps?.dragHandleProps && (
+          <div className="rdg-cell-drag-handle" {...selectedCellProps.dragHandleProps} />
```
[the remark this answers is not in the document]
amanmahajan7: It is difficult to style it from what I can tell, @nstepien is there a way to make it look like the old drag handle?

src/DataGrid.tsx
```diff
@@ -36,9 +42,20 @@ import {
...
-  SelectRowEvent
+  SelectRowEvent,
+  CommitEvent,
+  SelectedCellProps
 } from './common/types';
-import { CellNavigationMode, SortDirection } from './common/enums';
+import { CellNavigationMode, SortDirection, UpdateActions } from './common/enums';
+
+interface SelectCellState extends Position {
+  mode: 'SELECT';
```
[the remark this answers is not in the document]
amanmahajan7: I think it is fine as we only have 2 places where we define mode. It feels like we may be adding too many interfaces with this approach.

src/Cell.tsx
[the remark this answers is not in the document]
nstepien: We can't make it overflow without degrading performance, we could use a portal but then it wouldn't be in perfect sync when scrolling.

amanmahajan7:
I faced the focus issue again while profiling. I think React is not able to register the change in the focused element. I will try adding keydown on the root element

src/DataGrid.tsx
```diff
@@ -368,6 +576,105 @@ function DataGrid<R, K extends keyof R, SR>({
...
+      case 'ArrowLeft':
+        return { idx: idx - 1, rowIdx };
+      case 'ArrowRight':
+        return { idx: idx + 1, rowIdx };
+      case 'Tab':
+        if (selectedPosition.idx === -1 && selectedPosition.rowIdx === -1) {
+          return shiftKey ? { idx: columns.length - 1, rowIdx: rows.length - 1 } : { idx: 0, rowIdx: 0 };
+        }
+        return { idx: idx + (shiftKey ? -1 : 1), rowIdx };
+      case 'Home':
```
amanmahajan7, line 592: https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-for-data-grids

src/DataGrid.tsx
```diff
@@ -438,27 +755,17 @@ function DataGrid<R, K extends keyof R, SR>({
...
-          )}
-          <div style={{ height: Math.max(rows.length * rowHeight, clientHeight) }} />
+          <div
+            ref={rowsContainerRef}
+            tabIndex={0}
+            style={{
+              height: Math.max(rows.length * rowHeight, clientHeight),
+              outline: 0,
+              position: 'relative',
+              top: -totalHeaderHeight // Prevents scrolling when grid receives focus via keyboard
```
nstepien, line 765: Let's move these to `core.less`

src/DataGrid.tsx
```diff
@@ -268,6 +296,15 @@ function DataGrid<R, K extends keyof R, SR>({
     };
   }, [width]);
 
+  useLayoutEffect(() => {
+    if (selectedPosition === prevSelectedPosition.current || selectedPosition.mode === 'EDIT' || !isCellWithinBounds(selectedPosition)) return;
+    prevSelectedPosition.current = selectedPosition;
+    scrollToCell(selectedPosition);
+    if (document.activeElement !== rowsContainerRef.current) {
+      rowsContainerRef.current!.focus({ preventScroll: true });
```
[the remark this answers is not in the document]
qili26: Sticky css is not working well with Safari either. Not sure if we discussed that as well.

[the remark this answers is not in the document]
amanmahajan7: I implemented focus in a different way and `{ preventScroll }` is no longer needed as the focused element never needs to be scrolled. I think in the next PR we can push focusing logic to the next level

- Scroll the the last selected cell on focus
- Tabbing in an out of the grid can be further improved
- We can set `tabIndex=-1` on the checkbox selector and pass a flag `isCellActive = isSelected && isFocused` flag to formatters. Then the formatters can choose to focus the checkbox when the cell is active. This way keyboard selection would work. Same thing can be done for treeview example.

style/cell.less
```diff
@@ -23,8 +23,38 @@
...
+    border: 2px solid #66afe9;
+    background: #fff;
+  }
+}
+
+.rdg-cell-dragged-over {
+  background-color:#ccccff;
+  border-left: 1px dashed black;
+  border-right: 1px dashed black;
+  padding-left: 7px;
```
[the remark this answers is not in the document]
amanmahajan7: Each cell only has the `right` and the `bottom` border and we need to set `left` and `right` border when dragged over. This shifts the cell content so I had to change the left-padding to `8px- 1px (border-left-width)`

https://github.com/adazzle/react-data-grid/blob/6ffb06a3a59bb16f3e7201b3e8abf21afb5307be/style/cell.less#L5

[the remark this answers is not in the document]
amanmahajan7: The last dragged over cell should have a bottom border but we will fix it later

PR #2093, opened 2020-07-20
Add hooks to cleanup code

amanmahajan7:


src/hooks/useClickOutside.ts
```diff
@@ -70,13 +65,9 @@ export default function ClickOutside({ onClickOutside, children }: Props) {
     return () => {
       document.removeEventListener('click', handleDocumentClick);
     };
-  }, [onClickOutside]);
```
amanmahajan7, line 73: `commit` function always changes so this is not required

nstepien:
We could also inline `getVerticalRangeToRender` in that hook, even if we don't wrap it in `useMemo`.

amanmahajan7:
@nstepien let's inline the helpers in a follow up PR as we need to migrate the tests also

PR #2097, opened 2020-07-21
Fix for Copy&Paste events not working properly in non-English keyboard event

Yaron24:
Ctrl+V and Ctrl+C events are not detected when the keyboard language is not set on English. The condition for the copy/paste event is if event.code equals to 'c' or 'v'. This results in the non-English character showing the the selected cell instead of triggering a copy/paste event. For example 'C' in Hebrew is 'ב':

image

The suggested commit is using **event.keyCode** (cKey = 67, vKey = 86) instead of **event.code** to detect the event in non-English keyboard layouts.


nstepien (approved):

nstepien:
LGTM, thanks!

I'd rather not use `keyCode` as it's "obsolete" according to the mdn docs, but using `event.key` isn't good enough for Hebrew input, and `event.nativeEvent.code` isn't good enough for the English DVORAK layout either.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

amanmahajan7 (approved):
Thank you for the fix

PR #2102, opened 2020-07-24
Editor2

amanmahajan7:
## New API

```JSX
column.editor2?: React.ComponentType<Editor2Props<TRow, TSummaryRow>>;
column.editorOptions?: {
    /** Default: true for editor1 and false for editor2 */
    createPortal?: boolean;
    /** Default: false */
    editOnClick?: boolean;
    /** Prevent default to cancel editing */
    onCellKeyDown?: (event: React.KeyboardEvent<HTMLDivElement>) => void;
};

export interface Editor2Props<TRow, TSummaryRow = any> {
  row: Readonly<TRow>;
  rowIdx: number;
  column: Readonly<CalculatedColumn<TRow, TSummaryRow>>;
  rowHeight: number;
  onRowChange: (row: Readonly<TRow>, commitChanges?: boolean) => void;
  onClose: (commitChanges?: boolean) => void;
  top: number;
  left: number;
  editorPortalTarget: Element;
}
```

## How to use

```
1) Use `column.editor2`

   {
      key: 'country',
      name: 'Country',
      width: 180,
      editor2: p => (
        <SelectEditor
          value={p.row.country}
          onChange={value => p.onRowChange({ ...p.row, country: value })}
          options={countries.map(c => ({ value: c, label: c }))}
          rowHeight={p.rowHeight}
          menuPortalTarget={p.editorPortalTarget}
        />
      ),
      resizable: true,
      sortable: true
    },
    {
      key: 'contact',
      name: 'Contact',
      width: 160,
      editor2: p => (
        <TextEditor
          value={p.row.contact}
          onChange={value => p.onRowChange({ ...p.row, contact: value })}
          rowHeight={p.rowHeight}
        />
      ),
      resizable: true,
      sortable: true
    }

2) Handle `onRowsChange`

onRowsChange={updatedRows => setRows(updatedRows )}
```

`DataGrid` now keeps the updated row in a local state so it is easy to handle commit, cancel and validations (future).


src/DataGrid.tsx
```diff
@@ -426,10 +446,29 @@ function DataGrid<R, K extends keyof R, SR>({
...
-    if (canOpenEditor && (key === 'Enter' || isActivatedByUser)) {
+    const isActivatedByUser = isUsingEditor2
+      ? defaultCellInput(event) && column.editor2Props?.onCellInput?.(event) !== false
+      : defaultCellInput(event);
+
+    if (isCellEditable(selectedPosition) && isActivatedByUser) {
+      if (isUsingEditor2) {
+        if (key === 'Delete' || key === 'Backspace') {
+          setEditorValue('');
+        } else if (key === 'Enter' || key === 'F2') {
```
amanmahajan7, line 466: https://www.w3.org/TR/wai-aria-practices/#gridNav_inside

src/DataGrid.tsx
```diff
@@ -504,8 +548,10 @@ function DataGrid<R, K extends keyof R, SR>({
 
   function selectCell(position: Position, enableEditor = false): void {
     if (!isCellWithinBounds(position)) return;
+    handleCommit2();
```
amanmahajan7, line 551: I think it is better to have some control over when the value is committed. This way we can add validation in the future. Using an effect to commit the value is probably not the best way 
https://github.com/adazzle/react-data-grid/blob/c/src/editors/EditorContainer.tsx#L70

src/DataGrid.tsx
```diff
@@ -426,10 +446,29 @@ function DataGrid<R, K extends keyof R, SR>({
...
+    if (selectedPosition.mode === 'EDIT') {
+      if (isUsingEditor2 && (key === 'Enter' || key === 'F2')) {
+        handleCommit2();
+      }
+      return;
+    }
 
-    if (canOpenEditor && (key === 'Enter' || isActivatedByUser)) {
+    const isActivatedByUser = isUsingEditor2
+      ? defaultCellInput(event) && column.editor2Props?.onCellInput?.(event) !== false
```
amanmahajan7, line 459: I think we should open the editor on `Enter`, `F2`, or on any other alpha numeric key. If the users need to restrict it then they can use `onCellInput` . This is also the recommended wau
https://www.w3.org/TR/wai-aria-practices/#gridNav_inside

src/types.ts
```diff
@@ -33,11 +33,15 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
-  // TODO: Should we return an object to make it more powerful?
-  // { isEditable?: boolean, initialInput?: unknown } | undefined
-  unsafe_onCellInput?: (event: React.KeyboardEvent<HTMLDivElement>, row: TRow) => unknown;
   /** Editor to be rendered when cell of column is being edited. If set, then the column is automatically set to be editable */
   editor?: React.ComponentType<EditorProps<TRow[keyof TRow], TRow, TSummaryRow>>;
+  editor2?: React.ComponentType<Editor2Props<any, TRow, TSummaryRow>>;
+  editor2Props?: {
+    createPortal?: boolean;
+    editOnSingleClick?: boolean;
+    commitOnOutsideClick?: boolean;
```
amanmahajan7, line 42: This needs to be implemented.

stories/demos/TreeView.tsx
```diff
@@ -130,19 +130,21 @@ export default function TreeView() {
...
-            event.preventDefault();
-            dispatch({ id: row.id, type: 'toggleSubRow' });
-          }
-
-          if (event.key === 'Delete' && !hasChildren) {
-            event.preventDefault();
-            dispatch({ id: row.id, type: 'deleteSubRow' });
-          }
         }
+        // editor2Props: {
```
amanmahajan7, line 134: I think we can handle it differently using https://www.w3.org/TR/wai-aria-practices/#gridNav_focus

stories/demos/components/Editors/TextEditor.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useRef, useLayoutEffect } from 'react';
+import { Editor2Props } from '../../../../src/types';
+
+export function TextEditor<R>({ value, onChange, rowHeight }: Editor2Props<string, R>) {
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useLayoutEffect(() => {
```
amanmahajan7, line 7: We no longer have the bug where the input value is initially selected and typing another key deletes it

stories/demos/components/Editors/SelectEditor.tsx
```diff
@@ -0,0 +1,36 @@
...
+        control: (provided) => ({
+          ...provided,
+          height: rowHeight - 1,
+          minHeight: 30,
+          padding: 0,
+          lineHeight: 'normal'
+        })
+      }}
+      onKeyDown={event => {
+        if (['ArrowDown', 'ArrowUp', 'Enter'].includes(event.key) && selectRef.current!.getProp('menuIsOpen')) {
```
amanmahajan7, line 26: I think this is the best way to cancel keyboard navigation IMO

src/utils/keyboardUtils.ts
```diff
@@ -10,3 +10,7 @@ export function isKeyPrintable(keycode: number): boolean {
 export function isCtrlKeyHeldDown(e: React.KeyboardEvent): boolean {
   return (e.ctrlKey === true || e.metaKey === true) && e.key !== 'Control';
 }
+
+export function defaultCellInput(event: React.KeyboardEvent<HTMLDivElement>) {
```
qili26, line 14: `isDefaultCellInput`?

src/DataGrid.tsx
```diff
@@ -423,13 +441,34 @@ function DataGrid<R, K extends keyof R, SR>({
...
     const { key } = event;
     const column = columns[selectedPosition.idx];
     const row = rows[selectedPosition.rowIdx];
-    const canOpenEditor = selectedPosition.mode === 'SELECT' && isCellEditable(selectedPosition);
-    const isActivatedByUser = (column.unsafe_onCellInput ?? legacyCellInput)(event, row) === true;
+    const isUsingEditor2 = column.editor2 !== undefined;
 
-    if (canOpenEditor && (key === 'Enter' || isActivatedByUser)) {
+    if (selectedPosition.mode === 'EDIT') {
+      if (isUsingEditor2 && (key === 'Enter' || key === 'F2')) {
```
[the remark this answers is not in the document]
amanmahajan7: I think it is not possible to have one solution for every type of editor and we need some way to to tweak the default behavior. If the editor do not need to commit changes on `Enter` then it should handle the `keydown` event and stop propagation. The question is what default behavior we want to use? Right now we do the following 
- Open editor on `Enter`, `F2` and alphanumeric keys
- Cancel changes on `Escape`
- Commit changes on `Enter` and `F2`
- Commit changes when another cell is clicked
- Commit changes when header or summary row is clicked

Any suggestions?

src/Cell.tsx
```diff
@@ -65,16 +68,35 @@ function Cell<R, SR>({
...
       const gridTop = top + docTop;
 
+      if (column.editor2 !== undefined) {
+        const editor2 = (
+          <column.editor2
+            {...editor2Props}
+            rowIdx={rowIdx}
+            row={row}
+            column={column}
+            left={gridLeft}
```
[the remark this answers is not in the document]
amanmahajan7: Editor can choose not to use `left, right` props. The only way to override them is in the editor itself. We control `editor2Props` and not the user

src/editors/EditorContainer.tsx
```diff
@@ -133,7 +133,7 @@ export default function EditorContainer<R, SR>({
   function onKeyDown(e: KeyboardEvent) {
     if (preventDefaultNavigation(e.key)) {
       e.stopPropagation();
-    } else if (['Enter', 'Tab', 'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'].includes(e.key)) {
+    } else if (['Enter', 'F2', 'Tab', 'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'].includes(e.key)) {
```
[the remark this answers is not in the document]
amanmahajan7: We can remove it. I was following accessibility guidelines

src/SummaryRow.tsx
```diff
@@ -29,6 +32,7 @@ function SummaryRow<R, SR>({
       aria-rowindex={ariaRowIndex}
       className={`rdg-row rdg-row-${rowIdx % 2 === 0 ? 'even' : 'odd'} rdg-summary-row`}
       style={{ bottom }}
+      onClick={onClick}
```
amanmahajan7, line 35: I accidentally deleted your comment

> Why this change? I don't think we expose SummaryRow?
So we can commit the value when summary row is clicked. We need to decide on when to commit the values

src/DataGrid.tsx
```diff
@@ -223,6 +223,7 @@ function DataGrid<R, K extends keyof R, SR>({
   const [copiedPosition, setCopiedPosition] = useState<Position & { value: unknown } | null>(null);
   const [isDragging, setDragging] = useState(false);
   const [draggedOverRowIdx, setOverRowIdx] = useState<number | undefined>(undefined);
+  const [editorValue, setEditorValue] = useState<unknown>();
```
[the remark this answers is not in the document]
nstepien: I thought maybe the editor container could have it in its scope

src/DataGrid.tsx
```diff
@@ -48,8 +48,9 @@ interface SelectCellState extends Position {
   mode: 'SELECT';
 }
 
-interface EditCellState extends Position {
+interface EditCellState<R> extends Position {
   mode: 'EDIT';
+  updatedRow: R;
```
nstepien, line 53:
```suggestion
  row: R;
```
WDYT?

src/DataGrid.tsx
```diff
@@ -729,7 +785,7 @@ function DataGrid<R, K extends keyof R, SR>({
         onSelectedRowsChange={onSelectedRowsChange}
         sortColumn={sortColumn}
         sortDirection={sortDirection}
-        onSort={onSort}
+        onSort={handleSort}
```
nstepien, line 788: Since `onSort` is not optional anymore for `HeaderRow` and its children, we should update the types.

src/types.ts
```diff
@@ -36,11 +36,16 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
-  // { isEditable?: boolean, initialInput?: unknown } | undefined
-  unsafe_onCellInput?: (event: React.KeyboardEvent<HTMLDivElement>, row: TRow) => unknown;
   /** Editor to be rendered when cell of column is being edited. If set, then the column is automatically set to be editable */
   editor?: React.ComponentType<EditorProps<TRow[keyof TRow], TRow, TSummaryRow>>;
+  editor2?: React.ComponentType<Editor2Props<TRow, TSummaryRow>>;
+  editor2Options?: {
+    createPortal?: boolean;
+    singleClickEdit?: boolean;
+    commitOnOutsideClick?: boolean;
+    getInitialValue?: (event: React.KeyboardEvent<HTMLDivElement>, row: TRow) => unknown;
```
nstepien, line 46:
```suggestion
```

src/types.ts
```diff
@@ -36,11 +36,16 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
-  // TODO: Should we return an object to make it more powerful?
-  // { isEditable?: boolean, initialInput?: unknown } | undefined
-  unsafe_onCellInput?: (event: React.KeyboardEvent<HTMLDivElement>, row: TRow) => unknown;
   /** Editor to be rendered when cell of column is being edited. If set, then the column is automatically set to be editable */
   editor?: React.ComponentType<EditorProps<TRow[keyof TRow], TRow, TSummaryRow>>;
+  editor2?: React.ComponentType<Editor2Props<TRow, TSummaryRow>>;
+  editor2Options?: {
+    createPortal?: boolean;
+    singleClickEdit?: boolean;
+    commitOnOutsideClick?: boolean;
```
nstepien, line 45: Let's remove or comment this out if it won't be implemented in this PR.

src/DataGrid.tsx
```diff
@@ -89,6 +90,7 @@ export interface DataGridProps<R, K extends keyof R, SR = unknown> extends Share
    * 4. Update all cells under a given cell by double clicking the cell's fill handle.
    */
   onRowsUpdate?: <E extends RowsUpdateEvent>(event: E) => void;
+  onRowsUpdate2?: (rows: readonly R[]) => void;
```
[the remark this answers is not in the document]
amanmahajan7: Makes sense

src/types.ts
amanmahajan7: It was a quick implementation so I added it

src/DataGrid.tsx
amanmahajan7: Changed

src/DataGrid.tsx
```diff
@@ -718,6 +777,11 @@ function DataGrid<R, K extends keyof R, SR>({
       } as unknown as React.CSSProperties}
       ref={gridRef}
       onScroll={handleScroll}
+      onClickCapture={() => {
+        if (selectedPosition.mode === 'SELECT') return;
+        const column = columns[selectedPosition.idx];
+        return column.editor2 !== undefined && column.editor2Options?.commitOnOutsideClick ? onClickCapture : undefined;
```
nstepien, line 783: Shouldn't `onClickCapture` be called? Maybe we should do all this at the cell level?

src/DataGrid.tsx
```diff
@@ -490,6 +529,11 @@ function DataGrid<R, K extends keyof R, SR>({
     });
   }
 
+  function handleSort(columnKey: string, direction: SortDirection) {
+    handleCommit2();
```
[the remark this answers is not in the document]
amanmahajan7: `handleCommit2` handles it now

src/DataGrid.tsx
amanmahajan7: I don' t know how it was working during my testing

amanmahajan7: So we commit value on 
- Cell click
- Header click
- Footer click

The only thing left is outside grid click. Not sure how we can handle it at the cell level. Do you mean to always commit the value when we click outside the cell and remove the above logic? The reason I added individual click handlers is to have more control over when the value is committed

amanmahajan7: I have commented `commitOnOutsideClick` for now. Wdyt about adding another api method like `commitChanges`. Users can call `gridRef.current?.commitChanges();` however they like. I think `useClickOutside` is great but it is not flexible

src/HeaderRow.tsx
```diff
@@ -10,9 +10,8 @@ type SharedDataGridProps<R, K extends keyof R, SR> = Pick<DataGridProps<R, K, SR
   | 'onSelectedRowsChange'
   | 'sortColumn'
   | 'sortDirection'
-  | 'onSort'
   | 'rowKey'
->;
+> & Required<Pick<DataGridProps<R, K, SR>, 'onSort'>>;
```
nstepien, line 14: In `SortableHeaderCell` we can now remove the `if (!onSort) return;` line.

src/DataGrid.tsx
nstepien: I think catching all clicks outside the cell to commit would be great, would clicking on the datagrid border do nothing for example? 🤔
Having a method to commit at any time is also good, we could have both.

>The reason I added individual click handlers is to have more control over when the value is committed

Aren't they all "clicking outside the cell" anyway? Do we handle those in special ways?

src/types.ts
```diff
@@ -36,11 +36,15 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
-  // TODO: Should we return an object to make it more powerful?
-  // { isEditable?: boolean, initialInput?: unknown } | undefined
-  unsafe_onCellInput?: (event: React.KeyboardEvent<HTMLDivElement>, row: TRow) => unknown;
   /** Editor to be rendered when cell of column is being edited. If set, then the column is automatically set to be editable */
   editor?: React.ComponentType<EditorProps<TRow[keyof TRow], TRow, TSummaryRow>>;
+  editor2?: React.ComponentType<Editor2Props<TRow, TSummaryRow>>;
+  editor2Options?: {
+    createPortal?: boolean;
+    singleClickEdit?: boolean;
+    // commitOnOutsideClick?: boolean; // Do we need this?
```
nstepien, line 45: So we won't forget about it
```suggestion
    // commitOnOutsideClick?: boolean; // TODO: Do we need this?
```

src/DataGrid.tsx
```diff
@@ -156,7 +158,7 @@ export interface DataGridProps<R, K extends keyof R, SR = unknown> extends Share
    * Miscellaneous
    */
   /** The node where the editor portal should mount. */
-  editorPortalTarget?: Element;
+  editorPortalTarget?: HTMLElement;
```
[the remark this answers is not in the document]
amanmahajan7: Reverted. I changed it for the Select editor

src/Cell.tsx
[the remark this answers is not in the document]
qili26: If the cell is at the right bottom corner, it might end up with weird issue if the editor is customized width and height. That's why I think shall somehow make this left / top controllable by user.

src/DataGrid.tsx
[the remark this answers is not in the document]
qili26: I think for the `Enter` case, we shall leave a prop for the user to decide if we want to use `Enter` to close the the editor.
It is a common use case that people might want to put a save button with 2 inputs for example, `Enter` would mean something else for them. As for F2, maybe we should keep like excel to only open the cell rather than closing it. `Esc` is pretty straightforward, just close without saving / commit.

src/DataGrid.tsx
amanmahajan7: The issue is the order of the event handlers. Click event on the cell is fired before the outside click. So if we need to add validation and prevent commit for example then it may not be possible with one outside click handler. May be we can have a click handler on the root component to handle commits for header, footer etc 🤔

stories/demos/components/Editors/SelectEditor.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export function SelectEditor<R>({ value, onChange, options, rowHeight, menuPortalTarget }: SelectEditorProps) {
+  const selectRef = useRef<Select<OptionTypeBase>>(null);
+  return (
+    <Select
+      autoFocus
+      defaultMenuIsOpen
+      ref={selectRef}
+      value={options.find(o => o.value === value)}
+      onChange={o => onChange(o.value)}
```
qili26, line 20: When the value is selected, this select editor is not closed.

I am using my mac, haven't tested with Windows yet.

src/types.ts
```diff
@@ -36,11 +36,15 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
   sortDescendingFirst?: boolean;
-  // TODO: Should we return an object to make it more powerful?
-  // { isEditable?: boolean, initialInput?: unknown } | undefined
-  unsafe_onCellInput?: (event: React.KeyboardEvent<HTMLDivElement>, row: TRow) => unknown;
   /** Editor to be rendered when cell of column is being edited. If set, then the column is automatically set to be editable */
   editor?: React.ComponentType<EditorProps<TRow[keyof TRow], TRow, TSummaryRow>>;
+  editor2?: React.ComponentType<Editor2Props<TRow, TSummaryRow>>;
+  editor2Options?: {
+    createPortal?: boolean;
+    singleClickEdit?: boolean;
```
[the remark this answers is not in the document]
amanmahajan7: This prop is only for mouse click
https://github.com/adazzle/react-data-grid/blob/49e530bee69bcf18c96f8f1dae5ede1e8b3a84b9/src/Cell.tsx#L53

We can think about adding a prop to allow auto edit on cell navigation

stories/demos/components/Editors/SelectEditor.tsx
amanmahajan7: That is the default behavior but we can change the editor to call `onCommit` when the value is changed I believe. Or may be we can change `onChange(newValue, commit flag)`. Wdyt?

qili26: Maybe just keep the `dropdown` open after `onChange`?

src/DataGrid.tsx
qili26: For the validation, do you want to bind inside RDG? I feel for RDG itself, it doesn't need to handle the validation.

So, the user should control the validation state out of the RDG. And whenever any error style / message needs to be display, formatter can directly access outside validation data and render.

By doing so, they can just run validation check after committing and they can decide to eventually dispatch the value or not in `onCommit`.

amanmahajan7: We need some way to prevent default actions like committing value when a different cell is clicked. Editor needs to inform the grid somehow that editor value is invalid so do not commit. Not sure what is the best way to handle it but we need give users some control over grid's internal behavior

src/types.ts
```diff
@@ -36,11 +36,20 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
+  editorOptions?: {
+    /** Default: true for editor1 and false for editor2 */
+    createPortal?: boolean;
+    /** Default: false */
+    editOnClick?: boolean;
+    // TODO: Do we need these options
+    // editOnDoubleClick?: boolean;
+    // commitOnScroll?: boolean;
+    // commitOnOutsideClick?: boolean;
+    onCellInput?: (event: React.KeyboardEvent<HTMLDivElement>) => void; // Prevent default to cancel editing
```
amanmahajan7, line 51: Should we rename it to `onKeyDown` or `onCellKeyDown`?

nstepien: It's only used when the editor is closed right? Then `onCellKeyDown` makes more sense.

src/HeaderRow.tsx
amanmahajan7: Done

src/DataGrid.tsx
```diff
@@ -492,6 +532,28 @@ function DataGrid<R, K extends keyof R, SR>({
...
+  function handleOnClose(commitChanges?: boolean) {
+    if (selectedPosition.mode === 'SELECT') return;
+    if (commitChanges) {
+      handleRowChange(selectedPosition.row);
+    }
+    closeEditor();
+  }
+
+  function handleSort(columnKey: string, direction: SortDirection) {
+    handleCommit2();
```
amanmahajan7, line 553: This is still needed as the cell is unmounted before `useOutsideClick` runs

src/Row.tsx
```diff
@@ -49,21 +50,42 @@ function Row<R, SR = unknown>({
...
+            rowIdx={rowIdx}
+            column={column}
+            lastFrozenColumnIndex={lastFrozenColumnIndex}
+            row={row}
+            isCopied={copiedCellIdx === column.idx}
+            isDraggedOver={draggedOverCellIdx === column.idx}
+            isCellSelected={isCellSelected}
+            isRowSelected={isRowSelected}
+            eventBus={eventBus}
+            dragHandleProps={isCellSelected ? (selectedCellProps as SelectedCellProps)?.dragHandleProps : undefined}
```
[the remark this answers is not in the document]
amanmahajan7: Yeah, TS does not recognize it 

image

src/Row.tsx
```diff
@@ -49,21 +50,42 @@ function Row<R, SR = unknown>({
...
-          isDraggedOver={draggedOverCellIdx === column.idx}
-          isRowSelected={isRowSelected}
-          eventBus={eventBus}
-          selectedCellProps={selectedCellProps?.idx === column.idx ? selectedCellProps : undefined}
-          onRowClick={onRowClick}
-        />
-      ))}
+      {viewportColumns.map(column => {
+        const isCellSelected = selectedCellProps?.idx === column.idx;
+        if (selectedCellProps?.mode === 'EDIT' && isCellSelected) {
```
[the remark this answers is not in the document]
amanmahajan7: image

src/Row.tsx
```diff
@@ -49,21 +50,42 @@ function Row<R, SR = unknown>({
...
+            column={column}
+            lastFrozenColumnIndex={lastFrozenColumnIndex}
+            row={row}
+            isCopied={copiedCellIdx === column.idx}
+            isDraggedOver={draggedOverCellIdx === column.idx}
+            isCellSelected={isCellSelected}
+            isRowSelected={isRowSelected}
+            eventBus={eventBus}
+            dragHandleProps={isCellSelected ? (selectedCellProps as SelectedCellProps)?.dragHandleProps : undefined}
+            onKeyDown={isCellSelected ? selectedCellProps?.onKeyDown : undefined}
```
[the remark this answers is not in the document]
amanmahajan7: Same reason.

src/Row.tsx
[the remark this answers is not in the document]
nstepien: if `isCellSelected` is true, then we can assert that `selectedCellProps` isn't nullish, so we can add `!`.

src/utils/keyboardUtils.ts
amanmahajan7: Changed

PR #2104, opened 2020-07-28
Handle focus on focusable formatters

amanmahajan7:
https://www.w3.org/TR/wai-aria-practices/#gridNav_focus

> For assistive technology users, the quality of experience when navigating a grid heavily depends on both what a cell contains and on where keyboard focus is set. For example, if a cell contains a button and a grid navigation key places focus on the cell instead of the button, screen readers announce the button label but do not tell users a button is present.

>There are two optimal cell design and focus behavior combinations:
> - A cell contains one widget whose operation does not require arrow keys and grid navigation keys set focus on that widget. Examples of such widgets include link, button, menubutton, toggle button, radio button (not radio group), switch, and checkbox.
> - A cell contains text or a single graphic and grid navigation keys set focus on the cell.

https://www.loom.com/share/3a0e559f0ea04db98226c3169dffad04

### DIFF
--- a/src/formatters/SelectCellFormatter.tsx
+++ b/src/formatters/SelectCellFormatter.tsx
@@ -1,24 +1,22 @@
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import clsx from 'clsx';
 
-type SharedInputProps = Pick<React.InputHTMLAttributes<HTMLInputElement>,
-  | 'disabled'
-  | 'aria-label'
-  | 'aria-labelledby'
->;
-
-export interface SelectCellFormatterProps extends SharedInputProps {
+export interface SelectCellFormatterProps {
   value: boolean;
+  tabIndex?: number;
+  isCellSelected?: boolean;
+  disabled?: boolean;
   onChange: (value: boolean, isShiftClick: boolean) => void;
 }
 
-export function SelectCellFormatter({
-  value,
-  disabled,
-  onChange,
-  'aria-label': ariaLabel,
-  'aria-labelledby': ariaLabelledBy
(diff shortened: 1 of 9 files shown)


src/formatters/SelectCellFormatter.tsx
```diff
@@ -1,33 +1,31 @@
...
-  value,
-  disabled,
-  onChange,
-  'aria-label': ariaLabel,
-  'aria-labelledby': ariaLabelledBy
-}: SelectCellFormatterProps) {
+export function SelectCellFormatter({ value, tabIndex, isCellSelected, disabled, onChange }: SelectCellFormatterProps) {
+  const inputRef = useRef<HTMLInputElement>(null);
+  useEffect(() => {
+    if (isCellSelected) {
```
amanmahajan7, line 15: User can handle focus however they like and for complex components `event` can be stopped from propagating.

stories/demos/components/Formatters/ChildRowDeleteButton.tsx
```diff
@@ -1,17 +1,37 @@
...
+    }
+  }
+
   return (
     <>
       <div className="rdg-child-row-action-cross" />
       {isDeleteSubRowEnabled && (
         <div className="rdg-child-row-btn" onClick={onDeleteSubRow}>
-          ❌
+          <span
```
nstepien, line 28: Why add another element?
Also I think if you just set `onClick={onDeleteSubRow}` instead of handling `onKeyDown` manually it might work just as well.

amanmahajan7: Otherwise it looks like this

image

PR #2106, opened 2020-07-30
Grouping

amanmahajan7:
- [x] Finalize grouping API
  -  Added `groupBy`, `rowGrouper`, and `column.groupFormatter` props
- [x] `rowIdx` is no longer in sync with `rows` as the grid groups the data internally. How should it be handled?
  - Using `rawRows.indexOf(rows[rowIdx] as R)` to get the correct index
- [x] Should dragging be allowed
  - Disabled dragging and copy/paste when grouping is enabled
- [x] Should we freeze the groupBy column?
  - `groupBy` columns are set as frozen columns and moved to the left (after the select column)
- [x] Add `aria` attributes for `treegrid` and fix `arria-rowindex/aria-rowcount`
- [x] Should both row and cell be selected on the group row?
  - Added row selection logic similar to the treegrid [example ](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treegrid/treegrid-1.html)


src/Cell.tsx
```diff
@@ -78,16 +78,20 @@ function Cell<R, SR>({
...
-        column={column}
-        rowIdx={rowIdx}
-        row={row}
-        isCellSelected={isCellSelected}
-        isRowSelected={isRowSelected}
-        onRowSelectionChange={onRowSelectionChange}
-      />
-      {dragHandleProps && (
-        <div className="rdg-cell-drag-handle" {...dragHandleProps} />
+      {!column.rowGroup && (
```
amanmahajan7, line 81: Rendering an empty cell for `groupBy` columns

src/DataGrid.tsx
```diff
@@ -232,6 +238,9 @@ function DataGrid<R, K extends keyof R, SR>({
   const [isDragging, setDragging] = useState(false);
   const [draggedOverRowIdx, setOverRowIdx] = useState<number | undefined>(undefined);
 
+  // TODO: change it to props
```
amanmahajan7, line 241: Not sure how to change it to props as the the keys are generated internally. The format is `grandParentKey__parentKey__key__so on`

src/DataGrid.tsx
```diff
@@ -269,12 +280,23 @@ function DataGrid<R, K extends keyof R, SR>({
...
     scrollTop,
-    rows.length
-  );
+    expandedGroupIds
+  });
+
+  const hasGroups = groupBy.length > 0 && rowGrouper;
+
+  if (hasGroups) {
+    // TODO: finalize if these flags need to be supported on treegrid
```
amanmahajan7, line 296: Thoughts?

src/DataGrid.tsx
```diff
@@ -722,11 +785,41 @@ function DataGrid<R, K extends keyof R, SR>({
   }
 
   function getViewportRows() {
-    const rowElements = [];
+    // TODO: cleanup rowIndex/rowIdx logic
```
amanmahajan7, line 788: Need to revisit to make sure the algorithm is correct. Check `useViewportRows`

src/DataGrid.tsx
```diff
@@ -722,11 +785,41 @@ function DataGrid<R, K extends keyof R, SR>({
...
+    let startRowIndex = 0;
+    return viewportRows.map((row, index) => {
+      const rowIdx = startRowIdx + index;
+      const top = rowIdx * rowHeight + totalHeaderHeight;
+      if (isGroupedRow(row)) {
+        const isSelected = selectedPosition.rowIdx === rowIdx;
+        ({ startRowIndex } = row);
+        return (
+          <GroupRowRenderer<R, SR>
+            aria-level={row.level + 1} // aria-level is 1-based
```
amanmahajan7, line 798: https://www.w3.org/TR/wai-aria-practices-1.1/examples/treegrid/treegrid-1.html

src/GroupRow.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onKeyDown,
+  'aria-setsize': ariaSetSize,
+  'aria-posinset': ariaPosInSet,
+  'aria-rowindex': ariaRowIndex,
+  ...props
+}: GroupRowRendererProps<R, SR>) {
+  // Select is always the first column
+  const idx = viewportColumns[0].key === SELECT_COLUMN_KEY ? level + 1 : level;
+
+  function selectGroup() {
```
amanmahajan7, line 29: https://www.w3.org/TR/wai-aria-practices-1.1/examples/treegrid/treegrid-1.html

Check the `Example Usage Options` section. Should we focus on both rows and cells?

stories/demos/CommonFeatures.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useState, useCallback, useMemo } from 'react';
 import faker from 'faker';
 import { AutoSizer } from 'react-virtualized';
+import { groupBy as rowGrouper } from 'lodash';
```
amanmahajan7, line 4: I will create a separate story after the review

src/DataGrid.tsx
amanmahajan7: Changed it to props for now.

src/hooks/useViewportRows.ts
```diff
@@ -0,0 +1,93 @@
...
+    return groupRows(rawRows, groupBy, 0);
+  }, [groupBy, rowGrouper, rawRows]);
+
+  const rows = useMemo(() => {
+    if (!groupedRows) return rawRows;
+
+    const expandGroup = (rows: GroupByDictionary<R> | R[], parentKey: string | undefined, level: number): Array<GroupRow<R> | R> => {
+      if (Array.isArray(rows)) return rows;
+      const flattenedRows: Array<R | GroupRow<R>> = [];
+      Object.keys(rows).forEach((key, posInSet, keys) => {
```
qili26, line 52: Use native `for` loop?

stories/demos/Grouping.tsx
```diff
@@ -0,0 +1,170 @@
...
+  { value: 'year', label: 'Year' },
+  { value: 'sport', label: 'Sport' },
+  { value: 'athlete', label: 'athlete' }
+];
+
+export default function Grouping() {
+  const [rows] = useState(createRows);
+  const [selectedRows, setSelectedRows] = useState(() => new Set<number>());
+  const [selectedOptions, setSelectedOptions] = useState<ValueType<Option>>([options[0], options[1]]);
+  const [expandedGroupIds, setExpandedGroupIds] = useState<Set<unknown>>(new Set(['United States of America', 'United States of America__2015']));
```
qili26, line 120: `() => new Set()`

src/DataGrid.tsx
```diff
@@ -769,13 +909,13 @@ function DataGrid<R, K extends keyof R, SR>({
 
   return (
     <div
-      role="grid"
+      role={hasGroups ? 'treegrid' : 'grid'}
```
qili26, line 912: Perhaps we need to expose this `role` props outside so that people that are using other languages can set the value as well in different language.

src/hooks/useViewportColumns.ts
```diff
@@ -18,24 +24,66 @@ export function useViewportColumns<R, K extends keyof R, SR>({
...
+            if (isGroupRow(row)) {
+              return row.level === index;
+            }
+            return false;
+          }
+        }
+      }))
+      .sort((c1, c2) => groupBy.findIndex(k => k === c1.key) - groupBy.findIndex(k => k === c2.key));
+    const remaningColumns = rawColumns.filter(c => !groupBy.includes(c.key) && c.key !== SELECT_COLUMN_KEY);
+    const sortedColumns = [
```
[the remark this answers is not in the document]
amanmahajan7: They would be added after the groupBy columns.

https://github.com/adazzle/react-data-grid/blob/0b7fd8ce30156aca85625ed1a4d05c3d687c161e/src/hooks/useViewportColumns.ts#L59

src/hooks/useViewportRows.ts
amanmahajan7: We need index for aria-attributes. I think it is okay to use `forEach` here

src/hooks/useViewportColumns.ts
```diff
@@ -18,24 +24,66 @@ export function useViewportColumns<R, K extends keyof R, SR>({
...
-  defaultColumnOptions
+  defaultColumnOptions,
+  groupBy: rawGroupBy,
+  rowGrouper
 }: ViewportColumnsArgs<R, K, SR>) {
+  const [sortedColumns, groupBy] = useMemo(() => {
+    if (!rawGroupBy || !rowGrouper) return [rawColumns, []];
+
+    // Find valid groupBy columns
+    const groupBy: readonly string[] = rawGroupBy.filter(g => rawColumns.find(c => c.key === g) !== undefined);
```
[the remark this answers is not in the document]
amanmahajan7: I think it should be fine as the `rawGroupBy` should be small and js engines are really fast. The example uses 10000 rows

src/hooks/useViewportColumns.ts
```diff
@@ -18,24 +24,66 @@ export function useViewportColumns<R, K extends keyof R, SR>({
...
+        formatterOptions: {
+          focusable(row: R | GroupRow<R>) {
+            if (isGroupRow(row)) {
+              return row.level === index;
+            }
+            return false;
+          }
+        }
+      }))
+      .sort((c1, c2) => groupBy.findIndex(k => k === c1.key) - groupBy.findIndex(k => k === c2.key));
```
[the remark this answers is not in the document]
amanmahajan7: same here

src/DataGrid.tsx
amanmahajan7: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques

I don't think aria-roles need to be translated.

src/formatters/ToggleGroupFormatter.tsx
```diff
@@ -0,0 +1,37 @@
...
+      style={{ cursor: 'pointer' }}
+      onClick={toggleGroup}
+    >
+      {groupKey}{' '}
+      <span
+        ref={cellRef}
+        tabIndex={-1}
+        onKeyDown={handleKeyDown}
+      >
+        {isExpanded ? '\u25BC' : '\u25B6'}
```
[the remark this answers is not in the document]
amanmahajan7: We need some way to customize the ToggleGroupFormatter. I will try adding some props

src/types.ts
```diff
@@ -216,3 +245,26 @@ export interface SelectRowEvent {
...
+}
+
+export type GroupByDictionary<TRow> = Dictionary<{
+  childRows: TRow[];
+  childGroups: TRow[] | GroupByDictionary<TRow>;
+  startRowIndex: number;
+}>;
+
+export interface GroupRow<TRow> {
+  __isGroup: true;
```
nstepien, line 260: We could generate a `Set`/`WeakSet` that contains all the group rows instead of hiding it in plain sight.

```ts
export function isGroupRow<R>(row: unknown, groupRows: ReadonlySet<GroupRow<R>>): row is GroupRow<R> {
  return groupRows.has(row);
}
```

Alternatively we could use a private symbol, though I don't know if that works well with ts.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol
```ts
const rowGroupSymbol = Symbol();

const rowGroup: RowGroup<...> = {
  [rowGroupSymbol]: true
};
```

WDYT?

src/types.ts
```diff
@@ -216,3 +245,26 @@ export interface SelectRowEvent {
...
   isShiftClick: boolean;
 }
+
+export interface Dictionary<T> {
+  [index: string]: T;
+}
+
+export type GroupByDictionary<TRow> = Dictionary<{
+  childRows: TRow[];
+  childGroups: TRow[] | GroupByDictionary<TRow>;
```
nstepien, line 255:
```suggestion
  childRows: readonly TRow[];
  childGroups: readonly TRow[] | GroupByDictionary<TRow>;
```

src/types.ts
```diff
@@ -216,3 +245,26 @@ export interface SelectRowEvent {
...
+  startRowIndex: number;
+}>;
+
+export interface GroupRow<TRow> {
+  __isGroup: true;
+  childRows: TRow[];
```
nstepien, line 261:
```suggestion
  childRows: readonly TRow[];
```

src/utils/index.ts
```diff
@@ -1,9 +1,15 @@
+import { GroupRow } from '../types';
+
 export * from './domUtils';
 export * from './columnUtils';
 export * from './viewportUtils';
 export * from './keyboardUtils';
 export * from './selectedCellUtils';
 
+export function isGroupRow<R>(row: R | GroupRow<R>): row is GroupRow<R> {
+  return (row as GroupRow<R>).__isGroup === true;
```
nstepien, line 10: Not safe if we want to allow `R` to be really anything.

nstepien (changes requested):
Not done reviewing this.

src/hooks/useViewportRows.ts
qili26: Classic `for` would also support the index. However, we don't have any `break;`, so it's not critical. I am just thinking for whatever even slightly performance we can improve, perhaps we shall do that.

```typescript
const keys = Object.keys(rows);
for (let posInSet = 0; posInSet < keys.length; posInSet++) {
  ...
}
```

src/DataGrid.tsx
qili26: You are right. I was thinking about `aria-label` for screen reader.

src/types.ts
```diff
@@ -216,3 +245,26 @@ export interface SelectRowEvent {
   checked: boolean;
   isShiftClick: boolean;
 }
+
+export interface Dictionary<T> {
+  [index: string]: T;
+}
```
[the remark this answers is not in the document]
qili26: @nstepien 
wdyt?
https://fnune.com/typescript/2019/01/30/typescript-series-1-record-is-usually-not-the-best-choice/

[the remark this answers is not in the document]
nstepien: It's a valid point, really depends whether you know what the value is or not.

src/types.ts
amanmahajan7: We can use a `Set` that contains all the group rows. This way we won't need `__isGroup `

amanmahajan7: I can try a private symbol

nstepien: We could do the check outside the function?
```
focusable(row: R, isGroupRow: false) {
focusable(row: GroupRow<R>, isGroupRow: true) {
```

src/hooks/useViewportRows.ts
```diff
@@ -0,0 +1,102 @@
...
+  rawRows: readonly R[];
+  rowHeight: number;
+  clientHeight: number;
+  scrollTop: number;
+  groupBy: readonly string[];
+  rowGrouper?: (rows: readonly R[], columnKey: string) => Dictionary<readonly R[]>;
+  expandedGroupIds?: ReadonlySet<unknown>;
+}
+
+export function useViewportRows<R, SR>({
```
nstepien, line 16: Can we move `getViewportRows` in this hook? 🤔

src/hooks/useViewportColumns.ts
```diff
@@ -18,24 +24,63 @@ export function useViewportColumns<R, K extends keyof R, SR>({
   columnWidths,
   viewportWidth,
   scrollLeft,
-  defaultColumnOptions
+  defaultColumnOptions,
+  groupBy: rawGroupBy,
+  rowGrouper
 }: ViewportColumnsArgs<R, K, SR>) {
+  const [sortedColumns, groupBy] = useMemo(() => {
```
nstepien, line 31: Can we add a return type?

src/hooks/useViewportRows.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  const [rows, allGroupRows] = useMemo(() => {
+    const allGroupRows = new Set<unknown>();
+    if (!groupedRows) return [rawRows, allGroupRows];
+
+    const expandGroup = (rows: GroupByDictionary<R> | readonly R[], parentId: string | undefined, level: number): Array<GroupRow<R> | R> => {
+      if (Array.isArray(rows)) return rows;
+      const flattenedRows: Array<R | GroupRow<R>> = [];
+      Object.keys(rows).forEach((groupKey, posInSet, keys) => {
+        // TODO: should users have control over the gerenated key?
```
nstepien, line 54:
```suggestion
        // TODO: should users have control over the generated key?
```

src/hooks/useViewportRows.ts
```diff
@@ -0,0 +1,102 @@
...
+    return groupRows(rawRows, groupBy, 0);
+  }, [groupBy, rowGrouper, rawRows]);
+
+  const [rows, allGroupRows] = useMemo(() => {
+    const allGroupRows = new Set<unknown>();
+    if (!groupedRows) return [rawRows, allGroupRows];
+
+    const expandGroup = (rows: GroupByDictionary<R> | readonly R[], parentId: string | undefined, level: number): Array<GroupRow<R> | R> => {
+      if (Array.isArray(rows)) return rows;
+      const flattenedRows: Array<R | GroupRow<R>> = [];
```
nstepien, line 52: We can move this `flattenedRows` outside `expandGroup`, and just push to it directly.

```patch
diff --git a/src/hooks/useViewportRows.ts b/src/hooks/useViewportRows.ts
index b9769254..b27420e5 100644
--- a/src/hooks/useViewportRows.ts
+++ b/src/hooks/useViewportRows.ts
@@ -47,9 +47,14 @@ export function useViewportRows<R, SR>({
     const allGroupRows = new Set<unknown>();
     if (!groupedRows) return [rawRows, allGroupRows];

-    const expandGroup = (rows: GroupByDictionary<R> | readonly R[], parentId: string | undefined, level: number): Array<GroupRow<R> | R> => {
-      if (Array.isArray(rows)) return rows;
-      const flattenedRows: Array<R | GroupRow<R>> = [];
+    const flattenedRows: Array<R | GroupRow<R>> = [];
+
+    const expandGroup = (rows: GroupByDictionary<R> | readonly R[], parentId: string | undefined, level: number) => {
+      if (Array.isArray(rows)) {
+        flattenedRows.push(...rows);
+        return;
+      }
+
       Object.keys(rows).forEach((groupKey, posInSet, keys) => {
         // TODO: should users have control over the gerenated key?
         const id = parentId !== undefined ? `${parentId}__${groupKey}` : groupKey;
@@ -71,14 +76,13 @@ export function useViewportRows<R, SR>({
         allGroupRows.add(groupRow);

         if (isExpanded) {
-          flattenedRows.push(...expandGroup(childGroups, id, level + 1));
+          expandGroup(childGroups, id, level + 1);
         }
       });
-
-      return flattenedRows;
     };

-    return [expandGroup(groupedRows, undefined, 0), allGroupRows];
+    expandGroup(groupedRows, undefined, 0);
+    return [flattenedRows, allGroupRows];
   }, [expandedGroupIds, groupedRows, rawRows]);

   const isGroupRow = <R>(row: unknown): row is GroupRow<R> => {
```

src/hooks/useViewportRows.ts
```diff
@@ -0,0 +1,102 @@
...
+      Object.keys(rows).forEach((groupKey, posInSet, keys) => {
+        // TODO: should users have control over the gerenated key?
+        const id = parentId !== undefined ? `${parentId}__${groupKey}` : groupKey;
+        const isExpanded = expandedGroupIds?.has(id) ?? false;
+        const { childRows, childGroups, startRowIndex } = (rows as GroupByDictionary<R>)[groupKey]; // TODO: why is it failing?
```
nstepien, line 57: `Array.isArray` only asserts that `rows` is `any[]`, unfortunately.
https://github.com/microsoft/TypeScript/issues/33700
https://github.com/microsoft/TypeScript/issues/17002

We could wrap `Array.isArray` in our own type guard function, and use the `/*@__INLINE__*/` annotation 🤔 
https://github.com/terser/terser#annotations

nstepien (changes requested):

style/row.less
```diff
@@ -34,37 +34,25 @@
...
+  &::after {
+    content: "";
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    box-shadow: inset 0 0 0 2px @selectedBorderColor;
+    pointer-events: none;
+    z-index: 2;
```
qili26, line 51: I feel this might cause issue for other projects with `z-index`

nstepien: How so? We already have some z-index in rdg.
We also set a stacking context to avoid z-index conflicts.
https://github.com/adazzle/react-data-grid/blob/canary/style/core.less#L4-L5

src/hooks/useViewportRows.ts
amanmahajan7: Great idea.

src/hooks/useViewportColumns.ts
```diff
@@ -18,24 +24,63 @@ export function useViewportColumns<R, K extends keyof R, SR>({
...
+        frozen: true,
+        rowGroup: true,
+        groupFormatter: c.groupFormatter ?? ToggleGroupedFormatter,
+        formatterOptions: {
+          groupFocusable(row: GroupRow<R>) {
+            return row.level === index;
+          }
+        }
+      }))
+      .sort((c1, c2) => groupBy.findIndex(k => k === c1.key) - groupBy.findIndex(k => k === c2.key));
```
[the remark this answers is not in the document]
amanmahajan7: Made some change. Wdyt about the last commit?

src/hooks/useViewportRows.ts
amanmahajan7: I am not too sure but I do want to do some cleanup as `DataGrid` code is getting complex. Let's think about it and improve it in a followup PR

src/hooks/useViewportRows.ts
amanmahajan7: I think we can leave it for now as it is needed only in one plane. I will add a comment

src/utils/columnUtils.ts
```diff
@@ -34,16 +42,46 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
...
+    } else if (column.frozen) {
       lastFrozenColumnIndex++;
       columns.splice(lastFrozenColumnIndex, 0, column);
     } else {
       columns.push(column);
     }
   }
 
+  // Sort groupBy columns
+  if (groupBy.length >= 0) {
```
nstepien, line 71: We can do better than this.
I think implementing sorting with `.sort()` is more flexible, we could add more sorting features in it in the future.
`groupBy` is also not `O(n^2)` anymore, which should please @qili26 I hope.
```patch
diff --git a/src/utils/columnUtils.ts b/src/utils/columnUtils.ts
index a96fde03..c7a4c8c5 100644
--- a/src/utils/columnUtils.ts
+++ b/src/utils/columnUtils.ts
@@ -1,4 +1,4 @@
-import { Column, CalculatedColumn, FormatterProps, Omit, GroupRow } from '../types';
+import { Column, CalculatedColumn, FormatterProps, GroupRow } from '../types';
 import { ToggleGroupedFormatter } from '../formatters';
 import { SELECT_COLUMN_KEY } from '../Columns';
 
@@ -26,13 +26,10 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
   let allocatedWidths = 0;
   let unassignedColumnsCount = 0;
   let lastFrozenColumnIndex = -1;
-  type IntermediateColumn = Omit<Column<R, SR>, 'width'> & { width: number | undefined; rowGroup?: boolean };
-  let columns: IntermediateColumn[] = [];
+  type IntermediateColumn = Column<R, SR> & { width: number | undefined; rowGroup?: boolean };
+  const { rawGroupBy } = metrics;
 
-  // Find valid groupBy columns
-  const groupBy: readonly string[] = metrics.rawGroupBy?.filter(key => metrics.columns.find(c => c.key === key) !== undefined) ?? [];
-
-  for (const metricsColumn of metrics.columns) {
+  const columns = metrics.columns.map(metricsColumn => {
     let width = getSpecifiedWidth(metricsColumn, metrics.columnWidths, metrics.viewportWidth);
 
     if (width === undefined) {
@@ -42,45 +39,56 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
       allocatedWidths += width;
     }
 
-    let column: IntermediateColumn = { ...metricsColumn, width };
+    const column: IntermediateColumn = { ...metricsColumn, width };
 
-    if (groupBy.includes(column.key)) {
-      lastFrozenColumnIndex++;
-      const level = groupBy.indexOf(column.key);
-      column = {
-        ...column,
-        frozen: true,
-        rowGroup: true,
-        groupFormatter: column.groupFormatter ?? ToggleGroupedFormatter,
-        formatterOptions: {
-          groupFocusable(row: GroupRow<R>) {
-            return row.level === level;
-          }
+    if (rawGroupBy?.includes(column.key)) {
+      const level = rawGroupBy.indexOf(column.key);
+      column.frozen = true;
+      column.rowGroup = true;
+      column.groupFormatter = column.groupFormatter ?? ToggleGroupedFormatter;
+      column.formatterOptions = {
+        groupFocusable(row: GroupRow<R>) {
+          return row.level === level;
         }
       };
-      columns.push(column);
-    } else if (column.frozen) {
+    }
+
+    if (column.frozen) {
       lastFrozenColumnIndex++;
-      columns.splice(lastFrozenColumnIndex, 0, column);
-    } else {
-      columns.push(column);
     }
-  }
 
-  // Sort groupBy columns
-  if (groupBy.length >= 0) {
-    const selectColumn = columns.find(c => c.key === SELECT_COLUMN_KEY);
-    const groupByColumns = groupBy.map(key => columns.find(c => c.key === key)!);
-    const remaningColumns = columns.filter(c => !groupByColumns.includes(c) && c !== selectColumn);
-    columns = [
-      ...groupByColumns,
-      ...remaningColumns
-    ];
+    return column;
+  });
 
-    if (selectColumn) {
-      columns.unshift(selectColumn);
+  columns.sort(({ key: aKey, frozen: frozenA }, { key: bKey, frozen: frozenB }) => {
+    // Sort select column first:
+    if (aKey === SELECT_COLUMN_KEY) return -1;
+    if (bKey === SELECT_COLUMN_KEY) return 1;
+    // Sort grouped columns second, following the groupBy order:
+    if (rawGroupBy) {
+      if (rawGroupBy.includes(aKey)) {
+        if (rawGroupBy.includes(bKey)) {
+          return rawGroupBy.indexOf(aKey) - rawGroupBy.indexOf(bKey);
+        }
+        return -1;
+      }
+      if (rawGroupBy.includes(bKey)) {
+        return 1;
+      }
     }
-  }
+    // Sort frozen columns third:
+    if (frozenA) {
+      if (frozenB) {
+        return 0;
+      }
+      return -1;
+    }
+    if (frozenB) {
+      return 1;
+    }
+    // Sort other columns last:
+    return 0;
+  });
 
   const unallocatedWidth = metrics.viewportWidth - allocatedWidths;
   const unallocatedColumnWidth = Math.max(
@@ -88,6 +96,9 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
     metrics.minColumnWidth
   );
 
+  // Valid groupBy columns
+  const groupBy: string[] = [];
+
   const calculatedColumns: CalculatedColumn<R, SR>[] = columns.map((column, idx) => {
     // Every column should have a valid width as this stage
     const width = column.width ?? clampColumnWidth(unallocatedColumnWidth, column, metrics.minColumnWidth);
@@ -102,6 +113,9 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
     };
     totalWidth += width;
     left += width;
+    if (column.rowGroup) {
+      groupBy.push(column.key);
+    }
     return newColumn;
   });
```
Where's my raise?

nstepien (changes requested):

src/utils/columnUtils.ts
```diff
@@ -34,16 +42,46 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
...
-    if (column.frozen) {
+    if (groupBy.includes(column.key)) {
+      lastFrozenColumnIndex++;
+      const level = groupBy.indexOf(column.key);
+      column = {
+        ...column,
+        frozen: true,
+        rowGroup: true,
+        groupFormatter: column.groupFormatter ?? ToggleGroupedFormatter,
+        formatterOptions: {
```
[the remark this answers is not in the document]
qili26: I think we should.

style/row.less
qili26: For example if we have a page rendering rdg in the model while a side panel opens which covers the model, potentially there will be some possibilities that this selected border is rendering over the side panel while the model itself is beneath the side panel.

On another thought, maybe we just have to ask them to override the z-index themselves.

nstepien: >there will be some possibilities that this selected border is rendering over the side panel while the model itself is beneath the side panel.

Not possible:
https://github.com/adazzle/react-data-grid/blob/canary/style/core.less#L4-L5
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

src/utils/columnUtils.ts
```diff
@@ -62,6 +97,14 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
       resizable: column.resizable ?? metrics.defaultResizable,
       formatter: column.formatter ?? metrics.defaultFormatter
     };
+
+    if (newColumn.rowGroup) {
+      const level = groupBy.push(column.key) - 1;
+      newColumn.groupFormatter = column.groupFormatter ?? ToggleGroupFormatter;
+      newColumn.formatterOptions = newColumn.formatterOptions ?? {};
+      newColumn.formatterOptions.groupFocusable = newColumn.formatterOptions.groupFocusable ?? (row => row.level === level);
```
nstepien, line 105: If `formatterOptions` is already defined but `formatterOptions.groupFocusable` isn't, we're setting `groupFocusable` in the original column's `formatterOptions`, which could break things.

nstepien (changes requested):

style/row.less
qili26: nice, missed that comment.

src/types.ts
```diff
@@ -213,3 +241,23 @@ export interface SelectRowEvent {
...
+  childRows: readonly TRow[];
+  childGroups: readonly TRow[] | GroupByDictionary<TRow>;
+  startRowIndex: number;
+}>;
+
+export interface GroupRow<TRow> {
+  childRows: readonly TRow[];
+  id: string;
+  parentId: unknown;
+  groupKey: unknown;
```
nstepien, line 257: We should keep things consistent and either use `id` or `key` for those, not both.

src/utils/columnUtils.ts
```diff
@@ -36,22 +41,52 @@ export function getColumnMetrics<R, SR>(metrics: Metrics<R, SR>): ColumnMetrics<
...
+    return 0;
+  });
 
   const unallocatedWidth = metrics.viewportWidth - allocatedWidths;
   const unallocatedColumnWidth = Math.max(
     Math.floor(unallocatedWidth / unassignedColumnsCount),
     metrics.minColumnWidth
   );
 
+  const groupBy: string[] = [];
```
nstepien, line 89: Should add a comment explaining that this array is a sanitized version of `rawGroupBy`.

nstepien (changes requested):

src/formatters/ToggleGroupFormatter.less
```diff
@@ -0,0 +1,47 @@
+.caret-collapsed,
```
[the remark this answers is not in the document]
amanmahajan7: Done

src/formatters/ToggleGroupFormatter.less
```diff
@@ -0,0 +1,47 @@
...
+  border-left: 6px solid transparent;
+  border-right: 6px solid transparent;
+}
+
+.caret-collapsed:after {
+  content: '';
+  position: absolute;
+  left: 2px;
+  top: 2px;
+  border-bottom: 4px solid #fff;
```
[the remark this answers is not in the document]
amanmahajan7: Any quick fix otherwise we can revisit if needed. It can be overridden with a custom `groupFormatter` so should be okay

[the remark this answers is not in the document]
nstepien: Let's revisit when needed.

stories/demos/Grouping.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+interface Option {
+  value: string;
+  label: string;
+}
+
+const sports = ['Swimming', 'Gymnastics', 'Speed Skating', 'Cross Country Skiing', 'Short-Track Speed Skating', 'Diving', 'Cycling', 'Biathlon', 'Alpine Skiing', 'Ski Jumping', 'Nordic Combined', 'Athletics', 'Table Tennis', 'Tennis', 'Synchronized Swimming', 'Shooting', 'Rowing', 'Fencing', 'Equestrian', 'Canoeing', 'Bobsleigh', 'Badminton', 'Archery', 'Wrestling', 'Weightlifting', 'Waterpolo', 'Wrestling', 'Weightlifting'];
+
+const columns: Column<Row>[] = [
```
nstepien, line 28:
```suggestion
const columns: readonly Column<Row>[] = [
```

stories/demos/Grouping.tsx
```diff
@@ -0,0 +1,170 @@
...
+    formatter({ row }) {
+      return <>{row.gold + row.silver + row.bronze}</>;
+    },
+    groupFormatter({ childRows }) {
+      return <>{childRows.reduce((prev, row) => prev + row.gold + row.silver + row.bronze, 0)}</>;
+    }
+  }
+];
+
+function createRows(): Row[] {
```
nstepien, line 79:
```suggestion
function createRows(): readonly Row[] {
```

stories/demos/Grouping.tsx
```diff
@@ -0,0 +1,170 @@
...
+      country: faker.address.country(),
+      sport: sports[faker.random.number(sports.length - 1)],
+      athlete: faker.name.findName(),
+      gold: faker.random.number(5),
+      silver: faker.random.number(5),
+      bronze: faker.random.number(5)
+    });
+  }
+
+  return rows.sort((r1, r2) => r2.country.localeCompare(r1.country));
```
nstepien, line 94: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#Performance
If you want.

src/GroupRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+  level,
+  isExpanded,
+  selectedCellIdx,
+  isRowSelected,
+  eventBus,
+  onFocus,
+  onKeyDown,
+  'aria-setsize': ariaSetSize,
+  'aria-posinset': ariaPosInSet,
+  'aria-rowindex': ariaRowIndex,
```
nstepien, line 23: Don't need to extract all the props if we're going to spread `...props` back into the div anyway.

src/GroupCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+        'rdg-cell-selected': isCellSelected
+      })}
+      style={{
+        width: column.width,
+        left: column.left,
+        cursor: isLevelMatching ? 'pointer' : 'default'
+      }}
+      onClick={isLevelMatching ? toggleGroup : undefined}
+    >
+      {column.groupFormatter && (column.rowGroup ? groupColumnIndex === column.idx : true) && (
```
nstepien, line 61:
```suggestion
      {column.groupFormatter && (!column.rowGroup || groupColumnIndex === column.idx) && (
```

nstepien (changes requested):

nstepien (approved):

PR #2108, opened 2020-08-03
Update packages

amanmahajan7:


package.json
```diff
@@ -37,58 +37,58 @@
...
     "react-dnd-html5-backend": "^11.1.1",
     "react-dom": "^16.13.1",
     "react-popper": "^2.2.3",
     "react-select": "^3.1.0",
     "react-virtualized": "^9.21.2",
-    "ts-jest": "^26.0.0",
+    "ts-jest": "^26.1.4",
     "ts-loader": "^7.0.5",
-    "typescript": "~3.9.3"
+    "typescript": "^3.9.7"
```
nstepien, line 91: I would have prefered `~` instead of `^`, but since the next version will be 4.0.0 we can leave it as is.

PR #2112, opened 2020-08-08
Fix null ref error

amanmahajan7:


amanmahajan7:
We need a better fix, I found this issue while working on editor2. Let me check if there is a better fix

amanmahajan7:
It only happens in `editor2`.  `editor1` commits value before unmount so we won't have a case where edit flag is true on the first cell render. I will look for a better fix in the `editor2` PR

PR #2116, opened 2020-08-12
Looking to add styles to cells.

vinceprofeta:
Open to any other approach as well. ClassName props would work too.


nstepien:
Are you running into any limitation with using `Column.cellClass`, `Column.headerCellClass`, and `Column.summaryCellClass`?

PR #2117, opened 2020-08-14
Add defaultColumnOptions

amanmahajan7:


src/hooks/useViewportColumns.ts
```diff
@@ -16,21 +15,27 @@ interface ViewportColumnsArgs<R, K extends keyof R, SR> extends SharedDataGridPr
...
   viewportWidth,
-  defaultFormatter,
-  scrollLeft
+  scrollLeft,
+  defaultColumnOptions
 }: ViewportColumnsArgs<R, K, SR>) {
+  const minColumnWidth = defaultColumnOptions?.minWidth ?? 80;
+  const defaultFormatter = defaultColumnOptions?.formatter ?? ValueFormatter;
+  const defaultSortable = defaultColumnOptions?.sortable;
+  const defaultResizable = defaultColumnOptions?.resizable;
```
nstepien, line 26: Let's enforce non-nullish types when we can. We can simplify the calculated column type as well that way I think.
```suggestion
  const defaultSortable = defaultColumnOptions?.sortable ?? false;
  const defaultResizable = defaultColumnOptions?.resizable ?? false;
```

src/utils/columnUtils.ts
```diff
@@ -6,6 +6,8 @@ interface Metrics<R, SR> {
   columnWidths: ReadonlyMap<string, number>;
   minColumnWidth: number;
   viewportWidth: number;
+  defaultResizable?: boolean;
+  defaultSortable?: boolean;
```
nstepien, line 10:
```suggestion
  defaultResizable: boolean;
  defaultSortable: boolean;
```

stories/demos/CommonFeatures.tsx
```diff
@@ -280,6 +252,10 @@ export default function CommonFeatures() {
           rows={sortedRows}
           width={width}
           height={height}
+          defaultColumnOptions={{
```
nstepien, line 255: Let's move this to a constant outside the component so we avoid rerenders.

PR #2128, opened 2020-08-21
Cleanup

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -336,7 +338,10 @@ function DataGrid<R, K extends keyof R, SR>({
       current.scrollTop = rowIdx * rowHeight;
     },
     selectCell,
-    commitChanges: handleCommit2
+    commitChanges() {
```
nstepien, line 341: Do we use this api ourselves? I wonder what's the use case for this.

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: It was added as initially we were not committing the value on outside click. It is probably not needed now. Let's revisit it
https://github.com/adazzle/react-data-grid/pull/2102#discussion_r463855118

PR #2129, opened 2020-08-24
Use ResizeObserver to calculate grid width

amanmahajan7:
ResizeObserver has a pretty good browser support now
https://caniuse.com/#feat=resizeobserver
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

This can solve a few issues
- Grid container resize may not always trigger window resize but resize observer will catch it
- If the grid is initially hidden and then displayed then grid width will be calculated again

TS does not have definitions for `ResizeObserver` so I copied some from https://www.w3.org/TR/resize-observer/#resizeobserversize

### DIFF
--- a/src/hooks/useGridWidth.ts
+++ b/src/hooks/useGridWidth.ts
@@ -1,4 +1,4 @@
-import { useRef, useState, useLayoutEffect } from 'react';
+import { useRef, useState, useLayoutEffect, useMemo } from 'react';
 
 // https://github.com/microsoft/TypeScript/issues/37861
 interface ResizeObserverSize {
@@ -14,15 +14,14 @@ type ResizeObserverCallback = (entries?: ResizeObserverEntry[]) => void;
 
 type ResizeObserver = new (callback: ResizeObserverCallback) => {
   observe: (target: Element) => void;
-  unobserve: (target: Element) => void;
+  disconnect: () => void;
 };
-
-
-const { ResizeObserver } = window as (typeof window & { ResizeObserver?: ResizeObserver });
 
 export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement>, number] {
   const gridRef = useRef<HTMLDivElement>(null);
-  const [resizeObserver] = useState(() => {
+  const [gridWidth, setGridWidth] = useState(0);
+  const resizeObserver = useMemo(() => {
+    const { ResizeObserver } = window as (typeof window & { ResizeObserver?: ResizeObserver });
     if (ResizeObserver === undefined) {
       return null;
     }
@@ -33,8 +32,7 @@ export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement
         setGridWidth(newWidth);
       }
     });
-  });
-  const [gridWidth, setGridWidth] = useState(0);
+  }, []);
 
   useLayoutEffect(() => {
     // Do not calculate the width if width is provided
@@ -44,7 +42,7 @@ export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement
     if (resizeObserver !== null) {
       const node = gridRef.current!;
       resizeObserver.observe(node);
-      return () => resizeObserver.unobserve(node);
+      return () => resizeObserver.disconnect();
     }
 
     function onResize() {
--- a/src/hooks/useGridWidth.ts
+++ b/src/hooks/useGridWidth.ts
@@ -1,16 +1,7 @@
 import { useRef, useState, useLayoutEffect, useMemo } from 'react';
 
 // https://github.com/microsoft/TypeScript/issues/37861
-interface ResizeObserverSize {
-  inlineSize: number;
-}
-
-interface ResizeObserverEntry {
-  target: Element;
-  borderBoxSize?: ResizeObserverSize;
-}
-
-type ResizeObserverCallback = (entries: ResizeObserverEntry[]) => void;
+type ResizeObserverCallback = () => void;
 
 type ResizeObserver = new (callback: ResizeObserverCallback) => {
   observe: (target: Element) => void;
@@ -26,11 +17,8 @@ export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement
       return null;
     }
 
-    return new ResizeObserver((entries: ResizeObserverEntry[]) => {
-      if (entries.length > 0) {
-        const newWidth = entries[0].borderBoxSize?.inlineSize ?? entries[0].target.getBoundingClientRect().width;
-        setGridWidth(newWidth);
-      }
+    return new ResizeObserver(() => {
+      setGridWidth(gridRef.current!.getBoundingClientRect().width);
     });
   }, []);
 
--- a/src/hooks/useGridWidth.ts
+++ b/src/hooks/useGridWidth.ts
@@ -18,7 +18,7 @@ type ResizeObserver = new (callback: ResizeObserverCallback) => {
 };
 
 
-const { ResizeObserver } = window as (typeof window & { ResizeObserver: ResizeObserver });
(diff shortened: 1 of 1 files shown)


src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
+const { ResizeObserver } = window as (typeof window & { ResizeObserver: ResizeObserver });
+
 export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement>, number] {
   const gridRef = useRef<HTMLDivElement>(null);
+  const [resizeObserver] = useState(() => {
+    if (ResizeObserver === undefined) {
+      return null;
+    }
+
+    return new ResizeObserver((entries?: ResizeObserverEntry[]) => {
```
amanmahajan7, line 30: Not sure if `entries` can be undefined or not so added a safeguard

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
+      if (Array.isArray(entries) && entries.length > 0) {
+        const newWidth = entries[0].borderBoxSize?.inlineSize ?? entries[0].target.getBoundingClientRect().width;
```
amanmahajan7, line 32: `borderBoxSize` is not supported everywhere and it is marked as experimental
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/borderBoxSize

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
+  observe: (target: Element) => void;
+  unobserve: (target: Element) => void;
+};
+
+
+const { ResizeObserver } = window as (typeof window & { ResizeObserver?: ResizeObserver });
+
 export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement>, number] {
   const gridRef = useRef<HTMLDivElement>(null);
+  const [resizeObserver] = useState(() => {
```
nstepien, line 25: This could be a `useMemo`

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
+    if (ResizeObserver === undefined) {
+      return null;
+    }
+
+    return new ResizeObserver((entries?: ResizeObserverEntry[]) => {
+      if (Array.isArray(entries) && entries.length > 0) {
+        const newWidth = entries[0].borderBoxSize?.inlineSize ?? entries[0].target.getBoundingClientRect().width;
+        setGridWidth(newWidth);
```
nstepien, line 33: Can't we just do
```suggestion
        setGridWidth(gridRef.current!.getBoundingClientRect().width);
```

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
+
 export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement>, number] {
   const gridRef = useRef<HTMLDivElement>(null);
+  const [resizeObserver] = useState(() => {
+    if (ResizeObserver === undefined) {
+      return null;
+    }
+
+    return new ResizeObserver((entries?: ResizeObserverEntry[]) => {
+      if (Array.isArray(entries) && entries.length > 0) {
```
nstepien, line 31: In which cases is it not an array or an empty array?

src/hooks/useGridWidth.ts
nstepien, line 33: Also this references `setGridWidth` which isn't in scope yet, I'm worried about TDZ.
https://stackoverflow.com/questions/33198849/what-is-the-temporal-dead-zone

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
+
+type ResizeObserverCallback = (entries?: ResizeObserverEntry[]) => void;
+
+type ResizeObserver = new (callback: ResizeObserverCallback) => {
+  observe: (target: Element) => void;
+  unobserve: (target: Element) => void;
+};
+
+
+const { ResizeObserver } = window as (typeof window & { ResizeObserver?: ResizeObserver });
```
nstepien, line 21: Let's move this line inside the function that needs it.

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
 
   useLayoutEffect(() => {
     // Do not calculate the width if width is provided
     if (typeof width === 'number') return;
+
+    // Use ResizeObserver if supported
+    if (resizeObserver !== null) {
+      const node = gridRef.current!;
+      resizeObserver.observe(node);
+      return () => resizeObserver.unobserve(node);
```
nstepien, line 47:
```suggestion
      resizeObserver.observe(gridRef.current!);
      return () => resizeObserver.disconnect();
```

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,65 @@
...
+    }
+
     function onResize() {
-    // Immediately re-render when the component is mounted to get valid columnMetrics.
+      // Immediately re-render when the component is mounted to get valid columnMetrics.
       setGridWidth(gridRef.current!.getBoundingClientRect().width);
     }
+
+    // Fallback to window resize
     onResize();
```
nstepien, line 56: Will the resize observer work immediately? Do we not need to call this?

nstepien (changes requested):

nstepien:
Should we remove `AutoSizer` from the stories?

nstepien:
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/observe
>`box`: Sets which box model the observer will observe changes to. Possible values are content-box (the default), and border-box.

https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
>Note: The content box is the box in which content can be placed, meaning the border box minus the padding and border width. The border box encompasses the content, padding, and border. See The box model for further explanation.

Maybe we should use the border-box since we do `width - border widths`, or use the content-box and remove the border widths from the equation.

src/hooks/useGridWidth.ts
amanmahajan7: I thought about it but I was not sure if it can cause forced reflow / layout. I can change it

amanmahajan7: Moved it before `useMemo`

nstepien: It should be safe, if we get a resize event that means it has already computed the layout.

src/hooks/useGridWidth.ts
amanmahajan7: Yes, it does seem to work immediately

src/hooks/useGridWidth.ts
amanmahajan7: Done

src/hooks/useGridWidth.ts
amanmahajan7: Removed with gridRef

amanmahajan7:
> Should we remove AutoSizer from the stories?

We need it for the height calculation

src/hooks/useGridWidth.ts
```diff
@@ -1,23 +1,51 @@
...
+  observe: (target: Element) => void;
+  disconnect: () => void;
+};
 
 export function useGridWidth<T>(width?: number): [React.RefObject<HTMLDivElement>, number] {
   const gridRef = useRef<HTMLDivElement>(null);
   const [gridWidth, setGridWidth] = useState(0);
+  const resizeObserver = useMemo(() => {
+    const { ResizeObserver } = window as (typeof window & { ResizeObserver?: ResizeObserver });
+    if (ResizeObserver === undefined) {
```
nstepien, line 16: Let's avoid setting up a resize observer if we won't end up using it.
```suggestion
    if (typeof width === 'number' || ResizeObserver === undefined) {
```

nstepien (changes requested):

PR #2130, opened 2020-08-25
rm width/height props, use ResizeObserver to get the grid's content dimensions

nstepien:
I've only added `classNames` to steer devs toward using proper css sizing rather than more js-heavy alternatives.
We can always add support for `style={{ width, height }}` if there is demand for it.


src/hooks/useGridDimensions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+type ResizeObserver = new (callback: (entries: readonly ResizeObserverEntry[]) => void) => {
+  observe: (target: Element) => void;
+  disconnect: () => void;
+};
+
+export function useGridDimensions(): [React.RefObject<HTMLDivElement>, number, number] {
+  const gridRef = useRef<HTMLDivElement>(null);
+  const [gridWidth, setGridWidth] = useState(1);
+  const [gridHeight, setGridHeight] = useState(1);
```
nstepien, line 19: I've set to `1` by default just to avoid divisions by zero possibly breaking things.

src/hooks/useGridDimensions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  useLayoutEffect(() => {
+    const { ResizeObserver } = window as typeof window & { ResizeObserver: ResizeObserver };
```
nstepien, line 22: I think it's fair to ask devs to use a polyfill to get the grid working, though that'd only be on IE11/old Edge/Safari 13.
Chrome and Firefox have had it for a long time now.
I can add a `if (!ResizeObserver) return;` if you want, but the grid will look quite empty.

src/hooks/useGridDimensions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    const resizeObserver = new ResizeObserver(entries => {
+      const { width, height } = entries[0].contentRect;
```
nstepien, line 25: We can now properly support css-only dimensions, custom border sizes for the grid, and custom scrollbar sizes as well.

src/utils/domUtils.ts
```diff
@@ -1,23 +1,3 @@
...
-
-    scrollDiv.style.position = 'absolute';
-    scrollDiv.style.top = '-9999px';
-    scrollDiv.style.width = '50px';
-    scrollDiv.style.height = '50px';
-    scrollDiv.style.overflow = 'scroll';
-
-    document.body.appendChild(scrollDiv);
-    size = scrollDiv.offsetWidth - scrollDiv.clientWidth;
-    document.body.removeChild(scrollDiv);
```
nstepien, line 15: We won't manually trigger layout/reflow in the middle of rendering now 👌 👌 👌

style/core.less
```diff
@@ -4,6 +4,7 @@
   // https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context
   // We set a stacking context so internal elements don't render on top of external components.
   contain: strict;
+  height: 350px;
```
nstepien, line 7: Same default as before, users can override however they want with `height`/flex/grid/etc.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## `alpha` to `canary`
 - **Added:**
   - **Props:**
+    - `className`
```
amanmahajan7, line 6: Do we want to support the style prop?

src/hooks/useGridDimensions.ts
```diff
@@ -0,0 +1,41 @@
...
+export function useGridDimensions(): [React.RefObject<HTMLDivElement>, number, number] {
+  const gridRef = useRef<HTMLDivElement>(null);
+  const [gridWidth, setGridWidth] = useState(1);
+  const [gridHeight, setGridHeight] = useState(1);
+
+  useLayoutEffect(() => {
+    const { ResizeObserver } = window as typeof window & { ResizeObserver: ResizeObserver };
+
+    // don't break in jest/jsdom and browsers that don't support ResizeObserver
+    if (ResizeObserver == null) return;
```
amanmahajan7, line 25: So in this case we render an empty grid? I wonder if we should throw an exception

src/hooks/useGridDimensions.ts
amanmahajan7: May be we can add a note on the readme file that a polyfill is required for old browsers. Something like https://reactjs.org/docs/javascript-environment-requirements.html

src/utils/domUtils.ts
amanmahajan7: 👏

CHANGELOG.md
nstepien: Let's add it if there's demand for it, see my opening comment.

src/hooks/useGridDimensions.ts
nstepien: Almost.

image

We could log something with `console.error`, but it'd spam jest tests as well.

src/hooks/useGridDimensions.ts
nstepien: I've added a line in the readme.

src/utils/domUtils.ts
qili26: @nstepien 
Could you explain a little bit why we needed this originally? Is it for the scenario when the scroll bar appears and we need to get the new demension?

nstepien: We needed it to know what the scrollbar size was so we could correctly calculate the actual vertical/horizontal viewport.

ekeuus:
So there is no way of passing in the height variable as a prop? I am using css modules and seems like there is no way to really have a full height container anymore.

nstepien:
Use the `className` or `style` prop.

PR #2132, opened 2020-08-25
Add useFocusRef hook

amanmahajan7:
I think this can be a useful hook as many custom formatters would need it.

### DIFF
--- a/src/hooks/useFocusRef.tsx
+++ b/src/hooks/useFocusRef.tsx
@@ -1,0 +1,11 @@
+import { useRef, useLayoutEffect } from 'react';
+
+export function useFocusRef<T extends HTMLOrSVGElement>(isCellSelected: boolean | undefined) {
+  const ref = useRef<T>(null);
+  useLayoutEffect(() => {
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -6,3 +6,4 @@ export * from './formatters';
 export { SimpleTextEditor } from './editors';
 export * from './enums';
 export * from './types';
+export { useFocusRef } from './hooks';
```
qili26, line 9: Don't know if it worth to export.

amanmahajan7: You are probably right. I have removed it for now

src/hooks/useFocusRef.tsx
```diff
@@ -0,0 +1,11 @@
+import { useRef, useLayoutEffect } from 'react';
+
+export function useFocusRef<T extends HTMLOrSVGElement>(isCellSelected: boolean | undefined) {
```
nstepien, line 3: If we ever decide to export it out of rdg, we might want to consider a less generic name.

nstepien (approved):

PR #2136, opened 2020-08-26
Add content-visibility: auto to the root element

nstepien:
https://web.dev/content-visibility/
https://www.chromestatus.com/feature/4613920211861504
https://drafts.csswg.org/css-contain/#propdef-content-visibility

I've not added `content-visibility` to the rows/cells since we already apply strict containment, and it seems to reduce performance instead of improving it.

I've also added `contain: size layout style paint;` **in addition to** `contain: strict;`.
Firefox doesn't support `style` containment, so
- `contain: strict;` will be applied on Firefox
- `contain: size layout style paint;` will be applied on Chrome, and will override `contain: strict;`

### DIFF
--- a/style/core.less
+++ b/style/core.less
@@ -4,6 +4,7 @@
   // https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context
   // We set a stacking context so internal elements don't render on top of external components.
   contain: strict;
+  contain: size layout style paint;
   content-visibility: auto;
   height: 350px;
   border: 1px solid @borderColor;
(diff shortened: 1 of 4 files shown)


style/core.less
```diff
@@ -4,6 +4,8 @@
   // https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context
   // We set a stacking context so internal elements don't render on top of external components.
   contain: strict;
+  contain: size layout style paint;
+  content-visibility: auto;
```
amanmahajan7, line 8: Just so I understand, we will eventually remove the `contain` style once `content-visibility` has a better support. Correct?

nstepien: I think so yes.

PR #2138, opened 2020-08-31
Remove formatterOptions

amanmahajan7:
`onFocus` bubbles so we can use it to check if a cell was focused or not. This behavior is not changing in react 17. We won't need special handling for group row cells and conditional focusing 

https://reactjs.org/blog/2020/08/10/react-v17-rc.html#aligning-with-browsers

### DIFF
--- a/src/Row.tsx
+++ b/src/Row.tsx
@@ -80,8 +80,8 @@ function Row<R, SR = unknown>({
             isCellSelected={isCellSelected}
             isRowSelected={isRowSelected}
             eventBus={eventBus}
-            dragHandleProps={isCellSelected ? (selectedCellProps as SelectedCellProps)!.dragHandleProps : undefined}
-            onFocus={isCellSelected ? (selectedCellProps as SelectedCellProps)!.onFocus : undefined}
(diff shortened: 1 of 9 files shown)


src/Row.tsx
```diff
@@ -81,6 +81,7 @@ function Row<R, SR = unknown>({
             isRowSelected={isRowSelected}
             eventBus={eventBus}
             dragHandleProps={isCellSelected ? (selectedCellProps as SelectedCellProps)!.dragHandleProps : undefined}
+            onFocus={isCellSelected ? (selectedCellProps as SelectedCellProps)!.onFocus : undefined}
```
nstepien, line 84: non-nullish assertions are redundant.
```suggestion
            dragHandleProps={isCellSelected ? (selectedCellProps as SelectedCellProps).dragHandleProps : undefined}
            onFocus={isCellSelected ? (selectedCellProps as SelectedCellProps).onFocus : undefined}
```

nstepien (approved):

nstepien (approved):

PR #2140, opened 2020-09-01
Fix select column sorting

amanmahajan7:


src/Columns.tsx
```diff
@@ -11,6 +11,7 @@ export const SelectColumn: Column<any, any> = {
   width: 35,
   maxWidth: 35,
   resizable: false,
+  sortable: false,
```
amanmahajan7, line 14: This breaks if `defaultColumnOptions` sets `sortable` to true

src/DataGrid.tsx
```diff
@@ -323,11 +322,7 @@ function DataGrid<R, K extends keyof R, SR>({
       if (!current) return;
       current.scrollTop = rowIdx * rowHeight;
     },
-    selectCell,
-    commitChanges() {
```
amanmahajan7, line 327: Let's add it back if needed

nstepien (approved):

PR #2144, opened 2020-09-03
Add a quick animation to the caret

nstepien:
Makes grouping delightful to use.
https://css-tricks.com/animate-svg-path-changes-in-css/
Doesn't work in Firefox/Safari though.


qili26 (approved):

amanmahajan7 (approved):
👍

PR #2147, opened 2020-09-04
Add .rdg-row-selected to group rows

nstepien:


amanmahajan7 (approved):
Should we add css variables and remove https://github.com/adazzle/react-data-grid/blob/canary/style/variables.less?

nstepien:
We should, for a future PR.

PR #2149, opened 2020-09-04
Promote editor2

nstepien:
- [x] Changelog
- [x] Removed support for the old editor API completely


src/hooks/useCombinedRefs.ts
```diff
@@ -1,9 +1,17 @@
-import { useMemo } from 'react';
-import { wrapRefs } from '../utils';
+import { useCallback } from 'react';
 
 export function useCombinedRefs<T>(...refs: readonly React.Ref<T>[]) {
-  return useMemo(
```
nstepien, line 5: https://reactjs.org/docs/hooks-reference.html#usememo

>**You may rely on useMemo as a performance optimization, not as a semantic guarantee.** In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without `useMemo` — and then add it to optimize performance.

Looks like `useCallback` is safer than `useMemo` for this hook 🤔

amanmahajan7 (commented):
Still reviewing. A few comments so far

amanmahajan7 (approved):

PR #2151, opened 2020-09-11
Update deps

nstepien:


src/enums.ts
```diff
@@ -1,14 +1,14 @@
 export enum CellNavigationMode {
-  NONE = 'none',
-  CHANGE_ROW = 'changeRow',
-  LOOP_OVER_ROW = 'loopOverRow'
+  none = 'none',
+  changeRow = 'changeRow',
+  loopOverRow = 'loopOverRow'
 }
 
 export enum UpdateActions {
```
amanmahajan7, line 7: I wonder if we should use `type UpdateActions='COPY_PASTE' | 'CELL_DRAG' | 'COLUMN_FILL' | 'CELL_UPDATE'` instead of enum. We can change in a separate PR

stories/demos/Grouping.tsx
```diff
@@ -4,7 +4,7 @@ import Select, { components, ValueType, OptionsType, Props as SelectProps } from
 import { SortableContainer, SortableElement } from 'react-sortable-hoc';
 import faker from 'faker';
 
-import DataGrid, { Column, Row, SelectColumn } from '../../src';
+import DataGrid, { Column, SelectColumn } from '../../src';
```
amanmahajan7, line 7: Interesting TS did not throw an error before. One is component and other is a type

stories/demos/components/Editors/SelectEditor.tsx
```diff
@@ -9,7 +9,7 @@ interface SelectEditorProps {
   menuPortalTarget: Element;
 }
 
-export function SelectEditor<R>({ value, onChange, options, rowHeight, menuPortalTarget }: SelectEditorProps) {
+export function SelectEditor({ value, onChange, options, rowHeight, menuPortalTarget }: SelectEditorProps) {
```
amanmahajan7, line 12: Is there an eslint rule to catch unused types?

tsconfig.eslint.json
```diff
@@ -0,0 +1,9 @@
+{
```
amanmahajan7, line 1: Why do we need it? Does `"eslint": "eslint --ext mjs,ts,tsx --max-warnings 0 -f codeframe --cache --color src stories jest tools",` not work?

amanmahajan7 (commented):
One Q otherwise looks good

stories/demos/components/Editors/SelectEditor.tsx
nstepien: Yes, `@typescript-eslint/no-unused-vars` caught this one.

tsconfig.eslint.json
nstepien: This one's for the `parserOptions.project` setting. Without this the type-aware typescript-eslint rules can't work.

amanmahajan7 (approved):

PR #2152, opened 2020-09-11
Enable the prefer-nullish-coalescing rule

nstepien:
WDYT?
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md

### DIFF
--- a/stories/demos/components/Toolbar/Toolbar.tsx
+++ b/stories/demos/components/Toolbar/Toolbar.tsx
@@ -2,32 +2,19 @@ import React, { PropsWithChildren } from 'react';
 import './rdg-toolbar.less';
 
 interface Props {
-  onAddRow?: (arg: { newRowIndex: number }) => void;
-  onToggleFilter?: () => void;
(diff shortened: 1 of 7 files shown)


stories/demos/components/Toolbar/Toolbar.tsx
```diff
@@ -2,32 +2,19 @@ import React, { PropsWithChildren } from 'react';
 import './rdg-toolbar.less';
 
 interface Props {
-  onAddRow?: (arg: { newRowIndex: number }) => void;
-  onToggleFilter?: () => void;
-  enableFilter?: boolean;
+  onAddRow: (arg: { newRowIndex: number }) => void;
```
amanmahajan7, line 5: I wonder if we need to keep the `Toolbar` component. We can probably add the button directly on the parent component

amanmahajan7 (approved):
Looks like a good rule 👍

PR #2153, opened 2020-09-14
rdg height was fixed as 350px but it's not useful

sugitlab:
rdg height was fixed as 350px but height and width should be responsible in many cases, so it should be changed to 100%. I hope this change will be accepted, because I love this grid but I always make another wrapped css only for this changes.

### DIFF
--- a/style/core.less
+++ b/style/core.less
@@ -8,7 +8,7 @@
   contain: strict;
   contain: size layout style paint;
   content-visibility: auto;
-  height: 350px;
+  height: 100%;
   border: 1px solid @borderColor;
   box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


nstepien:
`height: 100%;` doesn't work well in many cases, it depends on the container's style.
You can set a `className` prop, and set the height in css, or just set a `style={{ height: ... }}` prop as well.

sugitlab:
Yes, I always use className instead. Thank you. I understand your thought.

karthikeyan676:
Can you try this. This works for me. 
            
            style={{ height: "100%", contain: "none" }}

PR #2155, opened 2020-09-14
Add a few more css variables, prepare dark mode support

nstepien:


style/cell.less
```diff
@@ -4,8 +4,8 @@
   position: absolute;
   height: inherit;
   padding: 0 8px;
-  border-right: 1px solid @borderColor;
-  border-bottom: 1px solid @borderColor;
+  border-right: 1px solid var(--border-color);
```
amanmahajan7, line 7: I wonder if we should use `--rdg` prefix? I am fine with either

amanmahajan7 (approved):
👍

style/cell.less
nstepien: It is scoped under `.rdg`, so it's probably fine I think.

PR #2177, opened 2020-10-19
Downgrade build modules to ES2019, add cjs and mjs bundles

nstepien:


rollup.config.mjs
```diff
@@ -0,0 +1,25 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+
+export default {
+  input: './lib/index.js',
+  output: [{
+    file: './lib/bundle.mjs',
```
nstepien, line 6: Node.js doesn't support importing esm files without a file extension, and TS doesn't support adding file extension to emitted files, so we still need an mjs bundle.
https://github.com/microsoft/TypeScript/issues/16577

package.json
```diff
@@ -12,9 +12,16 @@
...
   },
-  "main": "lib/index.js",
-  "module": "lib/index.js",
-  "types": "lib/index.d.ts",
+  "exports": {
+    "node": {
+      "import": "./lib/bundle.mjs",
+      "require": "./lib/bundle.cjs"
+    }
+  },
```
nstepien, line 20: This works for Node 12 (lts) and Node 14 (current):
https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_nested_conditions
https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_nested_conditions

package.json
```diff
@@ -12,9 +12,16 @@
...
+  "browser": "./lib/index.js",
+  "main": "./lib/bundle.cjs",
+  "module": "./lib/index.js",
```
nstepien, line 23: `"main"` will work for Node 10 (maintenance).
`"browser"` and `"module"` are ignored by Node.
https://nodejs.org/docs/latest-v10.x/api/esm.html
https://github.com/nodejs/Release#release-schedule

webpack and rollup should pick up `"browser"` and `"module"` before `"main"` by default.
https://webpack.js.org/configuration/resolve/#resolvemainfields
https://github.com/rollup/plugins/tree/master/packages/node-resolve#mainfields

rollup.config.mjs
```diff
@@ -0,0 +1,27 @@
...
+    sourcemap: true
+  }, {
+    file: './lib/bundle.cjs',
+    format: 'cjs',
+    preferConst: true,
+    sourcemap: true,
+    interop: false
+  }],
+  external: [
+    'clsx',
```
amanmahajan7, line 19: Is `clsx` external?

amanmahajan7 (commented):
Looks good

rollup.config.mjs
nstepien: What this means is that `clsx` won't be included in the bundle, and it will be imported at runtime.
image
image

amanmahajan7 (approved):

PR #2179, opened 2020-10-20
Run Node 15 in CI

nstepien:
https://nodejs.org/en/blog/release/v15.0.0/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,13 +11,13 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
-        node-version: '14.x'
+        node-version: '15.x'
     - uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -11,13 +11,13 @@ jobs:
...
       with:
-        node-version: '14.x'
+        node-version: '15.x'
     - uses: actions/cache@v2
       with:
         path: ~/.npm
         key: npm-${{ hashFiles('package.json') }}
     - name: npm install
-      run: npm i
+      run: npm i --legacy-peer-deps
```
nstepien, line 20: npm 7 wants matching peerDeps, but we're on React 17 and other dependencies don't list it in their peerDeps.
So for now we'll need this flag.

amanmahajan7 (approved):

PR #2181, opened 2020-10-20
Focus selected cell on every update

amanmahajan7:
When a selected cell is clicked then the formatter looses focus. This change fixes it

### DIFF
--- a/src/hooks/useFocusRef.tsx
+++ b/src/hooks/useFocusRef.tsx
@@ -5,7 +5,7 @@ export function useFocusRef<T extends HTMLOrSVGElement>(isCellSelected: boolean 
   useLayoutEffect(() => {
     if (!isCellSelected) return;
     ref.current?.focus({ preventScroll: true });
-  }, [isCellSelected]);
+  });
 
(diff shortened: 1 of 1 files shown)


src/hooks/useFocusRef.tsx
```diff
@@ -5,7 +5,7 @@ export function useFocusRef<T extends HTMLOrSVGElement>(isCellSelected: boolean
   useLayoutEffect(() => {
     if (!isCellSelected) return;
     ref.current?.focus({ preventScroll: true });
-  }, [isCellSelected]);
+  });
```
nstepien, line 8: Won't that always try and focus the cell? Even when we don't interact with the grid?

amanmahajan7: Only if the formatter is rerendered which should not happen if the grid props have not changed. But in the [TreeView ](https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--tree-view) example it steals the focus when `Allow Delete` is clicked. Need to think more. Great catch

PR #2184, opened 2020-10-22
Use the new jsx runtime, bump peerDeps

nstepien:
https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "declaration": true,
     "forceConsistentCasingInFileNames": true,
     "inlineSources": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
nstepien, line 7: Now that we support the `exports` field in `package.json`, we could have a prod bundles using `react-jsx`, and dev bundles using `react-jsxdev`. 🤔

amanmahajan7 (approved):

PR #2185, opened 2020-10-22
Storybook 6

nstepien:
https://storybook.js.org/docs/react/writing-stories/introduction
https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy


package.json
```diff
@@ -35,7 +35,7 @@
     "*.less"
   ],
   "scripts": {
-    "start": "start-storybook --quiet -p 6006",
+    "start": "start-storybook --quiet --no-dll -p 6006",
```
nstepien, line 38: https://github.com/storybookjs/storybook/issues/12408

amanmahajan7 (commented):
One comment otherwise looks good. I wonder if we should look into adding args so we can make the examples interactive

amanmahajan7 (approved):

PR #2190, opened 2020-10-27
rowKey -> rowKeyGetter

nstepien:
Pros:
- Rows can now be of any type, and we type-check that the row key is a `React.Key`.
- It's a lot more flexible, it doesn't have to be a value held in the row data for example.
- -1 generic

Cons:
- Users must implement the function.
  - Implementations should be trivial, I think this is a non-issue
  - It's still optional, unless row selection is enabled.
- The function should be static to avoid re-renders
  - A static function declaration outside the functional component should work, at worst users have to use `useCallback`.
  - Class members are still good.

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -88,7 +88,7 @@ export interface DataGridProps<R, SR = unknown> extends SharedDivProps {
    * Bottom horizontal scroll bar can move the row left / right. Or a customized row renderer can be used to disabled the scrolling support.
    */
   summaryRows?: readonly SR[];
-  /** The primary key property of each row */
+  /** The getter should return a unique key for each row */
   rowKeyGetter?: (row: R) => React.Key;
   /**
    * Callback called whenever row data is updated
@@ -346,7 +346,7 @@ function DataGrid<R, SR>({
     };
 
     return eventBus.subscribe('SelectRow', handleRowSelectionChange);
-  }, [eventBus, isGroupRow, onSelectedRowsChange, rowKeyGetter, rows, selectedRows]);
+  });
 
   useEffect(() => {
     return eventBus.subscribe('SelectCell', selectCell);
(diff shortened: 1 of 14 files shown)


src/DataGrid.tsx
```diff
@@ -308,44 +308,45 @@ function DataGrid<R, K extends keyof R, SR>({
...
-        newSelectedRows.delete(rowId);
+        newSelectedRows.delete(rowKey);
         lastSelectedRowIdx.current = -1;
       }
 
       onSelectedRowsChange(newSelectedRows);
     };
 
     return eventBus.subscribe('SelectRow', handleRowSelectionChange);
-  }, [eventBus, isGroupRow, onSelectedRowsChange, rowKey, rows, selectedRows]);
```
nstepien, line 348: `isGroupRow` changes on each render, so this was unnecessary.

src/DataGrid.tsx
```diff
@@ -857,11 +858,11 @@ function DataGrid<R, K extends keyof R, SR>({
...
       let isRowSelected = false;
-      if (rowKey !== undefined) {
-        const rowId = row[rowKey];
-        isRowSelected = selectedRows?.has(rowId) ?? false;
-        if (typeof rowId === 'string' || typeof rowId === 'number') {
-          key = rowId;
+      if (typeof rowKeyGetter === 'function') {
+        const rowKey = rowKeyGetter(row);
+        isRowSelected = selectedRows?.has(rowKey) ?? false;
+        if (typeof rowKey === 'string' || typeof rowKey === 'number') {
```
nstepien, line 864: Should we still manually typecheck this?

amanmahajan7: Probably not as we are using `React.Key`

amanmahajan7 (commented):
Nice cleanup

amanmahajan7 (approved):

PR #2196, opened 2020-10-31
Disable cell navigation while editing, add column.editorOptions.onNavigation

nstepien:


amanmahajan7:
Using tab during editing is quite common and I am not sure if we can completely disable it. Having an option to control the editing would be nice though

src/utils/columnUtils.ts
```diff
@@ -189,3 +189,16 @@ export function getColumnScrollPosition<R, SR>(columns: readonly CalculatedColum
 
   return 0;
 }
+
+/**
+ * By default, the following navigation keys are enabled while an editor is open, under specific conditions:
+ * - Tab:
+ *   - The editor must be an <input>, a <textarea>, or a <select> element.
+ *   - The editor element must be the only immmediate child of the editor container.
```
amanmahajan7, line 197: Would it break if input has a container element like a label?
div.rdg-editor-container > label > input

src/DataGrid.tsx
```diff
@@ -722,6 +723,10 @@ function DataGrid<R, SR>({
   }
 
   function navigate(event: React.KeyboardEvent<HTMLDivElement>) {
+    if (selectedPosition.mode === 'EDIT') {
+      const onNavigation = columns[selectedPosition.idx].editorOptions?.onNavigation ?? onEditorNavigation;
```
amanmahajan7, line 727: We can probably add an example of custom `onEditorNavigation` like 
- Navigate left if the cursor is at position 0
- Navigate right if the cursor is at position selection.length

amanmahajan7 (commented):
Looks good overall

amanmahajan7 (approved):

PR #2197, opened 2020-10-31
Fix crash when rows are undefined

nstepien:
Rows can be `undefined`, so `copiedCell?.row === row` can be `true` when `copiedCell` is nullish.

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -842,7 +842,7 @@ function DataGrid<R, SR>({
           onRowClick={onRowClick}
           rowClass={rowClass}
           top={top}
-          copiedCellIdx={copiedCell?.row === row ? columns.findIndex(c => c.key === copiedCell.columnKey) : undefined}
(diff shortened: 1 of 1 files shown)


src/DataGrid.tsx
```diff
@@ -842,7 +842,7 @@ function DataGrid<R, SR>({
           onRowClick={onRowClick}
           rowClass={rowClass}
           top={top}
-          copiedCellIdx={copiedCell?.row === row ? columns.findIndex(c => c.key === copiedCell.columnKey) : undefined}
+          copiedCellIdx={copiedCell !== null && copiedCell.row === row ? columns.findIndex(c => c.key === copiedCell.columnKey) : undefined}
```
amanmahajan7, line 845: We are looping through rows and rows type is `readonly R[];`. I am not sure when can row be undefined?

nstepien: `R` can be `undefined` like in our million cells demo.

amanmahajan7 (approved):

PR #2200, opened 2020-11-02
Only make scrollToRow smooth

nstepien:
`scroll-behavior: smooth;` made all scrolling smooth, including keyboard navigation, which is probably not desirable, especially when holding arrow keys.
I've made it so only `scrollToRow` is smooth.

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -354,7 +354,10 @@ function DataGrid<R, SR>({
     scrollToRow(rowIdx: number) {
       const { current } = gridRef;
       if (!current) return;
-      current.scrollTop = rowIdx * rowHeight;
+      current.scrollTo({
+        top: rowIdx * rowHeight,
+        behavior: 'smooth'
+      });
     },
     selectCell
   }));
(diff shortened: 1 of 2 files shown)


src/DataGrid.tsx
```diff
@@ -354,7 +354,10 @@ function DataGrid<R, SR>({
     scrollToRow(rowIdx: number) {
       const { current } = gridRef;
       if (!current) return;
-      current.scrollTop = rowIdx * rowHeight;
+      current.scrollTo({
```
amanmahajan7, line 357: Does it work in safari 
https://caniuse.com/?search=scrollTo

nstepien: https://developer.mozilla.org/en-US/docs/Web/API/ScrollToOptions
`ScrollToOptions` is supported, but the smooth option isn't supported yet so it'll be ignored.

>Last modified: Aug 21, 2019

The mdn page hasn't been updated for over a year so it's probably safe to use this.

amanmahajan7 (approved):

PR #2208, opened 2020-11-03
Add integration tests

nstepien:


test/column/cellClass.test.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+test('cellClass is undefined', () => {
+  const columns: readonly Column<Row>[] = [{
+    key: 'id',
+    name: 'ID',
+    cellClass: undefined
+  }];
+  setup({ columns, rows });
+  expect(getAllCellClassNames()).toStrictEqual([
```
amanmahajan7, line 22: Wdyt about using https://github.com/testing-library/jest-dom#tohaveclass?

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-assertion

nstepien: Sounds good. I've had this blog post bookmarked for a while now, I should get to it at some point.

nstepien: image
The error messages can be buggy 🤔

nstepien:
Not a fan that `toHaveTextContent` does not have an `exact` option like `toHaveClass`.
https://github.com/testing-library/jest-dom#tohavetextcontent

amanmahajan7 (approved):

PR #2211, opened 2020-11-05
Remove EventBus

amanmahajan7:
https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often

This is only recommended if there is no other solution and I think it is one of those situations where other patterns do not work. Tested and there was no performance impact


src/hooks/useLatest.ts
```diff
@@ -0,0 +1,10 @@
+import { useRef, useEffect } from 'react';
+
+export function useLatest<T>(value: T) {
```
nstepien, line 3:
```suggestion
// https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often
export function useLatest<T>(value: T) {
```

src/hooks/useLatest.ts
```diff
@@ -0,0 +1,10 @@
...
+  const ref = useRef<T>(value);
+  useEffect(() => {
```
nstepien, line 5: I wonder if using `useLayoutEffect` would a bit safer as it's the "sync" version of `useEffect`.

src/hooks/useLatest.ts
nstepien, line 3: Do we only use it for functions?
Maybe we could add a `useCallback` so we can skip the `current` part.

src/GroupRow.tsx
```diff
@@ -1,9 +1,26 @@
-import React, { memo } from 'react';
+import React, { memo, MutableRefObject } from 'react';
 import clsx from 'clsx';
-import { GroupRowRendererProps } from './types';
+
+import { CalculatedColumn, Position, SelectRowEvent } from './types';
```
nstepien, line 4:
```suggestion
import { CalculatedColumn, Position, SelectRowEvent, Omit } from './types';
```

nstepien (changes requested):

src/hooks/useLatest.ts
amanmahajan7: I am not 100% sure. React docs uses `useEffect` in the example. Also these function are only called when user performs an action (apart from `selectCell` which can be called using gridRef). I think we can use `useEffect` for now and revisit if we find bugs.

src/Cell.tsx
```diff
@@ -39,7 +40,7 @@ function Cell<R, SR>({
   );
 
   function selectCell(openEditor?: boolean) {
```
nstepien, line 42: Isn't this the `wrapper` instead? 🤔

src/hooks/useLatestFunc.ts
```diff
@@ -0,0 +1,14 @@
+import { useRef, useEffect, useCallback } from 'react';
+
+// https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function useLatestFunc<T extends (...args: any[]) => any>(fn: T) {
+  const ref = useRef<T>(fn);
```
nstepien, line 6:
```suggestion
  const ref = useRef(fn);

```

src/types.ts
```diff
@@ -145,9 +144,10 @@ export interface CellRendererProps<TRow, TSummaryRow = unknown> extends Omit<Rea
   isDraggedOver: boolean;
   isCellSelected: boolean;
   isRowSelected: boolean;
-  eventBus: EventBus;
   dragHandleProps?: Pick<React.HTMLAttributes<HTMLDivElement>, 'onMouseDown' | 'onDoubleClick'>;
   onRowClick?: (rowIdx: number, row: TRow, column: CalculatedColumn<TRow, TSummaryRow>) => void;
+  selectCell: (position: Position, enableEditor?: boolean) => void;
+  selectRow: ({ rowIdx, checked, isShiftClick }: SelectRowEvent) => void;
```
nstepien, line 150:
```suggestion
  selectRow: (selectRowEvent: SelectRowEvent) => void;
```

src/types.ts
```diff
@@ -158,26 +158,13 @@ export interface RowRendererProps<TRow, TSummaryRow = unknown> extends Omit<Reac
...
-  childRows: readonly TRow[];
-  rowIdx: number;
-  top: number;
-  level: number;
-  selectedCellIdx?: number;
-  isExpanded: boolean;
-  isRowSelected: boolean;
-  eventBus: EventBus;
+  selectCell: (position: Position, enableEditor?: boolean) => void;
+  selectRow: ({ rowIdx, checked, isShiftClick }: SelectRowEvent) => void;
```
nstepien, line 167:
```suggestion
  selectRow: (selectRowEvent: SelectRowEvent) => void;
```

nstepien (commented):
Just a few nitpicks, wdyt?

nstepien (approved):

src/Cell.tsx
amanmahajan7: Changed

nstepien (approved):

PR #2214, opened 2020-11-07
Publish bundles instead of modules, use babel only for transpilation

nstepien:
**Pros**:
- Babel's output is [better](https://babeljs.io/blog/2020/05/25/7.10.0#better-tree-shaking-for-react-components-11428httpsgithubcombabelbabelpull11428) and less [buggy](https://github.com/microsoft/TypeScript/issues/40774) than TS's.
- Can publish with the new React runtime sooner.
- More flexibility, can use fancy babel/rollup plugins:
  - `clsx` calls are optimized, users don't need to add an extra babel plugin.
  - We could do try doing css-in-js.
  - I'm thinking about perhaps adding a light terser config, or perhaps try the closure compiler. Not sure if it's worth it.
  - https://babeljs.io/docs/en/babel-plugin-transform-react-constant-elements 👀 
- Smaller `npm` footprint:
  - No mixed bundles/modules+sourcemaps in published files.
  - Code that's not exported or unused by exports will be tree-shaken.

We still use `tsc` to output declaration files.


.eslintrc.js
```diff
@@ -483,7 +483,7 @@ const rules = {
   '@typescript-eslint/consistent-indexed-object-style': 1,
   '@typescript-eslint/consistent-type-assertions': [2, { assertionStyle: 'as', objectLiteralTypeAssertions: 'never' }],
   '@typescript-eslint/consistent-type-definitions': [1, 'interface'],
-  '@typescript-eslint/consistent-type-imports': [1, { prefer: 'no-type-imports' }],
+  '@typescript-eslint/consistent-type-imports': 1,
```
nstepien, line 486: I've tweaked this so we'll now use `type` imports, this is to help babel elide type imports, otherwise it cannot know to remove some imports, especially in files like `src/index.ts`.

babel.config.json
```diff
@@ -4,8 +4,14 @@
       "bugfixes": true,
       "shippedProposals": true,
       "corejs": 3,
-      "useBuiltIns": "entry"
-    }]
+      "useBuiltIns": "entry",
+      "include": [
+        "@babel/proposal-nullish-coalescing-operator",
+        "@babel/proposal-optional-chaining"
```
nstepien, line 10: Fun fact: with our browserslist config, and our code, aside from these two transform plugins, `@babel/preset-env` does nothing.

babel.config.json
```diff
@@ -4,8 +4,14 @@
...
+      ]
+    }],
+    ["@babel/react", { "useSpread": true }],
```
nstepien, line 13: I'll change this to use the new runtime in #2184

babel.config.json
```diff
@@ -4,8 +4,14 @@
...
+    "@babel/typescript"
   ],
   "plugins": [
     ["@babel/transform-runtime", { "useESModules": true }],
```
nstepien, line 17: We might have to add `@babel/runtime` in our `dependencies` in the future, but currently no helpers get imported in our bundles.

rollup.config.mjs
```diff
@@ -15,13 +18,13 @@ export default {
     sourcemap: true,
     interop: false
   }],
-  external: [
-    'clsx',
-    'react',
-    'react-dom'
-  ],
+  external: id => !id.startsWith('.') && !isAbsolute(id),
```
nstepien, line 21: This will mark all non-relative imports as external, in other words all npm dependencies.

tsconfig.all.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "emitDeclarationOnly": false,
```
amanmahajan7, line 4: Do we need this if we have `noEmit`?

rollup.config.mjs
```diff
@@ -15,13 +18,17 @@ export default {
...
   plugins: [
-    sourcemaps(),
-    nodeResolve()
+    babel({
+      babelHelpers: 'runtime',
+      skipPreflightCheck: true,
```
amanmahajan7, line 25: Not sure if it is worth enabling it? Does it slow down the build by much?

amanmahajan7 (commented):
A few questions. LGTM

babel.config.json
```diff
@@ -1,11 +1,18 @@
 {
   "presets": [
     ["@babel/env", {
+      "loose": true,
```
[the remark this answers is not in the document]
nstepien: Original code:
```ts
ref.current?.focus({ preventScroll: true });
```
With `loose` enabled:
```js
(_ref$current = ref.current) == null ? void 0 : _ref$current.focus({
```
With `loose` disabled:
```js
(_ref$current = ref.current) === null || _ref$current === void 0 ? void 0 : _ref$current.focus({
```

In non-loose mode, the value is explicitely checked for both `null` and `undefined`, as it needs to work with `document.all`, but we don't use `document.all` anywhere so it's safe.
image

It might matter in the future for new js syntax features, but `loose` will only break edge cases so I think it's fine.

tsconfig.all.json
nstepien: Yes unfortunately.

```
> react-data-grid@7.0.0-canary.30 typecheck
> tsc -p tsconfig.all.json

tsconfig.all.json:4:5 - error TS5053: Option 'emitDeclarationOnly' cannot be specified with option 'noEmit'.

4     "noEmit": true
      ~~~~~~~~


Found 1 error.
```

rollup.config.mjs
nstepien: It builds in 1s in both cases, but I ran into issues before so I had to add this. I'll remove it now.

amanmahajan7 (approved):

PR #2248, opened 2020-12-07
remove memoization from Cell export statement

### LINKED ISSUE #2247: Necessary cell rerendering does not happen
Cells are not rerendered in the following scenario:

- initialize the grid with 3 columns and an empty row
- write a paste handler that maps the clipboard data from eg excel to the grid rows
- copy from Excel into the clipboard (1 row with three values)
- click into the first cell and paste

The selected cell shows the pasted value but the other two cells are not rerendered immediately. You need to focus them to force the rerender and show the row value

gernotkogler:
since it causes missing cell rerendering in certain scenarios.

Fixes #2247

### DIFF
--- a/src/Cell.tsx
+++ b/src/Cell.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, memo, useRef } from 'react';
+import { forwardRef, useRef } from 'react';
 import clsx from 'clsx';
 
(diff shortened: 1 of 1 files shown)


amanmahajan7:
Please check https://github.com/adazzle/react-data-grid/issues/2247#issuecomment-743391994

PR #2256, opened 2020-12-17
Migrate to css-in-js with linaria

nstepien:
Docs: https://github.com/callstack/linaria

Rollup is set up so we inject the styles, that way users don't have to worry about manually importing a stylesheet.

- [x] Investigate linaria configs
  - https://github.com/callstack/linaria/blob/master/docs/CONFIGURATION.md
  - Use `classNameSlug` to guarantee different generated class names after each rdg version update.
- [x] Investigate linaria webpack loader configs
  - https://github.com/callstack/linaria/blob/master/docs/BUNDLERS_INTEGRATION.md#webpack
- [x] Investigate linaria rollup plugin configs
  - https://github.com/callstack/linaria/blob/master/docs/BUNDLERS_INTEGRATION.md#rollup
- [x] Investigate postcss rollup plugin configs
  - https://github.com/egoist/rollup-plugin-postcss
- [x] Remove remaining references to `dist/`
- [x] Remove `less`
  - [x] Replace story styles
  - [x] Update `sideEffects` in `package.json`

---

I tried using [`polished`](https://polished.js.org/) with its [babel plugin](https://github.com/styled-components/babel-plugin-polished), but that didn't transpile everything away, for example:
```js
// transpiled away :^)
darken(0.2, '#fff');

// not transpiled away :^(
const bg = '#fff';
darken(0.2, bg);
````

I decided to use `hsl()` instead, close enough.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -18,31 +18,27 @@
...
   "files": [
-    "dist",
     "lib"
   ],
   "sideEffects": [
-    "*.css",
     "*.less"
   ],
   "scripts": {
-    "start": "start-storybook --quiet --no-dll -p 6006",
```
nstepien, line 37: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-dll-flags

src/formatters/ToggleGroupFormatter.tsx
```diff
@@ -20,12 +21,12 @@ export function ToggleGroupFormatter<R, SR>({
...
     <span
       ref={cellRef}
-      className="rdg-group-cell-content"
+      className={groupCellContentClassname}
       tabIndex={-1}
       onKeyDown={handleKeyDown}
     >
       {groupKey}
-      <svg viewBox="0 0 14 8" width="14" height="8" className="rdg-caret">
+      <svg viewBox="0 0 14 8" width="14" height="8" className={caretClassname}>
```
amanmahajan7, line 29: I think where possible, we can move the styles in the file that use them. This way we can encapsulate styles and I think that is what CSS-in-JS libraries promote. Wdyt?

src/style/editor.ts
```diff
@@ -1,9 +1,12 @@
-.rdg-editor-container {
+import { css } from '@linaria/core';
+
+const editorContainer = css`
```
amanmahajan7, line 3: For example this style can be move to `EditorContainer.tsx` and the next to `TextEditor`

src/SummaryCell.tsx
```diff
@@ -16,10 +17,10 @@ function SummaryCell<R, SR>({
 }: SummaryCellProps<R, SR>) {
   const { summaryFormatter: SummaryFormatter, summaryCellClass } = column;
   const className = clsx(
```
amanmahajan7, line 19: I wonder if we should create a helper component that can render the default styles for example

```
export function DefaultCell({ column, className, props }) {
className = clsx(
    cellClassname,
    {
      [cellFrozenClassname]: column.frozen,
      [cellFrozenLastClassname]: column.isLastFrozenColumn
    },
    className
  );

  return (
    <div
      role="gridcell"
      aria-colindex={column.idx + 1}
      className={className}
      {...props}
    >
     {children}
    </div>
  );
}
```

Wdyt?

amanmahajan7 (commented):
Looks really good. A few discussion items

src/Cell.tsx
```diff
@@ -1,8 +1,9 @@
 import { forwardRef, memo } from 'react';
 import clsx from 'clsx';
 
-import type { CellRendererProps } from './types';
+import { cellClassname, cellCopiedClassname, cellDraggedOverClassname, cellDragHandleClassname, cellFrozenClassname, cellFrozenLastClassname, cellSelectedClassname } from './style';
```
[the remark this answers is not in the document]
nstepien: I'd rather we keep individual variables for better tree shaking and no object access.

src/Row.tsx
```diff
@@ -33,10 +34,10 @@ function Row<R, SR = unknown>({
   }
 
   className = clsx(
-    'rdg-row',
+    rowClassname,
     `rdg-row-${rowIdx % 2 === 0 ? 'even' : 'odd'}`, {
```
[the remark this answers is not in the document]
nstepien: No? even/odd rows don't have a default style, and the generrated class names are for us only, it's not meant to be used by users.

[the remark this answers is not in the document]
amanmahajan7: ah okay

src/SummaryCell.tsx
nstepien: It's more stuff for react to render, not sure 🤔

amanmahajan7 (approved):
Looks great 👍

Issue #2173, opened 2020-10-09
--row-width is not correctly calculated


kimim:
--row-width is about 13px larger, which makes the horizontal scrollbar appear.

kimim:
Example:
http://40.73.112.202:3001

amanmahajan7:
The above link is not working

kimim:
sorry the server was restarted. Here is the new link: http://52.130.91.102:3001

FPDK:
@kimim did you figure out a solution?

FPDK:
The calculation of --row-width seems to cause an issue in Firefox (85.0b3 (64-bit)). It works fine in Chrome and Edge.

Chrome:
![image](https://user-images.githubusercontent.com/1729717/102688717-d10e4400-41f8-11eb-8713-4ac00432b27f.png)

Edge:
![image](https://user-images.githubusercontent.com/1729717/102688722-d79cbb80-41f8-11eb-86c9-c361707e425f.png)

Firefox:
![image](https://user-images.githubusercontent.com/1729717/102688735-e5ead780-41f8-11eb-98cf-bd41d3e00193.png)

kimim:
Not yet. But thank you for the solution!

PR #2259, opened 2020-12-19
Hotfix unnecessary horizontal scrollbar in Firefox

FPDK:
See issue #2173

If one or more column width's are auto calculated, the elements will overflow in Firefox, and thus create an unnecessary scrollbar.
This fix subtracts 16 pixels from the unallocatedWidth, before the column widths are calculated.

This is a bad solution as it assumes the scrollbar is always 16 pixels wide in Firefox.

### DIFF
--- a/src/hooks/useViewportColumns.ts
+++ b/src/hooks/useViewportColumns.ts
@@ -83,8 +83,14 @@ export function useViewportColumns<R, SR>({
       // Sort other columns last:
       return 0;
     });
+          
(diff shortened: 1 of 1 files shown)


nstepien:
Hey, thanks for the PR, I ended up fixing it differently in another PR of mine: https://github.com/adazzle/react-data-grid/pull/2272/commits/348b6392376e1f4110935ceaaf4d4039a2b66598#diff-2dbc2c9db9fe9009d9fe02a7a37f929629e684c70a2d02fff405bb30499ccb24

PR #2272, opened 2021-01-11
Use grid layout for rows, split column metrics from compute columns

nstepien:
What's the fun in making a grid component if you don't use `display: grid;`?

image


src/hooks/useViewportColumns.ts
```diff
@@ -84,51 +78,84 @@ export function useViewportColumns<R, SR>({
...
     }
 
-    return {
-      columns: calculatedColumns,
-      lastFrozenColumnIndex,
-      totalFrozenColumnWidth,
-      totalColumnWidth: totalWidth,
-      groupBy
+    const layoutCssVars: Record<string, string> = {
+      '--template-columns': templateColumns
```
nstepien, line 149: This is used for the `grid-template-columns` css property in all rows now.
That way we we the template on the root, and we don't have to update all rows/cells (in js).

src/hooks/useViewportColumns.ts
```diff
@@ -84,51 +78,84 @@ export function useViewportColumns<R, SR>({
...
     };
-  }, [columnWidths, defaultFormatter, defaultResizable, defaultSortable, minColumnWidth, rawColumns, rawGroupBy, viewportWidth]);
+
+    for (let i = 0; i <= lastFrozenColumnIndex; i++) {
+      const column = columns[i];
+      layoutCssVars[`--frozen-left-${column.key}`] = `${columnMetrics.get(column)!.left}px`;
```
nstepien, line 154: The frozen columns still need to know their `left` position otherwise `position: sticky` won't work.

src/types.ts
```diff
@@ -58,15 +58,19 @@ export interface Column<TRow, TSummaryRow = unknown> {
 
 export interface CalculatedColumn<TRow, TSummaryRow = unknown> extends Column<TRow, TSummaryRow> {
   idx: number;
-  width: number;
-  left: number;
   resizable: boolean;
   sortable: boolean;
-  isLastFrozenColumn?: boolean;
+  frozen: boolean;
+  isLastFrozenColumn: boolean;
```
nstepien, line 64: They're always booleans now, reducing megamorphism.
https://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html

src/utils/columnUtils.ts
```diff
@@ -1,37 +1,3 @@
-import type { CalculatedColumn } from '../types';
-
-export function getColumnScrollPosition<R, SR>(columns: readonly CalculatedColumn<R, SR>[], idx: number, currentScrollLeft: number, currentClientWidth: number): number {
```
nstepien, line 3: This did way too much unnecessary work.

src/utils/index.ts
```diff
@@ -8,3 +10,9 @@ export function assertIsValidKeyGetter<R>(keyGetter: unknown): asserts keyGetter
     throw new Error('Please specify the rowKeyGetter prop to use selection');
   }
 }
+
+export function getCellStyle<R, SR>(column: CalculatedColumn<R, SR>): React.CSSProperties {
+  return column.frozen
+    ? { left: `var(--frozen-left-${column.key})` }
+    : { gridColumnStart: column.idx + 1 };
```
nstepien, line 17: This "picks" the correct grid column from the `grid-template-columns`, regardless of its DOM position.

style/cell.less
```diff
@@ -1,8 +1,6 @@
 .rdg-cell {
   contain: strict;
   contain: size layout style paint;
-  position: absolute;
-  height: inherit;
```
nstepien, line 5: Height is set by the grid layout now.

style/row.less
```diff
@@ -1,11 +1,13 @@
 .rdg-row {
   contain: strict;
   contain: size layout style paint;
-  display: flex;
```
nstepien, line 4: We know that flex+sticky position cells can be buggy in Firefox, but I wonder if grid+sticky works fine or not.

style/row.less
```diff
@@ -1,11 +1,13 @@
...
+  display: grid;
+  grid-template-rows: var(--row-height);
```
nstepien, line 5: This basically means that we only have 1 row in this grid.

src/types.ts
```diff
@@ -58,15 +58,19 @@ export interface Column<TRow, TSummaryRow = unknown> {
 
 export interface CalculatedColumn<TRow, TSummaryRow = unknown> extends Column<TRow, TSummaryRow> {
   idx: number;
-  width: number;
```
amanmahajan7, line 61: I am not sure if any formatter/editor uses width/left or not.  What is the alternative if those values are needed? ref?

amanmahajan7 (commented):
Looks great overall

src/hooks/useViewportColumns.ts
```diff
@@ -84,51 +78,84 @@ export function useViewportColumns<R, SR>({
...
+        unassignedColumnsCount++;
+      } else {
+        width = clampColumnWidth(width, column, minColumnWidth);
+        allocatedWidth += width;
+        columnMetrics.set(column, { width, left: 0 });
+      }
+    }
+
+    const unallocatedWidth = viewportWidth - allocatedWidth;
     const unallocatedColumnWidth = Math.max(
```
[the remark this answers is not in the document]
nstepien: We don't, I've removed it now, we also don't need to `floor` it, now columns will always fill all the available space when possible.
For a before/after comparison, open the columns reordering story, and resize the browser, you'll see that with `Math.floor()` there can be some space between the last column and the scrollbar.

src/types.ts
nstepien: We could pass them in `editorProps` if it's ever needed.

amanmahajan7 (approved):

PR #2274, opened 2021-01-13
typescript@4.2

nstepien:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-2/
https://github.com/microsoft/TypeScript/issues/41601


test/setup.ts
```diff
@@ -1,11 +1,18 @@
-// @ts-expect-error
-window.ResizeObserver ??= function ResizeObserver(callback: () => void) {
-  callback();
-
-  return {
-    observe() {},
-    disconnect() {}
-  };
+import 'core-js/stable';
```
nstepien, line 1: TS complains without an import/export, so I added this. It doesn't hurt to have I guess.

```
test/setup.ts:1:1 - error TS1208: 'setup.ts' cannot be compiled under '--isolatedModules' because it is considered a global script file. Add an import, export, or an empty 'export {}' statement to make it a module.

1 window.ResizeObserver ??= class {
  ~~~~~~
```

src/hooks/useGridDimensions.ts
```diff
@@ -1,25 +1,12 @@
 import { useRef, useState, useLayoutEffect } from 'react';
 
-// https://github.com/microsoft/TypeScript/issues/37861
```
amanmahajan7, line 3: 👍

amanmahajan7 (approved):

PR #2275, opened 2021-01-14
:new: Add onRowChange method for a single row change

ekeuus:


ekeuus:
@nstepien Do you think this kind of an api would make sense? Not all users need/want the whole grid data out directly

nstepien:
We're thinking about doing something like
```js
onRowsChange(updatedRows, {
  indexes: [rowIdx], // will contain 1 or more indexes of the rows that have been updated
  column: column // column that has been updated/triggered an update on the rows
});
```
Would that work for you?

ekeuus:
> We're thinking about doing something like
> 
> ```js
> onRowsChange(updatedRows, {
>   indexes: [rowIdx], // will contain 1 or more indexes of the rows that have been updated
>   column: column // column that has been updated/triggered an update on the rows
> });
> ```
> 
> Would that work for you?

Yeah, that's even better since then we I can also detect which field I need to update instead of the whole row :)

ekeuus:
Should I implement it myself or is someone else already looking at it?

nstepien:
We'll look into it, probably tomorrow.

PR #2276, opened 2021-01-14
README: Add bundlephobia and type badges

nstepien:
Let's also add a code coverage badge once we have some real coverage.

### DIFF
--- a/browserslist
+++ b/browserslist
@@ -2,3 +2,4 @@ last 2 chrome versions
 last 2 edge versions
 last 2 firefox versions
 last 2 safari versions
+maintained node versions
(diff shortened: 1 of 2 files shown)


browserslist
```diff
@@ -2,3 +2,4 @@ last 2 chrome versions
 last 2 edge versions
 last 2 firefox versions
 last 2 safari versions
+maintained node versions
```
nstepien, line 5: I've also added this so we don't accidentally push js syntax that might break in jest tests ran with Node LTS versions.

amanmahajan7 (approved):
👍

PR #2277, opened 2021-01-14
Fix isAtBottom when browser is zoomed

stephenarra:
When zoomed, scrollTop is a decimal value, causing the equality check to fail.


amanmahajan7:
This is no longer needed as we add some buffer and created a separate infinite scrolling example
https://github.com/adazzle/react-data-grid/blob/main/stories/demos/InfiniteScrolling.tsx#L74

PR #2278, opened 2021-01-14
Augment onRowsChange with indexes and column data

nstepien:
cc @ekeuus


ekeuus:
I think it looks good to me! I can just easily create a wrapper to extract the changed column value now easily 😉

amanmahajan7 (approved):
Looks good 🚢

PR #2282, opened 2021-01-15
Genericize filters, make the api consistent with other rows

nstepien:
To help apply the new generic, I removed the generic defaults (`= unknown`/`= any`) until there were no type issues.
In some cases I did not add back these defaults, let's see if it'll cause any issues. Let me know if some of these should be added back.


amanmahajan7:
We won't need this change if we decide to move forward with removing filters
https://github.com/adazzle/react-data-grid/pull/2412

PR #2283, opened 2021-01-16
rm unused refs

nstepien:
`Cell` is now a stateless/hookless functional component!
FYI we removed the last usage of the internal cell ref in the [Editor2 PR](https://github.com/adazzle/react-data-grid/commit/ad19cc4fc7835585f706afd95170ab22970ace84#diff-6f5aa4063969484827f64cb234ad9aa341d702d6da7cd2b2414bf372d548703c) back in August!
`useCombinedRefs` will also be removed from the published bundles, smaller bundle size!


amanmahajan7 (approved):
👍

PR #2284, opened 2021-01-18
Minor type tweaks

nstepien:


src/DataGrid.tsx
```diff
@@ -267,7 +267,7 @@ function DataGrid<R, SR>({
     expandedGroupIds
   });
 
-  const hasGroups = groupBy.length > 0 && rowGrouper;
+  const hasGroups = groupBy.length > 0 && typeof rowGrouper === 'function';
```
nstepien, line 270: `hasGroups` is now always boolean, reducing type coercion.

src/hooks/useViewportColumns.ts
```diff
@@ -81,16 +83,15 @@ export function useViewportColumns<R, SR>({
     columns.forEach((column, idx) => {
       column.idx = idx;
 
-      if (idx === lastFrozenColumnIndex) {
```
nstepien, line 84: This didn't need to be checked on each iteration.

amanmahajan7 (approved):

PR #2289, opened 2021-01-19
Add basic keyboard navigation tests

nstepien:


src/DataGrid.tsx
```diff
@@ -375,6 +375,16 @@ function DataGrid<R, SR>({
     onExpandedGroupIdsChange(newExpandedGroupIds);
   }
 
+  // Tabbing into the grid should initiate keyboard navigation
```
amanmahajan7, line 378: 👍 We also need to scroll to the selected cell in case is not visible. Consider a case
- Select a cell
- Scroll the grid so cell is not visible
- Focus on any element outside the grid
- Tab into the grid
- Grid only scrolls when the selected position is changed

May be we can do something like 

```
const selectedPositionOnFocus = selectedPosition.idx === -1 ? { idx: 0, rowIdx: 0, mode: 'SELECT' } : { ...selectedPosition };
if (isCellWithinBounds(selectedPositionOnFocus)) {
    setSelectedPosition(selectedPositionOnFocus);
}
```

Wdyt?

amanmahajan7 (commented):
Good tests 👍

src/DataGrid.tsx
nstepien: Clicking anywhere on the grid triggers the `blur` event on the focus sink, which we then focus immediately after, triggering the focus handler, so that might not be desired.
We can prevent `blur` by `preventDefault()`ing the `mousedown` event, but that'll break other stuff as well.
Not sure if this is doable.

amanmahajan7 (approved):

PR #2294, opened 2021-01-22
Cleanup cell styles

amanmahajan7:


src/utils/index.ts
```diff
@@ -15,3 +18,12 @@ export function getCellStyle<R, SR>(column: CalculatedColumn<R, SR>): React.CSSP
...
     : { gridColumnStart: column.idx + 1 };
 }
+
+export function getCellClassname<R, SR>(column: CalculatedColumn<R, SR>): string {
+  return clsx(
+    `rdg-cell ${cell}`, {
+      [cellFrozenClassname]: column.frozen,
+      [cellFrozenLastClassname]: column.isLastFrozenColumn
+    }
+  );
```
nstepien, line 28:
```suggestion
export function getCellClassname<R, SR>(column: CalculatedColumn<R, SR>, ...extraClasses: Parameters<typeof clsx>): string {
  return clsx(
    `rdg-cell ${cell}`, {
      [cellFrozenClassname]: column.frozen,
      [cellFrozenLastClassname]: column.isLastFrozenColumn
    },
    ...extraClasses
  );
```
We can pass all the other `clsx` params in this function so we don't call it twice, WDYT?

nstepien (approved):

PR #2301, opened 2021-02-10
rollup type definitions

nstepien:
https://api-extractor.com/
Inspired by https://github.com/puppeteer/puppeteer/pull/6837

With this PR, we'll ship a single `lib/index.d.ts` file, instead of many `.d.ts` files.

Pros:
- Private values/types/APIs cannot be imported anymore
  - Will make it easier to avoid breaking changes with future releases.
- Cleaner intellisense since we'll have fewer types available to import image
- Smaller package size
- Could be used in the future to automatically generate documentation


src/Cell.tsx
```diff
@@ -129,4 +130,4 @@ function Cell<R, SR>({
   );
 }
 
-export default memo(forwardRef(Cell)) as <R, SR = unknown>(props: CellRendererProps<R, SR> & React.RefAttributes<HTMLDivElement>) => JSX.Element;
+export default memo(forwardRef(Cell)) as <R, SR = unknown>(props: CellRendererProps<R, SR> & RefAttributes<HTMLDivElement>) => JSX.Element;
```
nstepien, line 133: image
image
🤷

src/Row.tsx
```diff
@@ -1,4 +1,5 @@
 import { memo, forwardRef } from 'react';
+import type{ RefAttributes } from 'react';
```
amanmahajan7, line 2:
```suggestion
import type { RefAttributes } from 'react';
```

amanmahajan7 (dismissed):

amanmahajan7 (approved):

PR #2304, opened 2021-02-12
:new: Implement initialScrollTop and initialScrollLeft

ekeuus:
This PR implements a way to restore scroll of a Grid on mount when the previous scroll values have been stored somewhere - redux, a context somewhere, localstorage, etc.

I also added a story where you can scroll around down and right, look at another story and upon returning to the "Initial scroll" demo, the original scroll is restored during the initial mount.

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -158,6 +158,12 @@ export interface DataGridProps<R, SR = unknown> extends SharedDivProps {
   /** The node where the editor portal should mount. */
   editorPortalTarget?: Element;
   rowClass?: (row: R) => string | undefined;
+
+  /**
+   * Initial scrolling states
+   */
+  initialScrollTop?: number;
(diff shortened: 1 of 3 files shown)


ekeuus:
@nstepien @amanmahajan7 Let me know if this makes sense or not. IMO it is a very small change and should not break anything for existing users. However, it is a much needed use case and would be the best if it was in the library already instead of re-implementing it.

nstepien:
IMO that's something that can and should be done outside the component.

nstepien:
Considering you have access to the root via `ref=(obj => obj.element}`, this should be implemented as needed, instead being a built-in function. WDYT @amanmahajan7?

amanmahajan7:
I agree with @nstepien, we can use [gridRef ](https://github.com/adazzle/react-data-grid/blob/main/src/DataGrid.tsx#L370)to handle it outside the grid.

PR #2311, opened 2021-02-22
Allow customizing the key property type

danielpza:
Why? This fails:

```tsx
const [selected, setSelected] = new Set<string>([])

<DataGrid selectedRows={selected} onSelectedRowsChange={setSelected} />
```

because onSelectedRowsChange was expected to receive all the properties React.Key has, eg number, which is not possible/desired.


src/DataGrid.tsx
```diff
@@ -80,7 +80,7 @@ type SharedDivProps = Pick<React.HTMLAttributes<HTMLDivElement>,
   | 'style'
 >;
 
-export interface DataGridProps<R, SR = unknown> extends SharedDivProps {
+export interface DataGridProps<R, SR = unknown, K = React.Key> extends SharedDivProps {
```
nstepien, line 83:
```suggestion
export interface DataGridProps<R, SR = unknown, K extends React.Key = React.Key> extends SharedDivProps {
```

nstepien (changes requested):
I think it's a good idea.
Can you please also add the generic to the default export, it's at the end of `DataGrid.tsx`.

nstepien:
Please also add the generic to the default export at the end of `DataGrid.tsx`, it won't be usable otherwise.

src/DataGrid.tsx
```diff
@@ -1071,4 +1071,4 @@ function DataGrid<R, SR>({
   );
 }
 
-export default forwardRef(DataGrid) as <R, SR = unknown>(props: DataGridProps<R, SR> & RefAttributes<DataGridHandle>) => JSX.Element;
+export default forwardRef(DataGrid) as <R, SR = unknown, Key extends React.ReactKey = React.ReactKey>(props: DataGridProps<R, SR, Key> & RefAttributes<DataGridHandle>) => JSX.Element;
```
nstepien, line 1074:
```suggestion
export default forwardRef(DataGrid) as <R, SR = unknown, Key extends React.Key = React.Key>(props: DataGridProps<R, SR, Key> & RefAttributes<DataGridHandle>) => JSX.Element;
```

danielpza:
Thank you! :tada:

PR #2322, opened 2021-03-04
Add onRowDoubleClick callback

xQwexx:
I want to make an external editor for a the table row, and I tried, but with the current interface it is not possible.
So I proposing this new onRowDoubleClick callback very similarly with onRowClick.
The example image what I want to achieve with it.
image


crashnick16:
Is this going to be added back to the canary version ?

amanmahajan7:
Thank you for the PR. I have replaced it with https://github.com/adazzle/react-data-grid/pull/2468 to account for a few more changes

PR #2328, opened 2021-03-11
Topic/fix erratic column resizing

gernotkogler:
We experienced erratic column resizing on a non trivial page layout containing a react-data-grid. By simplifying the resize logic, the behaviour is now stable in our application, no more resizing of the wrong columns or big jumps in column width when moving the mouse only a few pixels. 

Additionally, I've introduced the onColumnResized callback to be able to process column resizing when the resizing is done, which is a big performance boost for, eg, storing the column width in the local storage


gernotkogler:
I will create an new pull request

PR #2335, opened 2021-03-16
Fix summary row position and remove callback dependency

amanmahajan7:
**Before**
image

**After**
image


nstepien (approved):
:shipit: nice!

PR #2338, opened 2021-03-17
Use context for selection

amanmahajan7:
Select all operation
**Before**
image
image

**After**
image
image


src/Row.tsx
```diff
@@ -71,23 +76,25 @@ function Row<R, SR = unknown>({
...
-            dragHandleProps={isCellSelected ? (selectedCellProps as SelectedCellProps).dragHandleProps : undefined}
-            onFocus={isCellSelected ? (selectedCellProps as SelectedCellProps).onFocus : undefined}
-            onKeyDown={isCellSelected ? selectedCellProps!.onKeyDown : undefined}
-            onRowClick={onRowClick}
-            onRowChange={onRowChange}
-            selectCell={selectCell}
-            selectRow={selectRow}
-          />
+          <RowSelectionChangeContext.Provider value={onRowSelectionChange}>
+            <RowSelectionContext.Provider value={isRowSelected}>
```
nstepien, line 80: Creating 2 Provider instance per cell seems overkill, can't we just have 2 per row by rendering them straight under the row's `div`?

nstepien (commented):
Can't we just render the providers in DataGrid once?
That way we can avoid `Row` re-renders perhaps.
We might trigger formatter re-renders more often, but it's probably still less heavy since it's only for the few formatters that are in the viewport, and are using the contexts.
Row selection values won't change often anyway.

src/Row.tsx
amanmahajan7: We should. Totally missed it

amanmahajan7:
**Before**
image


**After**
image

This profile is for selecting/deselecting all rows.

src/hooks/useRowSelection.ts
```diff
@@ -0,0 +1,31 @@
...
+function useRowSelectionChange() {
+  const context = useContext(RowSelectionChangeContext);
+  if (context === undefined) {
+    throw new Error('useRowSelectionChange must be used within a RowSelectionChangeContext');
+  }
+  return context;
+}
+
+export function useRowSelection(): readonly [boolean, (selectRowEvent: SelectRowEvent) => void] {
+  return [useSelection(), useRowSelectionChange()];
```
nstepien, line 30: Let's inline all this

we can have a single `if` check
```js
if (context1 === undefined || context2 === undefined) throw ....
```

I'd also change the error message to be `'useRowSelection must be used within DataGrid cells'` or something like that.
We don't expose the providers themselves, so the error message wouldn't be very useful to devs otherwise.

src/hooks/useRowSelection.ts
```diff
@@ -0,0 +1,31 @@
...
+
+function useSelection() {
+  const context = useContext(RowSelectionContext);
+  if (context === undefined) {
+    throw new Error('useRowSelection must be used within a RowSelectionContext');
+  }
+  return context;
+}
+
+export const RowSelectionChangeContext =
```
nstepien, line 16:
```suggestion
const RowSelectionChangeContext =
```

src/Row.tsx
```diff
@@ -28,7 +29,6 @@ function Row<R, SR = unknown>(
     height,
     onRowChange,
     selectCell,
-    selectRow,
     'aria-rowindex': ariaRowIndex,
     'aria-selected': ariaSelected,
     ...props
```
nstepien, line 34: Do these two props actually need to be destructured?

nstepien (changes requested):

src/Row.tsx
amanmahajan7: no, removed

src/hooks/useRowSelection.ts
```diff
@@ -0,0 +1,22 @@
...
+const RowSelectionContext = createContext<boolean | undefined>(undefined);
+
+export const RowSelectionProvider = RowSelectionContext.Provider;
+
+const RowSelectionChangeContext =
+  createContext<((selectRowEvent: SelectRowEvent) => void) | undefined>(undefined);
+
+export const RowSelectionChangeProvider = RowSelectionChangeContext.Provider;
+
+export function useRowSelection(): readonly [boolean, (selectRowEvent: SelectRowEvent) => void] {
```
nstepien, line 13: FWIW the tuple is created on each call, and we don't care about it being mutable or immutable, so we can remove `readonly` here.
```suggestion
export function useRowSelection(): [boolean, (selectRowEvent: SelectRowEvent) => void] {
```

nstepien (approved):

nstepien (approved):
:shipit:

PR #2349, opened 2021-04-01
:art: Set width from maxWidth if exists

ekeuus:
This PR fixes issues where table grid layouts break (not filling the whole table) if you only define a maxWidth, but no width.

Current result:
Screenshot 2021-04-01 at 15 24 36

Does not work now:
```
{
   key: 'key',
   maxWidth: 123,
}
```

Expected and now the result after this change:
Screenshot 2021-04-01 at 15 24 53

Works now:
```
{
   key: 'key',
   maxWidth: 123,
   width: 123,
}
```
Works after this PR:
```
{
   key: 'key',
   maxWidth: 123,
}
```

I can also make it work now, but it is a bit annoying to have to define `width` as well since it already works with `minWidth`.

### DIFF
--- a/src/hooks/useViewportColumns.ts
+++ b/src/hooks/useViewportColumns.ts
@@ -44,7 +44,8 @@ export function useViewportColumns<R, SR>({
         rowGroup,
         sortable: rawColumn.sortable ?? defaultSortable,
         resizable: rawColumn.resizable ?? defaultResizable,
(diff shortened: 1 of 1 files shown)


nstepien:
I don't think this would work well if had a big `maxWidth` value, the grid would be filled very quickly instead of having evenly positioned columns.

amanmahajan7:
Closing as there is no activity

Issue #2352, opened 2021-04-07
Cannot delete input content in NumericFilter


neotan:
**Describe the bug**
In NumericFilter, cannot delete (by pressing the Backspace or Delete button) input content.

**Current behavior**
In NumericFilter, cannot delete (by pressing the Backspace or Delete button) input content.

**Expected behavior**
In NumericFilter, can delete (by pressing the Backspace or Delete button) input content.

**Environment**
- `standard-version` version(s): Canary
- Node/npm version: v14.15.3
- OS: Windows 1

**Possible Solution**
Allow Backspace and Delete's behaviors.

**Additional context**
Example: https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--header-filters

![3814f5f7-de79-41fe-80aa-f799ad6e4a52](https://user-images.githubusercontent.com/10399087/113880472-17120680-97ee-11eb-83d2-5a595e1a430f.gif)

amanmahajan7:
We have removed `NumericFilter` example and I would recommend [copying the code](https://github.com/adazzle/react-data-grid/blob/2f6181d9234fd81ea2f4ad15b1145e4b2cdf08ab/stories/demos/components/Filters/NumericFilter.tsx) and tweaking it as needed.

PR #2353, opened 2021-04-07
fix: backspace and delete buttons doesn't work on NumericFilter

neotan:
For fixing #2352. Kindly review and comment. Thanks.


nstepien:
We've removed the NumericFilter now

PR #2355, opened 2021-04-09
POC: Export to CSV, XLSX, PDF

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -208,6 +209,7 @@ function DataGrid<R, SR>({
   // Toggles and modes
   enableFilterRow = false,
   cellNavigationMode = 'NONE',
+  enableVirtualization = true,
```
amanmahajan7, line 212: Not sure on what the prop name. `disableVirtualization` makes more sense but I kept this to be consistent with `enableHeaderRow` . We can also consider forwarding this prop to the formatters to users can use it during printing.

stories/demos/exportUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+export function exportToXlsx<R, SR>(gridElement: ReactElement<DataGridProps<R, SR>>, fileName: string) {
+  const { head, body, foot } = getGridContent(gridElement);
+  const wb = XLSX.utils.book_new();
+  const ws = XLSX.utils.aoa_to_sheet([...head, ...body, ...foot]);
+  XLSX.utils.book_append_sheet(wb, ws, 'Sheet 1');
+  XLSX.writeFile(wb, fileName);
+}
+
+
+export function exportToPdf<R, SR>(gridElement: ReactElement<DataGridProps<R, SR>>, fileName: string) {
```
amanmahajan7, line 28: https://github.com/simonbengtsson/jsPDF-AutoTable
Handy plugin to convert table to pdf

stories/demos/exportUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+export function exportToCsv<R, SR>(gridElement: ReactElement<DataGridProps<R, SR>>, fileName: string) {
+  const { head, body, foot } = getGridContent(gridElement);
+  const content = [...head, ...body, ...foot]
+    .map(cells => cells.map(serialiseCellValue).join(','))
+    .join('\n');
+
+  downloadFile(fileName, new Blob([content], { type: 'text/csv;charset=utf-8;' }));
+}
+
+export function exportToXlsx<R, SR>(gridElement: ReactElement<DataGridProps<R, SR>>, fileName: string) {
```
amanmahajan7, line 19: https://github.com/SheetJS/sheetjs
We can use the commercial version

stories/demos/exportUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+    foot,
+    horizontalPageBreak: true,
+    styles: { cellPadding: 1.5, fontSize: 8, cellWidth: 'wrap' },
+    tableWidth: 'wrap'
+  });
+  doc.save(fileName);
+}
+
+function getGridContent<R, SR>(gridElement: ReactElement<DataGridProps<R, SR>>) {
+  const grid = document.createElement('content');
```
amanmahajan7, line 47: The idea is to grab the text content of each cell and create an array of arrays. Both pdf and xlsx libraries handle array of arrays

src/DataGrid.tsx
nstepien: Prop name is good to me. I prefer avoiding "negative" prop/variable names as they make reading logic harder.

>We can also consider forwarding this prop to the formatters to users can use it during printing.

What use case do you have in mind?

src/hooks/useViewportColumns.ts
```diff
@@ -199,7 +204,7 @@ export function useViewportColumns<R, SR>({
     const colOverscanEndIdx = Math.min(lastColIdx, colVisibleEndIdx + 1);
 
     return [colOverscanStartIdx, colOverscanEndIdx];
-  }, [columns, columnMetrics, lastFrozenColumnIndex, scrollLeft, totalFrozenColumnWidth, viewportWidth]);
+  }, [columnMetrics, columns, enableVirtualization, lastFrozenColumnIndex, scrollLeft, totalFrozenColumnWidth, viewportWidth]);
```
nstepien, line 207: `enableVirtualization` is unlikely to change often, so we can move it last (micro-optimization)
```suggestion
  }, [columnMetrics, columns, lastFrozenColumnIndex, scrollLeft, totalFrozenColumnWidth, viewportWidth, enableVirtualization]);
```

src/hooks/useViewportRows.ts
```diff
@@ -91,6 +93,17 @@ export function useViewportRows<R>({
...
+  if (!enableVirtualization) {
+    return {
+      rowOverscanStartIdx: 0,
+      rowOverscanEndIdx: rows.length - 1,
+      rows,
+      rowsCount,
+      isGroupRow
+    };
+  }
+
```
nstepien, line 106: Can keep `overscanThreshold` after the `if`

```suggestion
  if (!enableVirtualization) {
    return {
      rowOverscanStartIdx: 0,
      rowOverscanEndIdx: rows.length - 1,
      rows,
      rowsCount,
      isGroupRow
    };
  }

  const overscanThreshold = 4;
```

stories/demos/exportUtils.tsx
nstepien, line 47:
```suggestion
  const grid = document.createElement('div');
```
image
https://developer.mozilla.org/en-US/docs/Web/HTML/Element

stories/demos/exportUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+    foot: getRows('.rdg-summary-row')
+  };
+
+  function getRows(selector: string) {
+    return Array.from(
+      grid.querySelectorAll<HTMLDivElement>(selector)
+    ).map(gridRow => {
+      return Array.from(
+        gridRow.querySelectorAll<HTMLDivElement>('.rdg-cell')
+      ).map(gridCell => gridCell.innerText);
```
nstepien, line 64: So the cell ends up empty when we display an image, a checkbox, or a `progress` element for example?

stories/demos/exportUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+    horizontalPageBreak: true,
+    styles: { cellPadding: 1.5, fontSize: 8, cellWidth: 'wrap' },
+    tableWidth: 'wrap'
+  });
+  doc.save(fileName);
+}
+
+function getGridContent<R, SR>(gridElement: ReactElement<DataGridProps<R, SR>>) {
+  const grid = document.createElement('content');
+  grid.innerHTML = renderToStaticMarkup(cloneElement(gridElement, {
```
nstepien, line 48: If we're rendering images, it might end up downloading a lot of images at once.
We could avoid this by rendering the grid in an anonymous document: https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createDocument

stories/demos/exportUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+      grid.querySelectorAll<HTMLDivElement>(selector)
+    ).map(gridRow => {
+      return Array.from(
+        gridRow.querySelectorAll<HTMLDivElement>('.rdg-cell')
+      ).map(gridCell => gridCell.innerText);
+    });
+  }
+}
+
+function serialiseCellValue(value: unknown) {
```
nstepien, line 69: Should we handle `\n` as well?

stories/demos/exportUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+function downloadFile(fileName: string, data: Blob) {
+  const downloadLink = document.createElement('a');
+  downloadLink.download = fileName;
+  const url = URL.createObjectURL(data);
+  downloadLink.href = url;
+  downloadLink.style.display = 'none';
+  document.body.appendChild(downloadLink);
+  downloadLink.click();
+  document.body.removeChild(downloadLink);
```
nstepien, line 85:
```suggestion
  downloadLink.click();
```
I believe we can click without adding the link to the tree. Not sure if this'll work in all browsers though.

src/DataGrid.tsx
amanmahajan7: If for example there is a cell that shows a complex data like an icon with a tooltip. It may not work out of the box and it would helpful to provide an alternate formatter for printing. We can revisit if needed

stories/demos/exportUtils.tsx
amanmahajan7: Ah, copy paste error

stories/demos/exportUtils.tsx
amanmahajan7: Correct. We can manually update cell value during export or provide a flag to the formatter so it can return an print friendly value

stories/demos/exportUtils.tsx
amanmahajan7: Good to know. We can add it if needed or I will create a followup PR

stories/demos/exportUtils.tsx
amanmahajan7: Works on Edge, FF, and Chrome

stories/demos/exportUtils.tsx
amanmahajan7: We can. It works for now so may be next PR if needed

src/DataGrid.tsx
nstepien: Sounds like it should be a separate thing, and it could be implemented by the user via a provider.

nstepien (approved):

PR #2356, opened 2021-04-13
Add column.colSpan prop

amanmahajan7:
```
colSpan?: (args: ColSpanArgs<TRow, TSummaryRow>) => number | undefined;

export type ColSpanArgs<R, SR> = {
  type: 'HEADER' | 'FILTER';
} | {
  type: 'ROW';
  row: R;
} | {
  type: 'SUMMARY';
  row: SR;
};

```

It can be used as 

```
colSpan(args) {
  if (args.type === 'ROW') {
    if (key === '2' && args.row === 2) return 3;
    if (key === '4' && args.row === 4) return 6; // Will not work as colspan includes both frozen and regular columns
    if (key === '0' && args.row === 5) return 5;
    if (key === '27' && args.row === 8) return 3;
    if (key === '6' && args.row < 8) return 2;
  }
  if (args.type === 'HEADER' && key === '8') {
    return 3;
  }
  return undefined;
},
```

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -813,6 +813,7 @@ function DataGrid<R, SR>({
   }
 
   function getColSpanPosition(nextPosition: Position) {
+    if (!isCellWithinBounds(nextPosition)) return nextPosition;
     const row = rows[nextPosition.rowIdx];
     if (isGroupRow(row)) return nextPosition;
     // If a cell within the colspan range is selected then move to the
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -294,6 +294,8 @@ function DataGrid<R, SR>({
     columns,
     colOverscanStartIdx,
     colOverscanEndIdx,
+    rowOverscanStartIdx,
+    rowOverscanEndIdx,
     rows,
     isGroupRow
   });
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -23,7 +23,8 @@ import {
   isSelectedCellEditable,
   canExitGrid,
   isCtrlKeyHeldDown,
-  isDefaultCellInput
+  isDefaultCellInput,
+  getColSpan
 } from './utils';
 
 import type {
@@ -797,6 +798,7 @@ function DataGrid<R, SR>({
 
     const ctrlKey = isCtrlKeyHeldDown(event);
     let nextPosition = getNextPosition(key, ctrlKey, shiftKey);
+    nextPosition = getColSpanPosition(nextPosition);
     nextPosition = getNextSelectedCellPosition({
       columns,
       rowsCount: rows.length,
@@ -805,6 +807,23 @@ function DataGrid<R, SR>({
     });
 
     selectCell(nextPosition);
+  }
+
+  function getColSpanPosition(nextPosition: Position) {
+    const row = rows[nextPosition.rowIdx];
+    if (isGroupRow(row)) return nextPosition;
+    // If a cell within the colspan range is selected then move to the
+    // previous or the next cell depending on the navigation direction
+    for (let colIdx = 0; colIdx <= nextPosition.idx; colIdx++) {
+      const column = columns[colIdx];
+      const colSpan = getColSpan<R, SR>(column, columns, { type: 'ROW', row });
+      if (colSpan && nextPosition.idx > column.idx && nextPosition.idx < colSpan + column.idx) {
+        nextPosition.idx = column.idx + (nextPosition.idx - selectedPosition.idx > 0 ? colSpan : 0);
+        break;
+      }
+    }
+
+    return nextPosition;
   }
 
   function getDraggedOverCellIdx(currentRowIdx: number): number | undefined {
@@ -826,7 +845,6 @@ function DataGrid<R, SR>({
         mode: 'EDIT',
         idx: selectedPosition.idx,
         onKeyDown: handleKeyDown,
-        prevIdx: prevSelectedPosition.current.idx,
         editorProps: {
           editorPortalTarget,
           rowHeight,
@@ -840,7 +858,6 @@ function DataGrid<R, SR>({
     return {
       mode: 'SELECT',
       idx: selectedPosition.idx,
-      prevIdx: prevSelectedPosition.current.idx,
       onFocus: handleFocus,
       onKeyDown: handleKeyDown,
       dragHandleProps: enableCellDragAndDrop && isCellEditable(selectedPosition)
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -808,6 +808,7 @@ function DataGrid<R, SR>({
         mode: 'EDIT',
         idx: selectedPosition.idx,
         onKeyDown: handleKeyDown,
+        prevIdx: prevSelectedPosition.current.idx,
         editorProps: {
           editorPortalTarget,
           rowHeight,
@@ -821,6 +822,7 @@ function DataGrid<R, SR>({
     return {
       mode: 'SELECT',
       idx: selectedPosition.idx,
+      prevIdx: prevSelectedPosition.current.idx,
       onFocus: handleFocus,
       onKeyDown: handleKeyDown,
       dragHandleProps: enableCellDragAndDrop && isCellEditable(selectedPosition)
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -23,8 +23,7 @@ import {
   isSelectedCellEditable,
   canExitGrid,
   isCtrlKeyHeldDown,
-  isDefaultCellInput,
-  getColSpan
+  isDefaultCellInput
 } from './utils';
 
 import type {
@@ -804,35 +803,19 @@ function DataGrid<R, SR>({
     event.preventDefault();
 
     const ctrlKey = isCtrlKeyHeldDown(event);
-    let nextPosition = getNextPosition(key, ctrlKey, shiftKey);
-    nextPosition = getNextSelectedCellPosition({
+    const nextPosition = getNextSelectedCellPosition({
       columns,
-      rowsCount: rows.length,
+      colSpanColumns,
+      rows,
+      lastFrozenColumnIndex,
       cellNavigationMode: mode,
-      nextPosition
+      currentPosition: selectedPosition,
+      nextPosition: getNextPosition(key, ctrlKey, shiftKey),
+      isCellWithinBounds,
+      isGroupRow
     });
-    nextPosition = getColSpanPosition(nextPosition);
 
     selectCell(nextPosition);
-  }
-
-  function getColSpanPosition(nextPosition: Position) {
-    if (!isCellWithinBounds(nextPosition)) return nextPosition;
-    const row = rows[nextPosition.rowIdx];
-    if (isGroupRow(row)) return nextPosition;
-    // If a cell within the colspan range is selected then move to the
-    // previous or the next cell depending on the navigation direction
-    for (const column of colSpanColumns) {
-      const colIdx = column.idx;
-      if (colIdx > nextPosition.idx) break;
-      const colSpan = getColSpan<R, SR>(column, lastFrozenColumnIndex, { type: 'ROW', row });
-      if (colSpan && nextPosition.idx > column.idx && nextPosition.idx < colSpan + column.idx) {
-        nextPosition.idx = column.idx + (nextPosition.idx - selectedPosition.idx > 0 ? colSpan : 0);
-        break;
-      }
-    }
-
-    return nextPosition;
   }
 
   function getDraggedOverCellIdx(currentRowIdx: number): number | undefined {
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -801,13 +801,13 @@ function DataGrid<R, SR>({
 
     const ctrlKey = isCtrlKeyHeldDown(event);
     let nextPosition = getNextPosition(key, ctrlKey, shiftKey);
-    nextPosition = getColSpanPosition(nextPosition);
     nextPosition = getNextSelectedCellPosition({
       columns,
       rowsCount: rows.length,
       cellNavigationMode: mode,
       nextPosition
     });
+    nextPosition = getColSpanPosition(nextPosition);
 
     selectCell(nextPosition);
   }
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -297,6 +297,7 @@ function DataGrid<R, SR>({
     rowOverscanStartIdx,
     rowOverscanEndIdx,
     rows,
+    summaryRows,
     isGroupRow
   });
 
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -296,6 +296,7 @@ function DataGrid<R, SR>({
     colSpanColumns,
     colOverscanStartIdx,
     colOverscanEndIdx,
+    lastFrozenColumnIndex,
     rowOverscanStartIdx,
     rowOverscanEndIdx,
     rows,
@@ -824,7 +825,7 @@ function DataGrid<R, SR>({
     for (const column of colSpanColumns) {
       const colIdx = column.idx;
       if (colIdx > nextPosition.idx) break;
-      const colSpan = getColSpan<R, SR>(column, columns, { type: 'ROW', row });
+      const colSpan = getColSpan<R, SR>(column, lastFrozenColumnIndex, { type: 'ROW', row });
       if (colSpan && nextPosition.idx > column.idx && nextPosition.idx < colSpan + column.idx) {
         nextPosition.idx = column.idx + (nextPosition.idx - selectedPosition.idx > 0 ? colSpan : 0);
         break;
@@ -932,6 +933,7 @@ function DataGrid<R, SR>({
           copiedCellIdx={copiedCell !== null && copiedCell.row === row ? columns.findIndex(c => c.key === copiedCell.columnKey) : undefined}
           draggedOverCellIdx={getDraggedOverCellIdx(rowIdx)}
           setDraggedOverRowIdx={isDragging ? setDraggedOverRowIdx : undefined}
+          lastFrozenColumnIndex={lastFrozenColumnIndex}
           selectedCellProps={getSelectedCellProps(rowIdx)}
           onRowChange={handleFormatterRowChangeWrapper}
           selectCell={selectCellWrapper}
@@ -986,6 +988,7 @@ function DataGrid<R, SR>({
         sortColumn={sortColumn}
         sortDirection={sortDirection}
         onSort={onSort}
+        lastFrozenColumnIndex={lastFrozenColumnIndex}
       />
       {enableFilterRow && (
         <FilterRow<R, SR>
@@ -1013,6 +1016,7 @@ function DataGrid<R, SR>({
               row={row}
               bottom={summaryRowHeight * (summaryRows.length - 1 - rowIdx)}
               viewportColumns={viewportColumns}
+              lastFrozenColumnIndex={lastFrozenColumnIndex}
             />
           ))}
         </>
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -261,6 +261,7 @@ function DataGrid<R, SR>({
 
   const {
     columns,
+    colSpanColumns,
     colOverscanStartIdx,
     colOverscanEndIdx,
     layoutCssVars,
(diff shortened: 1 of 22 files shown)


ekeuus:
Why not go for a non-manually calculated api? https://github.com/adazzle/react-data-grid/pull/2350

amanmahajan7:
@ekeuus `colSpan` is a generic api that can handle any type of row so not just header. We can even have both apis if makes sense. Thank your for the header groups PR. We will review it soon

src/DataGrid.tsx
```diff
@@ -279,6 +289,14 @@ function DataGrid<R, SR>({
     enableVirtualization
   });
 
+  const viewportColumns = useViewportColumns({
```
amanmahajan7, line 292: Had to split into two hooks as we need to calculate colspan for all the rows in the viewport. This is needed to ensure a column that spans a column in the viewport is also visible

src/HeaderRow.tsx
```diff
@@ -41,17 +41,25 @@ function HeaderRow<R, SR>({
...
 
+  let colSpan: number| undefined;
   return (
     <div
       role="row"
       aria-rowindex={1} // aria-rowindex is 1 based
       className={headerRowClassname}
     >
       {columns.map(column => {
+        if (colSpan && colSpan > 1) {
```
amanmahajan7, line 52: Do not render cells in the colspan range

src/hooks/useCalculatedColumns.ts
```diff
@@ -0,0 +1,252 @@
+import { useMemo } from 'react';
```
amanmahajan7, line 1: Same as previous `useViewportHook.` I moved the last `useMemo` to the new `useViewport` hook as it needed rows

src/utils/index.ts
```diff
@@ -28,3 +35,28 @@ export function getCellClassname<R, SR>(column: CalculatedColumn<R, SR>, ...extr
...
+export function getColSpan<R, SR>(column: CalculatedColumn<R, SR>, viewportColumns: readonly CalculatedColumn<R, SR>[], args: ColSpanArgs<R, SR>) {
+  const colSpan = typeof column.colSpan === 'function' ? column.colSpan(args) : column.colSpan;
+  if (colSpan && !areColSpanColumnsCompatible(column, viewportColumns, colSpan)) {
+    // ignore colSpan if it spans over frozen and regular columns
+    return undefined;
+  }
+  return colSpan;
+}
+
+function areColSpanColumnsCompatible<R, SR>(column: CalculatedColumn<R, SR>, viewportColumns: readonly CalculatedColumn<R, SR>[], colSpan: number) {
```
amanmahajan7, line 48: Ensure all the columns within the colspan range are either frozen or not

src/Row.tsx
```diff
@@ -57,12 +61,29 @@ function Row<R, SR = unknown>({
     >
       {viewportColumns.map(column => {
         const isCellSelected = selectedCellProps?.idx === column.idx;
+        if (colSpan && colSpan > 1) {
+          colSpan--;
+          if (isCellSelected) {
```
nstepien, line 66: Can't we handle cell selection in `DataGrid`?

src/SummaryRow.tsx
```diff
@@ -21,20 +22,30 @@ function SummaryRow<R, SR>({
   bottom,
   'aria-rowindex': ariaRowIndex
 }: SummaryRowProps<R, SR>) {
+  let colSpan: number| undefined;
```
nstepien:
```suggestion
  let colSpan: number | undefined;
```

src/utils/index.ts
```diff
@@ -13,10 +14,16 @@ export function assertIsValidKeyGetter<R>(keyGetter: unknown): asserts keyGetter
...
+export function getCellStyle<R, SR>(column: CalculatedColumn<R, SR>, colSpan?: number): React.CSSProperties {
+  const style: CSSProperties = {
+    gridColumnStart: column.idx + 1,
+    gridColumnEnd: colSpan ? `span ${colSpan}` : undefined
+  };
+  if (column.frozen) {
+    style.left = `var(--frozen-left-${column.key})`;
+  }
+
+  return style;
```
nstepien, line 26:
```suggestion
  return {
    gridColumnStart: column.idx + 1,
    gridColumnEnd: colSpan ? `span ${colSpan}` : undefined,
    left: column.frozen ? `var(--frozen-left-${column.key})` : undefined
  };
```

src/utils/index.ts
```diff
@@ -28,3 +35,28 @@ export function getCellClassname<R, SR>(column: CalculatedColumn<R, SR>, ...extr
...
+    return undefined;
+  }
+  return colSpan;
+}
+
+function areColSpanColumnsCompatible<R, SR>(column: CalculatedColumn<R, SR>, viewportColumns: readonly CalculatedColumn<R, SR>[], colSpan: number) {
+  const isFrozen = column.frozen;
+  if (!isFrozen) return true;
+  const startIdx = column.idx - viewportColumns[0].idx;
+  for (
```
nstepien, line 52: Don't we only need to check for starting column + `columns[starting column index + colspan]`?
Why check for the columns in between?
Shouldn't we also check for out of bounds spanning/negative values/non-integer values?

src/utils/index.ts
```diff
@@ -28,3 +35,28 @@ export function getCellClassname<R, SR>(column: CalculatedColumn<R, SR>, ...extr
     ...extraClasses
   );
 }
+
+export function getColSpan<R, SR>(column: CalculatedColumn<R, SR>, viewportColumns: readonly CalculatedColumn<R, SR>[], args: ColSpanArgs<R, SR>) {
+  const colSpan = typeof column.colSpan === 'function' ? column.colSpan(args) : column.colSpan;
+  if (colSpan && !areColSpanColumnsCompatible(column, viewportColumns, colSpan)) {
+    // ignore colSpan if it spans over frozen and regular columns
```
nstepien, line 42:
```suggestion
    // ignore colSpan if it spans over both frozen and regular columns
```

stories/demos/CommonFeatures.tsx
```diff
@@ -52,6 +56,8 @@ interface Row {
   available: boolean;
 }
 
+const colSpanRowIds: number[] = [2, 4, 6];
```
nstepien: Could be a `Set`

amanmahajan7: I have added a new `Column Spanning` example

src/SummaryRow.tsx
amanmahajan7: Changed the logic to use cells array and removed `colSpan` variable. It is a tiny bit performant as we do not loop all the columns

src/utils/index.ts
amanmahajan7: We are only checking columns within the colspan range. 

```
index < colSpan && startIdx + index < viewportColumns.length;
```

Had to add `viewportColumns` as an extra check as it was failing with undefined column. I can check again. 

> Shouldn't we also check for out of bounds spanning/negative values/non-integer values?
Added some validations

src/hooks/useViewportColumns.ts
```diff
@@ -1,254 +1,40 @@
...
-  minColumnWidth: number
-): number {
-  width = Math.max(width, minWidth ?? minColumnWidth);
-
-  if (typeof maxWidth === 'number') {
-    return Math.min(width, maxWidth);
-  }
-
-  return width;
+  }, [colOverscanEndIdx, colOverscanStartIdx, columns, isGroupRow, rows]);
```
nstepien, line 39: We should split this into two `useMemo`s so we can avoid recreating the `viewportColumns` array every time we change `rows`/`isGroupRow`.

`if ((colSpan + column.idx - 1) > startIdx) {` also had a bug where we would render a cell that should have in a span,

Before:
image
image
image
^ the 7x7 cell is rendered when it shouldn't.

After:
image
image
No funky behavior even when "offscreen".

---

We also still need to loop through the header and summary spans, as it can lead to bugs right now:
image
image
Header for column `10` appears when the cell position for 8 and 9 are "offscreen".

---

```ts
  const startIdx = useMemo(() => {
    let startIdx = colOverscanStartIdx;
    // find the column that spans over a column within the visible columns range and adjust colOverscanStartIdx
    for (const row of rows) {
      if (isGroupRow(row)) continue;
      for (let colIdx = 0; colIdx < colOverscanStartIdx; colIdx++) {
        const column = columns[colIdx];
        const colSpan = getColSpan(column, columns, { type: 'ROW', row });
        if (!colSpan) continue;
        if ((colIdx + colSpan) > startIdx) {
          startIdx = colIdx;
        }
      }
    }

    return startIdx;
  }, [rows, colOverscanStartIdx, isGroupRow, columns]);

  return useMemo((): readonly CalculatedColumn<R, SR>[] => {
    const viewportColumns: CalculatedColumn<R, SR>[] = [];
    for (let colIdx = 0; colIdx <= colOverscanEndIdx; colIdx++) {
      const column = columns[colIdx];

      if (colIdx < startIdx && !column.frozen) continue;
      viewportColumns.push(column);
    }

    return viewportColumns;
  }, [startIdx, colOverscanEndIdx, columns]);
```

nstepien (changes requested):

src/hooks/useViewportColumns.ts
nstepien: Also we should probably just loop through the visible rows, and not all rows.

src/Row.tsx
```diff
@@ -44,6 +45,62 @@ function Row<R, SR = unknown>({
...
+    if (colSpan !== undefined) {
+      index += colSpan - 1;
+      // If a cell within the colspan range is selected then move to the
+      // previous or the next cell depending on the navigation direction
+      const isColSpanCellSelected = selectedCellProps && selectedCellProps.idx > column.idx && selectedCellProps.idx < column.idx + colSpan;
+      if (isColSpanCellSelected) {
+        const { prevIdx } = selectedCellProps!;
+        selectCell({
+          rowIdx,
+          idx: column.idx + (column.idx - prevIdx >= 0 ? colSpan : 0)
```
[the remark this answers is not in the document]
amanmahajan7: Consider
colSpan = [2 - 4],
- `selectedIdx` === 5, navigateLeft should set `selectedIdx` to 2
- `selectedIdx` === 1, navigateRight should set `selectedIdx` to 2
- - `selectedIdx` === 2, navigateRight should set `selectedIdx` to 5

src/hooks/useViewportColumns.ts
```diff
@@ -1,254 +1,83 @@
...
-        }
-        return -1;
+  // find the column that spans over a column within the visible columns range and adjust colOverscanStartIdx
+  const startIdx = useMemo(() => {
+    let startIdx = colOverscanStartIdx;
+
+    const updateStartIdx = (colIdx: number, colSpan: number | undefined) => {
+      if (!colSpan) return;
+      const newStartIdx = colIdx + colSpan;
+      if (newStartIdx > colOverscanStartIdx && colIdx < startIdx) {
```
nstepien, line 38: `newStartIdx` is a confusing variable name here IMO.
```suggestion
      if ((colIdx + colSpan) > colOverscanStartIdx && colIdx < startIdx) {
```

src/hooks/useViewportColumns.ts
```diff
@@ -1,254 +1,83 @@
...
-
-      // Sort grouped columns second, following the groupBy order:
-      if (rawGroupBy?.includes(aKey)) {
-        if (rawGroupBy.includes(bKey)) {
-          return rawGroupBy.indexOf(aKey) - rawGroupBy.indexOf(bKey);
-        }
-        return -1;
+  // find the column that spans over a column within the visible columns range and adjust colOverscanStartIdx
+  const startIdx = useMemo(() => {
+    let startIdx = colOverscanStartIdx;
```
nstepien, line 33: Initial render + small grids + `enableVirtualization=true` optimization? 🤔 
Though `if (colIdx >= colOverscanStartIdx) break;` covers it already so probably not necessary.
```suggestion
    if (colOverscanStartIdx === 0) return 0;

    let startIdx = colOverscanStartIdx;
```

src/hooks/useViewportColumns.ts
```diff
@@ -1,254 +1,83 @@
...
-        columnMetrics.set(column, { width, left: 0 });
-      }
-    }
 
-    const unallocatedWidth = viewportWidth - allocatedWidth;
-    const unallocatedColumnWidth = unallocatedWidth / unassignedColumnsCount;
+    for (const column of colSpanColumns) {
+      // check header row
+      const colIdx = column.idx;
+      if (colIdx >= colOverscanStartIdx) break;
```
nstepien, line 46: No? 🤔 
```suggestion
      if (colIdx >= startIdx) break;
```

src/hooks/useViewportColumns.ts
```diff
@@ -1,254 +1,83 @@
...
-    // get the last visible non-frozen column index
-    let colVisibleEndIdx = colVisibleStartIdx;
-    while (colVisibleEndIdx < lastColIdx) {
-      const { left, width } = columnMetrics.get(columns[colVisibleEndIdx])!;
-      // if the right side of the column is beyond or equal to the right side of the available viewport,
-      // then it the last column that's at least partially visible, as the previous column's right side is not beyond the viewport.
-      if (left + width >= viewportRight) {
-        break;
+      // check summary rows
+      if (summaryRows) {
```
nstepien, line 62:
```suggestion
      if (summaryRows !== undefined) {
```

src/hooks/useViewportColumns.ts
```diff
@@ -1,254 +1,83 @@
...
-      colVisibleEndIdx++;
     }
 
-    const colOverscanStartIdx = Math.max(firstUnfrozenColumnIdx, colVisibleStartIdx - 1);
-    const colOverscanEndIdx = Math.min(lastColIdx, colVisibleEndIdx + 1);
-
-    return [colOverscanStartIdx, colOverscanEndIdx];
-  }, [columnMetrics, columns, lastFrozenColumnIndex, scrollLeft, totalFrozenColumnWidth, viewportWidth, enableVirtualization]);
+    return startIdx;
+  }, [colOverscanStartIdx, colSpanColumns, columns, enableFilterRow, isGroupRow, rowOverscanEndIdx, rowOverscanStartIdx, rows, summaryRows]);
```
nstepien, line 70: Roughly sorted by most likely to change.
```suggestion
  }, [rowOverscanStartIdx, rowOverscanEndIdx, rows, summaryRows, colOverscanStartIdx, columns, colSpanColumns, isGroupRow, enableFilterRow]);
```

src/types.ts
```diff
@@ -25,6 +25,7 @@ export interface Column<TRow, TSummaryRow = unknown> {
   groupFormatter?: React.ComponentType<GroupFormatterProps<TRow, TSummaryRow>>;
   /** Enables cell editing. If set and no editor property specified, then a textinput will be used as the cell editor */
   editable?: boolean | ((row: TRow) => boolean);
+  colSpan?: ((args: ColSpanArgs<TRow, TSummaryRow>) => number | undefined);
```
nstepien, line 28:
```suggestion
  colSpan?: (args: ColSpanArgs<TRow, TSummaryRow>) => number | undefined;
```

nstepien: I also wonder if we should just rename it to `span`.
Otherwise it can read as redundant: `column.col(umn)Span`

src/types.ts
```diff
@@ -226,3 +228,13 @@ export interface GroupRow<TRow> {
 
 export type CellNavigationMode = 'NONE' | 'CHANGE_ROW' | 'LOOP_OVER_ROW';
 export type SortDirection = 'ASC' | 'DESC' | 'NONE';
+
+export type ColSpanArgs<R, SR> = {
```
nstepien, line 232: Need to re-export it in `src/index.ts`

nstepien (changes requested):

src/hooks/useViewportColumns.ts
amanmahajan7: yeah as `startIdx<=colOverscanStartIdx`, good catch

src/hooks/useViewportColumns.ts
amanmahajan7: I thought about it but it is not necessary I think but does not harm to add it either

amanmahajan7: We should probably also break within the inner loop

src/types.ts
amanmahajan7: Should we change it to `cellSpan` to match `cellClass`? `colSpan` is a standard term (for tables and aria label) so I think it may be better than `span` and if we ever add `rowSpan` then it `colSpan` is better

test/column/colSpan.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    expect(getHeaderCells()).toHaveLength(13);
+
+    // rows
+    const gridRows = getRows();
+    const getCells = (rowIdx: number) => within(gridRows[rowIdx]).getAllByRole('gridcell');
+    expect(getCells(1)).toHaveLength(14);
+    // 7th-8th cells are merged
+    expect(getCells(1)[6]).toHaveAttribute('aria-colindex', '7');
+    expect(getCells(1)[7]).toHaveAttribute('aria-colindex', '9');
+    expect(getCells(1)[6]).toHaveStyle({
```
nstepien, line 58: Kinda random but can you check that these assertions works?
I remember trying `toHaveStyle` before and it'd just accept almost anything I passed into it without erroring out.

test/column/colSpan.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  it('should merges cells', () => {
+    setupColSpanGrid();
+    // header
+    expect(getHeaderCells()).toHaveLength(13);
+
+    // rows
+    const gridRows = getRows();
+    const getCells = (rowIdx: number) => within(gridRows[rowIdx]).getAllByRole('gridcell');
```
nstepien, line 53: Should we add `getCellsAtRowIndex(rowIdx)` in the test utils?

test/column/colSpan.test.ts
```diff
@@ -0,0 +1,117 @@
...
+        cellClass(row) {
+          if (
+            (key === '0' && row === 5)
+              || (key === '2' && row === 2)
+              || (key === '6' && row < 8)
+          ) {
+            return 'col-span';
+          }
+          return undefined;
+        }
```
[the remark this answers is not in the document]
amanmahajan7: We don't. I was gonna test the class but then ended up using other tests

test/column/colSpan.test.ts
amanmahajan7: Works as expected 
image

May be a bug that was fixed

test/column/colSpan.test.ts
amanmahajan7: Added

src/utils/index.ts
```diff
@@ -28,3 +30,16 @@ export function getCellClassname<R, SR>(column: CalculatedColumn<R, SR>, ...extr
     ...extraClasses
   );
 }
+
+export function getColSpan<R, SR>(column: CalculatedColumn<R, SR>, lastFrozenColumnIndex: number, args: ColSpanArgs<R, SR>) {
+  const colSpan = typeof column.colSpan === 'function' ? column.colSpan(args) : column.colSpan;
```
nstepien, line 35: `column.colSpan` can't be a number anymore 🤔 
```suggestion
  const colSpan = typeof column.colSpan === 'function' ? column.colSpan(args) : 1;
```
```suggestion
  const colSpan = typeof column.colSpan === 'function' ? column.colSpan(args) : null;
```

src/utils/selectedCellUtils.ts
```diff
@@ -49,21 +67,36 @@ export function getNextSelectedCellPosition<R, SR>({ cellNavigationMode, columns
...
+  if (!isCellWithinBounds(position)) return position;
+  const row = rows[position.rowIdx];
+  if (isGroupRow(row)) return position;
+  // If a cell within the colspan range is selected then move to the
+  // previous or the next cell depending on the navigation direction
+  for (const column of colSpanColumns) {
+    const colIdx = column.idx;
+    if (colIdx > position.idx) break;
+    const colSpan = getColSpan<R, SR>(column, lastFrozenColumnIndex, { type: 'ROW', row });
+    if (colSpan && position.idx > column.idx && position.idx < colSpan + column.idx) {
```
nstepien, line 93: We can reuse `colIdx` more.
We can add `posIdx` as well.

nstepien (approved):

nstepien (approved):
:shipit: 
Maybe now is a good time to update the PR title

PR #2361, opened 2021-04-19
POC: Use a single CSS grid

amanmahajan7:
Currently each row is using a separate CSS grid which is not efficient. We can make the whole grid one CSS grid and use `grid-template-rows` instead.  Rows do not need absolute position anymore as we we can use `grid-row-start`. To make this all work we need to set [`display: contents`](https://hiddedevries.nl/en/blog/2018-04-21-more-accessible-markup-with-display-contents) on the row.

This will also help implement features like `rowSpan`, [`auto column resizing`](https://github.com/adazzle/react-data-grid/pull/2793), `RTL (may be)` etc

As the row is using `display:contents` , it can be focused anymore and an extra div is added to the treegrid to handle row navigation.

**Possible breaking changes.** 
- As the row is using `display: contents` so `rowClass` would need to be adjusted to set the style on the cell instead
- Due to a bug in browsers, this will currently remove the element from the accessibility tree — screen readers will not look at what's inside. https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents
- We may have to re-introduce `CellRenderer` as row does not generate box anymore. For example `row dragging` example only works on the first cell now


src/HeaderRow.tsx
```diff
@@ -25,21 +26,22 @@ export interface HeaderRowProps<R, SR, K extends React.Key> extends SharedDataGr
...
-  position: sticky;
-  top: 0;
   background-color: var(--rdg-header-background-color);
   font-weight: bold;
-  z-index: 3;
   outline: none;
 
+  > .${cell} {
+    z-index: 3;
+    position: sticky;
```
amanmahajan7, line 37: Need to make cell sticky as row does not render box anymore

website/demos/NoRows.tsx
```diff
@@ -4,7 +4,7 @@ import type { Column } from '../../src';
 
 function EmptyRowsRenderer() {
   return (
-    <div style={{ textAlign: 'center' }}>
+    <div style={{ textAlign: 'center', gridColumn: '1/-1' }}>
```
amanmahajan7, line 7: Span all the columns

website/root.tsx
```diff
@@ -77,6 +77,7 @@ const mainClassname = css`
   box-sizing: border-box;
   height: 100vh;
   padding: 8px;
+  overflow: hidden;
```
amanmahajan7, line 80: Without this the container height is not restricted and all the columns are rendered

src/HeaderRow.tsx
```diff
@@ -25,21 +26,22 @@ export interface HeaderRowProps<R, SR, K extends React.Key> extends SharedDataGr
...
-  display: grid;
-  grid-template-columns: var(--rdg-template-columns);
-  grid-template-rows: var(--rdg-header-row-height);
-  height: var(--rdg-header-row-height); /* needed on Firefox */
+  display: contents;
   line-height: var(--rdg-header-row-height);
-  width: var(--rdg-row-width);
-  position: sticky;
-  top: 0;
   background-color: var(--rdg-header-background-color);
```
nstepien, line 31: I think we can remove this line
```suggestion
```

src/HeaderRow.tsx
```diff
@@ -25,21 +26,22 @@ export interface HeaderRowProps<R, SR, K extends React.Key> extends SharedDataGr
...
   font-weight: bold;
-  z-index: 3;
   outline: none;
```
nstepien, line 33: This one as well, don't think we can set the focus on a `display: contents` element
```suggestion
```

src/HeaderRow.tsx
```diff
@@ -93,6 +95,7 @@ function HeaderRow<R, SR, K extends React.Key>({
       ref={ref}
       tabIndex={tabIndex}
```
nstepien, line 96: Is the `tabIndex` still needed?

src/HeaderRow.tsx
```diff
@@ -25,21 +26,22 @@ export interface HeaderRowProps<R, SR, K extends React.Key> extends SharedDataGr
...
+    position: sticky;
+    top: 0;
+  }
+
+  > .${cellFrozen} {
+    z-index: 4;
+  }
+
   &[aria-selected='true'] {
     box-shadow: inset 0 0 0 2px var(--rdg-selection-color);
```
nstepien, line 46: I don't think this still works

src/SummaryRow.tsx
```diff
@@ -20,12 +20,11 @@ interface SummaryRowProps<R, SR> extends SharedRowRendererProps<R, SR> {
 }
 
 const summaryRow = css`
-  &.${row} {
+  z-index: 3;
+  line-height: var(--rdg-summary-row-height);
```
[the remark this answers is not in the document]
nstepien: We can give it a try.

src/SummaryRow.tsx
```diff
@@ -20,12 +20,11 @@ interface SummaryRowProps<R, SR> extends SharedRowRendererProps<R, SR> {
...
+
+  &.${row} > .${cell} {
```
nstepien, line 26: Isn't this good enough?
```suggestion
  & > .${cell} {
```

src/hooks/useRovingRowRef.ts
```diff
@@ -6,18 +6,17 @@ import { useFocusRef } from './useFocusRef';
...
-    box-shadow: inset 0 0 0 2px var(--rdg-selection-color);
-    pointer-events: none;
-    z-index: 2;
-  }
-
-  > .${cell}:first-child {
-    /* preserve left border on the first frozen cell after scrolling to the right */
-    box-shadow: inset 2px 0 0 0 var(--rdg-selection-color);
+  > .${cell} {
+    box-shadow: inset 0 2px 0 0 var(--rdg-selection-color),
```
nstepien, line 10: This might interfere with the last-frozen-cell shadow.

src/utils/index.ts
```diff
@@ -18,13 +19,24 @@ export function assertIsValidKeyGetter<R, K extends React.Key>(
...
+}
+
 export function getCellStyle<R, SR>(
   column: CalculatedColumn<R, SR>,
   colSpan?: number
 ): React.CSSProperties {
   return {
     gridColumnStart: column.idx + 1,
     gridColumnEnd: colSpan !== undefined ? `span ${colSpan}` : undefined,
+    gridRowStart: 'var(--rdg-grid-row-start)',
```
nstepien, line 39: Since it's static, can we move this in css`` blocks?

src/style/core.ts
```diff
@@ -38,6 +38,11 @@ const root = css`
   --rdg-selection-color: #66afe9;
   --rdg-font-size: 14px;
 
+  position: relative;
```
nstepien, line 41: The `position: relative;` below for Safari isn't necessary anymore.

src/style/core.ts
```diff
@@ -38,6 +38,11 @@ const root = css`
...
+  display: grid;
+  grid-template-columns: var(--rdg-template-columns);
+  grid-template-rows: var(--rdg-template-rows);
```
nstepien, line 44: It looks like we only use these variables once now, let's use the css properties directly instead.

src/SummaryCell.tsx
```diff
@@ -11,12 +11,14 @@ interface SharedCellRendererProps<R, SR>
 
 interface SummaryCellProps<R, SR> extends SharedCellRendererProps<R, SR> {
   row: SR;
+  bottom: number;
```
nstepien, line 14: image
Summary rows don't stick to the very bottom of the container anymore
It looks like adding `top` might be necessary.

src/HeaderRow.tsx
amanmahajan7: The header does not have a background color if this is removed 
image

src/hooks/useCalculatedColumns.ts
```diff
@@ -175,7 +175,7 @@ export function useCalculatedColumns<R, SR>({
     }
 
     const layoutCssVars: Record<string, string> = {
```
nstepien, line 177:
```suggestion
    const layoutCssVars: React.CSSProperties = {
```

src/SummaryRow.tsx
amanmahajan7: https://github.com/adazzle/react-data-grid/pull/2630

src/HeaderRow.tsx
amanmahajan7: I don't think it ever worked. We never set `aria-selected` on the header row 🤔

src/SummaryCell.tsx
amanmahajan7: image

image

src/hooks/useRovingRowRef.ts
amanmahajan7: Fixed

src/Row.tsx
```diff
@@ -109,8 +103,6 @@ function Row<R, SR>(
   );
 }
 
-export default memo(Row) as <R, SR>(props: RowRendererProps<R, SR>) => JSX.Element;
```
amanmahajan7, line 112: Without `forwardRef`, `ref` is set to `{}` and it fails at runtime. Previously `useCombinedRefs` was handling it

src/DataGrid.tsx
```diff
@@ -1052,6 +1072,21 @@ function DataGrid<R, SR, K extends Key>(
...
+      {/* extra div is needed for row navigation in a treegrid */}
+      {hasGroups && (
+        <div
+          ref={rowRef}
+          tabIndex={selectedPosition.idx === -1 && selectedPosition.rowIdx !== -2 ? 0 : -1}
+          style={{
+            gridColumnStart: 1,
+            gridRowStart: selectedPosition.rowIdx + 2,
+            position: 'sticky',
+            left: 0
```
nstepien, line 1084: Let's add css for the static styles.

src/HeaderRow.tsx
```diff
@@ -90,9 +84,10 @@ function HeaderRow<R, SR, K extends React.Key>({
     <div
       role="row"
       aria-rowindex={1} // aria-rowindex is 1 based
-      ref={ref}
-      tabIndex={tabIndex}
-      className={clsx(headerRowClassname, className)}
+      className={clsx(headerRowClassname, {
+        [rowSelectedClassname]: selectedCellIdx === -1
```
nstepien, line 88: Is it possible to select the whole header row?

src/SummaryRow.tsx
```diff
@@ -2,30 +2,29 @@ import { memo } from 'react';
...
+  top: number | undefined;
+  bottom: number | undefined;
   lastFrozenColumnIndex: number;
   selectedCellIdx: number | undefined;
   selectCell: (row: SR, column: CalculatedColumn<R, SR>) => void;
 }
 
 const summaryRow = css`
-  &.${row} {
+  z-index: 3;
```
nstepien, line 23: Does `z-index` actually work on `display: contents` elements? 🤔

src/SummaryRow.tsx
```diff
@@ -74,16 +73,22 @@ function SummaryRow<R, SR>({
...
-      ref={ref}
-      tabIndex={tabIndex}
       className={clsx(
         rowClassname,
         `rdg-row-${rowIdx % 2 === 0 ? 'even' : 'odd'}`,
         summaryRowClassname,
         { [summaryRowBorderClassname]: rowIdx === 0 },
-        className
+        {
+          [rowSelectedClassname]: selectedCellIdx === -1
```
nstepien, line 82: Are summary rows selectable?

src/HeaderRow.tsx
amanmahajan7: image

If up arrow is pressed on the first grouped row

src/DataGrid.tsx
```diff
@@ -1028,6 +1034,14 @@ function DataGrid<R, SR, K extends Key>(
     setDraggedOverRowIdx(undefined);
   }
 
+  let templateRows = `${headerRowHeight}px`;
+  if (rows.length > 0) {
+    templateRows += gridTemplateRows;
```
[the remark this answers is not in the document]
amanmahajan7: This adds `repeat(0, 35px)` and rows are not displayed correctly
image

src/SummaryRow.tsx
amanmahajan7: Same as header row

src/SummaryRow.tsx
amanmahajan7: It turned out zIndex is not even needed as summary row is the last row

src/style/core.ts
```diff
@@ -92,3 +105,8 @@ const viewportDragging = css`
 `;
 
 export const viewportDraggingClassname = `rdg-viewport-dragging ${viewportDragging}`;
+
+export const focusSinkClassname = css`
+  position: 'sticky';
```
nstepien, line 110:
```suggestion
  position: sticky;
```

amanmahajan7: 🤦‍♂️

nstepien (dismissed):
:shipit:

amanmahajan7:
Last commit fixes 
image

nstepien (approved):

PR #2365, opened 2021-04-21
Multi Column Sorting

jmitchell1132:


src/DataGrid.tsx
```diff
@@ -118,8 +119,10 @@ export interface DataGridProps<R, SR = unknown> extends SharedDivProps {
   sortColumn?: string;
   /** The direction to sort the sortColumn*/
   sortDirection?: SortDirection;
+  /**Used for multi column sorting */
+  sortData?: Sorts[];
   /** Function called whenever grid is sorted*/
-  onSort?: (columnKey: string, direction: SortDirection) => void;
+  onSort?: (columnKey: string, direction: SortDirection, ctrlClick: boolean) => void;
```
amanmahajan7, line 125: I think we can replace these 4 props with 2 new props

```
sortColumns?: readonly Readonly<SortColumn>[];
onSortColumnsChange?: (sortColumns: SortColumn[]) => void;
```

If the sortColumns.length === 1 then it is a single sort 

```
interface SortColumn {
  columnKey: string;
  direction: SortDirection;
}
````

File: src/headerCells/SortableHeaderCell.tsx [hunk 2]
```diff
@@ -34,17 +35,23 @@ export default function SortableHeaderCell<R, SR>({
...
+  
   let sortText = '';
   if (sortDirection === 'ASC') {
     sortText = '\u25B2';
   } else if (sortDirection === 'DESC') {
     sortText = '\u25BC';
   }
 
-  function onClick() {
+  function onClick(e: any) {
```
amanmahajan7, line 54:
```suggestion
  function onClick(event: MouseEvent<HTMLSpanElement>) {
```

src/types.ts
```diff
@@ -224,5 +224,12 @@ export interface GroupRow<TRow> {
...
 
+export interface Sorts {
+  sortColumn: string;
+  direction: SortDirection;
+}
+
 export type CellNavigationMode = 'NONE' | 'CHANGE_ROW' | 'LOOP_OVER_ROW';
 export type SortDirection = 'ASC' | 'DESC' | 'NONE';
+
+
```
amanmahajan7, line 235:
```suggestion
```

amanmahajan7 (changes requested):
A few initial comments

nstepien:
@jmitchell1132 Can you merge `canary` back into your branch to resolve the conflicts?

src/headerCells/SortableHeaderCell.tsx
```diff
@@ -59,12 +52,13 @@ export default function SortableHeaderCell<R, SR>({
...
     }
-    onSort(column.key, direction);
+    onSort(column.key, direction, event.ctrlKey);
   }
 
   return (
-    <span className={headerSortCellClassname} onClick={onClick}>
+    <span className={headerSortCellClassname} onClick={e => onClick(e)}>
       <span className={headerSortNameClassname}>{children}</span>
+      {sortText && <span>{priority}</span>}
```
amanmahajan7, line 61: I think we should only show the priority if there are 2 or more sort columns.

src/DataGrid.tsx
```diff
@@ -187,9 +184,8 @@ function DataGrid<R, SR>({
   // Feature props
   selectedRows,
   onSelectedRowsChange,
-  sortColumn,
-  sortDirection,
-  onSort,
+  sortColumns,
+  onSortColumnsChange,
```
amanmahajan7, line 188: Currently multisort is always enabled. Should we restrict it somehow (enableMultiSort prop)? @nstepien wdyt?

src/HeaderCell.tsx
```diff
@@ -89,13 +87,32 @@ export default function HeaderCell<R, SR>({
     window.addEventListener('pointermove', onPointerMove);
     window.addEventListener('pointerup', onPointerUp);
   }
+  const sortDirection = sortColumns?.find(item => item.columnKey === column.key)?.direction;
+  const index = sortColumns?.findIndex(sort => sort.columnKey === column.key);
+  const priority = index !== undefined ? index + 1 : undefined;
+
+  const onSort = (columnKey: string, direction: SortDirection, ctrlClick: boolean) => {
```
amanmahajan7, line 94: I have seen a few grids using `shift` instead of `ctrl` for multi-sorting. I wonder if we should use a generic parameter like `isMulti` instead of `ctrlClick` and then it can be customized using `headerRender` as needed

stories/demos/CommonFeatures.tsx
```diff
@@ -223,7 +224,7 @@ function createRows(): readonly Row[] {
 
 export function CommonFeatures() {
   const [rows, setRows] = useState(createRows);
-  const [[sortColumn, sortDirection], setSort] = useState<[string, SortDirection]>(['id', 'NONE']);
+  const [sortColumns, setSortColumns] = useState<SortColumn[]>([]);
```
amanmahajan7, line 227: I think we can create a separate example. In the future we can group examples like
- Column
  - ColSpan
  - Multisort
  - Reorder
- Row
  - Group
  - Reorder
  - NoRows
etc

amanmahajan7 (commented):
Looks great. A few comments and discussion items

src/headerCells/SortableHeaderCell.tsx
```diff
@@ -59,12 +52,13 @@ export default function SortableHeaderCell<R, SR>({
         direction = sortDescendingFirst ? 'DESC' : 'ASC';
         break;
     }
-    onSort(column.key, direction);
+    onSort(column.key, direction, event.ctrlKey);
   }
 
   return (
-    <span className={headerSortCellClassname} onClick={onClick}>
+    <span className={headerSortCellClassname} onClick={e => onClick(e)}>
```
amanmahajan7, line 59:
```suggestion
    <span className={headerSortCellClassname} onClick={onClick}>
```

src/headerCells/SortableHeaderCell.tsx
amanmahajan7: This is pending ☝️

src/HeaderCell.tsx
```diff
@@ -89,13 +87,32 @@ export default function HeaderCell<R, SR>({
...
+    if (!onSortColumnsChange) return;
+    if (ctrlClick) {
+      const newSorts = sortColumns ? [...sortColumns] : [];
+      const index = newSorts.findIndex((sort) => sort.columnKey === columnKey);
+      if (index > -1) {
+        const sort = newSorts.find(sort => sort.columnKey === columnKey);
+        if (sort?.direction === 'ASC') newSorts[index] = { columnKey, direction };
```
nstepien, line 101: Same idea as above
```suggestion
        if (sortDirection === 'ASC') newSorts[index] = { columnKey, direction };
```

nstepien (changes requested):
I'm not done reviewing this, I'll continue at another time.

src/HeaderCell.tsx
```diff
@@ -89,13 +87,32 @@ export default function HeaderCell<R, SR>({
...
+    } else if (direction === 'NONE') onSortColumnsChange([]);
+    else onSortColumnsChange([{ columnKey, direction }]);
+  };
 
   function getCell() {
     if (column.headerRenderer) {
       return (
         <column.headerRenderer
           column={column}
-          sortColumn={sortColumn}
```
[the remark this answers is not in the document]
amanmahajan7: I think we should pass `priority` so users can use custom sort renderer

src/HeaderCell.tsx
```diff
@@ -128,7 +145,7 @@ export default function HeaderCell<R, SR>({
     <div
       role="columnheader"
       aria-colindex={column.idx + 1}
-      aria-sort={sortColumn === column.key ? getAriaSort(sortDirection) : undefined}
+      aria-sort={sortDirection ? getAriaSort(sortDirection) : undefined}
```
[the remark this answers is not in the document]
amanmahajan7: I think we can do that. The only thing is `aria-sort` should only be applied to 1 column from what I understand. But let's just add it to any column that is sorted. It will be helpful for testing

https://www.w3.org/TR/wai-aria-1.2/#dom-ariamixin-ariasort
> Authors SHOULD only apply this property to table headers or grid headers. If the property is not provided, there is no defined sort order. For each table or grid, authors SHOULD apply aria-sort to only one header at a time.

https://www.w3.org/TR/wai-aria-practices-1.2/#gridAndTableProperties_sort
> It is important to note that ARIA does not provide a way to indicate levels of sort for data sets that have multiple sort keys. Thus, there is limited value to applying aria-sort with a value other than none to more than one column or row.

src/DataGrid.tsx
```diff
@@ -127,12 +127,9 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
...
   onSelectedRowsChange?: ((selectedRows: Set<K>) => void) | null;
-  /** The key of the column which is currently being sorted */
-  sortColumn?: string | null;
-  /** The direction to sort the sortColumn*/
-  sortDirection?: SortDirection | null;
-  /** Function called whenever grid is sorted*/
-  onSort?: ((columnKey: string, direction: SortDirection) => void) | null;
+  /**Used for multi column sorting */
+  sortColumns?: readonly Readonly<SortColumn>[];
+  onSortColumnsChange?: (sortColumns: SortColumn[]) => void;
```
nstepien, line 132: Let's allow `null` as well (more context: https://github.com/adazzle/react-data-grid/pull/2406)
```suggestion
  sortColumns?: readonly Readonly<SortColumn>[] | null;
  onSortColumnsChange?: ((sortColumns: SortColumn[]) => void) | null;
```

nstepien (changes requested):

src/types.ts
```diff
@@ -116,9 +116,9 @@ export interface EditorProps<TRow, TSummaryRow = unknown> extends SharedEditorPr
 
 export interface HeaderRendererProps<TRow, TSummaryRow = unknown> {
   column: CalculatedColumn<TRow, TSummaryRow>;
-  sortColumn: string | undefined | null;
-  sortDirection: SortDirection | undefined | null;
-  onSort: ((columnKey: string, direction: SortDirection) => void) | undefined | null;
+  sortDirection?: SortDirection | undefined | null;
+  priority?: number;
+  onSort?: (columnKey: string, direction: SortDirection, ctrlClick: boolean) => void;
```
nstepien, line 121: - in HeaderCell, we should always pass `sortDirection`/`priority`/`onSort` so they're not really optional (`?:`)
- `sortDirection` is never `null`
- `onSort` is never `undefined`
```suggestion
  sortDirection: SortDirection | undefined;
  priority: number | undefined;
  onSort: (columnKey: string, direction: SortDirection, ctrlClick: boolean) => void;
```

nstepien (changes requested):

src/HeaderCell.tsx
```diff
@@ -88,14 +87,33 @@ export default function HeaderCell<R, SR>({
     window.addEventListener('pointermove', onPointerMove);
     window.addEventListener('pointerup', onPointerUp);
   }
+  const sortDirection = sortColumns?.find((item) => item.columnKey === column.key)?.direction;
+  const index = sortColumns?.findIndex((sort) => sort.columnKey === column.key);
```
amanmahajan7, line 91: https://github.com/adazzle/react-data-grid/pull/2365#discussion_r627632766

src/HeaderCell.tsx
```diff
@@ -88,14 +87,33 @@ export default function HeaderCell<R, SR>({
...
+  const priority = index !== undefined ? index + 1 : undefined;
+  const multiSorts = sortColumns ? sortColumns.length > 1 : false;
+
+  const onSort = (columnKey: string, direction: SortDirection, ctrlClick: boolean) => {
+    if (!onSortColumnsChange) return;
+    if (ctrlClick) {
+      const newSorts = sortColumns ? [...sortColumns] : [];
+      if (index! > -1) {
```
amanmahajan7, line 99: Index can be `undefined`, no?

stories/demos/CommonFeatures.tsx
```diff
@@ -220,7 +221,7 @@ function createRows(): readonly Row[] {
...
```
nstepien, line 224:
```suggestion
  const [sortColumns, setSortColumns] = useState<readonly SortColumn[]>([]);
```

src/headerCells/SortableHeaderCell.tsx
```diff
@@ -57,12 +53,13 @@ export default function SortableHeaderCell<R, SR>({
...
     }
-    onSort(column.key, direction);
+    onSort(column.key, direction, event.ctrlKey);
   }
 
   return (
     <span className={headerSortCellClassname} onClick={onClick}>
       <span className={headerSortNameClassname}>{children}</span>
+      {sortText && <span>{priority}</span>}
       <span>{sortText}</span>
```
nstepien, line 63: This can be simplified. We also need to make sure we only pass a number `priority` when the column is sorted and we have more than one column sorted. Should be `undefined` otherwise.
```suggestion
      <span>
        {priority}
        {sortText}
      </span>
```

nstepien (changes requested):

stories/demos/ColumnsReordering.tsx
```diff
@@ -66,15 +66,10 @@ function createColumns(): Column<Row>[] {
 export function ColumnsReordering() {
   const [rows] = useState(createRows);
   const [columns, setColumns] = useState(createColumns);
-  const [[sortColumn, sortDirection], setSort] = useState<[string, SortDirection]>([
-    'task',
-    'NONE'
+  const [sortColumns, setSortColumns] = useState<readonly Readonly<SortColumn>[]>([
+    { columnKey: 'task', direction: 'NONE' }
```
nstepien, line 70: When we have `direction: 'NONE'`, it's possible to add another sort column, but we still display the other's priority.
IMO we should just remove `NONE` as a possible sort direction. Just remove the `SortColumn` instead.
Sounds good @amanmahajan7?

image

stories/demos/CommonFeatures.tsx
```diff
@@ -253,29 +250,42 @@ export function CommonFeatures() {
...
+          return a[sortColumn].localeCompare(b[sortColumn]);
+        };
       case 'available':
-        sortedRows = sortedRows.sort((a, b) =>
-          a[sortColumn] === b[sortColumn] ? 0 : a[sortColumn] ? 1 : -1
-        );
-        break;
+        return function (a: Row, b: Row) {
+          return a[sortColumn] === b[sortColumn] ? 0 : a[sortColumn] ? 1 : -1;
+        };
```
nstepien, line 259:
```suggestion
        return (a, b) => a[sortColumn] === b[sortColumn] ? 0 : a[sortColumn] ? 1 : -1;
```

stories/demos/CommonFeatures.tsx
```diff
@@ -253,29 +250,42 @@ export function CommonFeatures() {
...
       case 'id':
       case 'progress':
       case 'startTimestamp':
       case 'endTimestamp':
       case 'budget':
-        sortedRows = sortedRows.sort((a, b) => a[sortColumn] - b[sortColumn]);
-        break;
+        return function (a: Row, b: Row) {
+          return a[sortColumn] - b[sortColumn];
+        };
```
nstepien, line 267:
```suggestion
        return (a, b) => a[sortColumn] - b[sortColumn];
```

stories/demos/CommonFeatures.tsx
```diff
@@ -253,29 +250,42 @@ export function CommonFeatures() {
...
       default:
+        throw new Error(`unsupported sortColumn: "${sortColumn}"`);
     }
+  };
 
-    return sortDirection === 'DESC' ? sortedRows.reverse() : sortedRows;
-  }, [rows, sortDirection, sortColumn]);
+  const sortedRows: readonly Row[] = useMemo(() => {
```
nstepien, line 273: Nitpicky, but moving the type closer to where a type issue may occur will give us better type errors.
```suggestion
  const sortedRows = useMemo((): readonly Row[] => {
```

stories/demos/CommonFeatures.tsx
```diff
@@ -1,12 +1,13 @@
-import { useState, useCallback, useMemo } from 'react';
+import { useState, useMemo } from 'react';
 import { css } from '@linaria/core';
 import faker from 'faker';
 
 import DataGrid, { SelectColumn, TextEditor, SelectCellFormatter } from '../../src';
-import type { Column, SortDirection } from '../../src';
+import type { Column } from '../../src';
```
nstepien, line 6: `SortColumn` is already exported at the index:
```suggestion
import type { Column, SortColumn } from '../../src';
```

stories/demos/CommonFeatures.tsx
```diff
@@ -220,7 +221,7 @@ function createRows(): readonly Row[] {
 
 export function CommonFeatures() {
   const [rows, setRows] = useState(createRows);
-  const [[sortColumn, sortDirection], setSort] = useState<[string, SortDirection]>(['id', 'NONE']);
+  const [sortColumns, setSortColumns] = useState<readonly SortColumn[]>([]);
```
nstepien, line 224:
```suggestion
  const [sortColumns, setSortColumns] = useState<readonly Readonly<SortColumn>[]>([]);
```

nstepien (changes requested):
I found a bug:

stories/demos/ColumnsReordering.tsx
amanmahajan7: Sounds good to me

src/headerCells/SortableHeaderCell.tsx
```diff
@@ -1,7 +1,6 @@
 import { css } from '@linaria/core';
-import type { HeaderCellProps } from '../HeaderCell';
-import type { SortDirection } from '../types';
-
+import type { HeaderRendererProps, SortDirection } from '../types';
```
nstepien, line 2:
```suggestion
import type { HeaderRendererProps } from '../types';
```

src/HeaderCell.tsx
```diff
@@ -88,14 +87,47 @@ export default function HeaderCell<R, SR>({
...
+        onSortColumnsChange([...newSorts]);
+      }
+    } else if (sortDirection === 'DESC') {
+      if (sortDescendingFirst) onSortColumnsChange([{ columnKey, direction }]);
+      else onSortColumnsChange([]);
+    } else if (sortDirection === 'ASC') {
+      if (sortDescendingFirst) onSortColumnsChange([]);
+      else onSortColumnsChange([{ columnKey, direction }]);
+    } else onSortColumnsChange([{ columnKey, direction }]);
+  };
```
nstepien, line 122: WDYT? 🤔 
```suggestion

  const sortIndex = sortColumns?.findIndex((sort) => sort.columnKey === column.key);
  const sortColumn =
    sortIndex !== undefined && sortIndex > -1 ? sortColumns![sortIndex] : undefined;
  const sortDirection = sortColumn?.direction;
  const priority = sortColumn !== undefined && sortColumns!.length > 1 ? sortIndex! + 1 : undefined;
  const ariaSort =
    sortDirection && !priority ? (sortDirection === 'ASC' ? 'ascending' : 'descending') : undefined;

  function onSort(ctrlClick: boolean) {
    if (onSortColumnsChange == null) return;

    const { sortDescendingFirst } = column;

    if (sortColumn === undefined) {
      // not currently sorted
      const nextSort: SortColumn = {
        columnKey: column.key,
        direction: sortDescendingFirst ? 'DESC' : 'ASC'
      };
      onSortColumnsChange(sortColumns && ctrlClick ? [...sortColumns, nextSort] : [nextSort]);
    } else {
      const nextSortColumns = [...sortColumns!];
      if (
        (sortDescendingFirst && sortDirection === 'ASC') ||
        (!sortDescendingFirst && sortDirection === 'DESC')
      ) {
        // remove sort
        nextSortColumns.splice(sortIndex!, 1);
      } else {
        // swap direction
        nextSortColumns[sortIndex!] = {
          columnKey: column.key,
          direction: sortDirection === 'ASC' ? 'DESC' : 'ASC'
        };
      }
      onSortColumnsChange(nextSortColumns);
    }
  }
```

src/HeaderCell.tsx
```diff
@@ -88,14 +87,47 @@ export default function HeaderCell<R, SR>({
...
 
   function getCell() {
     if (column.headerRenderer) {
       return (
         <column.headerRenderer
           column={column}
-          sortColumn={sortColumn}
           sortDirection={sortDirection}
+          priority={multiSorts && sorted ? priority : undefined}
```
nstepien, line 130:
```suggestion
          priority={priority}
```

src/HeaderCell.tsx
```diff
@@ -108,8 +140,8 @@ export default function HeaderCell<R, SR>({
         <SortableHeaderCell
           column={column}
```
nstepien, line 141:
```suggestion
```

src/HeaderCell.tsx
```diff
@@ -108,8 +140,8 @@ export default function HeaderCell<R, SR>({
...
           onSort={onSort}
-          sortColumn={sortColumn}
           sortDirection={sortDirection}
+          priority={multiSorts && sorted ? priority : undefined}
```
nstepien, line 144:
```suggestion
          priority={priority}
```

src/HeaderCell.tsx
```diff
@@ -127,7 +159,7 @@ export default function HeaderCell<R, SR>({
     <div
       role="columnheader"
       aria-colindex={column.idx + 1}
-      aria-sort={sortColumn === column.key ? getAriaSort(sortDirection) : undefined}
+      aria-sort={!multiSorts ? getAriaSort(sortDirection) : undefined}
```
nstepien, line 162:
```suggestion
      aria-sort={ariaSort}
```

nstepien (changes requested):
IMO it doesn't make sense to check for `direction`+`sortDescendingFirst` both in `SortableHeaderCell.tsx` and in `HeaderCell.tsx`.
Let's simplify and refactor! Let me know what you think

src/HeaderCell.tsx
nstepien: In `onSort`:
before: 11 `if`s + 9 `else`s
after: 3 `if`s + 2 `else`s
I'd say that's an improvement 🤔

src/HeaderCell.tsx
```diff
@@ -88,14 +76,58 @@ export default function HeaderCell<R, SR>({
     window.addEventListener('pointermove', onPointerMove);
     window.addEventListener('pointerup', onPointerUp);
   }
+  const sortIndex = sortColumns?.findIndex((sort) => sort.columnKey === column.key);
+  const sortColumn =
+    sortIndex !== undefined && sortIndex > -1 ? sortColumns![sortIndex] : undefined;
+  const sortDirection = sortColumn?.direction;
+  const priority = sortColumn !== undefined && sortColumns!.length > 1 ? sortIndex! + 1 : undefined;
+  const ariaSort =
+    sortDirection && !priority ? (sortDirection === 'ASC' ? 'ascending' : 'descending') : undefined;
```
nstepien, line 85: Question for @amanmahajan7, must we avoid setting `aria-sort` when we have more than one sorted column?

test/multiColumnSort.test.tsx
```diff
@@ -0,0 +1,154 @@
+import userEvent from '@testing-library/user-event';
+import { getCellsAtColumn, getHeaderCells } from './utils';
+import type { Column, SortColumn } from '../src/types';
+import { StrictMode, useMemo, useState } from 'react';
```
nstepien, line 4:
```suggestion
import { StrictMode, useState } from 'react';
```

test/multiColumnSort.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+  { id: 6, area: 'Brand', assignee: 'Tony Reinger', budget: 700, transaction: 'invoice' }
+];
+
+const newColumns: readonly Column<Row>[] = [
+  { key: 'id', name: 'id' },
+  { key: 'area', name: 'area' },
+  { key: 'assignee', name: 'assignee' },
+  { key: 'budget', name: 'budget' },
+  { key: 'transaction', name: 'transaction' }
+];
```
nstepien, line 50:
```suggestion
const columns: readonly Column<SortColumn>[] = [
  { key: 'columnKey', name: 'columnKey' },
  { key: 'direction', name: 'direction' }
];
```

test/multiColumnSort.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+        sortable: true,
+        resizable: true
+      }}
+      onRowsChange={setRows}
+      sortColumns={sortColumns}
+      onSortColumnsChange={setSortColumns}
+      className="fill-grid"
+    />
+  );
+};
```
nstepien, line 109:
```suggestion
function TestGrid() {
  const [sortColumns, setSortColumns] = useState<readonly SortColumn[]>([]);

  return (
    <DataGrid
      columns={columns}
      rows={sortColumns}
      defaultColumnOptions={{ sortable: true }}
      sortColumns={sortColumns}
      onSortColumnsChange={setSortColumns}
    />
  );
}
```

test/multiColumnSort.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+function setupNewGrid(props: Props) {
+  return render(
+    <StrictMode>
+      <SortingContainer {...props} />
```
nstepien, line 114:
```suggestion
function setup() {
  return render(
    <StrictMode>
      <TestGrid />
```

test/multiColumnSort.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+    </StrictMode>
+  );
+}
+
+test('single column sort', async () => {
+  setupNewGrid({ newRows, newColumns });
```
nstepien, line 120:
```suggestion
  setup();
```

test/multiColumnSort.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  //Sort by Assignee - ASC
+  userEvent.click(getHeaderCells()[2].firstElementChild!);
+  expect(assigneeColumn[0]).toHaveTextContent('Brad Braun');
+  expect(assigneeColumn[1]).toHaveTextContent('Janice Stehr');
+  expect(assigneeColumn[2]).toHaveTextContent('Keith McGlynn');
+});
+
+test('multi column sort', async () => {
+  setupNewGrid({ newRows, newColumns });
```
nstepien, line 132:
```suggestion
  setup();
```

nstepien (changes requested):
IMO it's overkill to actually implement sorting in tests. We can get away with just passing `sortColumns` to `rows`. It's a bit meta, but it'll allow us to test/assert the exact value of `sortColumns`.
We just need to test that the header cells are rendered correctly, and that they behave as expected.

What do you think?

src/HeaderCell.tsx
amanmahajan7: https://github.com/adazzle/react-data-grid/pull/2365#discussion_r631951078

test/utils.tsx
```diff
@@ -57,3 +60,17 @@ export function validateCellPosition(columnIdx: number, rowIdx: number) {
...
+
+export function getCellsAtColumn({ grid, name }: { grid: HTMLElement; name: Name }) {
+  const headerCell = within(grid).getByRole('columnheader', { name }, { hidden: true });
+  const index = getColumnIndex(headerCell);
+  const rows = getRows();
+  return rows.map((row) => {
+    const cells = within(row).getAllByRole('gridcell', { hidden: true });
+    return cells.find((c) => getColumnIndex(c) === index);
+  });
+}
```
nstepien, line 76: We can remove these now as they're not used.

nstepien (changes requested):
LGTM, just 1 more change.

nstepien (approved):
🎉

src/HeaderCell.tsx
```diff
@@ -88,14 +76,58 @@ export default function HeaderCell<R, SR>({
...
+        nextSortColumn = {
+          columnKey: column.key,
+          direction: sortDirection === 'ASC' ? 'DESC' : 'ASC'
+        };
+      }
+      if (ctrlClick) {
+        const nextSortColumns = [...sortColumns!];
+        if (nextSortColumn) {
+          // swap direction
+          nextSortColumns[sortIndex!] = nextSortColumn;
```
amanmahajan7, line 112: There are many places where we are using `!` and it can be confusing when it not undefined. I usually prefer explicit checks but this is fine for now

src/HeaderCell.tsx
```diff
@@ -88,14 +76,58 @@ export default function HeaderCell<R, SR>({
...
+        } else {
+          // remove sort
+          nextSortColumns.splice(sortIndex!, 1);
```
amanmahajan7, line 115: There are many places where we are using `!` and it can be confusing when it not undefined. I usually prefer explicit checks but this is fine for now

test/multiColumnSort.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+});
+
+test('multi column sort', () => {
+  setup();
+  userEvent.click(getHeaderCells()[0].firstElementChild!);
+  userEvent.click(getHeaderCells()[1].firstElementChild!, { ctrlKey: true });
+  expect(JSON.parse(screen.getByTestId('sortColumnsValue')!.textContent!)).toStrictEqual([
+    { columnKey: 'columnKey', direction: 'ASC' },
+    { columnKey: 'direction', direction: 'ASC' }
+  ]);
```
amanmahajan7, line 53: We should also check if `DESC` and removing sort works also but we can do it in a separate PR

amanmahajan7 (approved):
👏 great work

PR #2370, opened 2021-04-22
Linaria: disable autoprefixing

nstepien:
https://github.com/callstack/linaria/issues/57#issuecomment-469611837

Linaria/Stylis add a LOT of prefixed css properties, when we don't need any of them, so I disabled them. I don't think we need any prefixes anymore nowadays. And if we do I'd rather we manage it ourselves. Prefixes is a thing of the past and browsers don't add new ones anymore.
The css properties in the devtools will be less polluted now.
We don't even have any prefixes in our codebase anymore afaict.

Build output example:
image
image
image

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -5,11 +5,7 @@
       "bugfixes": true,
       "shippedProposals": true,
       "corejs": 3,
-      "useBuiltIns": "entry",
-      "include": [
-        "@babel/proposal-nullish-coalescing-operator",
-        "@babel/proposal-optional-chaining"
-      ]
+      "useBuiltIns": "entry"
     }],
     ["@babel/react", { "runtime": "automatic" }],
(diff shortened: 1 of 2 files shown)


amanmahajan7 (dismissed):

babel.config.json
```diff
@@ -5,11 +5,7 @@
       "bugfixes": true,
       "shippedProposals": true,
       "corejs": 3,
-      "useBuiltIns": "entry",
-      "include": [
-        "@babel/proposal-nullish-coalescing-operator",
-        "@babel/proposal-optional-chaining"
```
nstepien, line 11: These are included anyway as long as we support "maintained node versions"
https://github.com/adazzle/react-data-grid/blob/canary/browserslist#L5

amanmahajan7 (approved):

PR #2379, opened 2021-04-27
Add aria-readonly to readonly cells 

amanmahajan7:
https://www.w3.org/TR/wai-aria-practices-1.2/#grid_roles_states_props

> If the grid provides content editing functionality and contains cells that may have edit capabilities disabled in certain conditions, aria-readonly may be set true on cells where editing is disabled. If edit functions are disabled for all cells, aria-readonly may be set true on the grid element. Grids that do not provide editing functions do not include the aria-readonly attribute on any of their elements.

### DIFF
--- a/src/Cell.tsx
+++ b/src/Cell.tsx
@@ -107,7 +107,7 @@ function Cell<R, SR>({
       aria-colindex={column.idx + 1} // aria-colindex is 1-based
       aria-selected={isCellSelected}
       aria-colspan={colSpan}
-      aria-readonly={isCellEditable(column, row) ? true : undefined}
+      aria-readonly={isCellEditable(column, row) || undefined}
       ref={ref}
       className={className}
       style={getCellStyle(column, colSpan)}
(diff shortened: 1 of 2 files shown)


src/Cell.tsx
```diff
@@ -107,6 +107,7 @@ function Cell<R, SR>({
       aria-colindex={column.idx + 1} // aria-colindex is 1-based
       aria-selected={isCellSelected}
       aria-colspan={colSpan}
+      aria-readonly={isCellEditable(column, row) ? true : undefined}
```
nstepien, line 110: No?
```suggestion
      aria-readonly={isCellEditable(column, row)}
```

amanmahajan7: IIRC, it is usually recommended to skip setting `false` value. I check a few grids and found 1 that uses `aria-readonly` and it does not set it when value is editable
https://www.grapecity.com/wijmo/demos/Grid/Editing/Read-only/purejs

nstepien: 🤔 
```suggestion
      aria-readonly={isCellEditable(column, row) || undefined}
```

nstepien (approved):

PR #2386, opened 2021-05-03
Use aria-selected to style selected cell/rows

amanmahajan7:
[Reach UI](https://github.com/reach/reach-ui/blob/main/packages/combobox/styles.css#L24) also uses a similar strategy and it makes sense IMO as it forces us to design accessible components.

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -880,12 +880,14 @@ function DataGrid<R, SR>({
       const top = rowIdx * rowHeight + totalHeaderHeight;
       if (isGroupRow(row)) {
         ({ startRowIndex } = row);
+        const isGroupRowSelected = isSelectable && row.childRows.every(cr => selectedRows?.has(rowKeyGetter!(cr)));
         rowElements.push(
           <GroupRowRenderer<R, SR>
             aria-level={row.level + 1} // aria-level is 1-based
             aria-setsize={row.setSize}
             aria-posinset={row.posInSet + 1} // aria-posinset is 1-based
             aria-rowindex={headerRowsCount + startRowIndex + 1} // aria-rowindex is 1 based
+            aria-selected={isSelectable ? isGroupRowSelected : undefined}
             key={row.id}
             id={row.id}
             groupKey={row.groupKey}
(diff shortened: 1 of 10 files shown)


src/style/grouprow.ts
```diff
@@ -1,9 +1,8 @@
 import { css } from '@linaria/core';
-import { rowSelected } from './row';
 import { cell, cellFrozenLast } from './cell';
 
 const groupRow = css`
-  &:not(.${rowSelected}) {
+  &:not(aria-selected="true") {
```
nstepien, line 5: I don't think this works without the brackets.
```suggestion
  &:not([aria-selected="true"]) {
```

nstepien (changes requested):

nstepien (approved):

amanmahajan7:
Found a bug with this approach 😎

src/DataGrid.tsx
```diff
@@ -880,12 +880,14 @@ function DataGrid<R, SR>({
...
       if (isGroupRow(row)) {
         ({ startRowIndex } = row);
+        const isGroupRowSelected = isSelectable && row.childRows.every(cr => selectedRows?.has(rowKeyGetter!(cr)));
         rowElements.push(
           <GroupRowRenderer<R, SR>
             aria-level={row.level + 1} // aria-level is 1-based
             aria-setsize={row.setSize}
             aria-posinset={row.posInSet + 1} // aria-posinset is 1-based
             aria-rowindex={headerRowsCount + startRowIndex + 1} // aria-rowindex is 1 based
+            aria-selected={isSelectable ? isGroupRowSelected : undefined}
```
nstepien, line 890: This will also avoid `aria-selected="false"`
```suggestion
            aria-selected={isGroupRowSelected || undefined}
```

nstepien (changes requested):

src/DataGrid.tsx
amanmahajan7: > If the grid supports selection, when a cell or row is selected, the selected element has aria-selected set true. If the grid supports column selection and a column is selected, all cells in the column have aria-selected set to true.
https://www.w3.org/TR/wai-aria-practices/#grid_roles_states_props

From what I understand it should only be set if `isSelectable ` is true

nstepien: Yes that's in agreement with my suggestion.

nstepien (approved):

PR #2387, opened 2021-05-05
Remove focusSink and attach keydown event on the root element

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -1027,6 +1027,24 @@ function DataGrid<R, SR>({
       )}
       {rows.length === 0 && EmptyRowsRenderer ? <EmptyRowsRenderer /> : (
         <>
+          {/*
+            Ideally the focus should be set on the selected cell
+            (https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_roving_tabindex)
+            but there are a few issues with this approach
+            - onKeyDown/useLayoutEffect on the cell is not fast enough and grid
+              looses focus if tab is pressed repeatedly
```
nstepien, line 1035: Would it work if we managed focus in a `ref` callback instead of `useLayoutEffect`? 🤔

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: Not sure. We can check

PR #2389, opened 2021-05-06
ResizeObserver: Add workaround for odd grid dimensions and zoom levels leading to "flickering" scrollbars

nstepien:


amanmahajan7 (approved):

ekeuus:
We are still seeing this in production for Windows users.

Should the fix rather be to round the width down/up always to prevent the flickering? The modulo of 0.5 seems kind of strange and highly specific imo.

I am unable to reproduce it even in a windows VM so I do not know the exact steps to repro. If either of you knows how to reproduce it code-wise, I would happily write tests and fix the issue.

@nstepien @amanmahajan7

PR #2393, opened 2021-05-11
Feature: Column Groups

amanmahajan7:


ekeuus:
I think in general the API looks good. I just think it would be nice to have more options on the "parent" header cells as well. For example the class option.

Should I go ahead with the implementation to also render the header rows again?

ekeuus:
Since there might be varying tree depths in various column groups, a flatter structure would help. What I think should maybe be the return value for the rendered header column definitions, should be like an array of arrays of cells to make iterating through them easier when rendering.

alexey-koren:
@amanmahajan7 @ekeuus Hello. Any updates about this feature?

PR #2146, opened 2020-09-04
Add a ResizeObserver polyfill for the storybook

nstepien:
https://www.npmjs.com/package/@juggle/resize-observer

PR #2398, opened 2021-05-13
Update deps

nstepien:
Nothing noteworthy.

I've left react-select on the previous major version for a while, let's get rid of it some day.

We could also remove the ResizeObserver polyfill, wdyt?
https://caniuse.com/resizeobserver
For reference we added it in September: https://github.com/adazzle/react-data-grid/pull/2146
We don't use the callback args, so we should be free of any browser discrepancy in that regard. And we only [conditionally use it](https://github.com/adazzle/react-data-grid/blob/canary/.storybook/preview.tsx#L5), so I wonder why we needed it in the first place.


amanmahajan7:
> We could also remove the ResizeObserver polyfill, wdyt?

I remember grid was not loading on safari/iphone safari. We can try removing it and check again

amanmahajan7 (approved):

PR #2406, opened 2021-05-14
Allow optional DataGrid props to be null

nstepien:
Allowing `null` for optional props means this pattern won't be necessary anymore:
```tsx
<DataGrid optionalProp={maybeNull ?? undefined} />
```
`??`, `??=`, `?.`, and `!= null` makes this trivial thankfully.
Only downside I found was with default function params, `param = default` doesn't replace `null`s.
Non-ts users should get fewer surprises as well.

I've also fixed some missed spots with the new `K` generic.

`Column` props, and maybe other props will be done in separate PRs.

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -103,78 +103,78 @@ export interface DataGridProps<R, SR = unknown, K extends React.Key = React.Key>
    * Rows to be pinned at the bottom of the rows view for summary, the vertical scroll bar will not scroll these rows.
    * Bottom horizontal scroll bar can move the row left / right. Or a customized row renderer can be used to disabled the scrolling support.
    */
-  summaryRows?: readonly SR[];
+  summaryRows?: readonly SR[] | null;
   /** The getter should return a unique key for each row */
-  rowKeyGetter?: (row: R) => K;
-  onRowsChange?: (rows: R[], data: RowsChangeData<R, SR>) => void;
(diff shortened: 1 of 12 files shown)


src/DataGrid.tsx
```diff
@@ -221,21 +221,34 @@ function DataGrid<R, SR>(
...
+  }: DataGridProps<R, SR, K>,
   ref: React.Ref<DataGridHandle>
 ) {
+  /**
+   * defaults
+   */
+  rowHeight ??= 35;
+  headerRowHeight ??= typeof rowHeight === 'number' ? rowHeight : 35;
+  headerFiltersHeight ??= 45;
+  const summaryRowHeight = rawSummaryRowHeight ?? (typeof rowHeight === 'number' ? rowHeight : 35);
```
nstepien, line 245: There is a little bit of nuance here vs default params, if a param isn't defaulted in the parameter list, then it keeps its original type in functions, as TS cannot know that those functions are only called after the `??=` line. I wish we could use `??=` in the param list.
```ts
function fnA(param = 1) {
  param; // type; number

  function test() {
    param; // type: number
  }
}

function fnB(param: number | undefined | null) {
  param; // type: number | undefined | null
  param ??= 1;
  param; // type: number

  function test() {
    param; // type: number | undefined | null
    // this is because we could call `test` before the line where we set it to 1
  }
}
```

amanmahajan7 (approved):

Issue #2414, opened 2021-05-19
auto commit on lost focus when editing?


peasonokay:
Hi, so when i allow users to edit fields they tab from column to column and the lost focus event fires which updates the cell value fine. However on the last box i'm having issues because people dont press enter on the last field (which would commit the change) but press the commit button ive added to the toolbar (to send changes to server)- result - the changes they made in last column are not saved.

So, how to force the automatic commit when moving to another control? is there some way to keep track of the active cell 

code sample:. Here i have a class "customeditor"  . i check the data type and if its timestamp call a datepicker otherwise use a simple input box. As you can see in the handleChange event im correctly calling the this.props.onCommit event every time the value changes and it works as long as i move the focus out of the editor but as above, if i click another control outside the editor/data grid then the commit is no happenning. whats bizaare is if the say type "123" then three handleChange events are fired so at worse i would expect the underlying cell value to contain the first 2 key presses but no it still contains the original value. So **is the commit only happening after lost focus**.
Any help on which event i need to track/replicate to force the commit would be VERY helpful.

    handleChange = (e) => {
        var type= this.state.cellType;
        console.log("editor type:" + type);
        if (type==="timestamp") {
            var conv = moment(e).format(CONSTS.DB2_TIMESTAMP);
            this.setState({dateTime: conv}, () => this.props.onCommit);
        } else {
            this.setState({stdText: e.target.value}, () => this.props.onCommit);
        }
        console.log(e);
    };

    dateEditor() {
        return (
            <>
            <DateTimePicker
                id="datetimepicker-1"
                value={this.fromTimestamp(this.state.dateTime)}
                onChange={this.handleChange}
                formatStyle="medium"
                // required
                // error="Date/time(stamp) is required"
                okLabel={"OK"}
                cancelLabel={"Cancel"}
            />
            </>
        );
    };

    standardEditor() {
        return (
            <input
                className="inputBox"
                id="simpedit-1"
                value={this.state.stdText}
                onChange={this.handleChange}
                >
            </input>
        );
    };

nstepien:
Could you try `7.0.0-canary.48`? I've tweaked some internals in #2415, hopefully that should address it.

nstepien:
I'd also suggest adding the `autoFocus` attribute on your `input`, + `onBlur={() => props.onClose(true)}` to automatically close and commit the editor when focusing away from it.

peasonokay:
[...]

PR #2415, opened 2021-05-19
Fix "click outside + immediate editor commit" edge case by swiching to checking for mousedown events

nstepien:
Related: https://github.com/adazzle/react-data-grid/issues/2414

I had also tried to fix this by adding an extra "sync" check on the bubble phase of the `click` event, but that wouldn't work as the `click` on the external save button would happen first, and thus wouldn't get the updated state.


amanmahajan7 (approved):
LGTM

PR #2416, opened 2021-05-20
Tweak codecov config

nstepien:
https://docs.codecov.io/docs/codecovyml-reference

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -9,7 +9,7 @@ coverage:
 # https://docs.codecov.io/docs/pull-request-comments
 comment:
   layout: 'diff, files'
-  require_changes: false
+  require_changes: true
 
 # https://docs.codecov.io/docs/node
 parsers:
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,6 +1,6 @@
 comment:
   layout: 'diff, files'
-  require_changes: true
+  require_changes: false
 
 parsers:
   javascript:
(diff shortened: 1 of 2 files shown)


.github/codecov.yml
```diff
@@ -1,7 +1,17 @@
+# https://docs.codecov.io/docs/commit-status
+coverage:
+  status:
+    project:
+      default:
+        informational: true
+    patch: off
```
nstepien, line 7: Annotations are a bit overkill, so I've disabled them with `patch: off`, and I don't think we care too much about the coverage of a PR's diff.

image

image

image

.github/codecov.yml
```diff
@@ -1,7 +1,17 @@
+# https://docs.codecov.io/docs/commit-status
+coverage:
+  status:
+    project:
+      default:
+        informational: true
```
nstepien, line 6: This will keep the status green, even if we drop coverage %, so it's less annoying.

image

.github/codecov.yml
```diff
@@ -1,7 +1,17 @@
...
+    patch: off
+
+# https://docs.codecov.io/docs/pull-request-comments
 comment:
-  layout: 'reach, diff, files'
+  layout: 'diff, files'
```
nstepien, line 11: `reach` isn't super pertinent IMO

image

amanmahajan7 (approved):
LGTM

PR #2418, opened 2021-05-21
Clean up generics

nstepien:


src/Cell.tsx
```diff
@@ -127,6 +127,6 @@ function Cell<R, SR>(
   );
 }
 
-export default memo(forwardRef(Cell)) as <R, SR = unknown>(
+export default memo(forwardRef(Cell)) as <R, SR>(
```
nstepien, line 130: I don't think we need a default here: we need to pass a `column` prop, which will allow TS to infer the `SR` from it, so it should be fine.

image

This also applies to other components in this PR like `Row` and `TextEditor`.

src/types.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
```
nstepien, line 1: 👍

src/types.ts
```diff
@@ -74,15 +73,15 @@ export interface Position {
   rowIdx: number;
 }
 
-export interface FormatterProps<TRow = any, TSummaryRow = any> {
+export interface FormatterProps<TRow, TSummaryRow = unknown> {
```
nstepien, line 76: I think it's fair to at least ask for `TRow`. Users can still use `<unknown>` or `<any>` if they really want to.

amanmahajan7 (approved):
Nice 👍

PR #2419, opened 2021-05-21
Add missing parameter type to cellRef in EditCell

nstepien:
I guess TS allowed it because the type `any` got inferred by `useCallback` 🤷‍♂️


src/hooks/useViewportRows.ts
```diff
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 import { ceil, floor, max, min } from '../utils';
 import type { GroupRow, GroupByDictionary, RowHeightArgs } from '../types';
 
-const RENDER_BACTCH_SIZE = 8;
```
amanmahajan7, line 5: We made so many changes to this hook and never caught it 😄

amanmahajan7 (approved):

PR #2420, opened 2021-05-21
Merge EditorContainer into EditCell + cleanups

nstepien:
-1 div!


src/EditCell.tsx
```diff
@@ -1,37 +1,38 @@
...
-import type { CellRendererProps, SharedEditorProps, Omit } from './types';
+import type { CellRendererProps, SharedEditorProps } from './types';
 
 const cellEditing = css`
   padding: 0;
 `;
 
-const cellEditingClassname = `rdg-cell-editing ${cellEditing}`;
+// TODO: remove 1 of the two classes
+const cellEditingClassname = `rdg-cell-editing rdg-editor-container ${cellEditing}`;
```
nstepien, line 14: WDYT? We have some internal code depending on both class names at the moment.

src/EditCell.tsx
```diff
@@ -1,37 +1,38 @@
...
   CellRendererProps<R, SR>,
   'rowIdx' | 'row' | 'column' | 'colSpan'
 >;
 
-interface EditCellProps<R, SR>
-  extends SharedCellRendererProps<R, SR>,
-    Omit<React.HTMLAttributes<HTMLDivElement>, 'style' | 'children'> {
+interface EditCellProps<R, SR> extends SharedCellRendererProps<R, SR> {
   editorProps: SharedEditorProps<R>;
+  onKeyDown: Required<React.HTMLAttributes<HTMLDivElement>>['onKeyDown'];
```
nstepien, line 23: `EditCell` is a strictly internal component with only specific props.

src/EditCell.tsx
```diff
@@ -41,30 +42,31 @@ export default function EditCell<R, SR>({
...
   );
 
-  function getCellContent() {
-    if (dimensions === null) return;
-    const { scrollTop: docTop, scrollLeft: docLeft } =
-      document.scrollingElement ?? document.documentElement;
-    const { left, top } = dimensions;
-    const gridLeft = left + docLeft;
-    const gridTop = top + docTop;
+  let content;
```
nstepien, line 51: I prefer doing this over having a function that is only ever called once, unconditionally.

src/EditCell.tsx
```diff
@@ -1,37 +1,38 @@
...
   column,
   colSpan,
   row,
   rowIdx,
-  editorProps,
-  ...props
+  onKeyDown,
+  editorProps
 }: EditCellProps<R, SR>) {
   const [dimensions, setDimensions] = useState<{ left: number; top: number } | null>(null);
```
nstepien, line 34: Really wonder if we still need this

amanmahajan7: I think this was needed when we were rendering the whole editor in a portal. I am okay with removing them add them back if needed.

src/EditCell.tsx
amanmahajan7: Should we also flatten the props and removed `editorProps`?

src/EditCell.tsx
amanmahajan7: I think we can remove one if possible

amanmahajan7 (dismissed):

stories/demos/components/Editors/SelectEditor.tsx
```diff
@@ -27,13 +20,13 @@ export function SelectEditor({
       styles={{
         control: (provided) => ({
           ...provided,
-          height: rowHeight - 1,
+          height: 'var(--row-height)',
```
amanmahajan7, line 23:
```suggestion
          height: 'calc(var(--row-height) - 1)',
```

image

We can probably remove this editor completely

amanmahajan7 (dismissed):

amanmahajan7 (approved):

PR #2422, opened 2021-05-21
Editor tests

amanmahajan7:


nstepien (changes requested):
Would be nice to have tests with an editor that passes `true`/`false` to `onRowChange` and `onClose`. Maybe for a following PR.

nstepien (approved):

PR #2423, opened 2021-05-24
Add dialog editor

amanmahajan7:
image

We should probably still render the cell otherwise it is empty

### DIFF
--- a/stories/demos/CommonFeatures.tsx
+++ b/stories/demos/CommonFeatures.tsx
@@ -13,6 +13,28 @@ const toolbarClassname = css`
   margin-bottom: 8px;
 `;
 
+const dialogContainerClassname = css`
+  position: absolute;
+  left: 0;
+  right: 0;
+  top: 0;
+  bottom: 0;
+  display: flex;
+  place-items: center;
+  background: rgba(0, 0, 0, 0.1);
+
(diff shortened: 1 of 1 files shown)


stories/demos/CommonFeatures.tsx
```diff
@@ -13,6 +13,28 @@ const toolbarClassname = css`
   margin-bottom: 8px;
 `;
 
+const dialogContainerClassname = css`
```
amanmahajan7, line 16: > The ::backdrop CSS pseudo-element can be used to style behind a <dialog> element when the dialog is displayed with HTMLDialogElement.showModal().

😞

nstepien: Would that work? 🤔 
```tsx
ref={dialog => dialog.showModal()}
```

nstepien (approved):

nstepien:
>We should probably still render the cell otherwise it is empty

Let's do that yeah, should be easy with the grid layout.

stories/demos/CommonFeatures.tsx
amanmahajan7: Tried but getting ref errors. May be I will try some other time

PR #2435, opened 2021-05-27
rm Cell export and RowRendererProps.cellRenderer

nstepien:
I'm not 100% sure we should do it, but we don't use it ourselves, and it's better than abusing custom renderers just to add class names.
We can revert this in any case.

### DIFF
--- a/src/Cell.tsx
+++ b/src/Cell.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, memo } from 'react';
+import { memo } from 'react';
 import type { RefAttributes } from 'react';
 import { css } from '@linaria/core';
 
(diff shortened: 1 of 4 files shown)


amanmahajan7 (approved):
We can add it back if needed
https://github.com/adazzle/react-data-grid/issues?q=is%3Aissue+cellrenderer+is%3Aclosed

mmasdivins:
If we want to add a custom classname to an specific cell how we can do it now that we don't have the property cellRenderer?

mmasdivins:
But that will be for all the cells of that column and I want to do it for a specific cell in a row

nstepien:
```js
Column.cellClass = (row) => ...
```

mmasdivins:
Thanks

amber-cd:
We're working on upgrading to v7 and we use the CellRenderer fairly extensively in order to display a variety of different custom components depending on the column in question. It's more extensive than just adding a class. What is the new way to do this, if CellRenderer has been removed?

PR #2436, opened 2021-05-27
FIx de/selecting all rows, non existent row keys should be preserved

nstepien:


src/DataGrid.tsx
```diff
@@ -257,6 +257,7 @@ function DataGrid<R, SR, K extends Key>(
    * The identity of the wrapper function is stable so it won't break memoization
    */
   const selectRowWrapper = useLatestFunc(selectRow);
+  const selectAllRowsWrapper = useLatestFunc(selectAllRows);
```
amanmahajan7, line 260: May be we should use `Latest` suffix instead of `Wrapper` 🤔

amanmahajan7 (approved):

src/DataGrid.tsx
nstepien: Yeah

PR #2440, opened 2021-05-27
Render the cell formatter in the EditCell when the editor uses a portal

### LINKED ISSUE #2522: When modal / popup based editors open the cell value is not displayed
Currently, when we use modal based editor (out of cell editors). For e.g. completion column in [storybook example](https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--common-features). If you open the editor for any cell on that column, the current displayed value of the cell goes blank. It would be nice to have the current value continue to be displayed in such cases.

nstepien:
Not sure if the aria attributes need to be tweaked in some way.
image

Fixes #2522


src/Row.tsx
```diff
@@ -58,19 +58,16 @@ function Row<R, SR>(
...
+    let isCellSelected = false;
+    let dragHandleProps;
+    let onFocus;
+    let onKeyDown;
+
+    if (selectedCellProps !== undefined) {
+      isCellSelected = selectedCellProps.idx === column.idx;
+      if (isCellSelected && selectedCellProps.mode === 'SELECT') {
+        ({ dragHandleProps, onFocus, onKeyDown } = selectedCellProps);
+      }
```
nstepien, line 70: I did this so we don't hit too many `if`s or ternaries on each loop/render.

src/Row.tsx
```diff
@@ -83,16 +80,27 @@ function Row<R, SR>(
...
     );
+
+    if (isCellSelected && selectedCellProps!.mode === 'EDIT') {
+      cells.push(
+        <EditCell
+          key={column.key}
+          rowIdx={rowIdx}
+          column={column}
+          colSpan={colSpan}
+          onKeyDown={(selectedCellProps as EditCellProps<R>).onKeyDown}
```
amanmahajan7, line 99: TS still does not recognizes the type?

amanmahajan7 (commented):
It does not look like anything is mentioned about cell at the same position. The only thing I found is 
> Authors MAY indicate that a focusable gridcell is selectable as the object of an action with the aria-selected attribute. If the grid allows multiple gridcells to be selected, the author SHOULD set aria-multiselectable to true on the element with role grid.

https://www.w3.org/TR/wai-aria-1.2/#grid

Wdyt about doing the following instead?
- Render both cells if `createPortal` is true otherwise render the edit cell only
- When both cells are rendered then do not add [aria attributes] (https://github.com/adazzle/react-data-grid/blob/main/src/EditCell.tsx#L100) and style

src/Row.tsx
nstepien: Nope

amanmahajan7:
The formatter was behind the editor so it was still not visible. We can switch the order of `Cell` and `EditCell` to fix this issue. I ended up rendering formatter in the `EditCell`  instead. Wdyt?

amanmahajan7 (approved):

PR #2459, opened 2021-06-11
Add support for React 18

nstepien:
https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html


amanmahajan7:
> should we drop support for React 16/17? We don't have the capacity to support them.
We may be able to support as we are not using any react 18 features yet

nstepien:
Seeing as we already have types for `createRoot` when imported from `react-dom/client`, let's go forward with this.

nstepien:
We have +1 missed line in the coverage for some reason. Do we have a test for this specific line?
https://app.codecov.io/gh/adazzle/react-data-grid/compare/2459/changes
image

amanmahajan7 (approved):

PR #2460, opened 2021-06-11
Remove row batching

nstepien:
I think it was a _smart_ idea, but it made scrolling performance less consistent/more random, and React 18 will do have some batch updates anyway 🤔

### DIFF
--- a/src/hooks/useViewportRows.ts
+++ b/src/hooks/useViewportRows.ts
@@ -1,8 +1,6 @@
 import { useMemo } from 'react';
-import { ceil, floor, max, min } from '../utils';
+import { floor, max, min } from '../utils';
 import type { GroupRow, GroupByDictionary, RowHeightArgs } from '../types';
-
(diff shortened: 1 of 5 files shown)


src/hooks/useViewportRows.ts
```diff
@@ -176,15 +174,9 @@ export function useViewportRows<R>({
 
   const overscanThreshold = 4;
```
amanmahajan7, line 175: Should we consider adding `overscanRowCount` and `overscanColumnCount` to customize this behavior similar to [react-window](https://react-window.vercel.app/#/api/FixedSizeGrid)?

nstepien: I'd rather not, it's more APIs we'd have to support, we should reserve the right to change them, like how `RENDER_BATCH_SIZE` is being removed in this PR.

amanmahajan7 (approved):

PR #2464, opened 2021-06-16
Remove rowIdx props from cell components

nstepien:


src/DataGrid.tsx
```diff
@@ -432,6 +438,7 @@ function DataGrid<R, SR, K extends Key>(
     if (checked) {
       newSelectedRows.add(rowKey);
       const previousRowIdx = lastSelectedRowIdx.current;
+      const rowIdx = rows.indexOf(row);
```
nstepien, line 441: When we use `indexOf`, should we use `rows` or `rawRows`?

amanmahajan7: We were previously using `const row = rows[rowIdx];` so I guess it should be `rows`

src/Row.tsx
```diff
@@ -33,6 +33,10 @@ function Row<R, SR>(
   }: RowRendererProps<R, SR>,
   ref: React.Ref<HTMLDivElement>
 ) {
+  const handleRowChange = useLatestFunc((newRow: R) => {
```
amanmahajan7, line 36: I wonder if there is a way to avoid `useLatestFunc` here

amanmahajan7 (approved):

PR #2501, opened 2021-06-28
SortableHeaderCell: show sort arrow with svg

nstepien:
Before
image

After
image

We can animate it, tweak it however we want, and it'll be consistent across platforms/fonts.

### DIFF
--- a/src/headerCells/SortableHeaderCell.tsx
+++ b/src/headerCells/SortableHeaderCell.tsx
@@ -1,5 +1,6 @@
 import { css } from '@linaria/core';
 import type { HeaderRendererProps } from '../types';
+
 const headerSortCell = css`
   cursor: pointer;
   display: flex;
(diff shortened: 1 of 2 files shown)


src/headerCells/SortableHeaderCell.tsx
```diff
@@ -30,18 +42,15 @@ export default function SortableHeaderCell<R, SR>({
...
-    sortText = '\u25BC';
-  }
-
   return (
     <span className={headerSortCellClassname} onClick={(e) => onSort(e.ctrlKey)}>
       <span className={headerSortNameClassname}>{children}</span>
       <span>
-        {sortText}
+        {sortDirection !== undefined && (
+          <svg viewBox="0 0 12 8" width="12" height="8" className={arrowClassname}>
```
amanmahajan7, line 50:
```suggestion
          <svg viewBox="0 0 12 8" width="12" height="8" className={arrowClassname} aria-hidden="true">
```

Not sure if we want to set `focusable` to false as we do not support IE 11
https://adrianroselli.com/2021/04/sortable-table-columns.html#SVG

nstepien: https://github.com/adazzle/react-data-grid/blob/main/src/formatters/ToggleGroupFormatter.tsx#L49
Should we do the same here?

amanmahajan7: I think so but I will double check.

amanmahajan7: I think we can make the same change

nstepien: Done

amanmahajan7 (approved):

PR #2505, opened 2021-06-30
Set fields as readonly

amanmahajan7:


src/utils/selectedCellUtils.ts
```diff
@@ -96,7 +99,6 @@ export function getNextSelectedCellPosition<R, SR>({
           idx: 0
         };
       }
-      setColSpan(true);
```
amanmahajan7, line 99: It is not required as the next position is going to be the first cell of a colSpan range

src/hooks/useViewportRows.ts
```diff
@@ -1,6 +1,15 @@
...
+import type { GroupRow, RowHeightArgs } from '../types';
+
+type GroupByDictionary<TRow> = Record<
+  string,
+  {
+    childRows: readonly TRow[];
+    childGroups: readonly TRow[] | Readonly<GroupByDictionary<TRow>>;
+    startRowIndex: number;
+  }
+>;
```
nstepien, line 12: Should work,, no?
```suggestion
type GroupByDictionary<TRow> = Readonly<Record<
  string,
  {
    readonly childRows: readonly TRow[];
    readonly childGroups: readonly TRow[] | GroupByDictionary<TRow>;
    readonly startRowIndex: number;
  }
>>;
```

src/hooks/useViewportRows.ts
```diff
@@ -38,7 +47,7 @@ export function useViewportRows<R>({
       rows: readonly R[],
       [groupByKey, ...remainingGroupByKeys]: readonly string[],
       startRowIndex: number
-    ): [GroupByDictionary<R>, number] => {
+    ): [Readonly<GroupByDictionary<R>>, number] => {
```
nstepien, line 50:
```suggestion
    ): [GroupByDictionary<R>, number] => {
```

src/types.ts
```diff
@@ -175,7 +170,7 @@ export interface RowRendererProps<TRow, TSummaryRow = unknown>
 }
 
 export interface RowsChangeData<R, SR = unknown> {
-  indexes: number[];
+  indexes: readonly number[];
```
nstepien, line 173: We generate it then pass it on to the users, it's fine to leave it mutable, users can manipulate if they want to.
```suggestion
  indexes: number[];
```

nstepien (changes requested):

src/hooks/useViewportRows.ts
amanmahajan7: We [mutate](https://github.com/adazzle/react-data-grid/blob/472d2d27444b1251a59623005ac6b686c0fdb8c5/src/hooks/useViewportRows.ts#L59) the value but I can set fields to readonly

src/hooks/useViewportRows.ts
```diff
@@ -1,6 +1,15 @@
 import { useMemo } from 'react';
 import { floor, max, min } from '../utils';
-import type { GroupRow, GroupByDictionary, RowHeightArgs } from '../types';
+import type { GroupRow, RowHeightArgs } from '../types';
+
+type GroupByDictionary<TRow> = Record<
+  string,
+  {
+    readonly childRows: readonly TRow[];
+    readonly childGroups: readonly TRow[] | GroupByDictionary<TRow>;
```
nstepien, line 9:
```suggestion
    readonly childGroups: readonly TRow[] | Readonly<GroupByDictionary<TRow>>;
```

src/utils/selectedCellUtils.ts
```diff
@@ -64,19 +63,23 @@ export function getNextSelectedCellPosition<R, SR>({
         if (colIdx > posIdx) break;
         const colSpan = getColSpan(column, lastFrozenColumnIndex, { type: 'ROW', row });
         if (colSpan && posIdx > colIdx && posIdx < colSpan + colIdx) {
-          position.idx = colIdx + (moveRight ? colSpan : 0);
-          break;
+          return {
+            ...position,
```
nstepien, line 67: It's just 1 property, so we can set it manually and avoid the iterating over the object by spreading it.
```suggestion
            rowIdx: position.rowIdx,
```

src/utils/selectedCellUtils.ts
```diff
@@ -51,9 +51,8 @@ export function getNextSelectedCellPosition<R, SR>({
   isGroupRow
 }: GetNextSelectedCellPositionOpts<R, SR>): Position {
   const rowsCount = rows.length;
-  let position = nextPosition;
 
-  const setColSpan = (moveRight: boolean) => {
+  const setColSpan = (position: Position, moveRight: boolean): Position => {
```
nstepien, line 55: Does the name `setColSpan` make sense anymore?

I wonder if we even need a position object? Could we have something like this?
```ts
let { idx, rowIdx } = nextPosition;
```
Then we update `idx`/`rowIdx`, and at the end
```ts
return { idx, rowIdx };
```

amanmahajan7: I will check

nstepien (approved):

PR #2513, opened 2021-07-06
Support commandKey for multisort

nakazavva:
for Mac multi sort is not working with Ctrl key.


nstepien (approved):
Thanks for the contribution!

PR #2514, opened 2021-07-06
Nesting headerRenderer inside draggable header to use with custom headers

gormonn:
- Nesting headerRenderer inside draggable header to use with custom headers
- Change hook to useRefComposer with safer implementation


amanmahajan7:
Thank you for the PR. This PR is however obsolete as RDG now supports draggable columns so closing it

PR #2537, opened 2021-07-28
Use an element prop for setting editor and drag handle

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -940,32 +950,40 @@ function DataGrid<R, SR, K extends Key>(
     return isDraggedOver ? selectedPosition.idx : undefined;
   }
 
-  function getSelectedCellProps(rowIdx: number): SelectedCellProps | EditCellProps<R> | undefined {
-    if (selectedPosition.rowIdx !== rowIdx) return;
+  function getSelectedCellProps(rowIdx: number): SelectedCellProps<R, SR> | undefined {
```
nstepien, line 953: Could destructure `selectedPosition` so we don't repeat property access on `idx` and `rowIdx`

src/DataGrid.tsx
```diff
@@ -940,32 +950,40 @@ function DataGrid<R, SR, K extends Key>(
...
+        onFocus: handleFocus,
         onKeyDown: handleKeyDown,
-        editorProps: {
-          editorPortalTarget,
-          row: selectedPosition.row,
-          onRowChange: handleEditorRowChange,
-          onClose: handleOnClose
-        }
+        selectedCellEditor: (
+          <EditCell
```
nstepien, line 967: It's pushed to an array so we still need a `key` prop here.
Check the jest output.
```suggestion
          <EditCell
            key={column.key}
```

src/DataGrid.tsx
```diff
@@ -940,32 +950,41 @@ function DataGrid<R, SR, K extends Key>(
...
 
-  function getSelectedCellProps(rowIdx: number): SelectedCellProps | EditCellProps<R> | undefined {
+  function getSelectedCellProps(rowIdx: number): SelectedCellProps<R, SR> | undefined {
     if (selectedPosition.rowIdx !== rowIdx) return;
 
     if (selectedPosition.mode === 'EDIT') {
+      const row = rows[selectedPosition.rowIdx];
+      if (isGroupRow(row)) return;
+      const column = columns[selectedPosition.idx];
+      const colSpan = getColSpan(column, lastFrozenColumnIndex, { type: 'ROW', row });
```
amanmahajan7, line 960: May be we can use `row: selectedPosition.row`

src/types.ts
```diff
@@ -153,13 +133,15 @@ export interface RowRendererProps<TRow, TSummaryRow = unknown>
   viewportColumns: readonly CalculatedColumn<TRow, TSummaryRow>[];
   row: TRow;
   rowIdx: number;
+  selectedCellIdx?: number;
```
nstepien, line 136:
```suggestion
  selectedCellIdx: number | undefined;
```

src/types.ts
```diff
@@ -153,13 +133,15 @@ export interface RowRendererProps<TRow, TSummaryRow = unknown>
...
   copiedCellIdx: number | undefined;
   draggedOverCellIdx: number | undefined;
   lastFrozenColumnIndex: number;
   isRowSelected: boolean;
   top: number;
   height: number;
-  selectedCellProps: EditCellProps<TRow> | SelectedCellProps | undefined;
+  selectedCellEditor?: ReactElement<EditorProps<TRow>>;
+  selectedCellDragHandle?: ReactElement<React.HTMLAttributes<HTMLDivElement>>;
```
nstepien, line 144:
```suggestion
  selectedCellEditor: ReactElement<EditorProps<TRow>> | undefined;
  selectedCellDragHandle: ReactElement<React.HTMLAttributes<HTMLDivElement>> | undefined;
```

nstepien (changes requested):

src/DataGrid.tsx
nstepien: Worth a try

src/Row.tsx
```diff
@@ -94,9 +79,9 @@ function Row<R, SR>(
           isCopied={copiedCellIdx === idx}
           isDraggedOver={draggedOverCellIdx === idx}
           isCellSelected={isCellSelected}
-          dragHandleProps={dragHandleProps}
-          onFocus={onFocus}
-          onKeyDown={onKeyDown}
+          dragHandle={isCellSelected ? selectedCellDragHandle : undefined}
+          onFocus={isCellSelected ? onFocus : undefined}
+          onKeyDown={isCellSelected ? onKeyDown : undefined}
```
nstepien, line 84: What if we pass `onFocus`/`onKeyDown` as is all the time? 🤔

amanmahajan7: It would rerender all the cells then

src/types.ts
amanmahajan7: Ah, I remember why I used `?:` instead. `Row` complains as `getSelectedCellProps` also returns undefined

nstepien (approved):

PR #2538, opened 2021-07-28
Move drag related functions to the DragHandle component

amanmahajan7:


src/DragHandle.tsx
```diff
@@ -19,6 +23,108 @@ const cellDragHandle = css`
...
+  columns,
+  selectedPosition,
+  draggedOverRowIdx,
+  onRowsChange,
+  onFill,
+  setDragging,
+  setDraggedOverRowIdx
+}: Props<R, SR>) {
+  const latestDraggedOverRowIdx = useRef(draggedOverRowIdx);
+  useEffect(() => {
```
amanmahajan7, line 45: I think `useEffect` is fine here but we can pass the `latestDraggedOverRowIdx` ref instead if you think it is more safe

nstepien: Could just do
```ts
function handleDragEnd() {
  setDraggedOverRowIdx((draggedOverRowIdx) => {
    // ...
    return undefined;
  });
}
```

src/DragHandle.tsx
```diff
@@ -19,6 +22,106 @@ const cellDragHandle = css`
 
 const cellDragHandleClassname = `rdg-cell-drag-handle ${cellDragHandle}`;
 
-export default function DragHandle(props: React.HTMLAttributes<HTMLDivElement>) {
-  return <div className={cellDragHandleClassname} {...props} />;
+interface Props<R, SR> extends Pick<DataGridProps<R, SR>, 'rows' | 'onRowsChange'> {
+  columns: readonly CalculatedColumn<R, SR>[];
+  selectedPosition: SelectCellState | EditCellState<R>;
```
nstepien, line 27: I don't think it's possible
```suggestion
  selectedPosition: SelectCellState;
```

src/DragHandle.tsx
```diff
@@ -19,6 +22,106 @@ const cellDragHandle = css`
...
+    function onMouseUp() {
+      window.removeEventListener('mouseover', onMouseOver);
+      window.removeEventListener('mouseup', onMouseUp);
+      setDragging(false);
+      handleDragEnd();
+    }
+  }
+
+  function handleDragEnd() {
+    if (!onRowsChange) return;
```
nstepien, line 64: Should disable the feature if `onRowsChange` isn't defined

amanmahajan7: I think we should allow it if `onFill != null`. It is like editor, we do not block editor if `onRowsChange` is not specified.

nstepien: `onFill` cannot be called when `onRowsChange` is nullish though right now, should remove the return guards.

nstepien (approved):

PR #2542, opened 2021-07-29
Copy/paste tests

amanmahajan7:


test/copyPaste.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+        col2: 'a2'
+      },
+      {
+        col1: 3,
+        col2: 'a3'
+      }
+    ];
+  });
+
+  const columns = useMemo((): readonly Column<Row>[] => {
```
nstepien, line 76: Could define the rows and columns outside the component

test/copyPaste.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+      }
+    ];
+  }, []);
+
+  function onPaste({ sourceColumnKey, sourceRow, targetColumnKey, targetRow }: PasteEvent<Row>) {
+    return { ...targetRow, [targetColumnKey]: sourceRow[sourceColumnKey as keyof Row] };
+  }
+
+  return (
+    <StrictMode>
```
nstepien, line 99: Should we render `CopyPasteTest` under `StrictMode` as well?

test/copyPaste.test.tsx
amanmahajan7: Rows need to be updated for tests so cannot define outside

test/copyPaste.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    editor() {
+      return null;
+    }
+  }
+];
+
+const copyCellClassName = 'rdg-cell-copied';
+
+function CopyPasteTest({ allowCopyPaste = true }: { allowCopyPaste?: boolean }) {
+  const [rows, setRows] = useState((): readonly Row[] => {
```
nstepien, line 28: I mean to suggest that we can do this
```ts
const initialRows: readonly Row[] = [
  {
    col1: 1,
    col2: 'a1'
  },
  {
    col1: 2,
    col2: 'a2'
  },
  {
    col1: 3,
    col2: 'a3'
  }
];

function CopyPasteTest({ allowCopyPaste = true }: { allowCopyPaste?: boolean }) {
  const [rows, setRows] = useState(initialRows);
```

nstepien:
Build is currently failing because of this:
https://github.com/facebook/jest/issues/11708
https://github.com/nodejs/node/pull/39593
We can downgrade node in `ci.yml` by pinning `node-version` to `16.5`, should fix it I think.

test/copyPaste.test.tsx
nstepien: We can also remove the `col1` values

test/copyPaste.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+test('should not allow paste on readonly cells', () => {
+  setup();
+  userEvent.click(getCellsAtRowIndex(1)[0]);
+  copySelectedCell();
+  expect(getSelectedCell()).toHaveClass(copyCellClassName);
+  userEvent.type(document.activeElement!, '{arrowdown}');
+  pasteSelectedCell();
+  expect(getCellsAtRowIndex(2)[0]).toHaveTextContent('a3');
+});
```
nstepien, line 109: Could add a test that we allow copying a readonly cell, and pasting that value into a writable cell.

.github/workflows/ci.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '16.x'
+          node-version: '16.5'
```
nstepien, line 42: The other one needed to be changed 👀

nstepien (approved):

PR #2543, opened 2021-07-29
Drag fill tests

amanmahajan7:


src/DragHandle.tsx
```diff
@@ -94,12 +96,18 @@ export default function DragHandle<R, SR>({
     event.stopPropagation();
     const { idx, rowIdx } = selectedPosition;
     const sourceRow = rows[rowIdx];
-    const targetRows = rows.slice(rowIdx + 1);
+    const targetRows: R[] = [];
```
amanmahajan7, line 99: I don't think we should update the row when the cell is readonly. This also makes me wonder if we should have the same syntax for `onPaste` and `onFill`. With `onFill` we can update 1 row at a time. We do not have to worry about indexes also. Wdyt?

nstepien:
https://nodejs.org/en/blog/release/v16.6.1/
We can unpin Node.js now

src/DataGrid.tsx
```diff
@@ -141,7 +141,7 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
   rowGrouper?: ((rows: readonly R[], columnKey: string) => Record<string, readonly R[]>) | null;
   expandedGroupIds?: ReadonlySet<unknown> | null;
   onExpandedGroupIdsChange?: ((expandedGroupIds: Set<unknown>) => void) | null;
-  onFill?: ((event: FillEvent<R>) => R[]) | null;
+  onFill?: ((event: FillEvent<R>) => R) | null;
   onPaste?: ((event: PasteEvent<R>) => R) | null;
```
nstepien, line 145: I wonder if we should just merge them into one `onPaste` handler.
Can you think of any cases where pasing and filling would behave differently?

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: For paste, columns may not be compatible so we need source and target column keys and this API may be awkward for `onFill`. Also, there won't be a way to disable them individually .

PR #2544, opened 2021-08-03
Test column resize

amanmahajan7:


test/column/resizable.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    super(type, rest);
+    this.pointerId = pointerId;
+    this.pointerType = pointerType;
+    this.clientX = clientX;
+    this.buttons = buttons;
+  }
+}
+
+// @ts-expect-error
+globalThis.PointerEvent = PointerEvent;
```
amanmahajan7, line 27: Tests fail without this change. Had to suppress type errors. I don't think it is worth fixing them

image

test/column/resizable.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+  setup({ columns, rows: [] });
+  const [column] = getHeaderCells();
+  expect(getGrid()).toHaveStyle({ '--template-columns': '100px 200px' });
+  resize({ column, clientXStart: 95, clientXEnd: 200, rect: { right: 100, left: 0 } });
+  expect(getGrid()).toHaveStyle({ '--template-columns': '100px 200px' });
+});
+
+test('should not resize column if cursor offset is not within the allowed range', () => {
+  setup({ columns, rows: [] });
+  const [, col1] = getHeaderCells();
```
nstepien, line 101: Isn't it `col2` here and in the other tests?
```suggestion
  const [, col2] = getHeaderCells();
```

test/column/resizable.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+  expect(getGrid()).toHaveStyle({ '--template-columns': '100px 200px' });
+  resize({ column: col1, clientXStart: 288, clientXEnd: 250, rect: { right: 300, left: 100 } });
+  expect(getGrid()).toHaveStyle({ '--template-columns': '100px 200px' });
+});
+
+test('should resize column if cursor offset is not within the allowed range', () => {
```
nstepien, line 107:
```suggestion
test('should resize column if cursor offset is within the allowed range', () => {
```

test/column/resizable.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+  setup({ columns, rows: [] });
+  const [, col1] = getHeaderCells();
+  expect(getGrid()).toHaveStyle({ '--template-columns': '100px 200px' });
+  resize({ column: col1, clientXStart: 289, clientXEnd: 250, rect: { right: 300, left: 100 } });
+  expect(getGrid()).toHaveStyle({ '--template-columns': '100px 161px' });
+});
+
+test('should resize column if resizable is true', () => {
```
nstepien, line 115: Isn't this redundant?

amanmahajan7: I added for completeness but we can remove it

nstepien (approved):

PR #2548, opened 2021-08-04
Test cell navigation in a treegrid

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -629,11 +629,10 @@ function DataGrid<R, SR, K extends Key>(
   }
 
   function handleCellInput(event: React.KeyboardEvent<HTMLDivElement>) {
-    if (!isCellWithinBounds(selectedPosition)) return;
+    if (!isCellWithinBounds(selectedPosition) || selectedPosition.idx === -1) return;
```
amanmahajan7, line 632: If a regular row is selected in a group and we enter a key which is not explicitly handled (like 'a') then we get the following error

image

amanmahajan7: I think we should revisit `isCellWithinBounds` and it should probably return false if `idx === -1`

nstepien (approved):

PR #2551, opened 2021-08-05
Row selection tests

amanmahajan7:


test/rowSelection.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+  userEvent.click(within(getCellsAtRowIndex(rowIdx)[0]).getByLabelText('Select'), { shiftKey });
+}
+
+// https://github.com/testing-library/react-testing-library/issues/624
+test.skip('row selection should throw error if rowKeyGetter is not specified', () => {
+  setup(false);
+  expect(() => {
+    userEvent.click(within(getCellsAtRowIndex(0)[0]).getByLabelText('Select'));
+  }).toThrow();
+});
```
nstepien, line 69:
```suggestion
```
Doesn't look like there's a plan to fix it upstream, so let's remove it?

nstepien (changes requested):
We need another test to make sure we don't remove row ids from the `Set` when toggling the header checkbox for rows that are not present in the grid.
This is important when for example I select some rows, then filter the rows, and toggle the header checkbox.

test/rowSelection.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+  userEvent.click(filterButton);
+  expect(getRows()).toHaveLength(1);
+  expect(headerCheckbox).toBeChecked();
+  userEvent.click(headerCheckbox);
+  userEvent.click(filterButton);
+
+  testSelection(0, true);
+  testSelection(1, false);
+  testSelection(2, true);
+  expect(headerCheckbox).not.toBeChecked();
```
nstepien, line 146: We also need to test the opposite: deselecting all rows should not be creating a new `Set` that ignores the previous `Set` values.

nstepien (commented):
https://github.com/adazzle/react-data-grid/blob/main/src/DataGrid.tsx#L278
Might be good to check that `allRowsSelected` is computed by checking all the provided rows, so if we provide a `selectedRows` Set with a bigger size than the arrays length, it doesn't necessarily mean that all rows are selected.

test/rowSelection.test.tsx
nstepien: https://github.com/adazzle/react-data-grid/commit/3d407a8684a6f342fb31e977da0c790b11ddd10b#diff-58e1e0e99f0f1a98a760b9cce29a92394f0814a9ae4cb55e7eeb786e997713c4L38
We used to just create a `new Set(undefined)` when deselecting all rows, just wanna prevent regressions like this.

nstepien (approved):

PR #2559, opened 2021-08-11
Fix focus logic and tweaks

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -846,19 +835,27 @@ function DataGrid<R, SR, K extends Key>(
     event.preventDefault();
 
     const ctrlKey = isCtrlKeyHeldDown(event);
-    const nextPosition = getNextSelectedCellPosition({
+    const nextPosition = getNextPosition(key, ctrlKey, shiftKey);
```
amanmahajan7, line 838: If `Ctrl + Home` is pressed twice then checkbox looses focus. It is because we re-render the cell but `isCellSelected` flag is not changed

nstepien (approved):
Got any tests for this? So we can at least avoid regressions

nstepien (dismissed):

nstepien (approved):

PR #2560, opened 2021-08-11
Select row on Shift + Space

amanmahajan7:
https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-for-data-grids

> Shift + Space: Selects the row that contains the focus. If the grid includes a column with checkboxes for selecting rows, this key can serve as a shortcut for checking the box when focus is not on the checkbox.

### DIFF
--- a/test/rowSelection.test.tsx
+++ b/test/rowSelection.test.tsx
@@ -14,10 +14,7 @@ const columns: readonly Column<Row>[] = [
   SelectColumn,
   {
     key: 'name',
-    name: 'Name',
-    formatter() {
-      return null;
-    }
+    name: 'Name'
   }
 ];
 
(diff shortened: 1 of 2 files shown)


test/rowSelection.test.tsx
```diff
@@ -10,7 +10,16 @@ interface Row {
...
 
-const columns: readonly Column<Row>[] = [SelectColumn];
+const columns: readonly Column<Row>[] = [
+  SelectColumn,
+  {
+    key: 'name',
+    name: 'Name',
+    formatter() {
+      return null;
+    }
```
nstepien, line 20: The default formatter is good enough.
```suggestion
    name: 'Name'
```

nstepien (dismissed):

nstepien (approved):

PR #2562, opened 2021-08-12
Implement focus using roving index strategy 

amanmahajan7:
**New focus strategy**
https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_roving_tabindex
- Set `tabIndex = 0` on the `focusSink` when page loads. This way grid will be part of the tab sequence. 
- on `grid focus`, first cell is selected if available 
- When a cell is selected, it's `tabIndex` is set to `0` and the `focusSink` is removed. This way we can directly tab into the selected cell next time
- If the formatter within the selected cell needs to be focused then it should follow the same strategy and set `tabIndex = 0`. In this case selected cell has `-1 tabIndex`.  This way we can directly tab into the formatter next time

**Focus vs selection** 
https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_focus_vs_selection
Focus and selection are now treated differently with different cell border<del>I have removed this for now. Not sure if we want to have different styles</del> As per the recommendation
> The selected state must be visually distinct from the focus indicator.

**Tab behavior**
https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_general_within
>  The tab sequence should include only one focusable element of a composite UI component. Once a composite contains focus, keys other than Tab and Shift + Tab enable the user to move focus among its focusable elements.

Currently on tab the selected cell is changed and with complex grids it makes sense to not leave the grid. We can tweak this behavior if needed in the future


src/DataGrid.tsx
```diff
@@ -490,7 +474,9 @@ function DataGrid<R, SR, K extends Key>(
     onExpandedGroupIdsChange(newExpandedGroupIds);
   }
 
-  function onGridFocus() {
+  function onGridFocus({ target }: React.FocusEvent<HTMLDivElement>) {
+    // Do not scroll if clicked within the grid
+    if (gridRef.current !== target && gridRef.current!.contains(target)) return;
```
amanmahajan7, line 479: I am not sure if this can be an issue. One workaround is to always render the selected column and row and remove `onGridFocus` handler

src/DataGrid.tsx
```diff
@@ -504,6 +490,11 @@ function DataGrid<R, SR, K extends Key>(
   }
 
   function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
+    if (!(event.target instanceof Element)) return;
+    const isCellEvent = event.target.closest('[role="gridcell"]') !== null;
```
amanmahajan7, line 494: Not very react way but works in our case

src/DataGrid.tsx
```diff
@@ -504,6 +490,11 @@ function DataGrid<R, SR, K extends Key>(
   }
 
   function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
```
nstepien, line 492: `handleKeyDown` is used twice now, is that alright?

src/DataGrid.tsx
nstepien: Any meaningful difference between the two?
```suggestion
    const isCellEvent = event.target.closest('.rdg-cell') !== null;
```

src/DataGrid.tsx
```diff
@@ -504,6 +490,11 @@ function DataGrid<R, SR, K extends Key>(
...
+    if (!(event.target instanceof Element)) return;
+    const isCellEvent = event.target.closest('[role="gridcell"]') !== null;
+    const isRowEvent = minColIdx === -1 && event.target.getAttribute('role') === 'row';
```
nstepien, line 495:
```suggestion
    const isRowEvent = hasGroups && event.target.getAttribute('role') === 'row';
```

src/DataGrid.tsx
```diff
@@ -880,51 +871,61 @@ function DataGrid<R, SR, K extends Key>(
     return isDraggedOver ? selectedPosition.idx : undefined;
   }
 
-  function getSelectedCellProps(rowIdx: number): SelectedCellProps<R, SR> | undefined {
-    if (selectedPosition.rowIdx !== rowIdx) return;
+  function getGridTabIndex() {
+    const { idx, rowIdx } = selectedPosition;
+    return isCellWithinBounds(selectedPosition) &&
+      (idx <= lastFrozenColumnIndex || idx >= colOverscanStartIdx) &&
```
nstepien, line 877: Isn't `isCellWithinBounds` redundant here?
```suggestion
    return (idx <= lastFrozenColumnIndex || idx >= colOverscanStartIdx) &&
```

src/DataGrid.tsx
amanmahajan7: No, this needs to be fixed. I have a fix and I will push it soon.

src/DataGrid.tsx
```diff
@@ -491,19 +481,21 @@ function DataGrid<R, SR, K extends Key>(
...
-    } else {
-      // otherwise if we already have a selected cell, we should scroll back to it when focusing the grid
-      scrollToCell(selectedPosition);
+    if (isCellWithinBounds(selectedPosition)) return;
+    // Tabbing into the grid should initiate keyboard navigation
+    const initialPosition: SelectCellState = { idx: 0, rowIdx: 0, mode: 'SELECT' };
+    if (isCellWithinBounds(initialPosition)) {
+      setSelectedPosition(initialPosition);
     }
+    // otherwise browser automatically scrolls to the selected cell
```
amanmahajan7, line 490: It works well most of the time but if the cell is behind the frozen cell then it is scrolled.

src/DataGrid.tsx
```diff
@@ -880,57 +872,93 @@ function DataGrid<R, SR, K extends Key>(
...
+      if (isRowOutsideViewport) {
+        // if the row is outside the viewport then only render the selected cell
+        rowColumns = [columns[selectedIdx]];
+      } else if (
+        // if the row is within the viewport and cell is not, add the selected column to viewport columns
+        selectedRowIdx === rowIdx &&
+        viewportColumns.length > lastFrozenColumnIndex + 1 && // there are non frozen columns
+        ((selectedIdx > lastFrozenColumnIndex &&
+          selectedIdx < viewportColumns[lastFrozenColumnIndex + 1].idx) ||
+          selectedIdx > viewportColumns[viewportColumns.length - 1].idx)
```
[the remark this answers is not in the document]
amanmahajan7: We can use this for sure. I was trying to avoid array traversal and use indexes instead.

src/DataGrid.tsx
```diff
@@ -880,57 +872,93 @@ function DataGrid<R, SR, K extends Key>(
...
+      const rowIdx = isRowOutsideViewport ? selectedRowIdx : viewportRowIdx;
+
+      let rowColumns = viewportColumns;
+      if (isRowOutsideViewport) {
+        // if the row is outside the viewport then only render the selected cell
+        rowColumns = [columns[selectedIdx]];
+      } else if (
+        // if the row is within the viewport and cell is not, add the selected column to viewport columns
+        selectedRowIdx === rowIdx &&
+        viewportColumns.length > lastFrozenColumnIndex + 1 && // there are non frozen columns
```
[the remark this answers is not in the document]
amanmahajan7: No longer needed after array.includes

src/DataGrid.tsx
```diff
@@ -880,57 +872,93 @@ function DataGrid<R, SR, K extends Key>(
...
 
   function getViewportRows() {
     const rowElements = [];
     let startRowIndex = 0;
-    for (let rowIdx = rowOverscanStartIdx; rowIdx <= rowOverscanEndIdx; rowIdx++) {
+
+    const { idx: selectedIdx, rowIdx: selectedRowIdx } = selectedPosition;
+    const startRowIdx =
+      isCellWithinBounds(selectedPosition) && selectedRowIdx < rowOverscanStartIdx
+        ? rowOverscanStartIdx - 1
```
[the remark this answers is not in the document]
amanmahajan7: Currently we loop the row indexes to generate viewport rows. The selected row can be anywhere so we cannot use the for loop. This is just a way to add an extra entry to we can render the selected row if it is outside the viewport. We can do something line `[indexes].forEach` but then we will have to probably use another loop  to generate indexes array

src/DataGrid.tsx
```diff
@@ -880,57 +872,93 @@ function DataGrid<R, SR, K extends Key>(
...
+    for (let viewportRowIdx = startRowIdx; viewportRowIdx <= endRowIdx; viewportRowIdx++) {
+      const isRowOutsideViewport =
+        viewportRowIdx === rowOverscanStartIdx - 1 || viewportRowIdx === rowOverscanEndIdx + 1;
+      const rowIdx = isRowOutsideViewport ? selectedRowIdx : viewportRowIdx;
+
+      let rowColumns = viewportColumns;
+      const selectedColumn = columns[selectedIdx];
+      // selectedIdx can be -1 if grouping is enabled
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      if (selectedColumn !== undefined) {
```
nstepien, line 945: Would this work?
```suggestion
      if (cellIsWithinBounds) {
```

amanmahajan7: No, because `cellIsWithinBounds` is true if `groupBy.length > 0 && idx === -1`

nstepien (approved):

nstepien (approved):
:shipit:

PR #2564, opened 2021-08-12
Test committing on enter when editor is in a portal

amanmahajan7:
This is failing in https://github.com/adazzle/react-data-grid/pull/2562 so I have added a test


test/column/editor.test.tsx
```diff
@@ -30,23 +30,38 @@ describe('Editor', () => {
     expect(screen.queryByLabelText('col1-editor')).not.toBeInTheDocument();
     userEvent.keyboard('{enter}');
     expect(screen.getByLabelText('col1-editor')).toHaveValue(1);
-    userEvent.type(document.activeElement!, '3');
-    userEvent.tab();
+    userEvent.keyboard('3{enter}');
```
amanmahajan7, line 33: We were not testing commit on enter

nstepien (approved):

PR #2565, opened 2021-08-13
Fix false positive test

amanmahajan7:


test/column/editor.test.tsx
```diff
@@ -42,12 +43,16 @@ describe('Editor', () => {
         }}
       />
     );
-    userEvent.click(getCellsAtRowIndex(0)[0]);
-    expect(screen.queryByLabelText('col1-editor')).not.toBeInTheDocument();
+    userEvent.click(getCellsAtRowIndex(0)[1]);
+    expect(screen.queryByLabelText('col2-editor')).not.toBeInTheDocument();
```
amanmahajan7, line 47: `editorOptions` is only applied to `col2-editor` and it had no affect on `col1-editor`

nstepien (approved):

PR #2572, opened 2021-08-17
Tweak flickering grid workaround

nstepien:
https://github.com/adazzle/react-data-grid/pull/2389#issuecomment-900192997

@ekeuus I've never reproduced the issue since we added the workaround, and I never had issues with 150% zoom level, so that's why it was `% 0.5`. I don't mind using `% 1` to catch more edge cases.

>Should the fix rather be to round the width down/up always to prevent the flickering?

`clientWidth` is always an integer.

### DIFF
--- a/src/hooks/useGridDimensions.ts
+++ b/src/hooks/useGridDimensions.ts
@@ -23,7 +23,7 @@ export function useGridDimensions(): [
       // TODO: remove once fixed upstream
       // we reduce width by 1px here to avoid layout issues in Chrome
       // https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
-      setGridWidth(clientWidth - (devicePixelRatio % 0.5 === 0 ? 0 : 1));
+      setGridWidth(clientWidth - (devicePixelRatio % 1 === 0 ? 0 : 1));
       setGridHeight(clientHeight);
     });
 


amanmahajan7 (approved):

ekeuus:
Yeah, I think our users might use 90% or 125% zoom so the 0.5 might have caused issues.

Thanks for a quick response! Any chance of getting a release with these changes any time soon?

nstepien:
`react-data-grid@7.0.0-beta.3` is out!

rvion:
after this, I keep experiencing flickering even more than before 😥 with regular zoom level.
I have a 27' screen, and the problem is even more present when debugger open. one of my friends experiences it almost 100% of the time on its laptop. 

📝 my datagrids are used in a split-screen view: they have a flex-grow:1, flex-basis: 0; along some other side panel so both datagrid and side panel takes ~50% of the width

rvion:
also, our setup is very simple, so issue is unlikely to be related to our stuff. just a flex-row, with one datagrid and a panel, both having flex-grow:1, and flex-basis0

nstepien:
@rvion providing a reproduction code would go a long way to help us debug and address this.

https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
Starring the Chromium issue could help too.

zakaryan2004:
I have this issue with 100% zoom levels on Windows on a 24' screen, with the browser window in any size. My website has a grid, with two flexboxes in it, one of which has the datagrid. It's a simple layout. 

With Ubuntu on the same device, I only have the issue with zoom levels other than 100%.

Is there any workaround? What additional info will help find the cause of this issue?

PR #2586, opened 2021-08-17
[POC] Frozen rows

amanmahajan7:


rvion:
just in case, please: do not force frozen rows uppon users: material-ui/datagrid does this, and it try to freeze data passed in no matter what, => this makes it incompatible with any list of mobx models that refuse to be frozen. converting a 10_000 items list of mobx models to json just because of internal decision when I  don't need to have editable cells is a bit sad. 😅

PR #2587, opened 2021-08-18
Fix raw row index when grouping is enabled

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -572,11 +572,12 @@ function DataGrid<R, SR, K extends Key>(
 
   function updateRow(rowIdx: number, row: R) {
     if (typeof onRowsChange !== 'function') return;
-    if (row === rawRows[rowIdx]) return;
+    const rawRowIdx = getRawRowIdx(rowIdx);
```
amanmahajan7, line 575: I moved this here as `updateRow` is used by `formatter.onRowChange` also. It also fixes target row index while pasting

nstepien (approved):

PR #2557, opened 2021-08-10
[POC] Include header and summary rows in cell navigation 

amanmahajan7:
https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-for-data-grids

In general all the rows/cell should be reachable via keyboard. I checked a few third party grids and they support it as well
https://material-ui.com/components/data-grid/
https://www.ag-grid.com/example.php?gclid=Cj0KCQjwpf2IBhDkARIsAGVo0D1BfZ-ISqPQGf6qG7ngaVzyxzUpeOUJnp5PlEdez971H8Pan7u_rFMaApaBEALw_wcB

PR #2588, opened 2021-08-18
Tweak onKeyDown to only handle viewport events, cleanup styles

amanmahajan7:
- [x] <del>Add `rdg-cell` class to viewport cells only</del>
- [x] <del>Add `rdg-row` class to viewport rows only</del>
- [x] Move linaria styles into the components where makes sense

Should we add `rdg-header-cell`, `rdg-summary-cell` etc classes?


src/style/core.ts
```diff
@@ -93,8 +93,6 @@ const focusSink = css`
   outline: 0;
 `;
 
-export const focusSinkClassname = `rdg-focus-sink ${focusSink}`;
```
amanmahajan7, line 96: I think we do not need an extra class as it will never be styles

src/DataGrid.tsx
```diff
@@ -494,8 +494,8 @@ function DataGrid<R, SR, K extends Key>(
 
   function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>, isEditorPortalEvent = false) {
     if (!(event.target instanceof Element)) return;
-    const isCellEvent = event.target.closest(`.${cellClassname}`) !== null;
-    const isRowEvent = hasGroups && event.target.classList.contains(rowClassname);
+    const isCellEvent = event.target.closest(`.${viewportCellClassname}`) !== null;
+    const isRowEvent = hasGroups && event.target.classList.contains(viewportRowClassname);
```
nstepien, line 498: I wonder if this would have been enough 🤔 
```suggestion
    const isCellEvent = event.target.closest('.rdg-row:not(.rdg-summary-row) > .rdg-cell') !== null;
    const isRowEvent = hasGroups && event.target.matches('.rdg-row:not(.rdg-summary-row)');
```

nstepien (dismissed):

src/DataGrid.tsx
amanmahajan7: This makes sense, the selector can be tweaked in https://github.com/adazzle/react-data-grid/pull/2557

nstepien (approved):
:shipit:

PR #2589, opened 2021-08-19
emptyRowsRenderer -> noRowsFallback

nstepien:


src/DataGrid.tsx
```diff
@@ -163,6 +157,8 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
   /**
    * Miscellaneous
    */
+  rowRenderer?: React.ComponentType<RowRendererProps<R, SR>> | null;
+  noRowsFallback?: React.ReactNode | null;
```
amanmahajan7, line 161:
```suggestion
  noRowsFallback?: React.ReactNode;
```

```
ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined
```
Do we need null?

amanmahajan7 (approved):

PR #2595, opened 2021-08-20
Refactor pointer events handling when resizing columns

nstepien:
https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture
https://developer.mozilla.org/en-US/docs/Web/API/Element/releasePointerCapture
With this API, we shouldn't need any special handling when the pointer is outside the target.
Except iframes...
I believe we also don't need to check for multi-touch, not 100% sure, can't really test.
We should try this with the drag handle as well.

### DIFF
--- a/src/HeaderCell.tsx
+++ b/src/HeaderCell.tsx
@@ -6,6 +6,8 @@ import SortableHeaderCell from './headerCells/SortableHeaderCell';
 import { getCellStyle, getCellClassname } from './utils';
 
 const cellResizable = css`
+  touch-action: none;
+
   &::after {
     content: '';
     cursor: col-resize;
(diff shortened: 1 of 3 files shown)


src/HeaderCell.tsx
```diff
@@ -54,27 +59,32 @@ export default function HeaderCell<R, SR>({
     }
 
     function onPointerMove(event: PointerEvent) {
-      if (event.pointerId !== pointerId) return;
       if (event.pointerType === 'mouse' && event.buttons !== 1) {
-        onPointerUp(event);
+        // handle case where the pointer `up`'d outside an iframe
+        // https://bugs.chromium.org/p/chromium/issues/detail?id=606896
+        // https://bugs.chromium.org/p/chromium/issues/detail?id=693494
+        onPointerUp();
```
nstepien, line 66: https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture
Actually the example on this page is in an iframe and it works fine, wonder what's different with the storybook.
We could remove this check 🤔

amanmahajan7 (dismissed):
😎

amanmahajan7 (approved):

PR #2597, opened 2021-08-23
Avoid grid re-renders when clicking the selected cell

### LINKED ISSUE #2388: Avoid grid rerender on same cell click multiple times
When the same cell is clicked again and again, the selected position state is getting update and which causes grid to get rerendered on each click. To avoid this a condition can be checked like this 

`function selectCell(position, enableEditor = false) {
    if (!isCellWithinBounds(position)) return;
    commitEditorChanges();

    if (enableEditor && isCellEditable(position)) {
      const row = rows[position.rowIdx];
      setSelectedPosition({ ...position,
        mode: 'EDIT',
        key: null,
        row,
        originalRow: row
      });
    } else if (position.idx !== selectedPosition.idx || position.rowIdx !== selectedPosition.rowIdx || selectedPosition.mode !== 'SELECT') {
      setSelectedPosition({ ...position,
        mode: 'SELECT'
      });
    }

    onSelectedCellChange == null ? void 0 : onSelectedCellChange({ ...position
    });
  }`

Please check the else if condition. Let me know your thoughts

nstepien:
Fixes #2388

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -686,6 +686,7 @@ function DataGrid<R, SR, K extends Key>(
     if (enableEditor && isCellEditable(position)) {
       const row = rows[position.rowIdx] as R;
       setSelectedPosition({ ...position, mode: 'EDIT', key: null, row, originalRow: row });
+      onSelectedCellChange?.(position);
     } else if (
       selectedPosition.mode !== 'SELECT' ||
       selectedPosition.idx !== position.idx ||
@@ -694,8 +695,8 @@ function DataGrid<R, SR, K extends Key>(
       // Avoid re-renders if the selected cell state is the same
       // TODO: replace with a #record? https://github.com/microsoft/TypeScript/issues/39831
       setSelectedPosition({ ...position, mode: 'SELECT' });
+      onSelectedCellChange?.(position);
     }
-    onSelectedCellChange?.(position);
   }
 
   function closeEditor() {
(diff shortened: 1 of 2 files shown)


src/DataGrid.tsx
```diff
@@ -686,10 +686,16 @@ function DataGrid<R, SR, K extends Key>(
...
+      selectedPosition.mode !== 'SELECT' ||
+      selectedPosition.idx !== position.idx ||
+      selectedPosition.rowIdx !== position.rowIdx
+    ) {
+      // Avoid re-renders if the selected cell state is the same
+      // TODO: replace with a #record? https://github.com/microsoft/TypeScript/issues/39831
       setSelectedPosition({ ...position, mode: 'SELECT' });
     }
-    onSelectedCellChange?.({ ...position });
+    onSelectedCellChange?.(position);
```
amanmahajan7, line 698: Should we call `onSelectedCellChange` if position is not changed?

src/DataGrid.tsx
```diff
@@ -686,10 +686,16 @@ function DataGrid<R, SR, K extends Key>(
     if (enableEditor && isCellEditable(position)) {
       const row = rows[position.rowIdx] as R;
       setSelectedPosition({ ...position, mode: 'EDIT', key: null, row, originalRow: row });
-    } else {
+    } else if (
+      selectedPosition.mode !== 'SELECT' ||
```
amanmahajan7, line 690: Can we add a test?

nstepien: How? 🤔

amanmahajan7: If we do not call `onSelectedCellChange` then we can use it to test this change

amanmahajan7 (approved):

PR #2599, opened 2021-08-24
Add TextEditor tests

nstepien:


test/TextEditor.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  userEvent.keyboard('Test{escape}');
+  expect(input).not.toBeInTheDocument();
+  input = null;
+  expect(getCells()[0]).toHaveTextContent(/^Tacitus Kilgore$/);
+
+  // blurring the input closes and commits the editor
+  userEvent.dblClick(getCells()[0]);
+  input = screen.getByRole('textbox') as HTMLInputElement;
+  userEvent.keyboard('Jim Milton');
+  fireEvent.blur(input);
```
amanmahajan7, line 50: Should we press tab 
```suggestion
  userEvent.tab();
```

or 
```suggestion
  userEvent.click(document.body);
```

nstepien: I thought about both of these, but we handle these internally in a special manner (keyboard navigation, outside click detection).
I wanted to explicatively test blurring, which can happen when you alt+tab for example.

amanmahajan7 (approved):

PR #2600, opened 2021-08-24
Add column order tests

nstepien:


amanmahajan7 (approved):
We probably have some overlapping tests in other files but that is okay

PR #2627, opened 2021-08-31
Remove core-js

nstepien:
We don't use any features that need to be polyfilled in the storybook/tests.
Running Babel with `debug: true`, it shows that only 3 polyfills are loaded, and we actually use none of these features:
```
[D:\repos\react-data-grid\test\setup.ts]
The corejs3 polyfill entry has been replaced with the following polyfills:
  es.math.hypot { "chrome":"91", "edge":"91", "firefox":"90", "node":"12.22", "safari":"14" }
  es.typed-array.sort { "chrome":"91", "edge":"91", "firefox":"90", "node":"12.22", "safari":"14" }
  web.immediate { "chrome":"91", "edge":"91", "firefox":"90", "node":"12.22", "safari":"14" }
```
I think going forward we can avoid relying on any polyfills at all. We can always add it back otherwise.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,4 +1,3 @@
-import 'core-js/stable';
 import { css } from '@linaria/core';
 
 css`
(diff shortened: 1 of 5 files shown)


amanmahajan7 (approved):
I wonder why the polyfills were loaded if we are not using them

nstepien:
With our settings it'd load all the stable features that aren't supported by our target platforms.
The idea was that it's possible for our dependencies to use these features too. We don't run babel on all our deps after all.

PR #2631, opened 2021-09-03
Use postcss as the only css preprocessor

nstepien:
Linaria uses Stylis as the default preprocessor, but compared to postcss it's not as well maintained, underdocumented, and [buggy](https://github.com/thysultan/stylis.js/issues/267).

We already use PostCSS to process and minify our stylesheets, so we may as well use it as the only preprocessor.
It might also improve build times, not that they were slow to begin with.

I've added the `nested` plugin so we keep the usual `&` nesting that we're used to.
There are a lot more plugins that we could look into if we ever wanted to, but I think we'll want to keep things simple. 
https://github.com/postcss/postcss/blob/main/docs/plugins.md

There is only one small difference in the build output:
image

Reference:
https://github.com/postcss/postcss-nested
https://github.com/callstack/linaria/blob/master/docs/BUNDLERS_INTEGRATION.md#options


amanmahajan7 (approved):
LGTM

PR #2632, opened 2021-09-03
Storybook: update config to use webpack 5

nstepien:


.storybook/package.json
```diff
@@ -0,0 +1 @@
+{}
```
nstepien, line 1: https://github.com/storybookjs/storybook/issues/15335
Some of the storybook packages don't have an `exports` field, so webpack ends up complaining.
This `package.json` is there so webpack ignores `"type": "module"` in the real `package.json`. 🤷‍♂️

amanmahajan7 (approved):

PR #2640, opened 2021-09-08
Use a forwardRef-less Row component internally by default

nstepien:
Before:
image
After:
image

-1 component per row by default!
`RowWithRef` will also be tree-shaken if left unused.

### DIFF
--- a/src/hooks/useCombinedRefs.ts
+++ b/src/hooks/useCombinedRefs.ts
@@ -6,7 +6,7 @@ export function useCombinedRefs<T>(...refs: readonly React.Ref<T>[]) {
       for (const ref of refs) {
         if (typeof ref === 'function') {
           ref(handle);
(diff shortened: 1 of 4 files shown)


src/hooks/useCombinedRefs.ts
```diff
@@ -6,7 +6,7 @@ export function useCombinedRefs<T>(...refs: readonly React.Ref<T>[]) {
       for (const ref of refs) {
         if (typeof ref === 'function') {
           ref(handle);
-        } else if (ref !== null) {
+        } else if (ref !== null && 'current' in ref) {
```
nstepien, line 9: Without `forwardRef`, the `ref` parameter is actually an empty, frozen object, and trying to assigning `current` to it crashes.

amanmahajan7 (approved):
Nice 👍

PR #2647, opened 2021-09-09
Add basic SSR test

nstepien:


src/hooks/index.ts
```diff
@@ -3,6 +3,7 @@ export * from './useCombinedRefs';
 export * from './useFocusRef';
 export * from './useGridDimensions';
 export * from './useLatestFunc';
+export * from './useLayoutEffect';
```
amanmahajan7, line 6: Should we call it `useIsomorphicLayoutEffect` just to avoid any confusion. It seems like a common name 
https://github.com/reach/reach-ui/blob/main/packages/utils/src/use-isomorphic-layout-effect.ts
https://github.com/reduxjs/react-redux/blob/d16262582b2eeb62c05313fca3eb59dc0b395955/src/components/connectAdvanced.js#L40

amanmahajan7 (approved):

src/hooks/index.ts
nstepien: Not sure I'd really call it "isomorphic", both `useEffect` and `useLayoutEffect` do nothing during SSR, it's just that `useLayoutEffect` has an extra warning for no good reason honestly.

PR #2654, opened 2021-09-13
Remove editorPortalTarget and editorOptions.createPortal

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -492,11 +485,11 @@ function DataGrid<R, SR, K extends Key>(
     onExpandedGroupIdsChange(newExpandedGroupIds);
   }
 
-  function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>, isEditorPortalEvent = false) {
+  function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
     if (!(event.target instanceof Element)) return;
     const isCellEvent = event.target.closest('.rdg-cell') !== null;
     const isRowEvent = hasGroups && event.target.matches('.rdg-row, .rdg-header-row');
-    if (!isCellEvent && !isRowEvent && !isEditorPortalEvent) return;
```
amanmahajan7, line 499: `isEditorPortalEvent` was added to handle `Enter` and `Escape`. I think we do need this for portal editors as portal editors would be usually complex like Modal etc

website/demos/CommonFeatures.tsx
```diff
@@ -149,28 +150,49 @@ function getColumns(countries: string[]): readonly Column<Row, SummaryRow>[] {
...
+                    <button onClick={() => onClose()}>Cancel</button>
+                    <button onClick={() => onClose(true)}>Save</button>
+                  </menu>
+                </dialog>
+              </div>,
+              document.body
+            )}
+            <div
+              className={css`
+                padding: 0 8px;
```
nstepien, line 184: This might be bothersome if we have to do it many times 🤔

nstepien (approved):

nstepien (approved):

website/demos/CommonFeatures.tsx
amanmahajan7: Yeah, may be we can return formattedValue with editorProps or formatterClassname or we can set a flag in editorProps like isPortal or something and render the formatter also.

PR #2657, opened 2021-09-14
lodash -> lodash-es

nstepien:
To save a few bytes


amanmahajan7 (approved):
Wonder if we should use https://www.npmjs.com/package/lodash.groupby instead as `groupBy` is the only function we use from lodash

nstepien:
The `lodash.*` packages re-bundle parts of lodash, which would lead to a lot of duplicate code

image
image
image

If we only keep `lodash-es` we can have tree-shaking and no duplicate code.
Would be nice to alias all the `lodash.*` packages as well just in case but I don't want to bother right now.

PR #2658, opened 2021-09-14
Add editorOptions.renderFormatter prop

amanmahajan7:


src/EditCell.tsx
```diff
@@ -84,7 +83,12 @@ export default function EditCell<R, SR>({
       onMouseDownCapture={cancelFrameRequest}
     >
       {column.editor != null && (
```
nstepien, line 85: Is it possible to render an `EditCell` for a column with no editor? 🤔

nstepien (approved):

src/EditCell.tsx
amanmahajan7: probably not. Let's revisit the types

PR #2660, opened 2021-09-14
Only preventDefault on Tab

amanmahajan7:
I don't think we need to prevent default for any other key


amanmahajan7:
It is needed for PageUp/Down also

PR #2661, opened 2021-09-14
Remove onSelectedCellChange prop

amanmahajan7:


src/hooks/useRovingRowRef.ts
```diff
@@ -3,7 +3,7 @@ import { css } from '@linaria/core';
 import { cell } from '../style';
 import { useFocusRef } from './useFocusRef';
 
-const rowSelected = css`
+export const rowSelected = css`
```
nstepien, line 6: Just use `rdg-row-selected` in the tests
```suggestion
const rowSelected = css`
```

nstepien (approved):

chilind:
I'm also using `onSelectedCellChange` and wondering if there is a replacement prop?

chilind:
I'm also using `onSelectedCellChange` and wondering if there is a replacement prop?

SergeyKatugin:
@nstepien we also used this props, mb have another prop to handle selected cell change?

Issue #2670, opened 2021-09-24
Cell selection not working correctly (scrolls back to previously selected cell)


chirayuchiripal:
## Describe the bug 
When a cell editor is open and the cell is scrolled out of view and another cell is selected, it results in grid getting scrolled weirdly.

## To Reproduce 

1. Open: https://adazzle.github.io/react-data-grid/#/common-features
2. Double click on `contact` cell of first row such that it opens the editor for that cell.
3. Scroll the grid such that cell (with open editor) is out of view bounds.
4. Select/Click another cell.
5. The focus will go back to original cell which had editor open and grid will scroll back to first row.

Link to code example:

## Expected behavior 
The focus should switch to the selected cell on step 4 of above mentioned steps.

## Environment

- `react-data-grid` version: 7.0.0-beta.5 and latest (whatever the website is running on)
- `react`/`react-dom` version: whatever the website is running on

PR #2672, opened 2021-09-24
Do not steal focus back to the cell if the editor is not in the viewport and another cell is clicked

amanmahajan7:
Fixes https://github.com/adazzle/react-data-grid/issues/2670

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -700,13 +700,11 @@ function DataGrid<R, SR, K extends Key>(
     if (enableEditor && isCellEditable(position)) {
       const row = rows[position.rowIdx] as R;
       setSelectedPosition({ ...position, mode: 'EDIT', row, originalRow: row });
-    } else if (
-      selectedPosition.mode !== 'SELECT' ||
-      selectedPosition.idx !== position.idx ||
-      selectedPosition.rowIdx !== position.rowIdx
-    ) {
+    } else if (isSamePosition(selectedPosition, position)) {
       // Avoid re-renders if the selected cell state is the same
       // TODO: replace with a #record? https://github.com/microsoft/TypeScript/issues/39831
+      scrollToCell(position);
+    } else {
       setSelectedPosition({ ...position, mode: 'SELECT' });
     }
   }
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -363,11 +363,12 @@ function DataGrid<R, SR, K extends Key>(
   useLayoutEffect(() => {
     if (
       !selectedCellIsWithinSelectionBounds ||
-      selectedPosition === prevSelectedPosition.current ||
-      selectedPosition.mode === 'EDIT'
+      isSamePosition(selectedPosition, prevSelectedPosition.current)
     ) {
+      prevSelectedPosition.current = selectedPosition;
       return;
     }
+
     prevSelectedPosition.current = selectedPosition;
     scrollToCell(selectedPosition);
   });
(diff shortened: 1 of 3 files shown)


src/DataGrid.tsx
```diff
@@ -363,11 +363,12 @@ function DataGrid<R, SR, K extends Key>(
   useLayoutEffect(() => {
     if (
       !selectedCellIsWithinSelectionBounds ||
-      selectedPosition === prevSelectedPosition.current ||
-      selectedPosition.mode === 'EDIT'
```
amanmahajan7, line 367: Not sure why we were checking `mode === 'EDIT'`

src/DataGrid.tsx
```diff
@@ -698,13 +700,11 @@ function DataGrid<R, SR, K extends Key>(
...
       setSelectedPosition({ ...position, mode: 'EDIT', row, originalRow: row });
-    } else if (
-      selectedPosition.mode !== 'SELECT' ||
-      selectedPosition.idx !== position.idx ||
-      selectedPosition.rowIdx !== position.rowIdx
-    ) {
+    } else if (isSamePosition(selectedPosition, position)) {
       // Avoid re-renders if the selected cell state is the same
       // TODO: replace with a #record? https://github.com/microsoft/TypeScript/issues/39831
+      scrollToCell(position);
```
amanmahajan7, line 706: We should scroll to the same cell again in case it is out of view
image

src/DataGrid.tsx
```diff
@@ -698,13 +700,11 @@ function DataGrid<R, SR, K extends Key>(
     if (enableEditor && isCellEditable(position)) {
       const row = rows[position.rowIdx] as R;
       setSelectedPosition({ ...position, mode: 'EDIT', row, originalRow: row });
-    } else if (
-      selectedPosition.mode !== 'SELECT' ||
```
amanmahajan7, line 702: Not sure why were we checking `selectedPosition.mode !== 'SELECT'`

src/DataGrid.tsx
```diff
@@ -363,11 +363,12 @@ function DataGrid<R, SR, K extends Key>(
   useLayoutEffect(() => {
     if (
       !selectedCellIsWithinSelectionBounds ||
-      selectedPosition === prevSelectedPosition.current ||
-      selectedPosition.mode === 'EDIT'
+      isSamePosition(selectedPosition, prevSelectedPosition.current)
     ) {
+      prevSelectedPosition.current = selectedPosition;
```
nstepien, line 368: Does it make sense to do this assignment if `isSamePosition(selectedPosition, prevSelectedPosition.current)`?

test/column/editor.test.tsx
```diff
@@ -219,30 +219,53 @@ describe('Editor', () => {
...
   });
 
-  it.skip('should not steal focus back to the cell after being closed by clicking outside the grid', async () => {
-    const column: Column<unknown> = {
-      key: 'col',
-      name: 'Column',
-      editor() {
-        return <input value="123" readOnly autoFocus />;
+  describe('editor focus', () => {
+    it('should not steal focus back to the cell if the editor is not in the viewport and another cell is clicked', () => {
```
nstepien, line 223: Does the test properly fail without the changes?

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: `isSamePosition` only checks for `idx` and `rowIdx` and mode can be different

test/column/editor.test.tsx
amanmahajan7: yeah, check the first commit 
https://github.com/adazzle/react-data-grid/runs/3702020217

PR #2675, opened 2021-09-27
Move some of the editing logic to the EditCell component

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -575,14 +573,7 @@ function DataGrid<R, SR, K extends Key>(
   }
 
   function commitEditorChanges() {
-    if (
-      columns[selectedPosition.idx]?.editor == null ||
-      selectedPosition.mode === 'SELECT' ||
-      selectedPosition.row === selectedPosition.originalRow
```
amanmahajan7, line 581: `updateRow` already checks the original row

src/DataGrid.tsx
```diff
@@ -532,7 +531,6 @@ function DataGrid<R, SR, K extends Key>(
     switch (event.key) {
       case 'Escape':
         setCopiedCell(null);
```
amanmahajan7, line 533: Not sure we should discard copied value when we press escape during editing

nstepien (approved):

PR #2722, opened 2021-10-26
Add: shouldCloseEditor prop

blancham:
## Context

React-data-grid works great in a local environment but when it comes to use it with a database, problems appear...

Imagine you have 2 tables on two different laptops. With the current implementation, if I'm editing a cell and that the other guy is modifying the grid data, my cell editor will be closed automatically independently of what row/ cell has been edited.

## Solution
`shouldCloseEditor` is a function and allows you to specify the behaviour when you are editing a cell and that new data comes from the outside of the table.

In my project, I close the editor only when the value of the cell I'm currently editing has changed.


amanmahajan7:
Thank you for the PR. We have added a new prop to handle this scenario `column.editorOptions.closeOnExternalRowChange`
https://github.com/adazzle/react-data-grid/pull/3759

PR #2730, opened 2021-11-03
Add onColumnResizeEnd props

vejvis:
Callback onColumnResizeEnd is called after column resizing is ended.


amanmahajan7:
Thank you for the PR. We have added a new `onColumnWidthsChange` prop which is fired when resizing is complete 
https://github.com/adazzle/react-data-grid/pull/3761

PR #2765, opened 2021-12-15
Add `TreeDataGrid` component

amanmahajan7:
Any new feature that we add to RDG must work with all the existing features and that makes a grid component quite complex.
For grouping feature we added many `isGroupRow` checks and updated `utils/hooks`. So the logic was spread and this increased the complexity of the codebase, It also makes it harder to improve grouping feature. This has also lead to bugs for example `Shift + Space` does not select the group row at the moment. 

With the new `TreeGridComponent`, all the grouping logic is moved to one place and we have more control over props. For example all the grouping props are now required and `onFill` is not supported so it is removed from `TreeDataGridProps`. It does not matter much but it will also reduce the bundle size if you are only using `DataGrid`.

I think in general this should improve maintainability and stability. Let me know wdyt and if you see any disadvantages with the new approach.


amanmahajan7:
Dependent on https://github.com/adazzle/react-data-grid/pull/2833 and https://github.com/adazzle/react-data-grid/pull/3117

src/DataGrid.tsx
```diff
@@ -88,7 +87,13 @@ export interface DataGridHandle {
 
 type SharedDivProps = Pick<
   React.HTMLAttributes<HTMLDivElement>,
-  'aria-label' | 'aria-labelledby' | 'aria-describedby' | 'className' | 'style'
+  | 'role'
+  | 'aria-label'
+  | 'aria-labelledby'
+  | 'aria-describedby'
+  | 'aria-rowcount'
```
amanmahajan7, line 94: `aria-rowcount` can be useful for paginated grid also

src/DataGrid.tsx
```diff
@@ -141,10 +146,6 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
   sortColumns?: Maybe<readonly SortColumn[]>;
   onSortColumnsChange?: Maybe<(sortColumns: SortColumn[]) => void>;
   defaultColumnOptions?: Maybe<DefaultColumnOptions<R, SR>>;
-  groupBy?: Maybe<readonly string[]>;
```
amanmahajan7, line 144: Most of the grouping logic has been removed from `DataGrid` to `TreeDataGrid`. The only thing remining is row focus logic

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,428 @@
...
+) {
+  const defaultRenderers = useDefaultRenderers<R, SR>();
+  const rawRenderRow = renderers?.renderRow ?? defaultRenderers?.renderRow ?? defaultRenderRow;
+  const headerAndTopSummaryRowsCount = 1 + (props.topSummaryRows?.length ?? 0);
+  const isRtl = props.direction === 'rtl';
+  const leftKey = isRtl ? 'ArrowRight' : 'ArrowLeft';
+  const rightKey = isRtl ? 'ArrowLeft' : 'ArrowRight';
+  const toggleGroupLatest = useLatestFunc(toggleGroup);
+
+  const { columns, groupBy } = useMemo(() => {
```
amanmahajan7, line 69: Extracted from `useCalculatedColumns`. Now we run an extra loop for columns. I don't think it should be an issue.

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,428 @@
...
+          renderGroupCell: column.renderGroupCell ?? renderToggleGroup,
+          editable: false
+        };
+      }
+    }
+
+    return { columns, groupBy };
+  }, [rawColumns, rawGroupBy]);
+
+  const [groupedRows, rowsCount] = useMemo(() => {
```
amanmahajan7, line 105: Extracted from `useViewportRows`

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,428 @@
...
+          return [parentRow, i] as const;
+        }
+      }
+
+      return undefined;
+    },
+    [isGroupRow, rows]
+  );
+
+  const rowKeyGetter = useCallback(
```
amanmahajan7, line 210: Added a wrapper so we can get keys for group and regular row

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,428 @@
...
+        if (isGroupRowSelected) {
+          selectedRows.add(row.id);
+        }
+      }
+    }
+
+    return selectedRows;
+  }, [isGroupRow, rawRowKeyGetter, rawSelectedRows, rows]);
+
+  function onSelectedRowsChange(newSelectedRows: Set<Key>) {
```
amanmahajan7, line 252: Select/unselect regular rows when group row selection changes

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,428 @@
...
+        const groupIndex = childRows.indexOf(row);
+        return startRowIndex + groupIndex + 1;
+      }
+
+      return rows.indexOf(row);
+    },
+    [getParentRowAndIndex, isGroupRow, rawRowKeyGetter, rows]
+  );
+
+  const selectedRows = useMemo((): Maybe<ReadonlySet<Key>> => {
```
amanmahajan7, line 232: Select group rows based on regular rows

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,428 @@
...
+    // If a group row is focused, and it is collapsed, move to the parent group row (if there is one).
+    if (idx === -1 && event.key === leftKey && !row.isExpanded && row.level !== 0) {
+      const parentRowAndIndex = getParentRowAndIndex(row);
+      if (parentRowAndIndex !== undefined) {
+        event.preventGridDefault();
+        selectCell({ idx, rowIdx: parentRowAndIndex[1] });
+      }
+    }
+
+    if (isCtrlKeyHeldDown(event) && (event.keyCode === 67 || event.keyCode === 86)) {
```
amanmahajan7, line 313: Copy/paste is not allowed on a group row

godon019:
what are the reasonings for separating the grouping(tree) feature from DataGrid?

godon019:
thank you for the explanation! it makes sense to separate complex logic. and easy to understand the codebase
now we are wrapping the DataGrid with Grouping.
and what are you thinking if we have another complex logic?
do we want to wrap on top of each other if user need both complex features together?

amanmahajan7:
> and what are you thinking if we have another complex logic?
do we want to wrap on top of each other if user need both complex features together?

I think it depends. We need to decide what are the core grid features and based on that make a decision. In general I think composing components help with maintainability

nstepien (commented):
Partial review

src/DataGrid.tsx
```diff
@@ -990,55 +908,19 @@ function DataGrid<R, SR, K extends Key>(
...
-            isRowSelected={isGroupRowSelected}
-            selectGroup={selectGroupLatest}
-            toggleGroup={toggleGroupLatest}
-          />
-        );
-        continue;
-      }
-
-      startRowIndex++;
       let key;
```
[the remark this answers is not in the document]
amanmahajan7: I tried earlier but does not work

image
image

src/hooks/useCalculatedColumns.ts
```diff
@@ -114,10 +94,6 @@ export function useCalculatedColumns<R, SR>({
     columns.forEach((column, idx) => {
```
godon019, line 94: what do you think about using `for...of` instead here?

src/GroupRow.tsx
```diff
@@ -42,43 +24,47 @@ const groupRow = css`
...
+  className,
   row,
-  gridRowStart,
-  height,
-  level,
-  isExpanded,
+  rowIdx,
+  viewportColumns,
   selectedCellIdx,
+  lastFrozenColumnIndex,
```
godon019, line 38: do we need this prop?

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,428 @@
...
+    const expandGroup = (
+      rows: GroupByDictionary<R> | readonly R[],
+      parentId: string | undefined,
+      level: number
+    ): void => {
+      if (isReadonlyArray(rows)) {
+        flattenedRows.push(...rows);
+        return;
+      }
+      Object.keys(rows).forEach((groupKey, posInSet, keys) => {
```
godon019, line 148: what do you think about using `for...of` here?

godon019 (commented):
a few comments

src/DataGrid.tsx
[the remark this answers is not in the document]
nstepien: This seems to work
```suggestion
      let key: K | number = rowIdx;
```

test/grouping.test.tsx
nstepien, line 1: Should we rename this file to `TreeDataGrid.tests.tsx`?

test/grouping.test.tsx
```diff
@@ -141,22 +133,20 @@ function getHeaderCellsContent() {
 }
 
 test('should not group if groupBy is not specified', () => {
```
nstepien, line 135:
```suggestion
test('should not group if groupBy is empty', () => {
```

src/hooks/useCalculatedColumns.ts
amanmahajan7: We need the index and can use `column.entries` but I think `forEach` is fine.

src/TreeDataGrid.tsx
amanmahajan7: Here we need access to the array ( `keys`) itself

src/GroupRow.tsx
amanmahajan7: Good catch. Does not look like it is needed on the group row

src/GroupRow.tsx
```diff
@@ -42,43 +24,49 @@ const groupRow = css`
...
     <RowSelectionProvider value={isRowSelected}>
       <div
         role="row"
-        aria-level={level}
-        aria-expanded={isExpanded}
+        aria-level={row.level + 1} // aria-level is 1-based
+        aria-setsize={row.setSize}
+        aria-posinset={row.posInSet + 1} // aria-posinset is 1-based
+        aria-expanded={row.isExpanded}
+        key={row.id}
```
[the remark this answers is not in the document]
amanmahajan7: Probably for performance. Each group does have a uniqueId.

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,430 @@
...
+        return -1;
+      }
+      if (rawGroupBy.includes(bKey)) return 1;
+
+      // Sort other columns last:
+      return 0;
+    });
+
+    const groupBy: string[] = [];
+    for (const [index, column] of columns.entries()) {
```
nstepien, line 89: 🤷‍♂️ 
```suggestion
    for (let index = 0; index < columns.length; index++ ) {
      const column = columns[index];
```

src/TreeDataGrid.tsx
nstepien: I wonder if this can break anything that expect `rowKeyGetter` to be defined when the user didn't actually pass one

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,430 @@
...
+    if (rawSelectedRows == null) return null;
+
+    assertIsValidKeyGetter<R, K>(rawRowKeyGetter);
+
+    const selectedRows = new Set<Key>(rawSelectedRows);
+    for (const row of rows) {
+      if (isGroupRow(row)) {
+        const isGroupRowSelected = row.childRows.every((cr) =>
+          rawSelectedRows.has(rawRowKeyGetter(cr))
+        );
```
nstepien, line 242: Does this handle more than one level of nested rows?

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,430 @@
...
+    assertIsValidKeyGetter<R, K>(rawRowKeyGetter);
+
+    const newRawSelectedRows = new Set(rawSelectedRows);
+    for (const row of rows) {
+      const key = rowKeyGetter(row);
+      if (selectedRows?.has(key) && !newSelectedRows.has(key)) {
+        if (isGroupRow(row)) {
+          for (const cr of row.childRows) {
+            newRawSelectedRows.delete(rawRowKeyGetter(cr));
+          }
```
nstepien, line 264: Same as above

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,430 @@
...
+        } else {
+          newRawSelectedRows.delete(key as K);
+        }
+      } else if (!selectedRows?.has(key) && newSelectedRows.has(key)) {
+        if (isGroupRow(row)) {
+          for (const cr of row.childRows) {
+            newRawSelectedRows.add(rawRowKeyGetter(cr));
+          }
```
nstepien, line 272: And here

src/TreeDataGrid.tsx
```diff
@@ -0,0 +1,430 @@
...
+  return (
+    <DataGrid<R, SR, Key>
+      {...props}
+      role="treegrid"
+      aria-rowcount={
+        rowsCount + 1 + (props.topSummaryRows?.length ?? 0) + (props.bottomSummaryRows?.length ?? 0)
+      }
+      ref={ref}
+      columns={columns}
+      rows={rows as R[]} // TODO: check types
```
nstepien, line 409: Can we fix this TODO?
```suggestion
      rows={rows as readonly R[]} // TODO: check types
```

src/TreeDataGrid.tsx
amanmahajan7: Why is this preferred over `columns.entries()`

nstepien: Both are fine I guess, I'd expect the `for...let` loop to be more optimized though since it's an older pattern and doesn't create an iterator + an array on each iteration.
We can keep it as it is.

src/TreeDataGrid.tsx
amanmahajan7: It does as `row.childRows` contains all the raw rows within that group. So for example if we have 

```
groupLevel1
  - subGroupLevel1
    - row1
    - row2
  - subGroupLevel2
    - row3
```

Then 

```
groupLevel1,childRows == [row1, row2, row3]
subGroupLevel1,childRows == [row1, row2]
subGroupLevel2,childRows == [row3]
```

src/TreeDataGrid.tsx
amanmahajan7: Same here

nstepien (approved):

src/TreeDataGrid.tsx
amanmahajan7: I tried but it requires updating column definition. Let's think about grouping props and do it in a separate PR

nstepien (approved):

PR #2793, opened 2022-01-21
Auto resize columns on double click

amanmahajan7:
Idea
- Set the grid-column-template to `max-content` for the resized column
- Browser would resize the column automatically
- Find the column and update the calculated width
- Remove `max-content`


src/DataGrid.tsx
```diff
@@ -378,6 +379,21 @@ function DataGrid<R, SR, K extends Key>(
     }
   });
 
+  useLayoutEffect(() => {
+    if (autoResizeColumn === null) return;
+    const columnElement = gridRef.current!.querySelector(
+      `[aria-colindex="${autoResizeColumn.idx + 1}"]`
+    )!;
+    const width = columnElement.clientWidth + 1;
```
amanmahajan7, line 387: Not entirely sure why is this needed but may be because of right border. May be we can use gap to set the borders 🤔

nstepien:
https://developer.mozilla.org/en-US/docs/Web/CSS/clamp()
Wonder if we can use min/max/clamp in CSS to handle min/max widths 🤔

amanmahajan7:
I will investigate but this is good idea otherwise column width may jump if `maxWidth < calculatedWidth`

nstepien (approved):
:shipit:

PR #2794, opened 2022-01-21
Remove lineHeight and use flex to vertically align items

amanmahajan7:
Blocker: `overflow: hidden; text-overflow: ellipsis;` does not work with `display: flex` without extra spans


nstepien (commented):
LGTM, just need to tweak the header filters demo

nstepien:
We can also center text vertically by removing the `line-height` on rows, and adding an empty pseudo-element in each cell like this.
image

This works because the pseudo-element will have the full height of the cell, and its text will be centered vertically, then the remaining text will be aligned on the baseline by default, thus following the centered text (if there was any).
I'm not sure how well this works with multi-line cells though.

amanmahajan7:
Closing this for now as it does not work as expected.

PR #2802, opened 2022-01-28
Fix column resize on header cell using colSpan

amanmahajan7:


src/HeaderCell.tsx
```diff
@@ -85,11 +85,25 @@ export default function HeaderCell<R, SR>({
       return;
     }
 
+    const colSpan = currentTarget.ariaColSpan;
```
nstepien, line 88:
```suggestion
    const colSpan = Number(currentTarget.ariaColSpan);
```

src/HeaderCell.tsx
```diff
@@ -85,11 +85,25 @@ export default function HeaderCell<R, SR>({
...
+
+    function getColSpanColumnsWidth() {
+      let colSpanColumnsWidth = 0;
+      if (colSpan === null) return colSpanColumnsWidth;
+      for (let index = 1; index < Number(colSpan); index++) {
+        // nextElementSibling won't be null as all the colSpan columns are rendered
+        colSpanColumnsWidth += currentTarget.nextElementSibling!.getBoundingClientRect().width;
```
nstepien, line 95: This just gets the next column's width
1. why? the next column is unrelated, it just doesn't work well
2. this won't work if there's no column on the right

Recording 2022-09-21 at 11 13 19

nstepien (changes requested):

src/HeaderCell.tsx
amanmahajan7: Yeah, this is not correct. I wanted to put this PR back to draft

Issue #2744, opened 2021-11-17
Pass rowIdx in Row to Cell and column formatter


adityatoshniwal:
## Use case 
I'm trying to add a column which shows row number. But there is no way currently. Currently I have to check the complete rows array and get the row index. This can be expensive for large number of rows.

## Proposed solution 
If the rowIdx prop of Row is passed to the Cell component and forwarded to column formatter, I can use the rowIdx to show the row number.
If required, I can raise the PR.

adityatoshniwal:
Even passing rowIdx to onRowClick can be helpful to get the selected cell position.

adityatoshniwal:
@amanmahajan7 @nstepien  The commit - https://github.com/adazzle/react-data-grid/commit/905074a0a4c171a8847703182ec923a618da1bec
has broken the working of having Context as suggested in #2844. It is not working anymore. How to pass extra data to Row formatters now ?

paustint:
I am overriding the default `SelectColumn` with my own checkbox, but my implementation requires a unique id. It would be really convenient if I could get access to the output of `rowKeyGetter` or the `rowIdx`. The data in my table is customer generated and dynamic, so I don't know what fields exist on each row, so I am unable to know what (if any) datapoints are unique on the row.

For now, the only thing I can think of is to modify the data in advance to add in my own rowIdx to the dataset, which is not ideal.

PR #2820, opened 2022-02-10
Allow columns selection

adityatoshniwal:
Hi Team,

First of all, thank you so much for this wonderful component.
I have implemented column selection, which will be selected/deselected by clicking on header cells. Multiple columns can be selected. The catch is, if we select any columns and then select rows using checkbox then columns will be de-selected and vice-versa. The purpose here is, allowing selected columns to be copied and pasted in a spreadsheet.
I have also added a demo page /columns-selection.

Can you please review ?

Thanks.

### DIFF
--- a/src/hooks/useGridDimensions.ts
+++ b/src/hooks/useGridDimensions.ts
@@ -20,6 +20,7 @@ export function useGridDimensions(): [
     const resizeObserver = new ResizeObserver(() => {
       // Get dimensions without scrollbars.
       // The dimensions given by the callback entries in Firefox do not substract the scrollbar sizes.
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1733042
       const { clientWidth, clientHeight } = gridRef.current!;
       // TODO: remove once fixed upstream
       // we reduce width by 1px here to avoid layout issues in Chrome


adityatoshniwal:
I will write the test cases once the code is reviewed.

adityatoshniwal:
I have also added change for #2744. Please review and let me know if I need to change anything.

adityatoshniwal:
I need this for pgAdmin data grid for SQL query output. I'm one of the dev of pgadmin.org.

adityatoshniwal:
Hi @amanmahajan7, I'll try to implement the way you suggested. But, how can I know in my ReactDataGrid caller which columns are selected ? That info is needed.

adityatoshniwal:
Hello @amanmahajan7 can you please confirm so that I can work on it ?

adityatoshniwal:
Hello @amanmahajan7, can you please ? It will save a lot of time.

amanmahajan7:
You should be able to do something like 

```
const [selectedColumns, setSelectedColumns] = useState(new Set());

const columns = useMemo(() => {
  return [
     { 
        key, 
        name, 
        headerRenderer(props) { return <div onClick={toggleSelectedColumn}><HeaderRenderer {...props}></div> } />
     ...
  ]
}, [...])
```
https://github.com/adazzle/react-data-grid/blob/main/website/demos/ColumnsReordering.tsx#L96

adityatoshniwal:
@amanmahajan7 Oh OK !! You mean there is no need to change the code and just use HeaderRenderer instead.
I'll give a try. Thanks. 👍

PR #2821, opened 2022-02-11
Replace most box-shadows with border/outline

nstepien:
`box-shadow` is typically less efficient to raster, so this might make the grid ever so slightly more efficient.
I think we can do something similar with the last-frozen-cell shadow by overlaying a single div over the whole grid with a linear gradient.

Cell before / after:
image
image

Row before / after:
image
image

Checked checkbox before / after (identical):
image
image

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -1033,6 +1033,7 @@ function DataGrid<R, SR, K extends Key>(
             viewportColumns={rowColumns}
             childRows={row.childRows}
             rowIdx={rowIdx}
+            lastFrozenColumnIndex={lastFrozenColumnIndex}
             row={row}
             gridRowStart={gridRowStart}
             height={getRowHeight(rowIdx)}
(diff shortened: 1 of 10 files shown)


src/DataGrid.tsx
```diff
@@ -1033,6 +1033,7 @@ function DataGrid<R, SR, K extends Key>(
             viewportColumns={rowColumns}
             childRows={row.childRows}
             rowIdx={rowIdx}
+            lastFrozenColumnIndex={lastFrozenColumnIndex}
```
amanmahajan7, line 1036: Is it possible to set the selected row style using the extra div instead?
https://github.com/adazzle/react-data-grid/blob/3847e2f1d15f5766f05c0945f2e00b49672d1889/src/DataGrid.tsx#L1150

nstepien: Not sure, would still need something to set a border for the first sticky column

amanmahajan7 (approved):

Issue #2724, opened 2021-10-28
Hidden header-row but has Keyboard Navigation


vivaslj1:
## Describe the bug 
I have to hide the header-row of the grid

but keyboard navigation still works, so Focus Element disappears on screen.

## To Reproduce 

1. Hide header-row, I have try like this
   a. .rdg-header-row { display : none } 
   b. <ReactDataGrid {...otherProps} headerRowHeight={0} />

2. then use keyboard navigation

## Expected behavior 

Choice: 1. I would need a prop to set if Header-Row is disabled navigational element
Choice: 2. If the component detects headerRowHeight=0, disable Header-Row navigation

## Environment

react-data-grid version: 7.0.0-beta.7
react/react-dom version: 16.14.0

Any help greatly received!

alamothe:
It's weird to have navigable header. E.g. Google Sheet or Microsoft Excel don't have it.

amanmahajan7:
Fixed in https://github.com/adazzle/react-data-grid/pull/2833

Issue #2520, opened 2021-07-13
Loop over columns cell navigation


rbutl0215:
I would like `CellNavigationMode` to support looping through a column in addition to looping through rows. This can allow for easier data entry on tasks like taking inventory. I made a branch that can support this feature and attached a video of the updated Storybook. 

I've read through the contribution docs, but I am still a little confused on how to open a PR for this feature. Is the only way to contribute through the adazzle npm account?

https://user-images.githubusercontent.com/67930786/125492488-a90a2643-9f9d-4847-9cf8-1d676635a8e5.mov

amanmahajan7:
Fixed in https://github.com/adazzle/react-data-grid/pull/2833

Issue #2463, opened 2021-06-14
Cell navigation to next row using the enter key


crashnick16:
The old version of the grid used to have a feature called onGridKeyDown that allowed me to provide a better cell navigation feature by enabling the enter key to scroll down to the next row. Since this was removed we cannot have that anymore it appears.

This is to enable the same feature as in excel.

amanmahajan7:
This can be implemented using the new `onCellKeyDown` prop now

Issue #2377, opened 2021-04-26
Disable Tab Key On Grid


blessonabrahamah:
We are using react-data-grid v6.1.0  and we have a requirement to disable tab key on grid.

Tried making changes directly to library like this https://github.com/blessonabrahamah/react-data-grid/commit/2dd2b10c35cae93eeb836a09fa63bc2f31d09409

Just wanted to know, if its the right approach and do we have any other solution?

amanmahajan7:
Fixed in https://github.com/adazzle/react-data-grid/pull/2833

PR #2833, opened 2022-02-17
Add `onCellKeyDown` prop

amanmahajan7:
Add a new `DataGrid.onCellKeyDown` prop. This replaces the following props

- `cellNavigationMode`
- `editorOptions.onCellKeyDown`
- `editorOptions.onNavigation`

Signature
```
  onCellKeyDown?: Maybe<
    (
      args:
        | {
            mode: 'SELECT';
            row;
            column,
            rowIdx: number;
            selectCell: DataGridHandle['selectCell'];
          }
        | {
            mode: 'EDIT';
            row;
            column,
            rowIdx: number;
            onClose: (commitChanges?: boolean) => void;
          },
      event: CellKeyboardEvent
    ) => void
```
```
interface CellKeyboardEvent extends KeyboardEvent<HTMLDivElement> {
  preventGridDefault: () => void;
  isGridDefaultPrevented: () => boolean;
}
```

- `event.preventDefault` prevents both default grid and browser behavior.
- `event.preventGridDefault` only prevents default grid behavior. This is useful when we need to add functionality like ignoring tabs in keyboard navigation  

Fixes 
https://github.com/adazzle/react-data-grid/issues/2724
https://github.com/adazzle/react-data-grid/issues/2520
https://github.com/adazzle/react-data-grid/issues/2463
https://github.com/adazzle/react-data-grid/issues/2377
https://github.com/adazzle/react-data-grid/issues/1715


src/DataGrid.tsx
```diff
@@ -580,21 +595,31 @@ function DataGrid<R, SR, K extends Key>(
...
+    onCellKeyDown?.(
+      {
+        mode: 'SELECT',
+        idx,
+        rowIdx,
+        selectCell
+      },
+      event
+    );
+    if (event.isDefaultPrevented()) return;
```
amanmahajan7, line 594: User can completely control the add/edit behavior now

src/types.ts
```diff
@@ -239,3 +235,19 @@ export interface Renderers<TRow, TSummaryRow> {
 }
 
 export type Direction = 'ltr' | 'rtl';
+
+interface SelectCellKeyDownArgs {
```
amanmahajan7, line 239: We can pass row/column if needed

website/demos/CellNavigation.tsx
```diff
@@ -74,6 +76,59 @@ export default function CellNavigation({ direction }: Props) {
   const [rows] = useState(createRows);
   const [cellNavigationMode, setCellNavigationMode] = useState<CellNavigationMode>('CHANGE_ROW');
 
+  function handleCellKeyDown(args: CellKeyDownArgs, event: React.KeyboardEvent<HTMLDivElement>) {
```
amanmahajan7, line 79: Does not handle colSpan but I think it is fine for now.

nstepien (dismissed):

src/EditCell.tsx
```diff
@@ -77,19 +85,16 @@ export default function EditCell<R, SR>({
     };
   }, [cancelFrameRequest, commitOnOutsideClick, commitOnOutsideMouseDown, skipCellFocusRef]);
 
-  function onKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
+  function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
+    onKeyDown?.({ mode: 'EDIT', row, column, rowIdx, onClose }, event);
+    if (event.isDefaultPrevented()) return;
     if (event.key === 'Escape') {
-      event.stopPropagation();
```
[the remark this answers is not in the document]
amanmahajan7: No, it was added because we do not want to invoke parent's `onKeyDown` handler. Now we return if the cell is in edit mode
https://github.com/adazzle/react-data-grid/pull/2833/files#diff-0f9c631e39f2adf9fe111a2bb9ca7fab29327c1d43978939e30d3d36f5177ac4R595

src/utils/eventUtils.ts
```diff
@@ -0,0 +1,16 @@
+import type { CellEvent } from '../types';
+
+export function createCellEvent<E extends React.SyntheticEvent<HTMLDivElement>>(
+  event: E
+): CellEvent<E> {
+  let defaultPrevented = false;
+  return {
+    ...event,
```
amanmahajan7, line 8: We cannot call `preventDefault()` when we need to skip tab key. Calling `preventDefault` will prevent the default behavior and next element won't be focused. Cannot think of another way other than extending the default event.

nstepien (approved):

src/utils/eventUtils.ts
```diff
@@ -4,13 +4,17 @@ export function createCellEvent<E extends React.SyntheticEvent<HTMLDivElement>>(
...
   let defaultPrevented = false;
-  return {
+  const cellEvent = {
     ...event,
     preventGridDefault() {
       defaultPrevented = true;
     },
     isGridDefaultPrevented() {
-      return defaultPrevented;
+      return this.isDefaultPrevented() || defaultPrevented;
```
nstepien, line 13: micro-optimisation 🤷‍♂️ 
```suggestion
      return defaultPrevented || this.isDefaultPrevented();
```

nstepien (approved):

nstepien (approved):

nstepien (approved):

nstepien (dismissed):

nstepien (approved):

Issue #2837, opened 2022-02-22
Default width in `defaultColumnOptions` prop


peetjvv:
## Use case 

In our implementation we are setting a default width on all our columns. We are having to set this manually in every column since it is not supported by the DataGrid component's `defaultColumnOptions` prop.

## Proposed solution 

Add support for a default width in the DataGrid component's `defaultColumnOptions` prop so that the width doesn't have to be set manually for every column.

amanmahajan7:
Fixed in https://github.com/adazzle/react-data-grid/pull/2838

PR #2838, opened 2022-02-22
Default column width for all columns in the table

peetjvv:
Allows the client to set a default width for all columns that don't have a specified width.
Related issue: #2837


nstepien (approved):
LGTM, thanks for your contribution!

PR #2839, opened 2022-02-22
Feature: Add support for flex column width

amanmahajan7:
Column now supports CSS grid units including `auto` (default), `fr`, `%`, `max-content` etc. To handle virtualization, a flex column  is assigned a width of (80px) and the actual width is calculated after the grid is rendered.


src/DataGrid.tsx
```diff
@@ -413,6 +420,24 @@ function DataGrid<R, SR, K extends Key>(
     }
   });
 
+  useLayoutEffect(() => {
```
amanmahajan7, line 423: This effect is run if the viewport has any flex columns.

src/DataGrid.tsx
```diff
@@ -456,6 +482,7 @@ function DataGrid<R, SR, K extends Key>(
         newColumnWidths.set(column.key, width);
         return newColumnWidths;
       });
+      setFlexColumnWidths(new Map());
```
amanmahajan7, line 485: Re-calculated flex-column widths on resize. This is to keep the existing behavior. Ideally this is only needed if grid shows all the columns otherwise there is no need to re-calculate the widths again

src/DataGrid.tsx
```diff
@@ -885,10 +912,18 @@ function DataGrid<R, SR, K extends Key>(
...
   function getLayoutCssVars() {
-    if (autoResizeColumn === null) return layoutCssVars;
+    if (autoResizeColumn === null && flexWidthViewportColumns.length === 0) return layoutCssVars;
     const { gridTemplateColumns } = layoutCssVars;
     const newSizes = gridTemplateColumns.split(' ');
-    newSizes[autoResizeColumn.idx] = 'max-content';
+    if (autoResizeColumn !== null) {
+      newSizes[autoResizeColumn.idx] = 'max-content';
+    }
+    if (flexWidthViewportColumns.length > 0) {
```
amanmahajan7, line 921: Update width from fixed width (80px) to whatever the flex width is.

qujck:
Hi, is this PR going to be merged?

gugu91:
Hey 👋 Thanks so much for all the great work on this PR. Any chances this PR will get merged soon?

danhab99:
I am also quite keen to see this merged in please

src/hooks/useCalculatedColumns.ts
```diff
@@ -148,38 +153,19 @@ export function useCalculatedColumns<R, SR>({
...
-        // avoid decimals as subpixel positioning can lead to cell borders not being displayed
-        const unallocatedWidth = viewportWidth - allocatedWidth;
-        const unallocatedColumnWidth = round(unallocatedWidth / unassignedColumnsCount);
-        width = clampColumnWidth(unallocatedColumnWidth, column);
-        allocatedWidth += width;
-        unassignedColumnsCount--;
-        columnMetrics.set(column, { width, left });
+        // This is a placeholder width so we can continue to use virtualization.
+        // The actual value is set after the column is rendered
+        width = DEFAULT_COLUMN_MIN_WIDTH;
```
nstepien, line 164: Shouldn't we do this instead?
```suggestion
        width = column.minWidth;
```

src/hooks/useGridDimensions.ts
```diff
@@ -29,17 +34,24 @@ export function useGridDimensions(): [
...
       const size = entries[0].contentBoxSize[0];
-      setInlineSize(handleDevicePixelRatio(size.inlineSize));
+      const newInlineSize = handleDevicePixelRatio(size.inlineSize);
+      setInlineSize(newInlineSize);
       setBlockSize(size.blockSize);
+      if (prevInlineSize.current !== newInlineSize) {
+        prevInlineSize.current = newInlineSize;
+        // clear existing flex widths, this will trigger recalculation of visible flex columns again
+        setFlexColumnWidths((widths) => (widths.size > 0 ? new Map() : widths));
+      }
```
nstepien, line 44: Let's wrap this in a `useLatestFunc` so we don't re-create/disconnect the resize observer more than once.

src/DataGrid.tsx
```diff
@@ -908,10 +935,18 @@ function DataGrid<R, SR, K extends Key>(
...
+      for (const column of flexWidthViewportColumns) {
+        newSizes[column.idx] = column.width.toString();
+      }
+    }
```
nstepien, line 948: Unnecessary condition, it's built into the loop already
```suggestion
    for (const column of flexWidthViewportColumns) {
      newSizes[column.idx] = column.width.toString();
    }
```

src/hooks/useViewportColumns.ts
```diff
@@ -104,15 +108,31 @@ export function useViewportColumns<R, SR>({
...
+      }
     }
 
-    return viewportColumns;
+    return { viewportColumns, flexWidthViewportColumns };
   }, [startIdx, colOverscanEndIdx, columns]);
+
+  return {
+    viewportColumns,
+    flexWidthViewportColumns: useMemo(() => {
```
nstepien, line 132: Can we return `flexWidthViewportColumns` with only the rendered columns?

src/DataGrid.tsx
```diff
@@ -908,10 +935,18 @@ function DataGrid<R, SR, K extends Key>(
...
+    if (autoResizeColumn === null && flexWidthViewportColumns.length === 0) return layoutCssVars;
     const { gridTemplateColumns } = layoutCssVars;
     const newSizes = gridTemplateColumns.split(' ');
-    newSizes[autoResizeColumn.idx] = 'max-content';
+    if (autoResizeColumn !== null) {
+      newSizes[autoResizeColumn.idx] = 'max-content';
+    }
+    if (flexWidthViewportColumns.length > 0) {
+      for (const column of flexWidthViewportColumns) {
+        newSizes[column.idx] = column.width.toString();
```
nstepien, line 946: Is `toString()` necessary? Can we assert that it's a string already?
```suggestion
        newSizes[column.idx] = column.width as string;
```

src/hooks/useCalculatedColumns.ts
```diff
@@ -15,26 +15,31 @@ interface ColumnMetric {
...
   rawColumns,
   columnWidths,
+  flexColumnWidths,
   viewportWidth,
   scrollLeft,
   defaultColumnOptions,
   rawGroupBy,
   enableVirtualization
 }: CalculatedColumnsArgs<R, SR>) {
   const defaultWidth = defaultColumnOptions?.width;
```
nstepien, line 41: Should we do this here instead?
```suggestion
  const defaultWidth = defaultColumnOptions?.width ?? DEFAULT_COLUMN_WIDTH;
```

src/hooks/useCalculatedColumns.ts
```diff
@@ -15,26 +15,31 @@ interface ColumnMetric {
...
-  const defaultMinWidth = defaultColumnOptions?.minWidth ?? 80;
+  const defaultMinWidth = defaultColumnOptions?.minWidth;
```
nstepien, line 42:
```suggestion
  const defaultMinWidth = defaultColumnOptions?.minWidth ?? DEFAULT_COLUMN_MIN_WIDTH;
```

src/types.ts
```diff
@@ -59,6 +59,7 @@ export interface CalculatedColumn<TRow, TSummaryRow = unknown> extends Column<TR
   readonly resizable: boolean;
   readonly sortable: boolean;
   readonly frozen: boolean;
+  readonly width: number | string;
```
nstepien, line 62: Let's move it close to `minWidth`

src/hooks/useCalculatedColumns.ts
```diff
@@ -148,38 +151,19 @@ export function useCalculatedColumns<R, SR>({
...
 
     for (const column of columns) {
-      let width = getSpecifiedWidth(column, columnWidths, viewportWidth);
-
-      if (width === undefined) {
-        unassignedColumnsCount++;
-      } else {
+      let width = columnWidths.get(column.key) ?? column.width;
+      if (typeof width === 'number') {
         width = clampColumnWidth(width, column);
```
nstepien, line 158: I wonder if we can replace this with setting the min/max widths on the header cells instead 🤔
Could min/max widths also accept strings?
[Could `clamp` replace min/max?](https://caniuse.com/css-math-functions)
Might not play well with virtualization?

src/DataGrid.tsx
```diff
@@ -432,6 +433,25 @@ function DataGrid<R, SR, K extends Key>(
...
   });
 
+  useLayoutEffect(() => {
+    if (!isWidthInitialized || flexWidthViewportColumns.length === 0) return;
+    const newColumnWidths = new Map<string, number>();
+    for (const column of flexWidthViewportColumns) {
+      const columnElement = gridRef.current!.querySelector<HTMLDivElement>(
+        `[aria-colindex="${column.idx + 1}"]`
+      );
+      if (columnElement) {
```
nstepien, line 443: `flexWidthViewportColumns` only contains rendered columns, so is this check necessary?
Can `newColumnWidths` ever be empty after this loop?

src/DataGrid.tsx
```diff
@@ -1146,7 +1172,8 @@ function DataGrid<R, SR, K extends Key>(
         rootClassname,
         {
           [viewportDraggingClassname]: isDragging,
-          [cellAutoResizeClassname]: autoResizeColumn !== null
+          [cellAutoResizeClassname]:
```
nstepien, line 1175: Maybe this a more appropriate name
```suggestion
          [autosizeColumnsClassname]:
```

src/hooks/useCalculatedColumns.ts
nstepien: I guess `clamp` wouldn't work with manually resizing columns

src/DataGrid.tsx
amanmahajan7: IIRC, it was failing in some cases. It is possible to have `flexWidthViewportColumns .length > renderedFlexWidthViewportColumns.length`

src/hooks/useViewportColumns.ts
```diff
@@ -104,15 +106,29 @@ export function useViewportColumns<R, SR>({
...
+      return flexWidthViewportColumns.filter((column) => !columnWidths.has(column.key));
+    }, [flexWidthViewportColumns, columnWidths])
```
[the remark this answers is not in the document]
amanmahajan7: I was considering removing this hook completely. It should not be too expensive. Wdyt?

src/hooks/useGridDimensions.ts
```diff
@@ -6,11 +6,13 @@ import { ceil } from '../utils';
...
   ref: React.RefObject<HTMLDivElement>,
   width: number,
-  height: number
+  height: number,
+  isWidthInitialized: boolean
 ] {
   const gridRef = useRef<HTMLDivElement>(null);
   const [inlineSize, setInlineSize] = useState(1);
   const [blockSize, setBlockSize] = useState(1);
+  const [isWidthInitialized, setWidthInitialized] = useState(false);
```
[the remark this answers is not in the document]
amanmahajan7: I can try in the next PR

src/hooks/useViewportColumns.ts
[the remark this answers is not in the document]
amanmahajan7: I have removed it. I can add it back if you think it is necessary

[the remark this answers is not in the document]
nstepien: Might run the layout effect a lot, let's keep it

src/DataGrid.tsx
nstepien: >It is possible to have flexWidthViewportColumns .length > renderedFlexWidthViewportColumns.length

Is it? `flexWidthViewportColumns` is a subset of `viewportColumns` though

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: This can fail if header cell has a colSpan

PR #2848, opened 2022-02-28
Use scrollIntoView to scroll to the selected cell

amanmahajan7:
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView
https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding
https://caniuse.com/?search=scroll-padding

https://caniuse.com/?search=scrollIntoView
Safari does not support `scrollIntoViewOptions`. It is fixed in TP so we can wait before merging this PR

```
element.scrollIntoView(scrollIntoViewOptions); 

interface ScrollIntoViewOptions extends ScrollOptions {
    block?: ScrollLogicalPosition;
    inline?: ScrollLogicalPosition;
}

type ScrollLogicalPosition = "center" | "end" | "nearest" | "start";
```

Edit: 
https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes
> Added support for the CSS scroll-behavior property and ScrollOptions, allowing smooth scrolling to anchors or via JavaScript.

We can revert if safari has any issues


src/DataGrid.tsx
```diff
@@ -730,18 +729,18 @@ function DataGrid<R, SR, K extends Key>(
...
       // TODO: replace with a #record? https://github.com/microsoft/TypeScript/issues/39831
-      scrollToCell(position);
+      scrollIntoView(gridRef.current?.querySelector('[tabindex="0"]'));
     } else {
       setSelectedPosition({ ...position, mode: 'SELECT' });
     }
   }
 
-  function scrollToCell({ idx, rowIdx }: Partial<Position>): void {
+  function scrollToColumn(idx: number): void {
```
amanmahajan7, line 738: We may be able to render an extra div, call `scrollIntoView` and remove it from the DOM.

src/DataGrid.tsx
```diff
@@ -1141,6 +1125,15 @@ function DataGrid<R, SR, K extends Key>(
       style={
         {
           ...style,
+          // handle sticky row/columns
```
amanmahajan7, line 1128: Do not need this on the sticky cells

test/column/colSpan.test.ts
```diff
@@ -147,7 +147,7 @@ describe('colSpan', () => {
     validateCellPosition(6, 11);
   });
 
-  it('should scroll to the merged cell when selected', () => {
+  it.skip('should scroll to the merged cell when selected', () => {
```
amanmahajan7, line 150: Not sure how to fix these this test as we no longer set `scrollLeft` manually. We can probably delete it

nstepien: Can we at least check that we call `scrollIntoView`?

src/DataGrid.tsx
```diff
@@ -1141,6 +1125,15 @@ function DataGrid<R, SR, K extends Key>(
       style={
         {
           ...style,
+          // set scrollPadding to correctly position non sticky cells
```
nstepien, line 1128:
```suggestion
          // set scrollPadding to correctly position non-sticky cells after scrolling
```

test/column/colSpan.test.ts
```diff
@@ -149,23 +148,28 @@ describe('colSpan', () => {
 
   it('should scroll to the merged cell when selected', () => {
     setupColSpanGrid(30);
-    const grid = screen.getByRole('grid');
     userEvent.click(getCellsAtRowIndex(8)[23]); // last visible cell (1920/80)
+    const spy = jest.spyOn(window.HTMLElement.prototype, 'scrollIntoView');
     navigate(3);
-    expect(grid.scrollLeft).toBe(240);
+    expect(spy).toHaveBeenCalled();
+    spy.mockReset();
```
nstepien, line 155: We could check that the number of calls is incremented instead of resetting each time, up to you
```suggestion
    expect(spy).toHaveBeenCalledTimes(1);
```

nstepien (dismissed):

src/DataGrid.tsx
```diff
@@ -730,18 +729,18 @@ function DataGrid<R, SR, K extends Key>(
     } else if (isSamePosition(selectedPosition, position)) {
       // Avoid re-renders if the selected cell state is the same
       // TODO: replace with a #record? https://github.com/microsoft/TypeScript/issues/39831
```
nstepien, line 731:
```suggestion
```

src/hooks/useRovingCellRef.ts
```diff
@@ -10,8 +11,9 @@ export function useRovingCellRef(isSelected: boolean) {
   }
 
   const ref = useCallback((cell: HTMLDivElement | null) => {
-    if (cell === null || cell.contains(document.activeElement)) return;
-
+    if (cell === null) return;
+    scrollIntoView(cell);
```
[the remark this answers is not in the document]
amanmahajan7: https://github.com/adazzle/react-data-grid/pull/2848/files#diff-0f9c631e39f2adf9fe111a2bb9ca7fab29327c1d43978939e30d3d36f5177ac4L408

We no longer scroll in DataGrid

test/column/editor.test.tsx
```diff
@@ -99,8 +99,9 @@ describe('Editor', () => {
     expect(screen.queryByLabelText('col1-editor')).not.toBeInTheDocument();
     userEvent.keyboard('123');
     expect(screen.getByLabelText('col1-editor')).toHaveValue(123);
+    const spy = jest.spyOn(window.HTMLElement.prototype, 'scrollIntoView');
     userEvent.keyboard('{enter}');
-    expect(getCellsAtRowIndex(0)).toHaveLength(2);
```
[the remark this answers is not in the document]
amanmahajan7: We cannot as the cell is scrolled using `scrollIntoView`. Previously we were manually calling `scrollToCell` on `Enter` but now the grid would unmount the editor, render the cell which in turn would call `scrollIntoView`

src/hooks/useRovingCellRef.ts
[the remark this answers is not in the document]
amanmahajan7: Only when we click on already selected cell. `cell` won't call `scrollIntoView` from the `ref` callback if it is selected and you click on it again.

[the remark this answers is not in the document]
amanmahajan7: Double checking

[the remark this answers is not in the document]
amanmahajan7: I can confirm it does not call it twice when the cell does not have focusable element. For a column like checkbox, it does call it twice but it is the same behavior without the current change

nstepien (approved):

src/DataGrid.tsx
```diff
@@ -405,10 +406,10 @@ function DataGrid<R, SR, K extends Key>(
     }
 
     prevSelectedPosition.current = selectedPosition;
-    scrollToCell(selectedPosition);
 
     if (selectedPosition.idx === -1) {
       rowRef.current!.focus({ preventScroll: true });
+      scrollIntoView(rowRef.current);
```
godon019, line 412: question: why do we not use
`rowRef.current!.focus()` only for scroll?
I am not sure why we preventScroll in focus and scroll with another function

PR #2849, opened 2022-02-28
Fix viewport columns for header and summary rows

amanmahajan7:
This is needed to tab into the grid. Currently the selected cell is rendered for the viewport rows but not for the header and summary rows

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -977,6 +977,7 @@ function DataGrid<R, SR, K extends Key>(
   function getRowViewportColumns(rowIdx: number) {
     const selectedColumn = columns[selectedPosition.idx];
     if (
+      // idx can be -1 if grouping is enabled
       // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
       selectedColumn !== undefined &&
       selectedPosition.rowIdx === rowIdx &&
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -974,6 +974,26 @@ function DataGrid<R, SR, K extends Key>(
     );
   }
 
+  function getRowViewportColumns(rowIdx: number) {
+    const selectedColumn = columns[selectedPosition.idx];
+    if (
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      selectedColumn !== undefined &&
+      selectedPosition.rowIdx === rowIdx &&
+      !viewportColumns.includes(selectedColumn)
+    ) {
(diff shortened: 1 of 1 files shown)


src/DataGrid.tsx
```diff
@@ -974,6 +974,27 @@ function DataGrid<R, SR, K extends Key>(
...
   }
 
+  function getRowViewportColumns(rowIdx: number) {
+    const selectedColumn = columns[selectedPosition.idx];
+    if (
+      // idx can be -1 if grouping is enabled
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      selectedColumn !== undefined &&
+      selectedPosition.rowIdx === rowIdx &&
+      !viewportColumns.includes(selectedColumn)
```
nstepien, line 984: Might be cheaper to just compare `selectedPosition.idx` with `colOverscanStartIdx`/`colOverscanEndIdx`/`lastFrozenColumnIndex`

src/DataGrid.tsx
```diff
@@ -974,6 +974,27 @@ function DataGrid<R, SR, K extends Key>(
     );
   }
 
+  function getRowViewportColumns(rowIdx: number) {
```
nstepien, line 977: Worth memoizing?

src/DataGrid.tsx
```diff
@@ -974,6 +974,27 @@ function DataGrid<R, SR, K extends Key>(
...
+    const selectedColumn = columns[selectedPosition.idx];
+    if (
+      // idx can be -1 if grouping is enabled
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      selectedColumn !== undefined &&
+      selectedPosition.rowIdx === rowIdx &&
+      !viewportColumns.includes(selectedColumn)
+    ) {
+      // Add the selected column to viewport columns if the cell is not within the viewport
+      return selectedPosition.idx > viewportColumns[viewportColumns.length - 1].idx
```
nstepien, line 987:
```suggestion
      return selectedPosition.idx > colOverscanEndIdx
```

src/DataGrid.tsx
amanmahajan7: Probably not. It is only for the selected row and if the selected idx is not in the viewport

src/DataGrid.tsx
amanmahajan7: https://github.com/adazzle/react-data-grid/pull/2562#discussion_r689614421

nstepien (approved):

PR #2851, opened 2022-03-01
Remove scrollToCell

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -958,25 +932,44 @@ function DataGrid<R, SR, K extends Key>(
...
+
+  function getRowViewportColumns(rowIdx: number) {
+    let rowViewportColumns = viewportColumns;
+    if (rowIdx === selectedPosition.rowIdx) {
+      rowViewportColumns = addOutOfBoundColumnToRowViewportColumns(
+        selectedPosition.idx,
+        viewportColumns
+      );
+    }
+    if (rowIdx === -1 && scrollToColumnIdx !== undefined) {
```
amanmahajan7, line 966: Same logic as the selected cell. We need to add the column if it is not in the viewport

src/DataGrid.tsx
```diff
@@ -734,44 +746,6 @@ function DataGrid<R, SR, K extends Key>(
       setSelectedPosition({ ...position, mode: 'SELECT' });
     }
   }
-
-  function scrollToColumn(idx: number): void {
```
amanmahajan7, line 738: This won't work with writing mode. Also, we do not need special handling for RTL if we use `scrollIntoView`

website/demos/ScrollToCell.tsx
```diff
@@ -0,0 +1,82 @@
+import { useRef, useState } from 'react';
+
+import DataGrid from '../../src';
+import type { Column, FormatterProps, DataGridHandle } from '../../src';
+import type { Props } from './types';
+
+type Row = number;
+const rows: readonly Row[] = [...Array(200).keys()];
```
nstepien, line 8: Optional, but this an optimized way to do this (probably)
```suggestion
const rows: readonly Row[] = Array.from({ length: 200 }, (_, i) => i);
```

website/demos/ScrollToCell.tsx
```diff
@@ -0,0 +1,82 @@
...
+            type="number"
+            value={idx}
+            onChange={(event) => {
+              const { valueAsNumber } = event.target;
+              setIdx(Number.isNaN(valueAsNumber) ? undefined : valueAsNumber);
+            }}
+          />
+        </label>
+        <button type="button" onClick={scrollToCell}>
+          Scroll to column
```
nstepien, line 47: Should we only have 1 button? Otherwise "Scroll to column" will also scroll to the row which I found confusing.

website/demos/ScrollToCell.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return (
+    <>
+      <div style={{ marginBlockEnd: 5 }}>
+        <label>
+          <span style={{ marginInlineEnd: 5 }}>Column index: </span>
+          <input
+            style={{ inlineSize: 50 }}
+            type="number"
+            value={idx}
```
nstepien, line 39: We're getting React warnings when it becomes undefined
```suggestion
            value={idx ?? ''}
```

website/demos/ScrollToCell.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </label>
+        <button type="button" onClick={scrollToCell}>
+          Scroll to column
+        </button>
+        <label>
+          <span style={{ marginInline: 5 }}>Row index: </span>
+          <input
+            style={{ inlineSize: 50 }}
+            type="number"
+            value={rowIdx}
```
nstepien, line 54:
```suggestion
            value={rowIdx ?? ''}
```

src/DataGrid.tsx
```diff
@@ -448,14 +448,15 @@ function DataGrid<R, SR, K extends Key>(
...
-        behavior: 'smooth'
-      });
+    scrollToCell({ idx, rowIdx }) {
+      const scrollToIdx =
+        idx !== undefined && idx > lastFrozenColumnIndex && idx < columns.length ? idx : undefined;
+      const scrollToRowIdx =
+        rowIdx !== undefined && isRowIdxWithinViewportBounds(rowIdx) ? rowIdx : undefined;
+
+      if (scrollToIdx !== undefined || scrollToRowIdx !== undefined) {
+        setScrollToPosition({ idx: scrollToIdx, rowIdx: scrollToRowIdx });
```
nstepien, line 458: If you want, an alternative could be:
- create a div with the same styles
- append it to the grid
- scroll to it
- immediately remove it

src/DataGrid.tsx
```diff
@@ -1151,11 +1114,12 @@ function DataGrid<R, SR, K extends Key>(
           ...style,
           // set scrollPadding to correctly position non-sticky cells after scrolling
           scrollPaddingInlineStart:
-            selectedPosition.idx > lastFrozenColumnIndex
+            selectedPosition.idx > lastFrozenColumnIndex || scrollToPosition?.idx !== undefined
```
nstepien, line 1117: Why do we need this?

amanmahajan7: This is needed otherwise cell will be behind the frozen columns or header/summary rows. T

https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-inline-start
> The scroll-padding-inline-start property defines offsets for the start edge in the inline dimension of the optimal viewing region of the scrollport: the region used as the target region for placing things in view of the user.

src/DataGrid.tsx
amanmahajan7: I was thinking the same but I need to set [styles ](https://github.com/adazzle/react-data-grid/pull/2851#discussion_r1204541214)on the grid also. May be more work to set the previous styles after removing div. Wdyt?

src/DataGrid.tsx
nstepien: Ah right I misunderstood.
Now my question is: when do we want `scrollPaddingInlineStart` to be `undefined`? Would it simplify things if it was always set? Even if it's `0px`

src/DataGrid.tsx
nstepien: Makes sense

nstepien (dismissed):

src/DataGrid.tsx
amanmahajan7: It seems like it is needed
Recording 2023-05-25 at 11 57 52

nstepien: Might be an issue in Chrome, I don't think I can reproduce the same issue in Firefox.
Should open a chromium issue if there isn't one already.

src/ScrollToCell.tsx
```diff
@@ -0,0 +1,34 @@
...
+}: {
+  scrollToPosition: PartialPosition;
+  onScroll: () => void;
+}) {
+  const ref = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
+    scrollIntoView(ref.current);
+    onScroll();
+  });
```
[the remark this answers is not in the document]
amanmahajan7: We can try

src/DataGrid.tsx
```diff
@@ -1189,6 +1153,27 @@ function DataGrid<R, SR, K extends Key>(
...
+      {scrollToPosition !== null && (
+        <ScrollToCell
+          scrollToPosition={scrollToPosition}
+          onScroll={() => {
+            if (scrollToPosition.idx !== undefined) {
+              const { key, width } = columns[scrollToPosition.idx];
+              if (
+                // with auto-sized column we need to keep scrolling
+                // until the column width is set and it is in the viewport
+                typeof width === 'string' &&
```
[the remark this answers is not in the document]
amanmahajan7: they will but I think if the width is set then we can safely assume the cell is visible 🤔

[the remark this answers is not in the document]
nstepien: Imagine this scenario:
- you've jumped to the end of the grid
- you `scrollToCell` to a column on the left
- scrolling to it would render other columns on its cell that may not have been sized before, or were invalidated

src/ScrollToCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+  setScrollToCellPosition: (cell: null) => void;
+}) {
+  const ref = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
+    function removeScrollToCell() {
+      setScrollToCellPosition(null);
+    }
+
+    const observer = new IntersectionObserver(removeScrollToCell, {
```
nstepien, line 27: Won't the callback be called when the element's visibility goes below the threshold too?
Not sure if it can be an issue.

What happens when the cell is already visible?

src/ScrollToCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    observer.observe(ref.current!);
+
+    return () => {
+      observer.disconnect();
+    };
+  }, [gridElement, setScrollToCellPosition]);
+
+  useLayoutEffect(() => {
```
nstepien, line 39: Should this effect run first?

amanmahajan7: I was thinking the same. I tried both and it did not make a difference. I can move it

src/ScrollToCell.tsx
amanmahajan7: > What happens when the cell is already visible?

It correctly calls the callback and remove the cell.

Not sure about the other question. I will check

amanmahajan7: May be we can check `intersectionRatio`

amanmahajan7: But like you said, it should not be an issue

nstepien (approved):

PR #2855, opened 2022-03-16
Fix min/max width

amanmahajan7:
Cherry picked changes from https://github.com/adazzle/react-data-grid/pull/2839


.eslintrc.cjs
```diff
@@ -533,21 +533,8 @@ const rules = {
...
-    },
-    {
-      selector: ['function', 'parameter', 'property'],
-      format: ['camelCase', 'PascalCase']
-    },
-    { selector: 'memberLike', format: ['camelCase'] },
-    { selector: 'typeLike', format: ['PascalCase'] }
-  ],
+  // TODO: fix
+  '@typescript-eslint/naming-convention': 0,
```
amanmahajan7, line 537: Not sure if this a bug or needs to be tweaked

src/HeaderCell.tsx
```diff
@@ -87,7 +87,7 @@ export default function HeaderCell<R, SR>({
       const { right, left } = currentTarget.getBoundingClientRect();
       const width = isRtl ? right + offset - event.clientX : event.clientX + offset - left;
       if (width > 0) {
-        onColumnResize(column, width);
+        onColumnResize(column, clampColumnWidth(width, column));
```
amanmahajan7, line 90: Otherwise `DataGrid.onColumnResize` is called with wrong width as it does not respect min/max width

nstepien (approved):

PR #2868, opened 2022-03-31
@testing-library/user-event@14

nstepien:
https://github.com/testing-library/user-event/releases


test/column/editor.test.tsx
```diff
@@ -13,64 +13,64 @@ interface Row {
...
-    userEvent.dblClick(getCellsAtRowIndex(0)[0]);
+    await userEvent.dblClick(getCellsAtRowIndex(0)[0]);
     expect(screen.getByLabelText('col1-editor')).toHaveValue(1);
-    userEvent.type(document.activeElement!, '2');
-    userEvent.tab();
+    await userEvent.keyboard('2');
+    await userEvent.tab();
     expect(screen.queryByLabelText('col1-editor')).not.toBeInTheDocument();
-    expect(getCellsAtRowIndex(0)[0]).toHaveTextContent(/^12$/);
+    expect(getCellsAtRowIndex(0)[0]).toHaveTextContent(/^21$/);
```
nstepien, line 25: Try loading
```
data:text/html,<!doctype html><input autofocus value="0">
```
in a new tab, `autofocus` sets the selection range to the beginning of the input.
Using `.type` types the value at the end of the input rather than where the caret currently is.
I guess the behavior for `.keyboard` changed.

amanmahajan7 (approved):

PR #2882, opened 2022-04-06
ResizeObserver: Use contentBoxSize

amanmahajan7:
[FF bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1733042) is fixed in 99. We can wait for a few weeks before merging this PR to give everyone some time

### DIFF
--- a/src/hooks/useGridDimensions.ts
+++ b/src/hooks/useGridDimensions.ts
@@ -44,5 +44,5 @@ export function useGridDimensions(): [
 // we reduce width by 1px here to avoid layout issues in Chrome
 // https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
 function handleDevicePixelRatio(size: number) {
-  return size - (devicePixelRatio % 1 === 0 ? 0 : 1);
+  return size - (devicePixelRatio === 1 ? 0 : ceil(devicePixelRatio));
 }
--- a/src/hooks/useGridDimensions.ts
+++ b/src/hooks/useGridDimensions.ts
@@ -17,9 +17,13 @@ export function useGridDimensions(): [
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     if (ResizeObserver == null) return;
 
-    const { clientWidth, clientHeight } = gridRef.current!;
-    setInlineSize(handleDevicePixelRatio(clientWidth));
-    setBlockSize(clientHeight);
+    const { clientWidth, clientHeight, offsetWidth, offsetHeight } = gridRef.current!;
+    const { width, height } = gridRef.current!.getBoundingClientRect();
+    const initialWidth = width - offsetWidth + clientWidth;
+    const initialHeight = height - offsetHeight + clientHeight;
+
+    setInlineSize(handleDevicePixelRatio(initialWidth));
+    setBlockSize(initialHeight);
 
     const resizeObserver = new ResizeObserver((entries) => {
       const size = entries[0].contentBoxSize[0];
--- a/src/hooks/useGridDimensions.ts
+++ b/src/hooks/useGridDimensions.ts
@@ -17,12 +17,16 @@ export function useGridDimensions(): [
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     if (ResizeObserver == null) return;
 
+    const { clientWidth, clientHeight } = gridRef.current!;
+    setInlineSize(handleDevicePixelRatio(clientWidth));
+    setBlockSize(clientHeight);
+
     const resizeObserver = new ResizeObserver((entries) => {
       const size = entries[0].contentBoxSize[0];
       // TODO: remove once fixed upstream
       // we reduce width by 1px here to avoid layout issues in Chrome
       // https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
-      setInlineSize(size.inlineSize - (devicePixelRatio % 1 === 0 ? 0 : 1));
+      setInlineSize(handleDevicePixelRatio(size.inlineSize));
       setBlockSize(size.blockSize);
     });
     resizeObserver.observe(gridRef.current!);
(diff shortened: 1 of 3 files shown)


src/hooks/useGridDimensions.ts
```diff
@@ -17,26 +17,20 @@ export function useGridDimensions(): [
...
       // TODO: remove once fixed upstream
       // we reduce width by 1px here to avoid layout issues in Chrome
       // https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
-      setGridWidth(clientWidth - (devicePixelRatio % 1 === 0 ? 0 : 1));
-      setGridHeight(clientHeight);
-    }
-
-    saveDimensions();
-    const resizeObserver = new ResizeObserver(saveDimensions);
+      setInlineSize(size.inlineSize - (devicePixelRatio % 1 === 0 ? 0 : 1));
```
amanmahajan7, line 25: As we are using `inlineSize/blockSize`, RDG is one step closer to handling writing mode
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentBoxSize#value

src/hooks/useGridDimensions.ts
```diff
@@ -17,26 +17,20 @@ export function useGridDimensions(): [
...
+    const resizeObserver = new ResizeObserver((entries) => {
+      const size = entries[0].contentBoxSize[0];
       // TODO: remove once fixed upstream
       // we reduce width by 1px here to avoid layout issues in Chrome
       // https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
-      setGridWidth(clientWidth - (devicePixelRatio % 1 === 0 ? 0 : 1));
-      setGridHeight(clientHeight);
-    }
-
-    saveDimensions();
```
amanmahajan7, line 32: We no longer initialize the value. Let's revisit in the react 18 PR

nstepien: Do we not care about this anymore? We could still get `clientWidth`/`clientHeight` early.
https://github.com/adazzle/react-data-grid/commit/32a20f43aac88f56ed06e47fc9c0f6e8295002c8

src/hooks/useGridDimensions.ts
nstepien: Can we check that these values are the same as `clientWidth`/`clientHeight`?
Something like
```ts
if (clientWidth !== inlineSize || clientHeight !== blockSize) {
  throw new Error('🤔');
}
```
Just to make sure, not to be commited.

amanmahajan7: image

src/hooks/useGridDimensions.ts
amanmahajan7: I can add it back but the value is not the same. Check the above comment

src/hooks/useGridDimensions.ts
nstepien: I guess we can give it a try, but I'm worried about decimal values leading to more shaky grids.

src/hooks/useGridDimensions.ts
nstepien: Can you compare with `grid.getBoundingClientRect()`?

nstepien: `getBoundingClientRect()` doesn't remove the scrollbar width, try
`$0.getBoundingClientRect().width - ($0.offsetWidth - $0.clientWidth)`

src/hooks/useGridDimensions.ts
amanmahajan7: If we have a reproducible example then we can give it a try

src/hooks/useGridDimensions.ts
amanmahajan7: Added the the above logic back. It is however using `width/height` instead of `inlineSize/blockSize` which okay for now I gues

src/hooks/useGridDimensions.ts
```diff
@@ -17,26 +17,35 @@ export function useGridDimensions(): [
...
+    const initialHeight = height - offsetHeight + clientHeight;
+
+    setInlineSize(handleDevicePixelRatio(initialWidth));
+    setBlockSize(initialHeight);
+
+    const resizeObserver = new ResizeObserver((entries) => {
+      const size = entries[0].contentBoxSize[0];
       // TODO: remove once fixed upstream
       // we reduce width by 1px here to avoid layout issues in Chrome
       // https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
```
nstepien, line 32: Let's not duplicate this comment
```suggestion
```

test/setup.ts
```diff
@@ -9,7 +9,9 @@ if (typeof window !== 'undefined') {
     }
 
     observe() {
-      this.callback([], this);
+      // patch inlineSize/blockSize to pretend we're rendering DataGrid at 1920p/1080p
+      // @ts-expect-error
+      this.callback([{ contentBoxSize: [{ inlineSize: 1920, blockSize: 1080 }] }], this);
```
nstepien, line 14: Can we remove the `clientWidth`/`clientHeight` patch below?

src/hooks/useGridDimensions.ts
```diff
@@ -17,26 +17,35 @@ export function useGridDimensions(): [
...
 
-  return [gridRef, gridWidth, gridHeight];
+  return [gridRef, inlineSize, blockSize];
+}
+
+// TODO: remove once fixed upstream
+// we reduce width by 1px here to avoid layout issues in Chrome
+// https://bugs.chromium.org/p/chromium/issues/detail?id=1206298
+function handleDevicePixelRatio(size: number) {
+  return size - (devicePixelRatio % 1 === 0 ? 0 : 1);
```
nstepien, line 50: Let's try this
https://github.com/adazzle/react-data-grid/pull/2870#issuecomment-1084687538
```suggestion
  return size - (devicePixelRatio === 1 ? 0 : ceil(devicePixelRatio));
```

test/setup.ts
amanmahajan7: It is still needed. We can also change the `window.ResizeObserver` to not call `callback`

nstepien (approved):

PR #2883, opened 2022-04-07
Refactor useRovingCellRef, fix test warnings

nstepien:
Extracted from #2459

### DIFF
--- a/src/hooks/useRovingCellRef.ts
+++ b/src/hooks/useRovingCellRef.ts
@@ -1,35 +1,30 @@
-import { useRef, useState } from 'react';
-import { useLayoutEffect } from './useLayoutEffect';
+import { useCallback, useState } from 'react';
 
 // https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_roving_tabindex
 export function useRovingCellRef(isSelected: boolean) {
-  const ref = useRef<HTMLDivElement>(null);
   // https://www.w3.org/TR/wai-aria-practices-1.1/#gridNav_focus
-  const isChildFocused = useRef(false);
-  const [, forceRender] = useState<unknown>({});
+  const [isChildFocused, setIsChildFocused] = useState(false);
 
-  useLayoutEffect(() => {
-    if (!isSelected) {
-      isChildFocused.current = false;
-      return;
-    }
+  if (isChildFocused && !isSelected) {
+    setIsChildFocused(false);
+  }
 
-    if (isChildFocused.current) {
-      // When the child is focused, we need to rerender
-      // the cell again so tabIndex is updated to -1
-      forceRender({});
-      return;
-    }
-    ref.current?.focus({ preventScroll: true });
-  }, [isSelected]);
+  const ref = useCallback(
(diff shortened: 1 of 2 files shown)


src/hooks/useRovingCellRef.ts
```diff
@@ -1,35 +1,30 @@
...
-      forceRender({});
-      return;
-    }
-    ref.current?.focus({ preventScroll: true });
-  }, [isSelected]);
+  const ref = useCallback(
+    (cell: HTMLDivElement | null) => {
+      if (cell === null || !isSelected || cell.contains(document.activeElement)) return;
+
+      cell.focus({ preventScroll: true });
```
amanmahajan7, line 16: My only concern is if it is possible to have a stale ref in a ref callback. Not sure why is `useLayoutEffect` is the recommended approach

src/hooks/useRovingCellRef.ts
```diff
@@ -1,35 +1,30 @@
...
+    if (event.target !== event.currentTarget) {
+      setIsChildFocused(true);
     }
   }
 
-  const isFocused = isSelected && !isChildFocused.current;
+  const isFocused = isSelected && !isChildFocused;
 
   return {
     ref,
```
amanmahajan7, line 30:
```suggestion
    ref: isSelected ? ref : null,
```
Wdyt?

src/hooks/useRovingCellRef.ts
```diff
@@ -1,35 +1,30 @@
...
-    }
-    ref.current?.focus({ preventScroll: true });
-  }, [isSelected]);
+  const ref = useCallback(
+    (cell: HTMLDivElement | null) => {
+      if (cell === null || !isSelected || cell.contains(document.activeElement)) return;
+
+      cell.focus({ preventScroll: true });
+    },
+    [isSelected]
```
amanmahajan7, line 18:
```suggestion
    []
```

We can remove this dependency then

src/hooks/useRovingCellRef.ts
nstepien: WDYM by stale ref?
`useLayoutEffect` would use that same ref anyway.

src/hooks/useRovingCellRef.ts
```diff
@@ -1,38 +1,30 @@
...
   }
 
-  const isFocused = isSelected && !isChildFocused.current;
+  const isFocused = isSelected && !isChildFocused;
 
   return {
-    ref,
+    ref: isSelected ? ref : undefined,
     tabIndex: isFocused ? 0 : -1,
     onFocus
```
amanmahajan7, line 29: `onFocus: isSelected ? onFocus: undefined` Would this work?

nstepien: It does, but then it does not catch programmatically focusing elements within a cell.
Then again we don't change the cell selection when we programmatically focus something within cells.

amanmahajan7 (dismissed):

amanmahajan7 (approved):

PR #2886, opened 2022-04-07
Fix cell focus stealing strategy

nstepien:
NVM we steal focus immediately on `mousedown` before we can focus on anything else it seems...


RafalWilinski:
Thanks for fixing this!

gkn:
@nstepien Will this be merged to main do you think? I have used this fix for a while in a project now and it works fine as far as I can see. Would be nice to avoid my forked dependency :)

nstepien:
@gkn it doesn't actually work though from my testing, the grid will steal focus when trying to focus something outside the grid while an editor is open
Recording 2022-09-29 at 18 34 12

gkn:
I use an older fork with this PR [https://github.com/gkn/react-data-grid](https://github.com/gkn/react-data-grid)

I need to click twice with the mouse on the input field outside the grid, and it will work.
With the unpatched version of the grid, there is no way to keep focus outside the grid.
Hope this might lead to a solution for you :)

PR #2932, opened 2022-06-09
Add onSelectedCellChange, cellRenderer, foucus on mouse down

Wroud:
In our product we need `onSelectedCellChange` to track focused cell.
Also we need `cellRenderer` to custumize cells.
"focus on click" related to some issues with "onClick" actions. (for example we have menu button in cells)

It will be cool if you be able to accept this editings.


Wroud:
separate pr will be created

Issue #2712, opened 2021-10-22
ResizeObserver loop limit exceeded


RMacfarlane:
## Describe the bug 

This doesn't seem to be impacting functional behavior, but is generating error noise. A "ResizeObserver loop limit exceeded" is fired on the window when resizing the window.

## To Reproduce 

Using this code base and google chrome > 64:

1. Add
```
window.addEventListener('error', (e) => {
  // eslint-disable-next-line no-console
  console.log(e)
})
```
on https://github.com/adazzle/react-data-grid/blob/ec2805865dab79f7abcda0208d87f12a5019dda4/src/hooks/useGridDimensions.ts#L3

2. Run `npm start`
3. Maximize the browser window and then resize it to be smaller on the common features example
4. Looking at the dev tools, `ResizeObserver loop limit exceeded` is logged repeatedly

## Expected behavior 

No window errors are generated

## Environment

- `react-data-grid` version: latest
- `react`/`react-dom` version: `^17.0.2`

## Additional context 
There's a good write up of this issue [here](https://github.com/WICG/resize-observer/issues/38) and more discussion [here](https://github.com/w3c/csswg-drafts/issues/5488#issuecomment-812279513)

The most common approach to preventing this error from being fired seems to be to use `requestAnimationFrame`

soichih:
I am having the same problem, and I confirm that the #2933 fixes it.

ofirelias:
I'm having the same issue, has anyone found an elegant fix for this?

amanmahajan7:
Duplicate issue https://github.com/adazzle/react-data-grid/issues/3314

PR #2933, opened 2022-06-19
add request animation frame to prevent resize observer limit loop exception

Karpengold:
Fix for "ResizeObserver loop limit exceeded" inspired by this topic: https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded

Related to https://github.com/adazzle/react-data-grid/issues/2712

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-data-grid",
-  "version": "7.0.0-beta.32",
+  "version": "7.0.0-beta.33",
   "license": "MIT",
   "description": "Feature-rich and customizable data grid React component",
   "keywords": [


Karpengold:
@nstepien @amanmahajan7 hi guys, how do you think is it possible to merge it in order to prevent this error in console?

damnitrahul:
Hey @nstepien @amanmahajan7
Can you guys please merge this? This will stop spamming sentry with issues. 
Thanks!

nstepien:
Could you create a reproducible demo first?
It's probably better to let it crash than have the resize observer trigger re-renders infinitely.

soichih:
This issue "observer limit loop exception" crashes the react-script dev instance so it's highly desirable that this issue is resolved. To reproduce this, you will need sufficiently complex UI so that each resize event takes long time to process (and observer loop will reach its limit).

The PR seems to be conflicting so I've updated to something like this.

```
const resizeObserver = new ResizeObserver(entries => {
    window.requestAnimationFrame(() => {
        if (!Array.isArray(entries) || !entries.length) {
            return;
        }
        flushSync(() => {
            setInlineSize(size.inlineSize);
            setBlockSize(size.blockSize);
        });
    });
});
```

I confirm that this patch fixes the issue and react-data-grid becomes usable again.

nstepien:
@soichih Could you please provide a reproducible demo?

mikeeus:
We're seeing this issue as well in a very large application and only on certain browsers. I'm not able to reliably reproduce this on a small demo but its definitely an issue and impacting a significant portion of our users.

davidjamesfriedman (approved):

mikeeus (approved):
Fantastic work!

Gwitchr (approved):

kcode20 (approved):

Karpengold:
Resolved conflicts in case maintainers would like to merge it :)

SarunasMaro:
I have this issue using the latest code base :) I don't think it was resolved yet

Issue #2937, opened 2022-06-23
Core features request


Wroud:
## Use case
In our product we need to have the ability to customize Cell rendering (https://github.com/adazzle/react-data-grid/issues/2799) for:
1. additional props on the cell root element
2. pass context to formatters
3. customize event handlers
4. modify props (for example `isCellSelected`)

Disable selection for specified columns/cells (https://github.com/adazzle/react-data-grid/issues/1859):
1. In our product we have the index column, the user can't edit it or anything else. We don't want to select these cells

Select the cell on the mouse down
1. it's the default behavior in many tables
2. prevent focus stealing in cases like cell component rerendering (in our case it leads to scrolling to a selected cell for no reason (if the cell is outside visible boundaries))

IsScrolling property for components
1. In our product we can simplify component rendering while scrolling to improve performance

onSelectCellChange
1. In our product we need to sync selection with a table (also we want to have the ability to manually manipulate selection)

Wroud:
@nstepien Hi! Please, can you give any feedback for this PR?

Wroud:
@nstepien hi, I've updated pull requests, may you take a look please?

Wroud:
I've updated the pull requests, may you take a look?
We are using this library in production and we want to migrate to the latest version. It will be cool to have this features as a part of the react-data-grid.

flipvh:
I would be very interested in this too. Is there still an interest by @Wroud too and/or maintainers? Thanks!

Wroud:
@amanmahajan7 hi, can you please take a look? I've updated both PRs.

PR #2941, opened 2022-06-29
Cell select callback

Wroud:
#2937


Wroud:
please take a look at this small improvement

amanmahajan7:
Thanks for the PR. This is available in RDG now
https://github.com/adazzle/react-data-grid/blob/main/src/DataGrid.tsx#L171

PR #2673, opened 2021-09-24
Do not steal focus back to the cell when editor is closing by clicking outside the grid

amanmahajan7:

PR #2943, opened 2022-06-29
Fix cell focus stealing

Wroud:
#2937

### DIFF
--- a/src/hooks/useRovingCellRef.ts
+++ b/src/hooks/useRovingCellRef.ts
@@ -1,32 +1,34 @@
-import { useCallback, useState } from 'react';
+import { useState } from 'react';
+import { useLayoutEffect } from './useLayoutEffect';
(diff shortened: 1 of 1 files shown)


gkn:
@nstepien I merged this in a fork a while back and it seems to work great. Will it be merged into main do you think?

nstepien:
This doesn't seem to work, the grid steals focus as usual
Recording 2022-09-29 at 18 40 27

Wroud:
I will check this case, because my original one was about focusing element on re-render. Can you please describe your case more detaily?
How can i reproduce it?

kcvgan (approved):

nstepien:
>How can i reproduce it?

Open an editor, then try to focus an input outside the grid, like in the gif above.

amanmahajan7:
Fixed in https://github.com/adazzle/react-data-grid/pull/2673

PR #2944, opened 2022-06-29
Feat add selectedCell to ref

Wroud:
#2937


amanmahajan7:
This is an internal state and we don't want to add it to the ref. You can use `onSelectedCellChange` and use a local state in your component

PR #2947, opened 2022-07-01
Replace component types in favor of render functions

### LINKED ISSUE #2872: RDG should use render funcs instead of components
Currently, RDG relies on components to customize rendering:
```tsx
  /** Formatter to be used to render the cell content */
  readonly formatter?: Maybe<ComponentType<FormatterProps<TRow, TSummaryRow>>>;
  /** Formatter to be used to render the summary cell content */
  readonly summaryFormatter?: Maybe<ComponentType<SummaryFormatterProps<TSummaryRow, TRow>>>;
  /** Formatter to be used to render the group cell content */
  readonly groupFormatter?: Maybe<ComponentType<GroupFormatterProps<TRow, TSummaryRow>>>;
```

This is problematic:
- Customization is more difficult, we have to create a component every time, but sometimes can be easily solved with a render func
- Error prone, as a lambda can be passed. This prevents React from reusing the DOM and causes a performance penalty.

Instead, RDG should require and use render funcs:
```tsx
  /** Formatter to be used to render the cell content */
  readonly formatter?: Maybe<(props: FormatterProps<TRow, TSummaryRow>) => React.ReactNode>;
  /** Formatter to be used to render the summary cell content */
  readonly summaryFormatter?: Maybe<(props: SummaryFormatterProps<TSummaryRow, TRow>) => React.ReactNode>;
  /** Formatter to be used to render the group cell content */
  readonly groupFormatter?: Maybe<(props: GroupFormatterProps<TRow, TSummaryRow>) => React.ReactNode>;
```
Note that with this approach, should one want, they can still use a component easily:
```tsx
formatter: (props) => <MyComponent {...props} />
```

React libraries have been moving to this approach. In particular, `react-router` has switched from 4 => 5. Please read more about it here: https://reactrouter.com/docs/en/v6/upgrading/v5#advantages-of-route-element

nstepien:
Resolves #2872


src/formatters/checkboxFormatter.tsx
```diff
@@ -50,27 +49,28 @@ const checkboxLabelDisabled = css`
...
+
+  return (
+    <label
+      className={clsx(checkboxLabelClassname, {
+        [checkboxLabelDisabledClassname]: props.disabled
+      })}
+    >
+      <input
+        type="checkbox"
+        ref={ref as React.RefObject<HTMLInputElement>}
```
nstepien, line 68: I'm not sure how to address this type issue 🤷‍♂️

README.md
```diff
@@ -218,16 +217,16 @@ For example, the default `<Row />` component can be wrapped via the `rowRenderer
 ```tsx
 import DataGrid, { Row, RowRendererProps } from 'react-data-grid';
 
-function MyRowRenderer(props: RowRendererProps<Row>) {
+function myRowRenderer(Key: React.Key, props: RowRendererProps<Row>) {
```
amanmahajan7, line 220:
```suggestion
function myRowRenderer(key: React.Key, props: RowRendererProps<Row>) {
```

src/formatters/checkboxFormatter.tsx
amanmahajan7: Should we change it to `React.RefObject<HTMLInputElement>`?

nstepien: The formatter can be implemented using other elements, so that would only be good for some cases 🤷‍♂️

amanmahajan7: It may be safe to assume checkbox formatter is always using a checkbox (may be a hidden element) for accessibility 🤔 We can revisit

amanmahajan7 (approved):
🚀 🚢

PR #2979, opened 2022-08-01
Fix disabled checkbox styles

### LINKED ISSUE #2946: SelectCellFormatter when disabled and checked does not look disabled.
## Describe the bug 
When `SelectCellFormatter` is both checked and disabled then it only looks like it is checked without being greyed out which does not indicate the state correctly that it cannot be unchecked.

## To Reproduce 
```
<SelectCellFormatter 
      isCellSelected
      value={true}
      disabled
      onChange={()=>{}}
/>
```

Link to code example:

## Expected behavior 
The background on the checkbox should be greyed out when disabled and checked as well.

## Environment

- `react-data-grid` version: `7.0.0-beta.7`
- `react`/`react-dom` version: `^16.14.0`

stbodurov:
Background color is now only applied on checked and **enabled** checkboxes.

Fixes #2946


src/formatters/CheckboxFormatter.tsx
```diff
@@ -28,8 +28,10 @@ const checkbox = css`
   block-size: 20px;
   border: 2px solid var(--rdg-border-color);
   background-color: var(--rdg-background-color);
-  .${checkboxInput}:checked + & {
+  .${checkboxInput}:checked:not(:disabled) + & {
```
nstepien, line 31:
```suggestion
  .${checkboxInput}:checked:enabled + & {
```

nstepien (commented):
Do you have screenshots showing the before/after appearances?

src/formatters/CheckboxFormatter.tsx
stbodurov: I inverted the changes in the last commit to use `:enabled` instead.

stbodurov:
> Do you have screenshots showing the before/after appearances?

Turns out the addition of CSS layers ([54effb8424](https://github.com/adazzle/react-data-grid/blob/54effb8424e231f897550e8b07b7c76286764f40/src/formatters/checkboxFormatter.tsx)) has made the styles for regular (i.e. not checked) **disabled** checkboxes not to be applied either. The reason is that the disabled styles are coming from the `CheckboxLabel`, which is in a separate layer now, and were being superseded by the styles in the `CheckboxIcon` layer. So this change now fixes the appearance of not only disabled+checked checkboxes, but regular disabled ones too.

The checkboxes in the `Available` column below are all disabled.

**Before:**
before

**After:**
after

You can add a `disabled` prop to [website/demos/CommonFeatures.tsx#L233](https://github.com/adazzle/react-data-grid/blob/54effb8424e231f897550e8b07b7c76286764f40/website/demos/CommonFeatures.tsx#L233) if you want to test locally. If you have a method of writing visual tests, please let me know.

nstepien (approved):
Thanks for looking into it

PR #2981, opened 2022-08-06
External copy paste handlers

adam-kuhn:
This PR adds two callback props for the copy and paste events. The new props allow developers to write their own logic for the copy and paste events. Thereby, allowing the devs to respond to the events as they see fit, by having all of the logic outside of react-data-grid.

The primary limitations of the internal copy/paste logic provided by react-data-grid, is the inability to copy/paste items from external sources by writing to the clipboard. As well as the limitation of only being able to copy one cell at a time. By letting developers write there own logic to these events, these use cases can be addressed by enabling the dev to write their own logic.

Possible use cases for these two props:
- copying values to the clipboard
- pasting values from the clipboard
   - allows the user to paste data that was copied from outside of react-data-grid
- copying from multiple cells
- pasting to multiples cells

### DIFF
--- a/test/copyPaste.test.tsx
+++ b/test/copyPaste.test.tsx
@@ -45,13 +45,19 @@ const summaryRows: readonly Row[] = [
 const copyCellClassName = 'rdg-cell-copied';
 const onPasteSpy = jest.fn();
 const onCopySpy = jest.fn();
+const externalPasteSpy = jest.fn();
+const externalCopySpy = jest.fn();
 
 function CopyPasteTest({
   onPasteCallback = true,
-  onCopyCallback = false
+  onCopyCallback = false,
+  externalCopy = false,
+  externalPaste = false,
 }: {
   onPasteCallback?: boolean;
   onCopyCallback?: boolean;
+  externalCopy?: boolean;
+  externalPaste?: boolean;
 }) {
   const [rows, setRows] = useState(initialRows);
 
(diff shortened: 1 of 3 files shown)


jotaoncode (approved):

test/copyPaste.test.tsx
```diff
@@ -45,13 +45,19 @@ const summaryRows: readonly Row[] = [
...
   onPasteCallback = true,
-  onCopyCallback = false
+  onCopyCallback = false,
+  externalCopy = false,
+  externalPaste = false,
 }: {
   onPasteCallback?: boolean;
   onCopyCallback?: boolean;
+  externalCopy?: boolean;
+  externalPaste?: boolean;
```
jotaoncode, line 60: I would try to keep the "on" convention they have in the previous callers, just to keep aligned with their code style.

Code looks good and thanks for this improvement, I am looking forward to this change too.

DVGY:
This is a feature Which I need

flipvh:
Hi @DVGY do you still need this feature or did you find another way? Thanks!!

DVGY:
> Hi @DVGY do you still need this feature or did you find another way? Thanks!!

Still need this feature

amanmahajan7:
Thank you for the PR.
Copy/Paste support added in https://github.com/adazzle/react-data-grid/pull/3667

PR #2984, opened 2022-08-07
Add CalculatedColumn.cellStyle and RowRendererProps.rowStyle attributes

zakaryan2004:
These two attributes allow for CSSProperties (JSX style) to be directly
passed to the table instead of through a class name.

The motivation for this change is that only Linaria's CSS seems to work when passing it to cellStyle and rowStyle attributes. Projects using alternatives such as Emotion can't style the table quickly.

Signed-off-by: Gegham Zakaryan <zakaryan.2004@outlook.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,10 @@ function MyGrid() {
 
 ###### `rowClass?: Maybe<(row: R) => Maybe<string>>`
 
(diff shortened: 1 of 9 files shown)


amanmahajan7:
Please check https://github.com/adazzle/react-data-grid/pull/3775

PR #2986, opened 2022-08-10
Type Column.key: 'keyof TRow' instead of 'string'

aagarciga:
I would like to request a review of this idea:
By changing the type of the field key from 'string' to 'keyof TRow' we get the possibility of code completion based on the type of the rows. This little simple change makes working on the configuration of the columns way more manageable than checking on the type of the row for every field we want to include as a column on the grid.

image


amanmahajan7:
We can have columns that do not match a field on the row. For example a `Fullname` column that combines `row.firstName` and `row.lastName`.

PR #2996, opened 2022-08-23
flush scroll state updates synchronously

nstepien:
Right now with React 18, using <kbd>PageUp</kbd> or <kbd>PageDown</kbd> for example, rows are rendered after a delay which can be quite distracting tbh. Using `flushSync` fixes that.
We can merge this later when we drop support for React <18.

Before:

Recording 2022-08-23 at 14 22 15

After:

Recording 2022-08-23 at 14 23 13


src/DataGrid.tsx
```diff
@@ -624,9 +625,11 @@ function DataGrid<R, SR, K extends Key>(
 
   function handleScroll(event: React.UIEvent<HTMLDivElement>) {
     const { scrollTop, scrollLeft } = event.currentTarget;
-    setScrollTop(scrollTop);
-    // scrollLeft is nagative when direction is rtl
-    setScrollLeft(abs(scrollLeft));
+    flushSync(() => {
```
amanmahajan7, line 628: Does the prod mode also have a delay?

amanmahajan7 (approved):

src/DataGrid.tsx
nstepien: Yes, you can test on https://adazzle.github.io/react-data-grid/#/common-features

Issue #3000, opened 2022-08-28
A way to customize `priority` number component on multi-sort


godon019:
## Use case 
When you customize sort icon through `<DataGridDefaultComponentsProvider />`, you might also want to customize/access to `priority` number shown up in multi-sort

For example
- you want to give some color that match to custom icon
- you want to give space between custom sort icon and `priority` number only when the `priority` number is shown

You can kind of see the problems [here](https://codesandbox.io/s/eager-butterfly-qp15x4?file=/src/App.tsx:1438-1471)

## Proposed solution 
maybe provide a prop `sortPriority` just like the way `sortIcon` does ?

amanmahajan7:
I think this makes sense. We can add `sortPriority` renderer

PR #3004, opened 2022-08-29
Add sortStatus renderer

amanmahajan7:
Fixes https://github.com/adazzle/react-data-grid/issues/3000

This replaces `sortIcon` renderer. Now both icon and priority can be customized using the new `sortStatus` renderer

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -213,6 +213,10 @@ export interface SortIconProps {
   sortDirection: SortDirection | undefined;
 }
 
+export interface SortPriorityProps {
+  priority: number | undefined;
+}
+
 export interface CheckboxFormatterProps
   extends Pick<
     React.InputHTMLAttributes<HTMLInputElement>,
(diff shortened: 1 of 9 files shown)


nstepien (dismissed):
I'm fine with the changes, but I wonder if it'd be better if we could render both at once instead, or if we could customize header cells as a whole.

amanmahajan7:
Wdyt about adding `priority` prop to `sortIcon`? We will have to re-export `sortIcon` I think

nstepien:
I don't mind re-exporting `sortIcon`.
I think rather than adding `priority` to `sortIcon`, we can add `sortStatus` that renders both `sortIcon` and `sortPriority`

src/types.ts
```diff
@@ -213,6 +213,12 @@ export interface SortIconProps {
   sortDirection: SortDirection | undefined;
 }
 
+export interface SortPriorityProps {
+  priority: number | undefined;
+}
+
+export type SortStatusProps = SortIconProps & SortPriorityProps;
```
nstepien, line 220: I believe this is better for 1. performance, 2. catching conflicting types
```suggestion
export interface SortStatusProps extends SortIconProps, SortPriorityProps {};
```

nstepien (approved):

nstepien (approved):

nstepien (approved):

PR #3006, opened 2022-08-30
Remove user-select

amanmahajan7:


src/style/core.ts
```diff
@@ -51,7 +51,6 @@ const root = css`
   border: 1px solid var(--rdg-border-color);
   box-sizing: border-box;
   overflow: auto;
-  user-select: none;
```
amanmahajan7, line 54: Double clicking on editable cell selects the text first and then the cell goes into edit mode and the editor value is selected again. Not the bet UI but I haven't found a good solution

nstepien:
Should we maybe disable text selection on editable cells?

amanmahajan7:
I thought about it but the formatter may display a different value for example if we have complex editor like a modal.

amanmahajan7:
I think we can merge this and revisit if there are any issues

nstepien (approved):

PR #3010, opened 2022-09-01
Fix a few styles

amanmahajan7:


src/style/core.ts
```diff
@@ -94,7 +94,7 @@ const root = css`
 export const rootClassname = `rdg ${root}`;
 
 const viewportDragging = css`
-  &.${row} {
+  & .${row} {
```
amanmahajan7, line 97: This style was never applied

nstepien (approved):

PR #3014, opened 2022-09-12
Use latest onRowClick/onRowDoubleClick

amanmahajan7:


src/hooks/useLatestFunc.ts
```diff
@@ -1,15 +1,19 @@
 import { useRef, useEffect, useCallback } from 'react';
+import type { Maybe } from '../types';
 
 // https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function useLatestFunc<T extends (...args: any[]) => any>(fn: T) {
+export function useLatestFunc<T extends Maybe<(...args: any[]) => any>>(fn: T): T {
```
amanmahajan7, line 6: Without `:T`, the type is 

image

src/hooks/useLatestFunc.ts
```diff
@@ -1,15 +1,19 @@
...
   const ref = useRef(fn);
 
   useEffect(() => {
     ref.current = fn;
   });
 
-  return useCallback((...args: Parameters<T>) => {
-    ref.current(...args);
+  const callbackFn = useCallback((...args: Parameters<NonNullable<T>>) => {
+    ref.current!(...args);
```
nstepien, line 14: Should we do that?
```suggestion
    return ref.current!(...args);
```

nstepien (approved):

src/hooks/useLatestFunc.ts
amanmahajan7: Looks like none of the event handlers return anything. We can revisit if needed

PR #3017, opened 2022-09-15
Add CSS layers to lib styles

nstepien:
https://css-tricks.com/css-cascade-layers/
https://caniuse.com/css-cascade-layers
https://github.com/postcss/postcss-nested#options

Need to review this with whitespace changes hidden.
I managed to reduce selector specificity in a few cases.


rollup.config.js
```diff
@@ -36,9 +36,8 @@ export default {
       }
     }),
     postcss({
-      plugins: [postcssNested],
-      minimize: true,
-      inject: { insertAt: 'top' }
```
nstepien, line 41: We can finally remove this odd setting, no more surprises

src/style/row.ts
```diff
@@ -1,39 +1,45 @@
...
-    background-color: var(--rdg-row-selected-background-color);
+    &[aria-selected='true'] {
+      background-color: var(--rdg-row-selected-background-color);
 
-    &:hover {
-      background-color: var(--rdg-row-selected-hover-background-color);
+      &:hover {
+        background-color: var(--rdg-row-selected-hover-background-color);
+      }
     }
```
nstepien, line 19: If it's ever needed, we could split this into `rdg.Row` and `rdg.RowSelected` layers for example, so that we can order them like this: `rdg.Row` -> `rdg.HeaderRow`/`rdg.SummaryRow`/ -> `rdg.RowSelected`

Summary rows doesn't try to change the `background-color` so it's fine, but I wonder about header rows 🤔

src/SummaryRow.tsx
```diff
@@ -2,7 +2,8 @@ import { memo } from 'react';
 import clsx from 'clsx';
 import { css } from '@linaria/core';
 
-import { cell, cellFrozen, row, rowClassname, rowSelectedClassname } from './style';
+import { cell, cellFrozen } from './style/cell';
```
nstepien, line 5: Had to import directly from `./style/cell` to fix linaria

src/style/core.ts
```diff
@@ -65,9 +65,8 @@ const root = css`
       color-scheme: var(--rdg-color-scheme, light dark);
 
       /* https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context */
-      /* We set a stacking context so internal elements don't render on top of external components. */
+      /* We set a stacking context so internal elements don't render on top of external elements. */
       contain: strict;
-      contain: size layout style paint;
```
nstepien: I've removed this line, as nowadays both Firefox and Safari support `style` containment, and `strict` is the same as listing all 4 of these.
https://caniuse.com/mdn-css_properties_contain_style
https://developer.mozilla.org/en-US/docs/Web/CSS/contain

amanmahajan7 (approved):

PR #3020, opened 2022-09-15
Fix column resizing in chrome

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -474,6 +481,9 @@ function DataGrid<R, SR, K extends Key>(
         newColumnWidths.set(column.key, width);
         return newColumnWidths;
       });
+      flushSync(() => {
+        setColumnResizing(true);
+      });
 
       onColumnResize?.(column.idx, width);
```
amanmahajan7, line 488: We can pass a `done` flag to `onColumnResize` now

src/HeaderCell.tsx
```diff
@@ -84,6 +84,8 @@ export default function HeaderCell<R, SR>({
     }
 
     function onPointerMove(event: PointerEvent) {
+      // prevents chrome from scrolling the grid and fixes re-size on an autosized column
```
nstepien, line 87:
```suggestion
      // prevents text selection in Chrome, which fixes scrolling the grid while dragging, and fixes re-size on an autosized column
```

nstepien (approved):

PR #3024, opened 2022-09-16
Add invisible cells to measure column widths regardless of column spanning

nstepien:


src/utils/renderMeasuringCells.tsx
```diff
@@ -0,0 +1,29 @@
+import { css } from '@linaria/core';
+import type { CalculatedColumn } from '../types';
+
+const measuringCellClassname = css`
+  @layer rdg.MeasuringCell {
+    contain: strict;
+    grid-row: 1;
+    visibility: hidden;
```
amanmahajan7, line 8: https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns
Nice!

src/utils/renderMeasuringCells.tsx
```diff
@@ -0,0 +1,29 @@
+import { css } from '@linaria/core';
+import type { CalculatedColumn } from '../types';
+
+const measuringCellClassname = css`
+  @layer rdg.MeasuringCell {
```
amanmahajan7, line 5: Does it need a layer? Does not hurt but wondering as we will never override the styles or it will conflict with other styles

amanmahajan7 (approved):
Nice. We can probably use the new cell to fix column resizing

src/utils/renderMeasuringCells.tsx
nstepien: It probably doesn't. Doesn't hurt to add a layer anyway 🤷‍♂️

PR #3025, opened 2022-09-16
Refactor column resizing

nstepien:
With the changes in this PR,
- instead of updating `grid-template-columns` by re-rendering in React, reading the width in a layout effect, then updating the state again,
- we set the size in the `grid-template-columns` manually, immediately read the width from the measuring cell, only then do we update the React state

-1 state
-1 layout effect
fewer React re-renders

Only downsides is that we can't have good tests like this, maybe we need to mock `getBoundingClientRect()`?

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -179,6 +179,7 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
    */
   renderers?: Maybe<Renderers<R, SR>>;
   rowClass?: Maybe<(row: R) => Maybe<string>>;
+  /** @default 'ltr' */
   direction?: Maybe<Direction>;
   'data-testid'?: Maybe<string>;
 }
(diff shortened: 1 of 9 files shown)


src/DataGrid.tsx
```diff
@@ -496,17 +480,34 @@ function DataGrid<R, SR, K extends Key>(
    * event handlers
    */
   function handleColumnResize(column: CalculatedColumn<R, SR>, width: number | 'max-content') {
-    if (width === 'max-content') {
-      setAutoResizeColumn(column);
-      return;
+    const { style } = gridRef.current!;
```
amanmahajan7, line 483: Wondering if we can run into any edge cases with this approach. Does not seem like but may be safer to use react api if there is no noticeable performance improvement with the new approach 
https://beta.reactjs.org/learn/manipulating-the-dom-with-refs#best-practices-for-dom-manipulation-with-refs-best-practices-for-dom-manipulation-with-refs

PR #3029, opened 2022-09-17
Resize flex columns on (column) resize

nstepien:
Recording 2022-09-17 at 17 40 58
Recording 2022-09-17 at 17 42 15

There is a bug that after resizing, then scrolling, flex columns can snap back to their newly measured width (content-dependent).
Maybe we need re-save measured widths after resizing.
This won't re-flex columns when toggling columns for example, maybe this can be controlled in a parent state.


src/DataGrid.tsx
```diff
@@ -487,21 +479,29 @@ function DataGrid<R, SR, K extends Key>(
   /**
    * callbacks
    */
-  const handleColumnResize = useCallback(
+  const handleColumnResize = useLatestFunc(
```
amanmahajan7, line 482: May be we can move the function down and add `handleColumnResizeLatest` for consistency 
https://github.com/adazzle/react-data-grid/blob/628d4e15168fd3e64dc7711a449b056dc5f1bbe4/src/DataGrid.tsx#L398

src/DataGrid.tsx
```diff
@@ -487,21 +479,29 @@ function DataGrid<R, SR, K extends Key>(
...
     (column: CalculatedColumn<R, SR>, width: number | 'max-content') => {
       if (width === 'max-content') {
-        setAutoResizeColumn(column);
+        flushSync(() => {
+          // TODO: only clear flex columns, if any?
+          setMeasuredColumnWidths(new Map());
```
amanmahajan7, line 487: Should we only clear the width of flex columns that are in the viewport (`flexWidthViewportColumns`)? Would that fix the bug?

src/DataGrid.tsx
```diff
@@ -487,21 +479,29 @@ function DataGrid<R, SR, K extends Key>(
...
+          setAutoResizeColumn(column);
+        });
         return;
       }
-      setColumnWidths((columnWidths) => {
-        const newColumnWidths = new Map(columnWidths);
-        newColumnWidths.set(column.key, width);
-        return newColumnWidths;
+
+      flushSync(() => {
```
amanmahajan7, line 493: Any difference if `flushSync` is not used?

src/hooks/useGridDimensions.ts
```diff
@@ -40,5 +38,36 @@ export function useGridDimensions(): [
     };
   }, []);
 
-  return [gridRef, inlineSize, blockSize, isWidthInitialized];
+  const observeMeasuringCell = useMemo(() => {
```
amanmahajan7, line 41: 🔥

src/hooks/useGridDimensions.ts
```diff
@@ -40,5 +38,36 @@ export function useGridDimensions(): [
...
+          const key = (entry.target as HTMLDivElement).dataset.measuringCellKey!;
+          newMeasuredColumnWidths.set(key, inlineSize);
+          hasChanges ||= measuredColumnWidths.get(key) !== inlineSize;
+        }
+
+        return hasChanges ? newMeasuredColumnWidths : measuredColumnWidths;
+      });
+    });
+
+    return (measuringCell: HTMLDivElement | null) => {
```
amanmahajan7, line 58: Does the `resizeObserver` need to be disconnected?

src/style/cell.ts
```diff
@@ -9,7 +9,11 @@ export const cell = css`
     contain: strict;
     */
     position: relative; /* needed for absolute positioning to work */
-    contain: size style;
+
+    /* max-content does not work when size containment is enabled *
+    /* "contain: content" leads to odd subpixel mismatches when combined with colSpan */
+    /* dynamically switching between different containment styles incurs a heavy relayout penalty */
```
amanmahajan7, line 15: Nice to know 👍

src/DataGrid.tsx
nstepien: It can make the bottom scroll bar flicker
Recording 2022-09-21 at 13 28 26

src/hooks/useGridDimensions.ts
nstepien: It should be garbage-collected correctly.

https://twitter.com/jaffathecake/status/1405437361643790337
https://bugs.chromium.org/p/chromium/issues/detail?id=1220041

src/DataGrid.tsx
nstepien: I'm not sure about using viewport columns, as resizing may display columns that were previously hidden on the next render.

src/DataGrid.tsx
```diff
@@ -659,6 +636,70 @@ function DataGrid<R, SR, K extends Key>(
...
+      }
+
+      return hasChanges ? newMeasuredColumnWidths : measuredColumnWidths;
+    });
+  }
+
+  function handleColumnResizeEnd() {
+    setMeasuredColumnWidths((measuredColumnWidths) => {
+      const newMeasuredColumnWidths = new Map(measuredColumnWidths);
+      const measuringCells = gridRef.current!.querySelectorAll('[data-measuring-cell-key]');
```
amanmahajan7, line 691: Should we not ignore columns that have fixed widths or `max-content`?

src/DataGrid.tsx
```diff
@@ -284,7 +289,8 @@ function DataGrid<R, SR, K extends Key>(
   /**
    * computed values
    */
-  const [gridRef, gridWidth, gridHeight, isWidthInitialized] = useGridDimensions();
+  const [gridRef, gridWidth, gridHeight] = useGridDimensions();
+  const [prevGridWidth, setPrevGridWidth] = useState(gridWidth);
```
nstepien, line 293: Should this be moved with the other states?

src/DataGrid.tsx
```diff
@@ -438,22 +444,23 @@ function DataGrid<R, SR, K extends Key>(
...
+        const measuringCell = gridRef.current!.querySelector(`[data-measuring-cell-key="${key}"]`)!;
+        const measuredWidth = measuringCell.getBoundingClientRect().width;
+        newMeasuredColumnWidths.set(key, measuredWidth);
       }
+    }
 
-      return newColumnWidths;
-    });
-  }, [isWidthInitialized, flexWidthViewportColumns, gridRef]);
+    if (newMeasuredColumnWidths.size !== measuredColumnWidths.size) {
```
nstepien, line 460: This doesn't seem like a safe check, should we explicitly check if we found an updated/new width in the loop?

src/DataGrid.tsx
```diff
@@ -483,6 +490,16 @@ function DataGrid<R, SR, K extends Key>(
   function handleColumnResize(column: CalculatedColumn<R, SR>, width: number | 'max-content') {
```
nstepien, line 490: Can we send `${width}px` to `onColumnResize` in `HeaderCell`?
```suggestion
  function handleColumnResize(column: CalculatedColumn<R, SR>, width: string) {
```

src/DataGrid.tsx
```diff
@@ -502,11 +519,17 @@ function DataGrid<R, SR, K extends Key>(
...
-    newColumnWidths.set(column.key, measuredWidth);
-    setColumnWidths(newColumnWidths);
+    const newResizedColumnWidths = new Map(resizedColumnWidths);
+    newResizedColumnWidths.set(column.key, measuredWidth);
+    setResizedColumnWidths(newResizedColumnWidths);
+
+    // We need to reset measured width otherwise grid used the already measured
+    // width in the next render and the new measured width is never set
+    if (resetMeasuredColumnWidths) {
+      setMeasuredColumnWidths(new Map());
```
nstepien, line 531: Rather than setting an empty map, can we immediately measure the new widths here?
Might save us an extra re-render

src/DataGrid.tsx
```diff
@@ -1138,6 +1166,13 @@ function DataGrid<R, SR, K extends Key>(
     return rowElements;
   }
 
+  if (gridWidth !== prevGridWidth) {
+    setPrevGridWidth(gridWidth);
+    if (columns.length === viewportColumns.length) {
+      setMeasuredColumnWidths(new Map());
```
nstepien, line 1172: 1. grid width changes (1st render)
2. we update the previous grid width state + reset measured col widths (2nd render)
3. we measure col widths and update the state (3rd render)

Can we perhaps combine 2. and 3.?

src/DataGrid.tsx
```diff
@@ -931,10 +954,15 @@ function DataGrid<R, SR, K extends Key>(
   }
 
   function getLayoutCssVars() {
```
nstepien, line 956: Rather than returning a new object, we could do
```suggestion
  function getGridTemplateColumns() {
```

```jsx
style={{
  ...,
  gridTemplateColumns: getGridTemplateColumns()
}}
```

src/DataGrid.tsx
amanmahajan7: Need to set the initial state using `gridWidth`

truk:
Anything I could do to help get this in?  It's exactly the feature I was needing.

amanmahajan7:
@truk, we haven't finalized if the current approach is the best solution. You can test it in the meantime if everything is working as expected

truk:
Yeah, I guess the first thing is just to understand the scope of what this is.  It looks like it's the ability to have the columns just auto-grow beyond the content width when there's space, though this doesn't cause columns to shrink and clip the text below the content width based on any relative sizes.  I think the latter would be a useful feature so you could have some columns that can flex up and down and some fixed.  And for this to be a column property rather than some default. 
 I'm also noticing the scroll bug still in this branch, so maybe it's not done yet.

JanBN:
Can this be merged ? I also need this feature

src/DataGrid.tsx
nstepien: The initial `gridWidth` is `1` anyway 🤔

amanmahajan7 (commented):
LGTM. @nstepien okay with merging it?

nstepien:
@amanmahajan7 I'll review it tomorrow

amanmahajan7 (approved):

PR #3035, opened 2022-09-21
Extract mergeable changes from #3029

amanmahajan7:
https://github.com/adazzle/react-data-grid/pull/3029

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -265,9 +265,7 @@ function DataGrid<R, SR, K extends Key>(
    */
   const [scrollTop, setScrollTop] = useState(0);
   const [scrollLeft, setScrollLeft] = useState(0);
-  const [resizedColumnWidths, setResizedColumnWidths] = useState(
-    (): ReadonlyMap<string, number> => new Map()
-  );
+  const [columnWidths, setColumnWidths] = useState((): ReadonlyMap<string, number> => new Map());
   const [selectedPosition, setSelectedPosition] = useState<SelectCellState | EditCellState<R>>(
     initialPosition
   );
@@ -330,7 +328,7 @@ function DataGrid<R, SR, K extends Key>(
     groupBy
   } = useCalculatedColumns({
     rawColumns,
-    resizedColumnWidths,
+    columnWidths,
     scrollLeft,
     viewportWidth: gridWidth,
     defaultColumnOptions,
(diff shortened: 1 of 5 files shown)


src/style/cell.ts
```diff
@@ -2,14 +2,14 @@ import { css } from '@linaria/core';
...
-    */
+    /* max-content does not work with size containment
+     * dynamically switching between different containment styles incurs a heavy relayout penalty
+     * Chromium bug: at odd zoom levels or subpixel positioning, layout/paint containment can make cell borders disappear
+     *   https://bugs.chromium.org/p/chromium/issues/detail?id=1326946
+     */
+    contain: style;
     position: relative; /* needed for absolute positioning to work */
-    contain: size style;
+    z-index: 0;
```
nstepien, line 12: I think I accidentally committed this in, I don't think we need it unless I'm missing something
```suggestion
```

src/DataGrid.tsx
```diff
@@ -266,7 +265,9 @@ function DataGrid<R, SR, K extends Key>(
    */
   const [scrollTop, setScrollTop] = useState(0);
   const [scrollLeft, setScrollLeft] = useState(0);
-  const [columnWidths, setColumnWidths] = useState<ReadonlyMap<string, number>>(() => new Map());
+  const [resizedColumnWidths, setResizedColumnWidths] = useState<ReadonlyMap<string, number>>(
+    () => new Map()
```
nstepien, line 269:
```suggestion
  const [resizedColumnWidths, setResizedColumnWidths] = useState(
    (): ReadonlyMap<string, number> => new Map()
```

src/DataGrid.tsx
```diff
@@ -440,18 +441,18 @@ function DataGrid<R, SR, K extends Key>(
...
+    setResizedColumnWidths((resizedColumnWidths) => {
+      const newResizedColumnWidths = new Map(resizedColumnWidths);
       const grid = gridRef.current!;
 
       for (const column of flexWidthViewportColumns) {
         const measuringCell = grid.querySelector(`[data-measuring-cell-key="${column.key}"]`)!;
         // Set the actual width of the column after it is rendered
         const { width } = measuringCell.getBoundingClientRect();
-        newColumnWidths.set(column.key, width);
+        newResizedColumnWidths.set(column.key, width);
```
nstepien, line 452: These cells aren't "resized", not sure we should rename this map

nstepien (approved):

PR #3059, opened 2022-10-24
Filtering empty arrays is used to resolve errors reported for column decomposition

Dingzhaocheng:
```
 Filtering empty arrays is used to resolve errors reported for column decomposition
```
截屏2022-10-24 18 36 21

### DIFF
--- a/src/hooks/useCalculatedColumns.ts
+++ b/src/hooks/useCalculatedColumns.ts
@@ -81,7 +81,7 @@ export function useCalculatedColumns<R, SR>({
       }
 
       return column;
-    });
+    }).filter(Boolean);
 
     columns.sort(({ key: aKey, frozen: frozenA }, { key: bKey, frozen: frozenB }) => {
       // Sort select column first:


src/hooks/useCalculatedColumns.ts
```diff
@@ -81,7 +81,7 @@ export function useCalculatedColumns<R, SR>({
       }
 
       return column;
-    });
+    }).filter(Boolean);
```
nstepien, line 84: This array never has falsy values, so this does nothing.

Dingzhaocheng: Sorry, it's my problem, I can handle the source data;
Actually, it's like this, for example Array.fill(10).map(item=>({. .item})), it will still return [empty * 10];
It's possible to filter out the damn troubleshooting time here, especially reading data from xlsx. 
Thank you for your reply!

nstepien: This should be the responsibility of the parent components, I don't think RDG should have to check for holey arrays.

Dingzhaocheng: 👌，Thank you for your reply

PR #3079, opened 2022-11-16
Tweak onEditorNavigation

amanmahajan7:


src/utils/keyboardUtils.ts
```diff
@@ -70,7 +70,7 @@ export function onEditorNavigation({ key, target }: React.KeyboardEvent<HTMLDivE
       target instanceof HTMLSelectElement)
   ) {
     return target.matches(
-      '.rdg-editor-container > :only-child, .rdg-editor-container > label:only-child > :only-child'
+      '.rdg-editor-container > :only-child, .rdg-editor-container > label:only-child > :only-child, .rdg-editor-container div:only-child > label:only-child > :only-child'
```
nstepien, line 73: The selector is too broad otherwise
```suggestion
      '.rdg-editor-container > :only-child, .rdg-editor-container > label:only-child > :only-child, .rdg-editor-container > div:only-child > label:only-child > :only-child'
```

amanmahajan7: Ah missed it. Thanks!

nstepien (approved):

PR #3113, opened 2022-12-23
Allow for arbitrary data attributes on upper element

hogansghost:
- Spread typesafe div props onto the upper element to allow for any data-attribute to be added to the top level.

Technically, this could also handle all of the aria tags as well by spreading the props but I was more concerned with adding arbitrary data-attributes over refactoring anything too heavily. This is to primarily allow for other tags such as `data-cy` for Cypress recommended naming selectors and to not require a specific prop for `data-testid`. The TS types for the `Div` element also prevent incorrect attributes being added:

Screenshot 2022-12-23 at 09 15 33


hogansghost:
I just noticed my prettier did a number on this one, I can change that back if you prefer 👍

hogansghost:
Any opinion on this one at all? @nstepien @amanmahajan7 

It would be super useful in my current project to target the multiple tables I have per page (rather than relying on `testid` for these elements only).

PR #3114, opened 2022-12-25
Add indeterminate state of header checkbox

### LINKED ISSUE #3058: Support indeterminate header level checkbox
## Use case
I would like to display an indeterminate checkbox in the header if some, but not all rows are selected.

## Proposed solution 

I am using a custom Checkbox header, so for my use-case I would only need an additional property `someRowsSelected: boolean;` added to `HeaderRendererProps` so that I could determine if some, but not all, rows are selected.

paustint:
Added someRowsSelected property to DataGrid to determine if indeterminate

Provide someRowsSelected as a new property to HeaderRendererProps

Automatically set indeterminate property on built-in or overridden checkboxFormatter as long as the ref exists and the ref component has a property named `indeterminate` (e.x. input component)

Fixed typo in comment

resolves #3058

**Note**: the built-in custom select checkbox does have the indeterminate property set, but there are no active styles for this state.

Here are screenshots of the outcome of this feature:

image

image

image

image

### DIFF
--- a/src/Columns.tsx
+++ b/src/Columns.tsx
@@ -50,6 +50,7 @@ export const SelectColumn: Column<any, any> = {
         aria-label="Select All"
         isCellSelected={props.isCellSelected}
         value={props.allRowsSelected}
+        isIndeterminate={props.someRowsSelected}
         onChange={props.onAllRowsSelectionChange}
       />
     );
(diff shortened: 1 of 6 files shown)


amanmahajan7:
It is supported now https://github.com/adazzle/react-data-grid/pull/3577

paustint:
Awesome - thank you!

PR #3119, opened 2023-01-05
Add onCellCllick and onCellDoubleClick events

amanmahajan7:


src/Cell.tsx
```diff
@@ -51,22 +52,26 @@ function Cell<R, SR>({
     typeof cellClass === 'function' ? cellClass(row) : cellClass
   );
 
-  function selectCellWrapper(openEditor?: boolean | null) {
```
amanmahajan7, line 54: Any reason for using `null` here?

package.json
```diff
@@ -97,7 +97,6 @@
     "postcss-nested": "^6.0.0",
     "prettier": "2.8.1",
     "react": "^18.1.0",
-    "react-contextmenu": "^2.14.0",
```
amanmahajan7, line 100: Added a custom context menu so this library is no longer needed

src/Cell.tsx
nstepien: No

src/DataGrid.tsx
```diff
@@ -151,10 +152,12 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
   /**
    * Event props
    */
-  /** Function called whenever a row is clicked */
-  onRowClick?: Maybe<(row: R, column: CalculatedColumn<R, SR>) => void>;
-  /** Function called whenever a row is double clicked */
-  onRowDoubleClick?: Maybe<(row: R, column: CalculatedColumn<R, SR>) => void>;
+  /** Function called whenever a cell is clicked */
+  onCellClick?: RowRendererProps<R, SR>['onCellClick'];
```
nstepien, line 156: Can we do the opposite and make `RowRendererProps` get these props from `DataGridProps` instead?

src/types.ts
```diff
@@ -112,21 +110,30 @@ export interface HeaderRendererProps<TRow, TSummaryRow = unknown> {
 }
 
 export interface CellRendererProps<TRow, TSummaryRow>
-  extends Pick<
-      RowRendererProps<TRow, TSummaryRow>,
-      'onRowClick' | 'onRowDoubleClick' | 'selectCell' | 'skipCellFocusRef'
-    >,
-    Omit<React.HTMLAttributes<HTMLDivElement>, 'style' | 'children'> {
+  extends Pick<RowRendererProps<TRow, TSummaryRow>, 'selectCell' | 'skipCellFocusRef'>,
```
nstepien, line 113: Can we add `onClick` and others here?

README.md
```diff
@@ -188,9 +188,9 @@ A number defining the height of summary rows.
 
 ###### `onPaste?: Maybe<(event: PasteEvent<R>) => R>`
 
-###### `onRowClick?: Maybe<(row: R, column: CalculatedColumn<R, SR>) => void>`
```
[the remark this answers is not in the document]
nstepien: I don't think so, you can't click a row without clicking a cell.

website/demos/ContextMenu.tsx
```diff
@@ -129,30 +52,32 @@ function rowKeyGetter(row: Row) {
...
-    insertRow(rowIdx);
-  }
-
-  function onRowInsertBelow(e: React.MouseEvent<HTMLDivElement>, { rowIdx }: { rowIdx: number }) {
-    insertRow(rowIdx + 1);
-  }
+    return () => {
+      removeEventListener('click', onClick);
+    };
+  }, [contextMenuProps]);
```
nstepien, line 80: Rather than depending on the object itself which could be updated easily, we could have `const isContextMenuOpen = contextMenuProps !== null` outside the effect and use it

website/demos/ContextMenu.tsx
```diff
@@ -171,27 +96,65 @@ export default function ContextMenuDemo({ direction }: Props) {
...
-      )}
+      {contextMenuProps !== null &&
+        createPortal(
+          <menu
+            ref={menuRef}
+            className={contextMenuClassname}
+            style={
+              {
+                '--top': `${contextMenuProps.top}px`,
+                '--left': `${contextMenuProps.left}px`
```
nstepien, line 118: These don't need to be variables

src/types.ts
amanmahajan7: The props have different names so not sure if we can do that

website/demos/ContextMenu.tsx
```diff
@@ -171,27 +95,65 @@ export default function ContextMenuDemo({ direction }: Props) {
...
-              {/* @ts-expect-error */}
-              <MenuItem onClick={onRowInsertAbove}>Above</MenuItem>
-              {/* @ts-expect-error */}
-              <MenuItem onClick={onRowInsertBelow}>Below</MenuItem>
-            </SubMenu>
-          </ContextMenu>
-        </div>,
-        document.body
-      )}
+      {contextMenuProps !== null &&
```
nstepien, line 109:
```suggestion
      {isContextMenuOpen &&
```

website/demos/ContextMenu.tsx
```diff
@@ -171,27 +95,65 @@ export default function ContextMenuDemo({ direction }: Props) {
...
+        createPortal(
+          <menu
+            ref={menuRef}
+            className={contextMenuClassname}
+            style={
+              {
+                top: `${contextMenuProps.top}px`,
+                left: `${contextMenuProps.left}px`
```
nstepien, line 117: This should work as well
```suggestion
                top: contextMenuProps.top,
                left: contextMenuProps.left
```

nstepien (approved):

PR #3139, opened 2023-01-26
Do not scroll back to cell when the editor is closed

### LINKED ISSUE #3137: Grid scrolls back to edited cell when clicking on another cell
## Describe the bug 

When editing a cell, if you scroll away from the editor and then click on another cell, the grid will scroll back to the edited cell. This can be frustrating and is unexpected behavior compared to e.g. Excel.

## To Reproduce 

1. Open https://adazzle.github.io/react-data-grid/#/common-features
2. Double click to edit a cell, e.g. under "Contact"
3. Scroll down in the grid and click on a different cell
4. Notice that the grid scrolls back up and the new cell is not selected.

## Expected behavior 

The editor should be closed but not steal focus. The clicked cell should gain focus instead and the grid should not scroll.

## Environment

- `react-data-grid` version: 18.0.0
- `react`/`react-dom` version: 7.0.0-beta.22

## Additional context 

I see a similar issue (#2943) was closed, but that refers specifically to clicking on contents outside of the grid. So it is unclear if this is expected behavior. If it is expected, I think it should at least be an option as it diverges from expected behavior in other grids.

kycutler:
Fixes #3137

### DIFF
--- a/src/hooks/useRovingCellRef.ts
+++ b/src/hooks/useRovingCellRef.ts
@@ -16,12 +16,12 @@ export function useRovingCellRef(
   const ref = useCallback(
     (cell: HTMLDivElement | null) => {
       if (cell === null) return;
-      scrollIntoView(cell);
-      if (cell.contains(document.activeElement)) return;
       if (skipCellFocusRef?.current) {
         skipCellFocusRef.current = false;
         return;
       }
+      scrollIntoView(cell);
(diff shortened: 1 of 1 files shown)


amanmahajan7 (approved):
Looks good. Thank you for the fix. 

Event order for future reference
image

kycutler:
Thank you for the quick response!

PR #3149, opened 2023-02-08
Flush `onRowsChange` changes to prevent `onRowsChange` getting called twice in some cases

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -996,8 +996,17 @@ function DataGrid<R, SR, K extends Key>(
...
-        updateRow(column, selectedPosition.rowIdx, row);
-        closeEditor();
+        // Fixes two issues when editor is closed by clicking on a different cell
+        //
+        // Sometimes commitEditorChanges is called before the cell state is changed to
+        // SELECT and this results in onRowChange getting called twice.
+        //
+        // Sometimes rows prop is changed before selectCell is called and this results in
+        // rows.indexOf(row) returning -1 so incorrect cell gets selected
+        flushSync(() => {
```
amanmahajan7, line 1006: Tried adding a test but could not create a failure scenario

src/DataGrid.tsx
```diff
@@ -996,8 +996,17 @@ function DataGrid<R, SR, K extends Key>(
 
     const onRowChange = (row: R, commitChanges?: boolean) => {
       if (commitChanges) {
-        updateRow(column, selectedPosition.rowIdx, row);
-        closeEditor();
+        // Fixes two issues when editor is closed by clicking on a different cell
```
nstepien, line 999: More forward-looking
```suggestion
        // Prevents two issues when editor is closed by clicking on a different cell
```

src/DataGrid.tsx
```diff
@@ -996,8 +996,17 @@ function DataGrid<R, SR, K extends Key>(
...
+        //
+        // Sometimes commitEditorChanges is called before the cell state is changed to
```
nstepien, line 1001:
```suggestion
        // Otherwise commitEditorChanges may be called before the cell state is changed to
```

src/DataGrid.tsx
```diff
@@ -996,8 +996,17 @@ function DataGrid<R, SR, K extends Key>(
...
+        // SELECT and this results in onRowChange getting called twice.
+        //
+        // Sometimes rows prop is changed before selectCell is called and this results in
+        // rows.indexOf(row) returning -1 so incorrect cell gets selected
```
nstepien, line 1005:
```suggestion
        // rows.indexOf(row) returning -1 so no cell can be selected
```

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: It does select the header or summary row if present

PR #3151, opened 2023-02-08
Add `navigate` function to `EditCellKeyDownArgs`

amanmahajan7:


src/EditCell.tsx
```diff
@@ -93,9 +93,22 @@ export default function EditCell<R, SR>({
...
     if (onKeyDown) {
       const cellEvent = createCellEvent(event);
-      onKeyDown({ mode: 'EDIT', row, column, rowIdx, onClose }, cellEvent);
+      onKeyDown(
+        {
+          mode: 'EDIT',
+          row,
+          column,
+          rowIdx,
+          navigate() {
```
amanmahajan7, line 102: Currently there is no way to continue navigation if default event is prevented. We need to pass this function to allow users to control navigation

src/types.ts
```diff
@@ -133,25 +133,26 @@ export type CellMouseEvent = CellEvent<React.MouseEvent<HTMLDivElement>>;
 
 export type CellKeyboardEvent = CellEvent<React.KeyboardEvent<HTMLDivElement>>;
 
-export interface CellClickArgs<TRow, TSummaryRow> {
+export interface CellClickArgs<TRow, TSummaryRow = unknown> {
```
amanmahajan7, line 136: So we do not have to specify `unknown` when using this type

nstepien (approved):

PR #3153, opened 2023-02-09
Do not check `isDefaultPrevented` in `isGridDefaultPrevented`

amanmahajan7:


src/utils/eventUtils.ts
```diff
@@ -10,7 +10,7 @@ export function createCellEvent<E extends React.SyntheticEvent<HTMLDivElement>>(
       defaultPrevented = true;
     },
     isGridDefaultPrevented() {
-      return defaultPrevented || this.isDefaultPrevented();
```
amanmahajan7, line 13: This can cause issues if the editor is calling `preventDefault`

nstepien (approved):
Is there a test worth adding that's justifying this change?

amanmahajan7:
> Is there a test worth adding that's justifying this change?

Not sure but I will double check and create a follow up PR

PR #3181, opened 2023-03-24
fix: focus lost when inside shadow dom

noru:
When RDG is inside a shadowroot, https://github.com/adazzle/react-data-grid/blob/f511e7bb3ba0d806b3688537834ce5caf551441f/src/hooks/useRovingCellRef.ts#L24 would fail, cuz `document.activeElement` should be `shadowRoot.activeElement`.

Fix it by calling `getRootNode()` to get the correct root element.


amanmahajan7:
RDG does not have `document.activeElement` anymore. Can you check with the latest beta version and create an issue if it does not work

noru:
Hi @amanmahajan7, my issue still exists: I have to click twice to be able to edit a cell

Preferably with one click.

But I think it is due to the removal of `isCellSelected` https://github.com/adazzle/react-data-grid/pull/3237

PR #3184, opened 2023-03-28
Randomize tests

amanmahajan7:
https://jestjs.io/docs/cli#--randomize


.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           npm run build
           npm run build:types
       - name: Test
-        run: npm t -- --coverage --colors
+        run: npm t -- --coverage --colors --randomize --showSeed
```
nstepien, line 40: Could set these in the config file, and not only in CI, wdyt?
https://jestjs.io/docs/configuration#randomize-boolean
https://jestjs.io/docs/configuration#showseed-boolean

nstepien (approved):

PR #3185, opened 2023-03-29
Rome 12.0.0

nstepien:
https://rome.tools/blog/2023/03/28/rome12/
https://docs.rome.tools/lint/rules/

I've switched all lint rules from `warn` to `error`, otherwise Rome wouldn't fail the CI on warnings, see: https://github.com/rome/tools/discussions/4328

Rome supports JSON formatting/parsing now, but it does not handle JSONC like our eslint config, so I had to ignore all .json files for now.


website/root.css
```diff
@@ -0,0 +1,43 @@
...
+
+.rdg.fill-grid {
+  block-size: 100%;
+}
+
+.rdg.small-grid {
+  block-size: 300px;
+}
+
+.rdg.big-grid {
```
amanmahajan7, line 35: looks like these styles are only used once. Should we move them

website/root.css
```diff
@@ -0,0 +1,43 @@
...
+  block-size: 600px;
+}
+
+.rdg-cell .Select {
```
amanmahajan7, line 39: Is this used anywhere?

rome.json
```diff
@@ -18,76 +18,99 @@
...
+        "noNonNullAssertion": "off",
+        "noShoutyConstants": "error",
+        "noUnusedTemplateLiteral": "error",
+        "noVar": "error",
         "useBlockStatements": "off",
-        "useFragmentSyntax": "warn",
-        "useSelfClosingElements": "warn",
-        "useShorthandArrayType": "warn",
+        "useConst": "error",
+        "useDefaultParameterLast": "error",
```
amanmahajan7, line 79: I think TS would enforce it, no? Does not hurt to add it though.

nstepien: TS does not enforce it, it's valid js, for example:

```ts
function test(a = 'a', b: string) {}
```
has the type
```ts
function test(a: string | undefined, b: string): void
```

Maybe we should keep this rule disabled actually, wdyt? 🤔
I know I have to disable it in our biggest repo.

website/root.css
nstepien: Done

website/root.css
nstepien: I think it was used with `react-select` which we removed a long time ago?

rome.json
amanmahajan7: Ah, I thought the default parameters should be the last 🤔 May be I am confusing it with something else. I am okay with either. We can disable it for consistency with other repo if you prefer.

amanmahajan7 (approved):

rome.json
amanmahajan7: May be I am confusing it with rest parameters 😄

PR #3186, opened 2023-03-29
Allow rowClass to be applied positionally by row index.

MichaelDimmitt:
This pull request allows rowClass to send an additional parameter `row index`, this will allow classes to be added positionally. The current capability of adding classes by looking at the values will continue to work with no breaking changes.
    
An example of adding rowClass based on row index has been added to the AllFeatures demo.

Previous to this change I had to use a css pseudo class like .rdg-row:nth-child(14) to make the 14th row have a certain css behavior. And was unable to get the last row to have a css class as I could not get nth-last-child or last-child to work.

@amanmahajan7 @nstepien , please let me know if anything else needs to be updated as I am happy to make changes.
Thank you for your work on this great component!

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -179,7 +179,7 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
    * Miscellaneous
    */
   renderers?: Maybe<Renderers<R, SR>>;
-  rowClass?: Maybe<(row: R) => Maybe<string>>;
+  rowClass?: Maybe<(row: R, rowIdx: number) => Maybe<string>>;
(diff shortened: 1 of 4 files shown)


amanmahajan7 (commented):
Need to run prettier otherwise looks good. Thanks for the PR.

MichaelDimmitt:
@amanmahajan7 , thanks for pointing out I am updating now!

MichaelDimmitt:
@amanmahajan7 , change has been pushed fixing prettier. Do I need to make any change to the package.json version number? Let me know if anything else I can assist with. Thank you!

nstepien (approved):
Thank you!

PR #3187, opened 2023-03-31
Separate focus and selection logic to prevent grid stealing focus 

amanmahajan7:


src/Cell.tsx
```diff
@@ -34,15 +34,15 @@ function Cell<R, SR>({
   row,
   rowIdx,
   dragHandle,
-  skipCellFocusRef,
+  shouldFocusCellRef,
```
amanmahajan7, line 37: Now we only focus if `shouldFocusCellRef.current` is `true`. Wonder if this should be a state instead? But then resetting it would trigger a re-render

src/hooks/useFocusRef.ts
```diff
@@ -1,14 +1,13 @@
...
-
-  useLayoutEffect(() => {
-    if (!isSelected) return;
-    ref.current?.focus({ preventScroll: true });
-  }, [isSelected]);
+export function useFocusRef<T extends HTMLOrSVGElement>(isSelected: boolean, isFocused: boolean) {
+  const ref = useCallback(
+    (el: T | null) => {
+      if (el === null || !isFocused) return;
+      el.focus({ preventScroll: true });
```
nstepien, line 7: Do we not want to update the ref here after focusing?

nstepien (commented):
I wonder if we should rename `cellIsFocused` to `willFocusCell` 🤔

src/hooks/useFocusRef.ts
amanmahajan7: I will be updated in the [cell ref](https://github.com/adazzle/react-data-grid/pull/3187/files#diff-322374a1392f0a82977b2cfe82a7e08639d0eb211395073067129f39fd9be244R23). It does make me wonder about the timing. What if the flag is set to false before the ref in the formatter is initialized. I think children are mounted first I believe and the code seems to work.

src/Cell.tsx
```diff
@@ -113,6 +113,7 @@ function Cell<R, SR>({
             column,
             row,
             isCellSelected,
+            isCellFocused: isCellSelected && shouldFocusCellRef.current,
```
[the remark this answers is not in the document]
amanmahajan7: We can pass the ref to the formatter but then it can lead to bugs if not set properly. May be a readonly ref 🤔

src/DataGrid.tsx
```diff
@@ -1267,6 +1279,15 @@ function DataGrid<R, SR, K extends Key>(
   );
 }
 
+function getCellToFocus(gridEl: HTMLDivElement) {
```
amanmahajan7, line 1282: Any other name? We do not always focus the cell

amanmahajan7: Rename to `getCellToScroll`

nstepien (approved):
😎 so much better

nstepien (approved):

PR #3196, opened 2023-04-15
Update linaria, work around linaria bug

nstepien:
See https://github.com/callstack/linaria/issues/1206#issuecomment-1509945297
I've removed the barrel file to work around the Linaria bug causing the crashes.
I'm keeping the other PR in draft to help the Linaria maintainer debug the issue.


nstepien:
@amanmahajan7 this should be good now.

amanmahajan7 (approved):

PR #3199, opened 2023-04-18
Fix css escape for static pages

amanmahajan7:
https://github.com/adazzle/react-data-grid/actions/runs/4734177622/jobs/8402735152

> ReferenceError: CSS is not defined

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -9,7 +9,10 @@ export * from './selectedCellUtils';
 export * from './styleUtils';
 
 export const { min, max, round, floor, sign, abs } = Math;
-export const { escape: cssEscape } = CSS;
+
+export function cssEscape(ident: string): string {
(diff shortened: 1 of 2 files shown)


src/utils/index.ts
```diff
@@ -9,7 +9,10 @@ export * from './selectedCellUtils';
 export * from './styleUtils';
 
 export const { min, max, round, floor, sign, abs } = Math;
-export const { escape: cssEscape } = CSS;
```
nstepien, line 12: I had done this to skip the property access on the `CSS` object,
so instead of creating a `cssEscape` function we can just use `CSS.escape()` 🤷‍♂️

amanmahajan7: 🤦‍♂️

nstepien (approved):

PR #3204, opened 2023-04-19
Add failing test regarding formatter stealing focus back to the grid when the cell is recreated

amanmahajan7:


test/column/formatter.test.tsx
```diff
@@ -114,3 +114,41 @@ describe('Custom formatter component', () => {
...
+    const [inputValue, setInputValue] = useState('');
+    const [rows, setRows] = useState((): readonly Row[] => [{ id: 1 }]);
+
+    function onChange(event: React.ChangeEvent<HTMLInputElement>) {
+      setInputValue(event.target.value);
+      setRows([{ id: 2 }]);
+    }
+
+    return (
+      <StrictMode>
```
nstepien, line 126: Won't be needed with the new `render`
```suggestion
```

test/column/formatter.test.tsx
```diff
@@ -114,3 +114,41 @@ describe('Custom formatter component', () => {
     });
   });
 });
+
+describe('formatter focus', () => {
```
nstepien, line 114: Doesn't matter too much, but isn't this about `cell` focus rather than `formatter` focus?

test/column/formatter.test.tsx
```diff
@@ -110,3 +110,41 @@ describe('Custom formatter component', () => {
...
+          columns={columns}
+          rows={rows}
+          onRowsChange={setRows}
+          rowKeyGetter={(row) => row.id}
+        />
+      </StrictMode>
+    );
+  }
+
+  it.skip('should not steal focus back to the cell if the focus is outside the grid and formatter is recreated', async () => {
```
nstepien, line 138: With the describe name it becomes

>formatter focus should not steal focus back to the cell if the focus is outside the grid and formatter is recreated

the describe name should be changed, or removed, wdyt?

test/column/formatter.test.tsx
```diff
@@ -110,3 +110,41 @@ describe('Custom formatter component', () => {
...
+    const [rows, setRows] = useState((): readonly Row[] => [{ id: 1 }]);
+
+    function onChange(event: React.ChangeEvent<HTMLInputElement>) {
+      setInputValue(event.target.value);
+      setRows([{ id: 2 }]);
+    }
+
+    return (
+      <StrictMode>
+        <input aria-label="outer-input" value={inputValue} onChange={onChange} />
```
nstepien, line 127: Could be a button, focus should remain on the button (if we didn't have the bug)

test/keyboardNavigation.test.tsx
```diff
@@ -296,21 +295,13 @@ test('reset selected cell when column is removed', async () => {
...
-    </StrictMode>
-  );
+  const { rerender } = render(<Test columns={columns} />);
 
   await userEvent.tab();
   await userEvent.keyboard('{arrowdown}{arrowright}');
   validateCellPosition(1, 1);
 
-  rerender(
-    <StrictMode>
```
nstepien, line 310: We need to replace `rerender` so it keeps re-rendering with `StrictMode`

test/column/formatter.test.tsx
```diff
@@ -110,3 +110,37 @@ describe('Custom formatter component', () => {
     });
   });
 });
+
+test.skip('Cell should not steal focus when the focus is outside the grid and cell is recreated', async () => {
```
nstepien, line 114: FWIW we can also do this, it'll run the test.
```suggestion
test.failing('Cell should not steal focus when the focus is outside the grid and cell is recreated', async () => {
```


If we ever accidentally fix this test, it'll make jest log this:
image
and so we could remove `.failing` to ensure the test keeps passing and isn't ignored.

amanmahajan7: Nice 👍

nstepien (approved):

PR #3213, opened 2023-05-02
useColumnWidths: trigger fewer re-renders when resizing columns

nstepien:
We may be able to adopt a similar strategy to update `grid-template-columns` when the grid resizes, would need some more refactoring though. Not as important IMO anyway.

### DIFF
--- a/src/hooks/useColumnWidths.ts
+++ b/src/hooks/useColumnWidths.ts
@@ -41,13 +41,12 @@ export function useColumnWidths<R, SR>(
 
   useLayoutEffect(() => {
     prevGridWidthRef.current = gridWidth;
-
-    if (columnsToMeasure.length === 0) return;
-
     updateMeasuredWidths(columnsToMeasure);
   });
 
   function updateMeasuredWidths(columnsToMeasure: readonly string[]) {
+    if (columnsToMeasure.length === 0) return;
+
     setMeasuredColumnWidths((measuredColumnWidths) => {
       const newMeasuredColumnWidths = new Map(measuredColumnWidths);
       let hasChanges = false;
--- a/src/hooks/useColumnWidths.ts
+++ b/src/hooks/useColumnWidths.ts
@@ -84,7 +84,8 @@ export function useColumnWidths<R, SR>(
     }
 
     gridRef.current!.style.gridTemplateColumns = newTemplateColumns.join(' ');
-    const measuredWidth = measureColumnWidth(gridRef, resizingKey)!;
+    const measuredWidth =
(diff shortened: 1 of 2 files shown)


src/hooks/useColumnWidths.ts
```diff
@@ -17,120 +17,96 @@ export function useColumnWidths<R, SR>(
...
     setMeasuredColumnWidths((measuredColumnWidths) => {
       const newMeasuredColumnWidths = new Map(measuredColumnWidths);
       let hasChanges = false;
 
-      for (const { key } of columnsToMeasure) {
+      for (const key of columnsToMeasure) {
         const measuredWidth = measureColumnWidth(gridRef, key);
         hasChanges ||= measuredWidth !== measuredColumnWidths.get(key);
-        newMeasuredColumnWidths.set(key, measuredWidth);
+        if (measuredWidth === undefined) {
```
amanmahajan7, line 57: When can it be undefined?

src/hooks/useColumnWidths.ts
```diff
@@ -17,120 +17,96 @@ export function useColumnWidths<R, SR>(
...
+        const width = typeof nextWidth === 'number' ? `${nextWidth}px` : nextWidth;
+        newTemplateColumns[idx] = width;
+      } else if (columnsCanFlex && typeof width === 'string' && !resizedColumnWidths.has(key)) {
+        newTemplateColumns[idx] = width;
+        columnsToMeasure.push(key);
+      }
+    }
 
-      const measuredWidth = measureColumnWidth(gridRef, key);
+    gridRef.current!.style.gridTemplateColumns = newTemplateColumns.join(' ');
```
amanmahajan7, line 83: Should we skip this step if `nextWidth` is a number and `columnsCanFlex` is false?

amanmahajan7 (commented):
A few quick comments. I will test it tomorrow

src/hooks/useColumnWidths.ts
nstepien: Columns that are outside the viewport.
Previously it was resize -> remove columns from state -> re-render -> re-measure
Now it's resize -> re-measure -> update state with new size, delete previous measurement otherwise

src/hooks/useColumnWidths.ts
nstepien: We could, but resizing columns is actually already smooth when columns cannot flex, so it would add complexity for not much gain.

https://user-images.githubusercontent.com/567105/235890802-a55c31ca-f017-411d-95c1-9bc5ec1c95b3.mp4

(`auto` layout is quite slow when rendering a lot of content that can grow 🤔)

amanmahajan7 (approved):

PR #3222, opened 2023-05-08
Vitest: use browser mode

amanmahajan7:
https://vitest.dev/guide/browser.html
https://vitest.dev/guide/browser/interactivity-api.html

I think we should check all the tests if we can improve them now that they are running in a real browser.


test/column/draggable.test.ts
```diff
@@ -1,31 +1,36 @@
...
 
 import type { Column } from '../../src';
 import { getHeaderCells, setup } from '../utils';
 
 const columns: readonly Column<never>[] = [
   {
     key: 'col1',
-    name: 'col1'
+    name: 'col1',
+    width: 100
```
amanmahajan7, line 11: Tests fail with `width`

test/column/resizable.test.tsx
```diff
@@ -68,34 +56,87 @@ const columns: readonly Column<never>[] = [
...
+test('should use the minWidth if specified', async () => {
+  setup<Row, unknown>({ columns, rows: [] });
   const [, col2] = getHeaderCells();
   expect(getGrid()).toHaveStyle({ gridTemplateColumns: '100px 200px' });
-  resize({ column: col2, clientXStart: 295, clientXEnd: 100, rect: { right: 300, left: 100 } });
+  await resize({ column: col2, resizeBy: -150 });
   expect(getGrid()).toHaveStyle({ gridTemplateColumns: '100px 100px' });
 });
+
+test('should not auto resize column if resizable is not specified', async () => {
```
amanmahajan7, line 96: Now we can test auto column

vite.config.ts
```diff
@@ -1,10 +1,25 @@
 import react from '@vitejs/plugin-react';
 import wyw from '@wyw-in-js/vite';
 import { defineConfig } from 'vite';
+import type { BrowserCommand } from 'vitest/node';
 
 const isCI = process.env.CI === 'true';
 const isTest = process.env.NODE_ENV === 'test';
 
+// TODO: remove when `userEvent.pointer` is supported
+const resizeColumn: BrowserCommand<[resizeBy: number]> = async (context, resizeBy) => {
```
amanmahajan7, line 10: https://vitest.dev/guide/browser/commands.html#custom-commands

vite.config.ts
```diff
@@ -1,10 +1,25 @@
...
+  const page = context.page;
+  const frame = await context.frame();
+  const resizeHandle = frame.locator('[role="columnheader"][aria-colindex="2"]  div');
+  const { x, y } = (await resizeHandle.boundingBox())!;
+  await resizeHandle.hover({
+    position: { x: 5, y: 5 }
```
amanmahajan7, line 16: Need to use a specific position otherwise the resized width is incorrect

vite.config.ts
```diff
@@ -13,7 +28,8 @@ export default defineConfig({
   },
   resolve: {
     alias: {
-      lodash: 'lodash-es'
+      lodash: isTest ? 'lodash' : 'lodash-es',
```
amanmahajan7, line 31: Is this needed?

test/virtualization.test.ts
```diff
@@ -104,7 +104,7 @@ test('virtualization is enabled', async () => {
   assertRows(38, 3, 40);
 
   await scrollGrid({ scrollTop: 250 });
-  assertRows(39, 3, 41);
+  assertRows(38, 3, 40);
```
nstepien, line 107: We scrolled from 249 to 250 to ensure rendering stays consistent even when 1px changes.
If we render the same rows between these two points then we should find new positions 1px apart.

test/virtualization.test.ts
```diff
@@ -116,7 +116,7 @@ test('virtualization is enabled', async () => {
   assertRows(38, 11, 48);
 
   await scrollGrid({ scrollTop: 530 });
-  assertRows(39, 11, 49);
+  assertRows(38, 11, 48);
```
nstepien, line 119: Same here

test/virtualization.test.ts
```diff
@@ -131,8 +131,8 @@ test('virtualization is enabled', async () => {
   assertCells(66, 18, 0, 17);
 
   await scrollGrid({ scrollLeft: 91 });
-  assertHeaderCells(19, 0, 18);
-  assertCells(66, 19, 0, 18);
+  assertHeaderCells(18, 0, 17);
+  assertCells(66, 18, 0, 17);
```
nstepien, line 135: And here

test/utils.tsx
```diff
@@ -1,10 +1,22 @@
...
+export function setup<R, SR, K extends React.Key = React.Key>(props: DataGridProps<R, SR, K>) {
+  render(
+    <DataGrid
+      {...props}
+      className={css`
+        width: 1920px;
+        height: 1080px;
+        &::-webkit-scrollbar {
+          display: none;
+        }
```
nstepien, line 16: Does this work?
```suggestion
        scrollbar-width: none;
```

.github/workflows/ci.yml
```diff
@@ -48,6 +48,8 @@ jobs:
           node --run build:types
       - name: Build website
         run: node --run build:website
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps
```
nstepien, line 52: https://playwright.dev/docs/browsers
Should we specify to speed things up?
Is `--with-deps` necessary?
```suggestion
        run: npx playwright install chromium --with-deps
```

test/rowHeight.test.ts
```diff
@@ -21,37 +20,24 @@ function setupGrid(rowHeight: DataGridProps<Row>['rowHeight']) {
...
-  expect(grid).toHaveStyle({ 'grid-template-rows': 'repeat(1, 40px) repeat(50, 40px)' });
-  expect(getRows()).toHaveLength(31);
-
-  await userEvent.tab();
-  expect(grid.scrollTop).toBe(0);
-
-  // Go to the last cell
-  const spy = vi.spyOn(window.HTMLElement.prototype, 'scrollIntoView');
-  await userEvent.keyboard('{Control>}{end}');
-  expect(spy).toHaveBeenCalled();
```
nstepien, line 37: Should we keep this?
We still want to check that we can navigate and scroll to cells outside the current viewport, regardless of whether we use a static or dynamic row height.

test/rowHeight.test.ts
```diff
@@ -21,37 +20,24 @@ function setupGrid(rowHeight: DataGridProps<Row>['rowHeight']) {
...
+  expect(grid).toHaveStyle({
+    'grid-template-rows':
+      '40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px 40px'
```
nstepien, line 29: Why does it not use the `repeat()` syntax? Is it because the "resolved" value is different than the input?

vite.config.ts
nstepien: `@testing-library/jest-dom` installs `lodash`, can't hurt to keep it.

.github/workflows/ci.yml
amanmahajan7: We can try. I followed the CI guide. I can try removing `--with-deps`
https://playwright.dev/docs/ci-intro#on-pushpull_request

amanmahajan7: The time reduced from ~50 seconds to ~6 seconds

nstepien: Nice!

test/rowHeight.test.ts
amanmahajan7: I think it is because we are using `toHaveStyle`. `expect(grid.style.gridTemplateRows).toBe('repeat(1, 40px) repeat(50, 50px)');` works fine.

amanmahajan7: https://github.com/testing-library/jest-dom/blob/main/src/to-have-style.js#L60

nstepien: Maybe we can reduce the amount of rows we render then 🙂

test/rowHeight.test.ts
amanmahajan7: We can reduce the number of rows then.

test/virtualization.test.ts
amanmahajan7: Done

test/column/draggable.test.ts
nstepien: Tests pass without `width` if you add `viewport: { width: 1920, height: 1080 }`

image

It might be good to add it so we test something that better represent the average viewport.

test/column/renderEditCell.test.tsx
```diff
@@ -100,15 +100,15 @@ describe('Editor', () => {
...
-
     await scrollGrid({ scrollTop: 2000 });
     expect(getCellsAtRowIndex(0)).toHaveLength(1);
     expect(screen.queryByRole('spinbutton', { name: 'col1-editor' })).not.toBeInTheDocument();
+    expect(getRows()[1]).toHaveTextContent('5353');
     await userEvent.keyboard('123');
-    expect(screen.getByRole('spinbutton', { name: 'col1-editor' })).toHaveValue(1230);
-    const spy = vi.spyOn(window.HTMLElement.prototype, 'scrollIntoView');
-    await userEvent.keyboard('{enter}');
-    expect(spy).toHaveBeenCalled();
```
nstepien, line 111: Why did we remove these checks?

```ts
expect(screen.getByRole('spinbutton', { name: 'col1-editor' })).toHaveValue(1230);
```
this checked that when typing on a cell, it opens the editor and the editor receives focus, and we continue typing in the editor

```ts
    const spy = vi.spyOn(window.HTMLElement.prototype, 'scrollIntoView');
    await userEvent.keyboard('{enter}');
    expect(spy).toHaveBeenCalled();
```
this checked that we scrolled back to the cell even if we had scrolled past it (see `await scrollGrid({ scrollTop: 2000 });` above)

Maybe we need more comments.

test/column/renderEditCell.test.tsx
```diff
@@ -100,15 +100,15 @@ describe('Editor', () => {
...
+    expect(getCellsAtRowIndex(0)).toHaveLength(2);
+    expect(getRows()[1]).toHaveTextContent('11');
+    const editor = screen.getByRole('spinbutton', { name: 'col1-editor' });
+    expect(editor).toHaveValue(1230);
```
nstepien, line 111: Not sure what the new checks are for

vite.config.ts
```diff
@@ -1,10 +1,25 @@
...
+const resizeColumn: BrowserCommand<[resizeBy: number]> = async (context, resizeBy) => {
+  const page = context.page;
+  const frame = await context.frame();
+  const resizeHandle = frame.locator('[role="columnheader"][aria-colindex="2"]  div');
+  const { x, y } = (await resizeHandle.boundingBox())!;
+  await resizeHandle.hover({
+    position: { x: 5, y: 5 }
+  });
+  await page.mouse.down();
+  await page.mouse.move(x + resizeBy + 5 + 0, y);
```
nstepien, line 19:
```suggestion
  await page.mouse.move(x + resizeBy + 5, y);
```

vite.config.ts
```diff
@@ -24,23 +40,29 @@ export default defineConfig({
         plugins: [['optimize-clsx', { functionNames: ['getCellClassname'] }]]
       }
     }),
-    !isTest && wyw({ preprocessor: 'none' })
+    wyw({ preprocessor: 'none' })
   ],
   server: {
-    open: true
+    // TODO: open bug
+    open: !isTest
```
nstepien, line 47: seems fine now
We could even enable `browser.headless: true`
```suggestion
    open: true
```

test/column/renderEditCell.test.tsx
amanmahajan7: I think it is better to check the actual row in the view instead of spying on `scrollIntoView`

test/column/renderEditCell.test.tsx
nstepien: `etRows()[1]` gets the second _rendered_ row, no? Not necessarily the true "second" row.

amanmahajan7: May be we can use something like `toBeVisible` 🤔

amanmahajan7: Maybe we can just check `grid.scrollTop`

test/column/renderEditCell.test.tsx
```diff
@@ -100,15 +100,13 @@ describe('Editor', () => {
...
     await scrollGrid({ scrollTop: 2000 });
     expect(getCellsAtRowIndex(0)).toHaveLength(1);
     expect(screen.queryByRole('spinbutton', { name: 'col1-editor' })).not.toBeInTheDocument();
     await userEvent.keyboard('123');
+    expect(getCellsAtRowIndex(0)).toHaveLength(2);
     expect(screen.getByRole('spinbutton', { name: 'col1-editor' })).toHaveValue(1230);
-    const spy = vi.spyOn(window.HTMLElement.prototype, 'scrollIntoView');
-    await userEvent.keyboard('{enter}');
-    expect(spy).toHaveBeenCalled();
+    expect(getGrid().scrollTop).toBe(0);
```
nstepien, line 109: Should we check `scrollTop` before typing?

test/column/resizable.test.tsx
```diff
@@ -68,34 +51,85 @@ const columns: readonly Column<never>[] = [
...
+      }
+    ]
+  });
+  const [, col2] = getHeaderCells();
+  expect(getGrid()).toHaveStyle({ gridTemplateColumns: '100px 200px' });
+  await autoResize(col2);
+  expect(getGrid()).toHaveStyle({ gridTemplateColumns: '100px 327.703px' });
+});
+
+test('should use the maxWidth if specified on auto resize', async () => {
```
nstepien, line 121: Could also test minWidth

nstepien (approved):

test/column/resizable.test.tsx
amanmahajan7: I will add it in the nest PR
https://github.com/adazzle/react-data-grid/pull/3547

nstepien (approved):

PR #3236, opened 2023-05-23
Add `tabIndex` prop to cell formatter and header renderer

amanmahajan7:


src/types.ts
```diff
@@ -74,13 +74,15 @@ export interface FormatterProps<TRow, TSummaryRow = unknown> {
   row: TRow;
   isCellSelected: boolean;
```
amanmahajan7, line 75: May be we no longer need `isCellSelected` prop

nstepien: If we don't use it at all then let's remove it.
We can add it back later if we find a use for it.

nstepien (dismissed):

nstepien (approved):
Approving if you want to merge it now

amanmahajan7:
> If we don't use it at all then let's remove it.

I will create a follow-up PR

PR #3239, opened 2023-05-26
Rename render functions

nstepien:
# Changes

| before                                 | after                                     |
| -------------------------------------- | ----------------------------------------- |
| `column.headerRenderer`                | `column.renderHeaderCell`                 |
| `column.formatter`                     | `column.renderCell`                       |
| `column.summaryFormatter`              | `column.renderSummaryCell`                |
| `column.groupFormatter`                | `column.renderGroupCell`                  |
| `column.editor`                        | `column.renderEditCell`                   |
| `column.editorOptions.renderFormatter` | `column.editorOptions.displayCellContent` |
| `HeaderRendererProps`                  | `RenderHeaderCellProps`                   |
| `FormatterProps`                       | `RenderCellProps`                         |
| `SummaryFormatterProps`                | `RenderSummaryCellProps`                  |
| `GroupFormatterProps`                  | `RenderGroupCellProps`                    |
| `EditorProps`                          | `RenderEditCellProps`                     |
| `headerRenderer`                       | `renderHeaderCell`                        |
| `checkboxFormatter`                    | `renderCheckbox`                          |
| `toggleGroupFormatter`                 | `renderToggleGroup`                       |
| `renderers.rowRenderer`                | `renderers.renderRow`                     |
| `renderers.checkboxFormatter`          | `renderers.renderCheckbox`                |
| `renderers.sortStatus`                 | `renderers.renderSortStatus`              |
| `RowRendererProps`                     | `RenderRowProps`                          |
| `CheckboxFormatterProps`               | `RenderCheckboxProps`                     |
| `sortIcon`                             | `renderSortIcon`                          |
| `sortPriority`                         | `renderSortPriority`                      |
| `SortIconProps`                        | `RenderSortIconProps`                     |
| `SortPriorityProps`                    | `RenderSortPriorityProps`                 |
| `SortStatusProps`                      | `RenderSortStatusProps`                   |

# Why

Mostly consistency and clarity.

Render functions can be used in a multitude of ways.
"Formatter" isn't accurate as render functions are not necessarily used to "format" data.
A checkbox is not formatted data for example, while a date or cost can be formatted.
`editor` sounds like a component but it was also a render function.

The render functions' names were also inconsistent.
Some of them are `render*Cell` now. They all start with `render` so it's better for discoverability via text editor completion.
Although `*CellRenderer` might sound more natural, wdyt?

What might be confusing is that the render functions are used to render cell *content* rather the cell itself. But I'm not sure it's worth making the name too long like `renderCellContent`.

---

I've also removed some unnecessary react fragments.


src/types.ts
```diff
@@ -39,16 +45,17 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
   readonly editorOptions?: Maybe<{
-    /** @default false */
-    readonly renderFormatter?: Maybe<boolean>;
+    /**
+     * Render the cell content in addition to the edit cell.
+     * Enable this option when the editor is rendered outside the grid, like a modal for example.
+     * By default, the cell content is not rendered when the edit cell is open.
+     * @default false
+     */
+    readonly displayCell?: Maybe<boolean>;
```
nstepien, line 55: `renderFormatter` would have been a bit confusing I think, so I renamed it to `displayCell`.
Maybe `displayCellContent` is better though, wdyt?

website/demos/TreeView.tsx
```diff
@@ -120,7 +120,8 @@ const defaultRows = createRows();
 export default function TreeView({ direction }: Props) {
   const [rows, dispatch] = useReducer(reducer, defaultRows);
   const [allowDelete, setAllowDelete] = useState(true);
-  const columns: Column<Row>[] = useMemo(() => {
+
+  const columns = useMemo((): readonly Column<Row>[] => {
```
nstepien, line 124: FYI, because of where the type was set, doing the global rename did not automatically rename `formatter` here.
Another reason to always define the type as close as possible to the actual values.

src/types.ts
```diff
@@ -22,12 +22,18 @@ export interface Column<TRow, TSummaryRow = unknown> {
...
-  /** Formatter to be used to render the cell content */
-  readonly formatter?: Maybe<(props: FormatterProps<TRow, TSummaryRow>) => ReactNode>;
-  /** Formatter to be used to render the summary cell content */
-  readonly summaryFormatter?: Maybe<(props: SummaryFormatterProps<TSummaryRow, TRow>) => ReactNode>;
-  /** Formatter to be used to render the group cell content */
-  readonly groupFormatter?: Maybe<(props: GroupFormatterProps<TRow, TSummaryRow>) => ReactNode>;
+  /** Render function used to render the content of the column's header cell */
+  readonly renderHeaderCell?: Maybe<(props: RenderHeaderCellProps<TRow, TSummaryRow>) => ReactNode>;
+  /** Render function used to render the content of cells */
+  readonly renderCell?: Maybe<(props: RenderCellProps<TRow, TSummaryRow>) => ReactNode>;
```
amanmahajan7, line 28: Wonder if this is confusing if we ever add `cellRenderer`. If we ever add `cellRenderer` then `renderGridCell` may be used

src/DataGrid.tsx
```diff
@@ -245,10 +245,10 @@ function DataGrid<R, SR, K extends Key>(
   const rowHeight = rawRowHeight ?? 35;
   const headerRowHeight = rawHeaderRowHeight ?? (typeof rowHeight === 'number' ? rowHeight : 35);
   const summaryRowHeight = rawSummaryRowHeight ?? (typeof rowHeight === 'number' ? rowHeight : 35);
-  const rowRenderer = renderers?.rowRenderer ?? defaultRenderers?.rowRenderer ?? defaultRowRenderer;
+  const renderRow = renderers?.renderRow ?? defaultRenderers?.renderRow ?? defaultRenderRow;
```
amanmahajan7, line 248: May be then this should be called `renderGridRow` but probably too verbose.

src/DataGrid.tsx
```diff
@@ -245,10 +245,10 @@ function DataGrid<R, SR, K extends Key>(
...
   const sortStatus = renderers?.sortStatus ?? defaultRenderers?.sortStatus ?? defaultSortStatus;
```
amanmahajan7, line 249: Should we rename `sortStatus`?

src/types.ts
amanmahajan7: yeah, let's rename it to `displayCellContent`

src/DataGrid.tsx
nstepien: What about `renderRowWrapper`/`wrapRow`? 🤔

amanmahajan7: May be `renderRow` is fine for now.

src/DataGrid.tsx
nstepien: Done

src/types.ts
nstepien: Done

amanmahajan7 (approved):
:shipit:

amanmahajan7:
Should we merge this?

nstepien:
YOLO

shashi-drishya:
Hello @nstepien,

I had been using an older version of react-data-grid where I customized columns using the groupFormatter. In that version, I used a Boolean rowGroup key to distinguish between columns that should be shown in the groupBy functionality and those that shouldn't.

Now, I'm in the process of migrating to the new version of react-data-grid (7.0.0-beta.40), and I still want to perform customization. However, I can't seem to find the same keyword, rowGroup, in the column object when using renderGroupCell. 

I'm a bit confused about whether you've introduced a new keyword or if you've completely eliminated it in the new version. Your prompt response would be greatly appreciated as this is quite urgent.

amanmahajan7:
https://github.com/adazzle/react-data-grid/pull/2765/files#diff-32603821d3ad53db0ffc221dfcc065a457ab378bed4e1298809d36f1a5da16adL59

This is how the `rowGroup` prop was set. We should be able to add a similar check outside the grid.

shashi-drishya:
okay thanks for quick response. I have added following code in my grid container code to track. 
image

PR #3247, opened 2023-05-31
Use a single `RowSelectionChangeProvider`

amanmahajan7:


nstepien (approved):

godon019:
why need this change?

amanmahajan7:
It is not strictly needed. This just wraps all the rows with one provider instead of wrapping header and viewport rows with separate providers.

PR #3248, opened 2023-05-31
Validate `rowKeyGetter`

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -992,15 +992,19 @@ function DataGrid<R, SR, K extends Key>(
       const gridRowStart = headerAndTopSummaryRowsCount + rowIdx + 1;
       if (isGroupRow(row)) {
         ({ startRowIndex } = row);
-        const isGroupRowSelected =
```
amanmahajan7, line 995: image

this fails on the grouping demo if `rowKeyGetter` is not specified

nstepien (approved):

PR #3250, opened 2023-06-01
Add `shouldFocusCell` param to the `onClose` prop

amanmahajan7:


src/types.ts
```diff
@@ -104,7 +104,7 @@ export interface RenderEditCellProps<TRow, TSummaryRow = unknown> {
   column: CalculatedColumn<TRow, TSummaryRow>;
   row: TRow;
   onRowChange: (row: TRow, commitChanges?: boolean) => void;
-  onClose: (commitChanges?: boolean) => void;
+  onClose: (commitChanges?: boolean, shouldFocusCell?: boolean) => void;
```
amanmahajan7, line 107: Wonder if there is better name?

nstepien (approved):

PR #3261, opened 2023-06-14
add request animation frame to prevent resize observer limit loop exception

Karpengold:
Fix for "ResizeObserver loop limit exceeded" inspired by this topic: https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded

Related to https://github.com/adazzle/react-data-grid/issues/2712


nstepien:
Can you provide a minimal reproducible example?

The grid doesn't resize itself after its dimensions change, unless you perhaps apply custom styles to make it fit its columns/rows content? Maybe you're getting the error from some other place? Does this only happen on older unsupported browsers? We need to confirm whether this is a real issue with react-data-grid.

Using `requestAnimationFrame` may still lead to infinite resize loops.
Maybe it'd be better to throttle the callback instead?

soichih:
This is very much a real issue affecting the latest version of chrome, firefox, and probably all other browsers.

I believe this issue only occurs if react-data-grid is placed inside an app with complex components that takes a sufficiently long time to render. For a simpler app, the resize observers will never "pile up" thus never running into this issue.

Lorenzras:
I am also experiencing the same issue with complex hierarchy. In my case, if the parent's width is 100%, and the width of the datagrid exeeded that width of the parent, it infinitely throws the error, and the datagrid breaks into infite loop resizing between the parent's width and the width of its content.

**Solution**

I set the maxwidth of the parent relative to the size of the screen. I also tested that fixed width will do like `900px`, and just set it dynamically when needed.

```
maxWidth: `calc(100vw - ${offset})`
```

`offset ` is the width of the persistent menu, and to compensate for margins and paddings.

nstepien:
@Karpengold @soichih @Lorenzras 
Could you please try `7.0.0-beta.34`? I've just published it with a fix which I believe should resolve your issues.

Lorenzras:
> @Karpengold @soichih @Lorenzras Could you please try `7.0.0-beta.34`? I've just published it with a fix which I believe should resolve your issues.

I checked it with 7.0.0-beta.34, and the error is not happening anymore in my usecase.

soichih:
> > @Karpengold @soichih @Lorenzras Could you please try `7.0.0-beta.34`? I've just published it with a fix which I believe should resolve your issues.
> 
> I checked it with 7.0.0-beta.34, and the error is not happening anymore in my usecase. Awesome!

Same here. The issue seems to be resolved with 7.0.0-beta.34

alecdesjardins:
I am on `7.0.0-beta.34` and am still getting the warning/error in React.

nstepien:
@alecdesjardins Can you provide a reproducible demo?

alecdesjardins:
> @alecdesjardins Can you provide a reproducible demo?

It is odd. I can't reproduce in a sandbox environment with the same amount of data. In my development app, I get the error.

soichih:
> > > @Karpengold @soichih @Lorenzras Could you please try `7.0.0-beta.34`? I've just published it with a fix which I believe should resolve your issues.
> > 
> > 
> > I checked it with 7.0.0-beta.34, and the error is not happening anymore in my usecase. Awesome!
> 
> Same here. The issue seems to be resolved with 7.0.0-beta.34

Sorry I take this back. I am still seeing this error on 7.0.0-beta.34..

BenJBailey:
I'm attaching a basic app that shows this error occurring consistently in `7.0.0-beta34`. It's resizing on an interval to simply show that this error is firing (though it's not bubbling up to console.error, for some reason):

(You'll need to download into a directory, `npm install` and `npm run dev`. Sorry, but I couldn't get codepen to run the beta easily).

package.json:

```json
{
  "name": "grid-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.jsx",
  "scripts": {
    "dev": "vite"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "vite": "^4.3.9"
  },
  "dependencies": {
    "react": "^17.0.2",
    "react-dom": "^17.0.2",
    "react-data-grid": "^7.0.0-beta.34"
  }
}
```
index.html
```html
<!DOCTYPE html>
<html>
  <head>
    <title>Grid test</title>

    <script type="module" src="index.jsx"></script>
    <script>
      window.onerror = (event, source, lineno, colno, error) => {
        console.log(event, source, lineno, colno, error);
        return false;
      };
      let style = "width: 50%";
      const resizeContainer = () => {
        const container = document.getElementById("app");
        container.setAttribute("style", style);
        style = style == "width: 50%" ? "width: 100%" : "width: 50%";
      };
      window.addEventListener("DOMContentLoaded", () => {
        setInterval(resizeContainer, 1000);
      });
    </script>
  </head>
  <body>
    <div
      id="app"
      style="margin: 20px; max-height: 200px; overflow: scroll"
    ></div>
    <link
      href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css"
      rel="stylesheet"
      integrity="sha384-1q8mTJOASx8j1Au+a5WDVnPi2lkFfwwEAa8hDDdjZlpLegxhjVME1fgjWPGmkzs7"
      crossorigin="anonymous"
    />
  </body>
</html>
```

index.jsx: 

```jsx
import React from "react";
import DataGrid, { textEditor, SelectColumn } from "react-data-grid";
import { render } from "react-dom";
import "react-data-grid/lib/styles.css";
render(
  <div style={{ width: "100% " }}>
    {" "}
    <DataGrid
      columns={[
        SelectColumn,
        { key: "Result", name: "Result", editor: textEditor },
        { key: "Details", name: "Details", editor: textEditor },
      ]}
      rows={[
        {
          id: 1,
          Result: "Value 1",
          Details: "Value 2",
        },
      ]}
      onRowsChange={() => false}
      rowHeight={30}
      rowKeyGetter={(row) => row.id}
      selectedRows={new Set([])}
    />
  </div>,
  document.querySelector("#app")
);


```

nstepien:
@BenJBailey Thanks, I managed to reduce this down to
```html
<!doctype html>
<html>
  <head>
    <script type="module" src="index.tsx"></script>
  </head>
  <body>
    <div id="app"></div>
  </body>
</html>
```
```tsx
import { createRoot } from 'react-dom/client';

import DataGrid from 'react-data-grid';

createRoot(document.getElementById('app')).render(
  <DataGrid columns={[{ key: '', name: '' }]} rows={[]} />
);

window.addEventListener('error', (event) => {
  console.log(event.error);
  console.log(event);
});
```

1. I can reproduce this issue in Chrome/Edge when sizing down the browser window
1. There is no actual error object or stack trace, only an error message: `ResizeObserver loop limit exceeded`
1. The issue does not happen if I do not use `window.addEventListener('error', ...)`/`window.onerror`
1. The issue does not happen in Firefox
1. We only resize at most 1 more time after resizing the grid:
   - sizing the grid down adds scrollbars
   - we resize the columns to fit the available space, and scrollbars get removed
   - thus, this isn't enough to exceed any real loop limits

I googled around and found [this](https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded).

>This error means that ResizeObserver was not able to deliver all observations within a single animation frame. It is benign (your site will not break).

In conclusion, RDG does not exceed any loop limits, the issue should be fixed in Chromium, and you can safely ignore this "error".
Adding `requestAnimationFrame()` would only make the grid worse by making it less responsive to user actions.

soichih:
@nstepien This issue also occurs on the latest version of Firefox. This is not a Chromium issue, as far as I know.  Also, this issue raises exceptions on our production build so I can not simply ignore this - unless we can add try/catch block on RDG?

aovchinn:
@nstepien
> The issue does not happen if I do not use window.addEventListener('error', ...)/window.onerror
it is separate issue that window error are "invisible" to the developer, error is still happening just not visible without the listener
https://bugs.chromium.org/p/chromium/issues/detail?id=1040031#c15

SarunasMaro:
Can someone please fix this? It has been a few months already...

PR #3262, opened 2023-06-17
Do not use size containment on the grid

nstepien:


src/style/core.ts
```diff
@@ -70,7 +70,8 @@ const root = css`
 
       /* https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context */
       /* We set a stacking context so internal elements don't render on top of external elements. */
-      contain: strict;
+      /* size containment is not used as it could break "width: min-content" for example, and the grid would infinitely resize on Chromium browsers */
+      contain: layout paint style;
```
amanmahajan7, line 74:
```suggestion
      contain: content;
```

amanmahajan7 (approved):

src/style/core.ts
godon019: do we not need `style`?

godon019: ah, it is included. nvm

PR #3263, opened 2023-06-19
Update eslint config

nstepien:


test/column/resizable.test.tsx
```diff
@@ -44,10 +44,12 @@ function resize<K extends keyof DOMRect>({
     ...original(),
     ...rect
   });
+  // eslint-disable-next-line testing-library/prefer-user-event
```
amanmahajan7, line 47: Did we ever try `userEvent.pointer` with JSDOM? Not sure if it is worth

test/utils.tsx
```diff
@@ -76,13 +76,15 @@ export function validateCellPosition(columnIdx: number, rowIdx: number) {
 }
 
 export function copySelectedCell() {
+  // eslint-disable-next-line testing-library/prefer-user-event
```
amanmahajan7, line 79: Is it worth exporting `fireEvent` from `testUtils` to avoid disabling this rule everywhere?

nstepien: I'd rather we did replace fireEvent whenever possible rather than allowing it, what's the point of enabling the rule otherwise

test/column/renderEditCell.test.tsx
```diff
@@ -71,17 +71,20 @@ describe('Editor', () => {
...
     expect(onSave).not.toHaveBeenCalled();
     const saveButton = screen.getByRole('button', { name: 'save' });
-    fireEvent.mouseDown(saveButton);
+
     // await userEvent.click() triggers both mousedown and click, but without delay,
     // which isn't realistic, and isn't enough to trigger outside click detection
+    await user.pointer([{ keys: '[MouseLeft>]', target: saveButton }]);
     await act(async () => {
       await new Promise(requestAnimationFrame);
     });
```
godon019, line 85: questions
1. is this `act` for saving animation?
2. how did we ensure that the commit is fast enough?

nstepien: 1. The `act` is to wrap the other [`requestAnimationFrame`](https://github.com/adazzle/react-data-grid/blob/d5120d82e54548f80280d907166e97b76839e312/src/EditCell.tsx#L73) call from the grid
2. we test that `onSave` was called with updated data with this assertion:
   ```ts
     expect(onSave).toHaveBeenCalledWith([
       { col1: 2341, col2: 'a1' },
   ```

godon019: thanks for the answer. I think I was curious on how RDG implemented for 2. but the answer of 1. helped me understand

PR #3264, opened 2023-06-19
Tweak z-index

amanmahajan7:
Fixes row selection styles

image

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -291,7 +291,7 @@ function DataGrid<R, SR, K extends Key>(
   const prevSelectedPosition = useRef(selectedPosition);
   const latestDraggedOverRowIdx = useRef(draggedOverRowIdx);
   const lastSelectedRowIdx = useRef(-1);
-  const rowRef = useRef<HTMLDivElement>(null);
+  const focusSinkRef = useRef<HTMLDivElement>(null);
   const shouldFocusCellRef = useRef(false);
 
   /**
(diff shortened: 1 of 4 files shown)


src/DataGrid.tsx
```diff
@@ -1236,10 +1220,36 @@ function DataGrid<R, SR, K extends Key>(
             </>
           )}
         </RowSelectionChangeProvider>
-
-        {/* render empty cells that span only 1 column so we can safely measure column widths, regardless of colSpan */}
-        {renderMeasuringCells(viewportColumns)}
       </DataGridDefaultRenderersProvider>
+
+      {/* render empty cells that span only 1 column so we can safely measure column widths, regardless of colSpan */}
+      {renderMeasuringCells(viewportColumns)}
```
amanmahajan7, line 1226: Moved it outside the provider.

src/DataGrid.tsx
```diff
@@ -1236,10 +1220,36 @@ function DataGrid<R, SR, K extends Key>(
...
+
+      {/* extra div is needed for row navigation in a treegrid */}
+      {hasGroups && (
```
amanmahajan7, line 1229: Move this div to the end. This way we can use a lower value of `zIndex` and it already uses `pointer-events: none`. Can you think of any issues with this change?

src/DataGrid.tsx
```diff
@@ -1236,10 +1220,36 @@ function DataGrid<R, SR, K extends Key>(
...
+        <div
+          ref={rowRef}
```
nstepien, line 1231: Should re rename `rowRef`?

src/HeaderRow.tsx
```diff
@@ -32,14 +32,14 @@ const headerRow = css`
     font-weight: bold;
 
     & > .${cell} {
-      /* Should have a higher value than 0 to show up above regular cells */
-      z-index: 1;
+      /* Should have a higher value than 1 to show up above regular cells */
```
nstepien, line 35:
```suggestion
      /* Should have a higher value than 1 to show up above regular cells and the focus sink */
```

nstepien (dismissed):

nstepien (approved):

PR #3267, opened 2023-06-26
Fix `Home/End` key when header/summary row is selected and validate `rowIdx` in before calling `onCellKeyDown`

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -1019,7 +1017,7 @@ function DataGrid<R, SR, K extends Key>(
         <RowSelectionChangeProvider value={selectRowLatest}>
           <RowSelectionProvider value={allRowsSelected}>
             <HeaderRow
-              columns={getRowViewportColumns(-1)}
+              columns={getRowViewportColumns(minRowIdx)}
```
amanmahajan7, line 1020: We missed it when we added `topSummaryRows`

src/TreeDataGrid.tsx
```diff
@@ -292,8 +292,26 @@ function TreeDataGrid<R, SR, K extends Key>(
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     const idx = column?.idx ?? -1;
     const row = rows[rowIdx];
-    if (!isGroupRow(row)) return;
```
amanmahajan7, line 295: `row` is actually undefined on the header and summary row. I will create a follow-up PR to fix types
https://github.com/adazzle/react-data-grid/blob/main/src/DataGrid.tsx#L515

amanmahajan7: I will also add more tests in a follow up PR

src/TreeDataGrid.tsx
```diff
@@ -305,24 +323,12 @@ function TreeDataGrid<R, SR, K extends Key>(
         toggleGroup(row.id);
```
amanmahajan7, line 323:
```suggestion
        toggleGroup(row.id);
        return;
```

nstepien (approved):

PR #3270, opened 2023-07-01
Include KeyboardEvent to onCopy and onPaste callbacks

paustint:
Users may need to stop propagation or prevent default or otherwise utilize the source event

There is a bug where I have a cell  renderer with a button which opens a modal and shows another table in the modal. The `onCopy` event is fired in the table from the modal then fired again from the base table, so the base table is what gets added to the clipboard.

Having the ability to `stopPropagation` and `preventDefault` solve this issue by making sure that the event does not get fired on both tables.


amanmahajan7:
RDG now supports `onCellCopy/onCellPaste` handlers

PR #3286, opened 2023-07-17
Update linaria, use `@linaria/vite` in vite

nstepien:
Looks like it's working fine now.

https://github.com/callstack/linaria/releases


amanmahajan7 (approved):
Nice

PR #3295, opened 2023-08-01
Tweak `onEditorNavigation` and `getCellToScroll`

nstepien:


src/utils/keyboardUtils.ts
```diff
@@ -69,8 +69,9 @@ export function onEditorNavigation({ key, target }: React.KeyboardEvent<HTMLDivE
       target instanceof HTMLTextAreaElement ||
       target instanceof HTMLSelectElement)
   ) {
-    return target.matches(
-      '.rdg-editor-container > :only-child, .rdg-editor-container > label:only-child > :only-child, .rdg-editor-container > div:only-child > label:only-child > :only-child'
```
nstepien, line 73: We can't rely on `:only-child` anymore for our use cases

PR #3309, opened 2023-08-15
Add support for dragging columns

amanmahajan7:
Added `column.draggable` and `onColumnsReorder` props


TucciAlessandro:
Hi @amanmahajan7, are you still working on this? 
I would really love to have this functionality in my project.
When do you think this would be avaiable to the public?

Thank you

src/HeaderCell.tsx
```diff
@@ -38,6 +39,14 @@ export const resizeHandleClassname = css`
   }
 `;
 
+const draggingClassname = css`
+  opacity: 0.5;
```
amanmahajan7, line 43: This also changes the opacity of the border. Wonder if we should fix that

src/DataGrid.tsx
```diff
@@ -82,7 +82,7 @@ interface EditCellState<R> extends Position {
 
 type DefaultColumnOptions<R, SR> = Pick<
   Column<R, SR>,
-  'renderCell' | 'width' | 'minWidth' | 'maxWidth' | 'resizable' | 'sortable'
+  'renderCell' | 'width' | 'minWidth' | 'maxWidth' | 'resizable' | 'sortable' | 'draggable'
```
amanmahajan7, line 85: If the column is both frozen and draggable then I think users can decide how to order them. RDG will still sort frozen columns first

amanmahajan7:
Need to fix https://stackoverflow.com/questions/7110353/html5-dragleave-fired-when-hovering-a-child-element.

amanmahajan7:
@TucciAlessandro we are still working on it. Hoping to release it in the next few days

godon019:
I think there is a bit of cursor style issue on entering/leaving [resizeHandle element
](https://github.com/adazzle/react-data-grid/blob/main/src/HeaderCell.tsx#L210)
https://github.com/adazzle/react-data-grid/assets/10363850/80e37c6d-6f95-4567-9b0e-75a4612a2892

src/HeaderCell.tsx
nstepien: Should we change the background color only instead?
Or maybe we can decrease the contrast with a css filter.

src/HeaderCell.tsx
```diff
@@ -191,13 +246,22 @@ export default function HeaderCell<R, SR>({
...
-      onKeyDown={column.sortable ? onKeyDown : undefined}
+      onKeyDown={sortable ? onKeyDown : undefined}
+      /* events fired on the draggable target */
+      onDragStart={draggable ? handleDragStart : undefined}
+      onDragEnd={draggable ? handleDragEnd : undefined}
+      /* events fired on the drop targets */
+      onDragOver={draggable ? handleDragOver : undefined}
+      onDragEnter={draggable ? handleDragEnter : undefined}
+      onDragLeave={draggable ? handleDragLeave : undefined}
+      onDrop={draggable ? handleDrop : undefined}
```
nstepien, line 264: Should we just have 1 `draggableProps` object with only 1 `draggable` check?

src/types.ts
```diff
@@ -43,6 +43,8 @@ export interface Column<TRow, TSummaryRow = unknown> {
   readonly resizable?: Maybe<boolean>;
   /** Enable sorting of a column */
   readonly sortable?: Maybe<boolean>;
+  /** Enable dragging of a column */
+  readonly draggable?: Maybe<boolean>;
```
nstepien, line 47: I wonder if this is the best API for this feature, what if we want some columns to never move? i.e. are there use cases where we'd want columns to move only in a portion of the grid.

website/demos/ColumnsReordering.tsx
```diff
@@ -29,72 +28,81 @@ function createRows(): Row[] {
...
+    resizable: true,
+    sortable: true,
+    draggable: true
+  }
+];
 
 export default function ColumnsReordering({ direction }: Props) {
   const [rows] = useState(createRows);
-  const [columns, setColumns] = useState(createColumns);
+  const [columnsOrder, setColumnsOrder] = useState(() => {
```
nstepien, line 69:
```suggestion
  const [columnsOrder, setColumnsOrder] = useState((): ReadonlyMap<string, number> => {
```

website/demos/ColumnsReordering.tsx
```diff
@@ -29,72 +28,81 @@ function createRows(): Row[] {
...
+    const map = new Map<string, number>();
+    for (const [index, column] of columns.entries()) {
+      map.set(column.key, index);
+    }
+    return map;
+  });
```
nstepien, line 75: Why not store the order as an array of indexes?
```suggestion
  const [columnsOrder, setColumnsOrder] = useState((): readonly number[] => {
    return columns.map((column, index) => index);
  });
```
then we can do
```ts
const reorderedColumns = /* useMemo => */ columnsOrder.map((index) => columns[index])
```

website/demos/ColumnsReordering.tsx
```diff
@@ -29,72 +28,81 @@ function createRows(): Row[] {
...
-      setColumns(reorderedColumns);
-    }
-
-    return columns.map((c) => {
-      if (c.key === 'id') return c;
-      return { ...c, renderHeaderCell };
+  const onColumnsReorder = useCallback((sourceKey: string, targetKey: string) => {
+    setColumnsOrder((columnsOrder) => {
+      const sourceColumnIndex = columnsOrder.get(sourceKey)!;
+      const targetColumnIndex = columnsOrder.get(targetKey)!;
```
nstepien, line 84: If we'd send the CalculatedColumn we could access `column.idx`

website/demos/ColumnsReordering.tsx
```diff
@@ -29,72 +28,81 @@ function createRows(): Row[] {
...
+      if (sourceColumnIndex === targetColumnIndex) return columnsOrder;
```
nstepien, line 85: Does RDG not handle this internally?

website/demos/ColumnsReordering.tsx
amanmahajan7: https://github.com/adazzle/react-data-grid/blob/ba9f62117e74c61c4a8f0db825c80a167ba79603/src/HeaderCell.tsx#L222

We can add a function in `DataGrid.tsx` that can find the column based on the key. I think it is okay to have keys here but I can change it if you prefer

amanmahajan7: https://github.com/adazzle/react-data-grid/blob/ba9f62117e74c61c4a8f0db825c80a167ba79603/src/HeaderCell.tsx#L222

We can add a function in `DataGrid.tsx` that can find the column based on the key. I think it is okay to have keys here but I can change it if you prefer

src/types.ts
amanmahajan7: I think for now users can decide whether to move a column or not. We can make `frozen` columns not draggable but there may be a use cases where frozen columns needs to reordered

website/demos/ColumnsReordering.tsx
amanmahajan7: Simplified a bit but using column keys array. How would we use array of indexes?

website/demos/ColumnsReordering.tsx
```diff
@@ -29,72 +28,67 @@ function createRows(): Row[] {
...
-
-    function handleColumnsReorder(sourceKey: string, targetKey: string) {
-      const sourceColumnIndex = columns.findIndex((c) => c.key === sourceKey);
-      const targetColumnIndex = columns.findIndex((c) => c.key === targetKey);
-      const reorderedColumns = [...columns];
+  const onColumnsReorder = useCallback((sourceKey: string, targetKey: string) => {
+    setColumnsOrder((columnsOrder) => {
+      const sourceColumnIndex = columnsOrder.indexOf(sourceKey)!;
+      const targetColumnIndex = columnsOrder.indexOf(targetKey)!;
+      const newColumnsOrder = [...columnsOrder];
```
nstepien, line 81: 🤔 
```suggestion
      const newColumnsOrder = columnsOrder.toSpliced(sourceColumnIndex, 1);
      newColumnsOrder.splice(targetColumnIndex, 0, sourceKey);
      return newColumnsOrder;
```

website/demos/ColumnsReordering.tsx
nstepien: The array would contain indexes of the columns in the original array, i.e. the original order.
That way
`[0, 1, 2]` -> `[col A, col B, col C]`
`[1, 2, 0]` -> `[col B, col C, col A]`

nstepien (approved):

amanmahajan7:
> think there is a bit of cursor style issue on entering/leaving [resizeHandle element](https://github.com/adazzle/react-data-grid/blob/main/src/HeaderCell.tsx#L210)

@godon019 this looks like a browser bug. Works find in FF

nstepien (approved):

PR #3316, opened 2023-08-22
Handle column sorting events directly on the header cell

nstepien:


src/HeaderCell.tsx
```diff
@@ -138,15 +157,18 @@ export default function HeaderCell<R, SR>({
     }
   }
 
-  function onClick() {
+  function onClick(event: React.MouseEvent<HTMLSpanElement>) {
+    if (ignoreClickRef.current) return;
```
nstepien, line 161: Maybe we should just have a real DOM element with `stopPropagation()` on it, would simplify the JS a little bit.

website/root.tsx
```diff
@@ -33,7 +33,7 @@ const mainClassname = css`
   box-sizing: border-box;
   block-size: 100vh;
   padding: 8px;
-  overflow: hidden;
+  overflow: clip;
```
amanmahajan7, line 36: Looks like this breaks the navigation style
image

src/HeaderCell.tsx
amanmahajan7: Yeah. let's do it

src/HeaderCell.tsx
```diff
@@ -62,28 +71,36 @@ export default function HeaderCell<R, SR>({
...
-    const offset = isRtl ? event.clientX - left : right - event.clientX;
+    const offset = getPointerOffset(event);
 
     if (offset > 11) {
       // +1px to account for the border size
       return;
     }
 
+    // prevent selecting/sorting if we started resizing the column, including double-clicking
+    ignoreClickRef.current = true;
```
amanmahajan7, line 98: We do not use a ref currently. Is there a bug?

nstepien: Without it, after resizing the column, the `click` event is triggered on the header cell.
This wasn't an issue before as the `click` event handler was on a child element

amanmahajan7 (dismissed):

amanmahajan7 (approved):

PR #3326, opened 2023-08-28
Use native drag/drop api to implement column reordering

amanmahajan7:


website/demos/ColumnsReordering.tsx
```diff
@@ -1,6 +1,4 @@
 import { useCallback, useMemo, useState } from 'react';
-import { DndProvider } from 'react-dnd';
-import { HTML5Backend } from 'react-dnd-html5-backend';
```
nstepien, line 3: Can we uninstall these package?

website/demos/components/DraggableHeaderRenderer.tsx
```diff
@@ -11,31 +11,52 @@ export function DraggableHeaderRenderer<R>({
...
+    toggleIsOver(false);
+  }
 
   return (
     <div
-      ref={(ref) => {
-        drag(ref);
-        drop(ref);
-      }}
+      draggable
```
nstepien, line 51: Should we implement all this as a DataGrid/column feature?

website/demos/components/DraggableHeaderRenderer.tsx
```diff
@@ -11,31 +11,52 @@ export function DraggableHeaderRenderer<R>({
...
-    drop({ key }: { key: string }) {
-      onColumnsReorder(key, column.key);
-    },
-    collect: (monitor) => ({
-      isOver: monitor.isOver(),
-      canDrop: monitor.canDrop()
-    })
-  });
+  const [isDragging, toggleDragging] = useState(false);
+  const [isOver, toggleIsOver] = useState(false);
```
nstepien, line 15: I don't mind too much, but why `toggle` instead of `set`? It's not consistent with pretty much everything else

website/demos/components/DraggableHeaderRenderer.tsx
```diff
@@ -11,31 +11,52 @@ export function DraggableHeaderRenderer<R>({
...
+    event.preventDefault();
+  }
+
+  function handleDrop(event: React.DragEvent<HTMLDivElement>) {
+    const sourceKey = event.dataTransfer.getData('text/plain');
+    if (sourceKey !== column.key) {
+      event.preventDefault();
+      onColumnsReorder(sourceKey, column.key);
+    }
+    toggleIsOver(false);
```
nstepien, line 46: Small tweak: we can call `toggleIsOver(false);` at the top of the function,
that way if for example `onColumnsReorder()` throws an error, then we'll at least end up with the correct state.

website/demos/ColumnsReordering.tsx
amanmahajan7: Still need to migrate https://github.com/adazzle/react-data-grid/blob/main/website/demos/components/DraggableRowRenderer.tsx#L1

website/demos/components/DraggableHeaderRenderer.tsx
amanmahajan7: https://github.com/adazzle/react-data-grid/pull/3309

nstepien (approved):

PR #3327, opened 2023-08-30
use Biome

godon019:
use Biome instead of Rome


biome.json
```diff
@@ -151,5 +150,8 @@
         "useValidTypeof": "error"
       }
     }
+  },
+  "organizeImports": {
+    "enabled": false
   }
```
godon019, line 156: Biome organize imports by default and it conflicts with prettier so turned it off. https://biomejs.dev/analyzer/#imports-sorting

should we use Biome to sort import?

biome.json
```diff
@@ -42,7 +42,6 @@
       },
       "complexity": {
         "noExtraBooleanCast": "error",
-        "noExtraSemicolon": "error",
```
godon019, line 45: this emits error. Biome doesn't seem to have it.
and the rule seems to be removed from Rome as well

it doesn't exists when we click the link to [lint/complexity/noExtraSemicolon](https://docs.rome.tools/lint/rules/noExtraSemicolon)
from https://rome.tools/blog/2023/05/10/rome12_1/

godon019: removed from Rome https://github.com/rome/tools/pull/4576

amanmahajan7 (approved):

PR #3329, opened 2023-08-31
Column Groups: Fix cell navigation 

amanmahajan7:


src/utils/selectedCellUtils.ts
```diff
@@ -119,12 +137,58 @@ export function getNextSelectedCellPosition<R, SR>({
     }
   };
 
+  const getParentRowIdx = (parent: CalculatedColumnParent<R, SR>) => {
+    return parent.level + mainHeaderRowIdx;
+  };
+
+  const setHeaderGroupColAndRowSpan = (moveRight: boolean) => {
+    if (moveRight) {
```
amanmahajan7, line 145: Needs special handling for these right and top navigation

amanmahajan7 (commented):
There may be some edge cases but this should cover basic navigation

src/utils/selectedCellUtils.ts
```diff
@@ -119,12 +137,58 @@ export function getNextSelectedCellPosition<R, SR>({
     }
   };
 
+  const getParentRowIdx = (parent: CalculatedColumnParent<R, SR>) => {
+    return parent.level + mainHeaderRowIdx;
+  };
+
+  const setHeaderGroupColAndRowSpan = (moveRight: boolean) => {
```
nstepien, line 144: Is `moveRight` always about moving right? What about right-to-left layouts? Should this be `moveNext` instead?

nstepien (approved):

src/utils/selectedCellUtils.ts
amanmahajan7: good point

nstepien (approved):

PR #3334, opened 2023-09-05
Move drag handle outside the cell

amanmahajan7:
Pros:
- Cleans up row and cell props

Cons
- Needs special handling for col spans

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -803,7 +803,7 @@ function DataGrid<R, SR, K extends Key>(
     if (
       onFill == null ||
       selectedPosition.mode === 'EDIT' ||
-      !isCellWithinViewportBounds(selectedPosition) ||
+      !isCellWithinViewportBounds(selectedPosition)
     ) {
       return;
     }
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -796,13 +796,18 @@ function DataGrid<R, SR, K extends Key>(
     return isDraggedOver ? selectedPosition.idx : undefined;
   }
 
-  function getDragHandle(rowIdx: number) {
-    if (selectedPosition.rowIdx !== rowIdx || selectedPosition.mode === 'EDIT' || onFill == null) {
+  function renderDragHandle() {
+    if (
+      !isCellWithinViewportBounds(selectedPosition) ||
+      selectedPosition.mode === 'EDIT' ||
+      onFill == null
+    ) {
       return;
     }
 
     return (
       <DragHandle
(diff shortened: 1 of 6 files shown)


src/DragHandle.tsx
```diff
@@ -105,7 +121,15 @@ export default function DragHandle<R, SR>({
 
   return (
     <div
-      className={cellDragHandleClassname}
+      style={{
+        gridColumnStart: column.idx + 1,
+        gridRowStart,
+        insetInlineStart:
+          column.frozen && typeof column.width === 'number'
```
amanmahajan7, line 128: Not sure if this can cause any issues

src/DragHandle.tsx
```diff
@@ -105,7 +121,15 @@ export default function DragHandle<R, SR>({
 
   return (
     <div
-      className={cellDragHandleClassname}
+      style={{
+        gridColumnStart: column.idx + 1,
```
amanmahajan7, line 125: Need to handle `colSpan`. I will check in the next PR

amanmahajan7: Fix in this PR

src/DataGrid.tsx
```diff
@@ -799,13 +799,18 @@ function DataGrid<R, SR, K extends Key>(
     return isDraggedOver ? selectedPosition.idx : undefined;
   }
 
-  function getDragHandle(rowIdx: number) {
-    if (selectedPosition.rowIdx !== rowIdx || selectedPosition.mode === 'EDIT' || onFill == null) {
+  function renderDragHandle() {
+    if (
```
nstepien, line 803: Do we need to check whether the selected cell is not in a summary row?

src/DragHandle.tsx
```diff
@@ -1,30 +1,40 @@
...
-    position: absolute;
-    inset-inline-end: 0;
-    inset-block-end: 0;
-    inline-size: 8px;
-    block-size: 8px;
+    inline-size: var(--rdg-drag-handle-size);
+    block-size: var(--rdg-drag-handle-size);
     background-color: var(--rdg-selection-color);
+    align-self: end;
+    justify-self: end;
```
nstepien, line 17:
```suggestion
    place-self: end;
```

src/DataGrid.tsx
amanmahajan7: `isCellWithinViewportBounds` would return false for header and summary rows. There is another function `isCellWithinSelectionBounds` that can be used to check header and summary rows as well

nstepien (approved):

PR #3341, opened 2023-09-13
Focus cell when drag handle is clicked

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -831,6 +825,9 @@ function DataGrid<R, SR, K extends Key>(
         isCellEditable={isCellEditable}
         latestDraggedOverRowIdx={latestDraggedOverRowIdx}
         onRowsChange={onRowsChange}
+        onClick={() => {
+          focusCellOrCellContent(gridRef.current!);
+        }}
```
nstepien, line 830: Maybe we can keep the function in the component, that way we don't need to pass a parameter
```suggestion
        onClick={focusCellOrCellContent}
```

nstepien (approved):

Issue #3351, opened 2023-09-29
Cell drag handle is rendered at the wrong place


9at8:
## Describe the bug 

Cell drag handle (blue square) is rendered at the incorrect position after column with a "fixed" width is changed.

## To Reproduce 

1. Change the width of the "Last Name" column
2. Horizontally scroll
3. Cell drag handle should moves around

Link to code example: https://adazzle.github.io/react-data-grid/#/all-features

## Expected behavior 

Cell drag handle should not move around

## Environment

- `react-data-grid` version: same as the demo website
- `react`/`react-dom` version: same as the demo website

![drag-handle-bad-position](https://github.com/adazzle/react-data-grid/assets/8235156/6d9e8f43-4b60-49b0-8428-ec3a99b86fda)

PR #3352, opened 2023-10-02
Fix drag handle position when frozen column is resized

amanmahajan7:
Fixes https://github.com/adazzle/react-data-grid/issues/3351


src/DragHandle.tsx
```diff
@@ -127,8 +127,8 @@ export default function DragHandle<R, SR>({
         ...style,
         gridRowStart,
         insetInlineStart:
-          style.insetInlineStart && typeof column.width === 'number'
```
amanmahajan7, line 130: We were not using the calculated width

nstepien (approved):

Issue #3349, opened 2023-09-29
Columns reordering: demo crash


d9k:
https://adazzle.github.io/react-data-grid/#/columns-reordering

```
react-dom.production.min.js:189 TypeError: x.toSpliced is not a function
    at ColumnsReordering.tsx:110:44
    at Ur (react-dom.production.min.js:169:46)
    at _1 (react-dom.production.min.js:170:451)
    at Object.useState (react-dom.production.min.js:186:528)
    at au.useState (react.production.min.js:25:394)
    at Qmu (ColumnsReordering.tsx:69:52)
    at Qo (react-dom.production.min.js:167:137)
    at _4 (react-dom.production.min.js:197:258)
    at T5 (react-dom.production.min.js:292:88)
    at P5 (react-dom.production.min.js:280:389)
```

PR #3353, opened 2023-10-02
Do not allow dropping text on the columns

amanmahajan7:
Fixes https://github.com/adazzle/react-data-grid/issues/3349

### DIFF
--- a/src/HeaderCell.tsx
+++ b/src/HeaderCell.tsx
@@ -39,6 +39,8 @@ export const resizeHandleClassname = css`
   }
 `;
 
+const cellDraggableClassname = 'rdg-cell-draggable';
+
 const cellDragging = css`
   opacity: 0.5;
 `;
@@ -50,6 +52,8 @@ const cellOver = css`
 `;
 
 const cellOverClassname = `rdg-cell-drag-over ${cellOver}`;
+
+const dragDropKey = 'column-key';
 
 type SharedHeaderRowProps<R, SR> = Pick<
   HeaderRowProps<R, SR, React.Key>,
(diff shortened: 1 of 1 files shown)


src/HeaderCell.tsx
```diff
@@ -201,7 +206,7 @@ export default function HeaderCell<R, SR>({
   }
 
   function onDragStart(event: React.DragEvent<HTMLDivElement>) {
-    event.dataTransfer.setData('text/plain', column.key);
+    event.dataTransfer.setData(dragDropKey, column.key);
```
amanmahajan7, line 209: https://web.dev/datatransfer/#custom-data-formats

src/HeaderCell.tsx
```diff
@@ -51,6 +53,8 @@ const cellOver = css`
 
 const cellOverClassname = `rdg-cell-drag-over ${cellOver}`;
 
+const dragDropKey = 'column-key';
```
amanmahajan7, line 56: We should probably make this unique for each grid instance

amanmahajan7: Using `useId` to generate uniqueKey for each grid

src/HeaderCell.tsx
```diff
@@ -218,10 +223,12 @@ export default function HeaderCell<R, SR>({
 
   function onDrop(event: React.DragEvent<HTMLDivElement>) {
     setIsOver(false);
-    const sourceKey = event.dataTransfer.getData('text/plain');
-    if (sourceKey !== column.key) {
-      event.preventDefault();
-      onColumnsReorder?.(sourceKey, column.key);
+    if (event.dataTransfer.types.includes(dragDropKey)) {
+      const sourceKey = event.dataTransfer.getData(dragDropKey);
```
amanmahajan7, line 227: I can probably add an undefined check instead but need to ignore the types
image

Issue #3350, opened 2023-09-29
Cell drag handle should not be visible for columns that are not editable


9at8:
## Describe the bug 

The cell drag handle (blue square) should not be visible if the column cannot be edited.

## To Reproduce 

1. Click on a cell that's not editable
2. Drag handle appears

Link to code example: https://adazzle.github.io/react-data-grid/#/all-features

## Expected behavior 

Cell drag handle should not appear.

## Environment

- `react-data-grid` version: same as the demo website
- `react`/`react-dom` version: same as the demo website

![drag-handle-uneditable](https://github.com/adazzle/react-data-grid/assets/8235156/c68b3589-1059-440a-bf10-753efe5caa66)

PR #3354, opened 2023-10-02
Hide drag handle if all the cells in a column are readonly

amanmahajan7:
Fixes https://github.com/adazzle/react-data-grid/issues/3350

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -839,6 +839,10 @@ function DataGrid<R, SR, K extends Key>(
     }
 
     const column = columns[selectedPosition.idx];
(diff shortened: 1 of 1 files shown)


src/DataGrid.tsx
```diff
@@ -839,6 +839,10 @@ function DataGrid<R, SR, K extends Key>(
     }
 
     const column = columns[selectedPosition.idx];
+    if (column.renderEditCell == null || column.editable === false) {
```
amanmahajan7, line 842: if `editable` is a function then we can check all the rows but I am not sure if it is worth

nstepien (approved):

PR #3355, opened 2023-10-02
Update dragHandle to display at the intersection of column and row

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -847,10 +848,13 @@ function DataGrid<R, SR, K extends Key>(
 
     return (
       <DragHandle
-        gridRowStart={headerAndTopSummaryRowsCount + selectedPosition.rowIdx + 1}
+        gridRowStart={headerAndTopSummaryRowsCount + rowIdx + 1}
         rows={rows}
         column={column}
         columnWidth={columnWidth}
+        rowHeight={getRowHeight(rowIdx)}
+        isLastColumn={column.idx === maxColIdx}
```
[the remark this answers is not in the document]
amanmahajan7: > What if we the cell spans to the end?

`place-self` fixes it I think. but we still need to handle colSpan on a frozen column 🤔

src/DragHandle.tsx
```diff
@@ -118,19 +122,26 @@ export default function DragHandle<R, SR>({
...
+    const isLastColumn = column.idx + colSpan - 1 === maxColIdx;
+
+    return {
+      ...style,
+      gridRowStart,
+      marginInlineEnd: isLastColumn ? undefined : marginEnd,
+      marginBlockEnd: isLastRow ? undefined : marginEnd,
+      insetInlineStart: insetInlineStart
+        ? `calc(${insetInlineStart} + ${columnWidth}px + var(--rdg-drag-handle-size) * -0.5 - 1px)`
+        : undefined
```
nstepien, line 138: What does this do exactly?

nstepien (approved):

src/DragHandle.tsx
amanmahajan7: It sets the correct left position for the drag handle when it is frozen.

PR #3356, opened 2023-10-05
bug: name collision in isCellEditable breaks edit feature

PenguinOfWar:
# Issue
Under certain conditions when bundled, two identically named functions (`isCellEditable`) can clash.

## How to replicate
1. Import the library into `next@13.5.4` or later
2. Configure an editable cell per the common example
3. `npm run dev` shows expected functionality of cell becoming editable after double click or carriage return
4. `npm run build && npm run start` shows an unexpected outcome, cell edit becomes unreachable in production

Going through the compiled code with some strategic logging shows that the production build version of the code prefers the utility `isCellEditable` rather than the function inside `DataGrid`. As they accept two different arguments sets, this breaks the edit feature.

## Fix
Renaming the utility to `isCellEditableUtil` resolves this issue. If this name isn't preferred, I am open to suggestion and happy to change it.

### DIFF
--- a/src/Cell.tsx
+++ b/src/Cell.tsx
@@ -2,7 +2,7 @@ import { memo } from 'react';
 import { css } from '@linaria/core';
 
 import { useRovingTabIndex } from './hooks';
-import { createCellEvent, getCellClassname, getCellStyle, isCellEditable } from './utils';
+import { createCellEvent, getCellClassname, getCellStyle, isCellEditableUtil } from './utils';
 import type { CellRendererProps } from './types';
 
 const cellCopied = css`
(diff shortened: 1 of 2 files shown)


nstepien:
Have you opened an issue in the Next.js repo?

DarrylWalkerJacobs:
> Have you opened an issue in the Next.js repo?

No, but I shall. However as the fix is a minor utility name change in the source this felt like the best way to get a quick fix into production and hopefully prevent it from happening on any other bundlers.

PenguinOfWar:
Logged a bug on next for visibility. vercel/next.js#56480

nstepien (approved):

nstepien:
Thanks!

PenguinOfWar:
Thanks for the merge @nstepien!

PR #3357, opened 2023-10-07
Resize column via keyboard

### LINKED ISSUE #3320: [Accessibility] Make columns resizable via keyboard
## Use case 

Non-mouse users should be able to resize columns using the keyboard.

## Proposed solution 

1. Add a keyboard shortcut (such as Ctrl + Alt + Right / Left) that nudges the column size up / down
    - Alternatively make the resize handle focusable and enable a similar shortcut (but maybe a simpler one)
3. Expose a column resize method via the API to allow consumers to enable such functionality

godon019:
Closes #3320


src/DataGrid.tsx
```diff
@@ -568,6 +569,19 @@ function DataGrid<R, SR, K extends Key>(
       }
     }
 
+    if (column.resizable && isCtrlKeyHeldDown(event) && event.shiftKey) {
```
godon019, line 572: need some insights here @amanmahajan7 @nstepien 
1. should we provide 'resizeViaKeyboard' flag? so that user can turn it on and off?
2. what is the best way to provide handler via API? not sure if I can use `onCellKeyDown`

godon019: I decided to go with using the `onCellKeyDown` API. but still not sure as this doesn't seem to be very intrinsic

src/DataGrid.tsx
```diff
@@ -525,18 +527,22 @@ function DataGrid<R, SR, K extends Key>(
 
   function handleKeyDown(event: KeyboardEvent<HTMLDivElement>) {
     const { idx, rowIdx, mode } = selectedPosition;
+    const column = columns[idx];
+
     if (mode === 'EDIT') return;
 
-    if (onCellKeyDown && isRowIdxWithinViewportBounds(rowIdx)) {
```
godon019, line 530: why do we need to check `isRowIdxWithinViewportBounds` before call `onCellKeyDown` ?

kycutler:
Hi folks, thanks for working on this! Please let me know if I can help at all.

One bit of feedback on the changes so far: ideally `resizeColumn` and `getColumnWidth` would be accessible from anywhere and not limited to specific handlers like `onCellKeyDown` so consumers have flexibility in how it is exposed (for example, column widths could be rendered within the grid and made editable via input elements). Perhaps the methods could be present on the `CalculatedColumn`s themselves, or exposed via imperative handle on the grid?

Thanks for all the work on this fantastic project!

src/DataGrid.tsx
amanmahajan7: I think we can always allow keyboard resizing if the column can be resized. I don't think we need an api to disable it unless there is a use case. We allow keyboard navigation on the header cell and can allow resizing when the cell is focused. 

- Check if the cell is focused and column is resizable 
- Check if a particular key combination is pressed (Ctrl + arrow right/left?)

Questions:
- Which key to use. Are there any grid examples we can find?
- Do we need some sort of timer if they keep combination is pressed and increment the width by 1px every few ms?

src/DataGrid.tsx
amanmahajan7: I think because row would be undefined otherwise. Same for summary row. We may have to add a type `HEADER | ROW | SUMMARY`

amanmahajan7:
Should we complete and merge this PR?

matanAlltra:
please merge this this behavior is an accessibility requirement

amanmahajan7:
Replaced by https://github.com/adazzle/react-data-grid/pull/3754

PR #3367, opened 2023-10-16
Enable `throwSuggestions` RTL configuration 

amanmahajan7:
https://testing-library.com/docs/dom-testing-library/api-configuration/#throwsuggestions-experimental

### DIFF
--- a/test/column/renderEditCell.test.tsx
+++ b/test/column/renderEditCell.test.tsx
@@ -16,25 +16,25 @@ describe('Editor', () => {
   it('should open editor on double click', async () => {
     render(<EditorTest />);
     await userEvent.click(getCellsAtRowIndex(0)[0]);
-    expect(screen.queryByLabelText('col1-editor')).not.toBeInTheDocument();
+    expect(screen.queryByRole('spinbutton', { name: 'col1-editor' })).not.toBeInTheDocument();
     await userEvent.dblClick(getCellsAtRowIndex(0)[0]);
-    expect(screen.getByLabelText('col1-editor')).toHaveValue(1);
+    expect(screen.getByRole('spinbutton', { name: 'col1-editor' })).toHaveValue(1);
     await userEvent.keyboard('2');
     await userEvent.tab();
-    expect(screen.queryByLabelText('col1-editor')).not.toBeInTheDocument();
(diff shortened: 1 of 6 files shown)


test/setup.ts
```diff
@@ -1,4 +1,7 @@
 import { act } from 'react-dom/test-utils';
+import { configure } from '@testing-library/react';
+
+configure({ throwSuggestions: true });
```
nstepien, line 4: >Duration  67.79s (transform 1.28s, setup 7.69s, collect 7.75s, tests 29.41s, environment 12.71s, prepare 2.69s)

The tests take over 1 minute now, I think it's fine to try it from time to time to improve tests, but I'm 👎 on keeping this enabled.

test/column/renderEditCell.test.tsx
```diff
@@ -16,25 +16,25 @@ describe('Editor', () => {
   it('should open editor on double click', async () => {
     render(<EditorTest />);
     await userEvent.click(getCellsAtRowIndex(0)[0]);
-    expect(screen.queryByLabelText('col1-editor')).not.toBeInTheDocument();
+    expect(screen.queryByRole('spinbutton', { name: 'col1-editor' })).not.toBeInTheDocument();
```
nstepien, line 19: What element are we actually querying here? I don't really understand what a spinbutton is

amanmahajan7: `input type=number`

test/setup.ts
amanmahajan7: How long does it take without this change?

nstepien: NVM I thought the GH CI was faster than that

nstepien: We can add it back

test/sorting.test.tsx
```diff
@@ -35,9 +35,9 @@ function setup() {
 }
 
 function testSortColumns(expectedValue: readonly SortColumn[]) {
-  expect(JSON.parse(screen.getByTestId('sortColumnsValue').textContent!)).toStrictEqual(
-    expectedValue
-  );
+  expect(
+    JSON.parse(screen.getByTestId('sortColumnsValue', { suggest: false }).textContent!)
+  ).toStrictEqual(expectedValue);
```
nstepien, line 40: Can we use `toHaveTextContent`?

Issue #3369, opened 2023-10-18
Console error display for the Row Grouping feature  (Cannot read properties of undefined (reading 'renderEditCell'))


odini85:
## Describe the bug 

When using the Row Grouping feature, if you click on the grouped element and press the command + c keys, the following error occurs in the console.

```
selectedCellUtils.ts:28 Uncaught TypeError: Cannot read properties of undefined (reading 'renderEditCell')
    at isCellEditable (selectedCellUtils.ts:28:12)
    at isSelectedCellEditable (selectedCellUtils.ts:23:10)
```

This issue is also reproducible on the Row Grouping example page(https://adazzle.github.io/react-data-grid/#/row-grouping)

![image](https://github.com/adazzle/react-data-grid/assets/29395995/4603f97a-da54-4801-bd70-ab30329a078a)

## To Reproduce 

1. Set up according to the Row Grouping feature.
2. Click on the toggle area that is grouped.
3. Press the "command" + "c" keys on the keyboard.
4. An error message is displayed.

Link to code example: https://adazzle.github.io/react-data-grid/#/row-grouping

## Environment

- `react-data-grid` version: ^7.0.0-beta.40
- `react`/`react-dom` version: ^18.2.0

PR #3372, opened 2023-10-20
Fix `undefined` error due to `columnIdx` out of bounds 

amanmahajan7:
Closes https://github.com/adazzle/react-data-grid/issues/3369

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -675,10 +675,6 @@ function DataGrid<R, SR, K extends Key>(
     return idx >= minColIdx && idx <= maxColIdx;
   }
 
-  function isColIdxWithinEditBounds(idx: number) {
-    return idx >= 0 && idx <= maxColIdx;
-  }
-
   function isRowIdxWithinViewportBounds(rowIdx: number) {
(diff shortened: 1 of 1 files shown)


src/DataGrid.tsx
```diff
@@ -690,6 +690,7 @@ function DataGrid<R, SR, K extends Key>(
   function isCellEditable(position: Position): boolean {
     return (
       isCellWithinViewportBounds(position) &&
+      position.idx !== -1 &&
```
nstepien, line 693: `isCellWithinViewportBounds` doesn't catch this? 🤔 
Should we add a test?

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: It uses https://github.com/adazzle/react-data-grid/blob/main/src/DataGrid.tsx#L674 and `minColIdx === -1`

amanmahajan7: May be we need to add a new function `withinEditBounds`

amanmahajan7: Not sure how to test it. I will check

nstepien (approved):

PR #3400, opened 2023-12-06
addition of enums in types.ts

vijaytembugade:
### Additions the enums for 
- DIRECTIONS, 
- SORT_DIRECTION,
- CELL_NAVIGATION_MODE,
- TYPES_ENUMS


abenhamdine (approved):

amanmahajan7:
Thank you for the PR but we would prefer to keep [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)

PR #3407, opened 2023-12-14
Linaria 6

amanmahajan7:
https://github.com/callstack/linaria/releases/tag/linaria%406.0.0
https://wyw-in-js.dev/


src/Cell.tsx
```diff
@@ -1,5 +1,5 @@
 import { memo } from 'react';
-import { css } from '@linaria/core';
+import { css } from '@wyw-in-js/core';
```
nstepien, line 2: Can't we keep these imports as `@linaria/core`? The package still exists

amanmahajan7: yeah, was just fixing it

amanmahajan7:
`styles.css` is the same but `bundle.js` is completely different. Were we packaging it incorrectly before?

image

nstepien:
There shouldn't be any remaining css`` in the output 🤔

amanmahajan7:
May be I incorrectly generated  the bundle in the main branch
image

amanmahajan7:
Anyway looks like the output is the same

package.json
```diff
@@ -64,8 +64,7 @@
     "@biomejs/biome": "1.4.1",
     "@faker-js/faker": "^8.0.0",
     "@ianvs/prettier-plugin-sort-imports": "^4.0.2",
-    "@linaria/rollup": "^5.0.3",
-    "@linaria/vite": "^5.0.3",
+    "@linaria/core": "^6.0.0",
```
amanmahajan7, line 67: This needs to be installed now

nstepien (approved):

PR #3417, opened 2024-01-04
Modify TreeDataGrid to use `column.frozen` value if present

jwasilgeo:
Hello maintainers 👋. I'd like to introduce the ability to toggle on/off frozen group columns in the `TreeDataGrid`. Please let me know if there is anything else I can do in this PR to make it ready for review and your consideration to merge in to this library. Thanks! 

This PR:

- Modifies the `TreeDataGrid` to rely on `column.frozen` values if present, rather than being hard-coded to `true` always.
- Updates the `RowGrouping` live demo to demonstrate this functionality with a checkbox to toggle it on/off for the grouping columns.

https://github.com/adazzle/react-data-grid/assets/4933392/30d48d90-afc9-4c38-bfce-3bc7297c3edd


src/TreeDataGrid.tsx
```diff
@@ -96,7 +96,7 @@ function TreeDataGrid<R, SR, K extends Key>(
         groupBy.push(column.key);
         columns[index] = {
           ...column,
-          frozen: true,
+          frozen: column.frozen ?? true,
```
amanmahajan7, line 99: What happens if the data is grouped on two columns and only the first column has `frozen: false`?

PR #3418, opened 2024-01-04
Remove `postcss-nested`

nstepien:
Some output diff:
image
image
image

- Layers order is now at the top of the stylesheet, which I believe is actually the correct way to use layers.
- We will now ship native CSS nesting

https://caniuse.com/css-nesting


amanmahajan7 (approved):
LGTM

PR #3423, opened 2024-01-11
Added mouse down and enter events for columns and cells

MrYogurt:
Added events for custom cells selecting in matrix case


amanmahajan7:
Is this still needed? What is the use case of these events? Btw grid now supports `onCellMouseDown`
https://github.com/adazzle/react-data-grid/blob/main/src/DataGrid.tsx#L190

PR #3431, opened 2024-01-17
Allow copying highlighted text and prevent editing when pressing ctrl+input key

jj49411:
It should prevent editing when we press
- ctrl+c on highlighted text
- ctrl+input key, apart from ctrl+v

Before
Recording 2024-01-17 at 15 09 08

After
Recording 2024-01-17 at 15 08 14


src/DataGrid.tsx
```diff
@@ -556,8 +562,6 @@ function DataGrid<R, SR, K extends Key>(
     ) {
       // event.key may differ by keyboard input language, so we use event.keyCode instead
       // event.nativeEvent.code cannot be used either as it would break copy/paste for the DVORAK layout
```
nstepien, line 564: Let's move these comments above `const cKey`

src/DataGrid.tsx
```diff
@@ -548,6 +548,12 @@ function DataGrid<R, SR, K extends Key>(
     if (!isCellEvent && !isRowEvent) return;
 
     const { keyCode } = event;
+    const cKey = 67;
+    const vKey = 86;
+
+    if (isCtrlKeyHeldDown(event) && keyCode === cKey && window.getSelection()?.toString() !== '') {
```
nstepien, line 554: Can/should we check that the selection starts/ends in the cell?
https://developer.mozilla.org/en-US/docs/Web/API/Selection#instance_properties
I wonder if we can do something like `cell.contains(selection.anchorNode)` for example.

How should we handle the `isRowEvent` case?

test/copyPaste.test.tsx
```diff
@@ -200,3 +200,10 @@ test('should not allow paste on header or summary cells', async () => {
   expect(getSelectedCell()).toHaveTextContent('s1');
   expect(onPasteSpy).not.toHaveBeenCalled();
 });
+
+test('should not allow editing when pressing ctrl+<input key>', async () => {
```
nstepien, line 204:
```suggestion
test('should not start editing when pressing ctrl+<input key>', async () => {
```

src/DataGrid.tsx
jj49411: > Can/should we check that the selection starts/ends in the cell?

Why do we need to check this? When the text selection is across multiple cells, it doesn't open the editor and performs the browser behaviour.
Recording 2024-01-18 at 10 37 19

nstepien: I can click on a cell, select text elsewhere without losing focus on the cell, then press ctrl+c, what should happen then? Right now we open the selected cell and copy the selected cell's text.

Recording 2024-01-18 at 11 19 14

I can also select text in a cell, tab to the next cell, and the selection will remain on the previous cell. Same issue as above.

Recording 2024-01-18 at 11 17 41

@amanmahajan7 any ideas?

jj49411: For both scenarios, it doesn't open the editor and copies the selected text, is it not desirable?
Recording 2024-01-18 at 11 37 36

nstepien: Mmh, yes that seems correct, maybe I'm confusing myself.

jj49411: We don't have special handling for `isRowEvent` and it copies text as expected, is there anything we need to consider?

src/DataGrid.tsx
```diff
@@ -548,16 +548,20 @@ function DataGrid<R, SR, K extends Key>(
     if (!isCellEvent && !isRowEvent) return;
 
     const { keyCode } = event;
+    // event.key may differ by keyboard input language, so we use event.keyCode instead
+    // event.nativeEvent.code cannot be used either as it would break copy/paste for the DVORAK layout
+    const cKey = 67;
+    const vKey = 86;
+
+    if (isCtrlKeyHeldDown(event) && keyCode === cKey && window.getSelection()?.toString() !== '') {
```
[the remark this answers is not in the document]
jj49411: Do you mean checking `window.getSelection() !== null`? When a cell is focused without highlighted text, `window.getSelection()` returns a selection object, if we check for `null`, it will just return and the cell won't be copied.

src/DataGrid.tsx
```diff
@@ -548,16 +548,20 @@ function DataGrid<R, SR, K extends Key>(
...
       selectedCellIsWithinViewportBounds &&
       (onPaste != null || onCopy != null) &&
       isCtrlKeyHeldDown(event)
     ) {
-      // event.key may differ by keyboard input language, so we use event.keyCode instead
-      // event.nativeEvent.code cannot be used either as it would break copy/paste for the DVORAK layout
-      const cKey = 67;
-      const vKey = 86;
       if (keyCode === cKey) {
         handleCopy();
```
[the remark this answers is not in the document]
jj49411: If I understand correctly, `selectedCellIsWithinViewportBounds` already checks if a cell is selected/focused, why do we need to check `window.selection`?

src/DataGrid.tsx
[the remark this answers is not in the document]
amanmahajan7: I meant 

```
const selection = window.getSelection();
if (isCtrlKeyHeldDown(event) && keyCode === cKey && selection !== null && selection.toString() !== '') {
```

`window.getSelection()?.toString()` would return `undefined` if `window.getSelection()` is `null` and the condition will be true, no?

src/DataGrid.tsx
```diff
@@ -568,6 +572,10 @@ function DataGrid<R, SR, K extends Key>(
       }
     }
 
+    if (isCtrlKeyHeldDown(event) && isDefaultCellInput(event) && keyCode !== vKey) {
```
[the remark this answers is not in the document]
jj49411: If we check `isDefaultCellInput(event)` here and return `false` when ctrl+v is pressed in `isDefaultCellInput`, it does not paste on the cell. Also, it still prevents selection on `ctrl + shift + space`. Or should we add a condition `&& keyCode !== spaceKey`?

src/DataGrid.tsx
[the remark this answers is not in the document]
jj49411: When `handleKeyDown` is triggered, `window.getSelection()` seems to always return a selection object 🤔

src/DataGrid.tsx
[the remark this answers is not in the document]
amanmahajan7: > If I understand correctly, selectedCellIsWithinViewportBounds already checks if a cell is selected/focused

`selectedCellIsWithinViewportBounds` checks cell selection which is a state inside RDG and it is different from `window.getSelection()`. I was thinking of doing this 

```
if (keyCode === cKey) {
  const selection = window.getSelection();
  if (selection !== null && selection.toString() !== '') return;
  handleCopy();
}
```

src/DataGrid.tsx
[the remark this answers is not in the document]
amanmahajan7: I meant modifying `isDefaultCellInput` directly

```
function isDefaultCellInput(event) {
  if (isCtrlKeyHeldDown(event) && keyCode !== vKey) return;
}
```

src/DataGrid.tsx
```diff
@@ -548,16 +548,27 @@ function DataGrid<R, SR, K extends Key>(
...
+    const cKey = 67;
+    const vKey = 86;
+
+    const selection = window.getSelection();
+    // pressing ctrl+c on highlighted text should copy text instead of starting editing
+    if (
+      isCtrlKeyHeldDown(event) &&
+      keyCode === cKey &&
+      selection !== null &&
+      selection.toString() !== ''
```
nstepien, line 562: 🤔 
https://developer.mozilla.org/en-US/docs/Web/API/Selection/isCollapsed
```suggestion
      window.getSelection()?.isCollapsed === false
```

amanmahajan7 (approved):
LGTM. Wonder if there is a way to emulate selection and test the new behavior 🤔

nstepien (approved):

PR #3432, opened 2024-01-18
Fix a couple lint issues

nstepien:


website/demos/ColumnsReordering.tsx
```diff
@@ -102,10 +102,10 @@ export default function ColumnsReordering({ direction }: Props) {
     setColumnsOrder((columnsOrder) => {
       const sourceColumnOrderIndex = columnsOrder.findIndex(
         (index) => columns[index].key === sourceKey
-      )!;
```
amanmahajan7, line 105: What fixed them?

amanmahajan7 (approved):

website/demos/ColumnsReordering.tsx
nstepien: https://typescript-eslint.io/rules/no-unnecessary-type-assertion/
Fixed automatically

PR #3433, opened 2024-01-18
Use `:nth-child/:nth-last-child` selector to set frozen and summary row styles

amanmahajan7:


src/GroupRow.tsx
```diff
@@ -16,7 +16,8 @@ const groupRow = css`
       background-color: var(--rdg-header-background-color);
     }
 
-    > .${cell}:not(:last-child):not(.${cellFrozenLast}) {
+    > .${cell}:not(:last-child):not(.${cellFrozen}),
```
nstepien, line 19: https://developer.mozilla.org/en-US/docs/Web/CSS/:not#description
>You can negate several selectors at the same time. Example: `:not(.foo, .bar)` is equivalent to `:not(.foo):not(.bar)`.

```suggestion
    > .${cell}:not(:last-child, .${cellFrozen}),
```

src/style/cell.ts
```diff
@@ -35,15 +35,12 @@ export const cellFrozen = css`
...
-  }
-`;
-
-export const cellFrozenClassname = `rdg-cell-frozen ${cellFrozen}`;
 
-export const cellFrozenLast = css`
-  @layer rdg.Cell {
-    box-shadow: calc(2px * var(--rdg-sign)) 0 5px -2px rgba(136, 136, 136, 0.3);
+    /* Add box-shadow on the last frozen cell */
+    &:has(+ :not(&)) {
```
nstepien, line 40: Does this work if the last frozen cell is also the last cell in the row?

src/style/core.ts
```diff
@@ -88,6 +89,18 @@ const root = css`
         ${darkTheme}
       }
     }
+
+    :nth-last-child(1 of .${topSummaryRowClassname}) {
```
nstepien, line 93: Can we specify `>`?
```suggestion
    > :nth-last-child(1 of .${topSummaryRowClassname}) {
```

src/style/core.ts
```diff
@@ -88,6 +89,18 @@ const root = css`
...
+      > .${cell} {
+        border-block-end: 2px solid var(--rdg-summary-border-color);
+      }
+    }
+
+    :nth-child(1 of .${bottomSummaryRowClassname}) {
```
nstepien, line 99:
```suggestion
    > :nth-child(1 of .${bottomSummaryRowClassname}) {
```

src/style/cell.ts
amanmahajan7: It did not work. Fixed with `:nth-last-child`

src/style/core.ts
```diff
@@ -88,6 +89,23 @@ const root = css`
         ${darkTheme}
       }
     }
+
+    /* Add box-shadow on the last frozen cell */
+    :nth-last-child(1 of .${cellFrozen}) {
```
amanmahajan7, line 94: Added here as header does not use `row` class. I can instead add it to both `row` and `header-row` if you prefer

amanmahajan7: Added a variable instead. I think it is better to use more specific styles

src/style/row.ts
```diff
@@ -17,6 +19,11 @@ export const row = css`
         background-color: var(--rdg-row-selected-hover-background-color);
       }
     }
+
+    /* Add box-shadow on the last frozen cell */
+    > :nth-last-child(1 of .${cellFrozen}) {
```
nstepien, line 24: Nice!

src/style/core.ts
```diff
@@ -29,6 +31,7 @@ const darkTheme = `
   --rdg-row-hover-background-color: hsl(0deg 0% 9%);
   --rdg-row-selected-background-color: hsl(207deg 76% 42%);
   --rdg-row-selected-hover-background-color: hsl(207deg 76% 38%);
+  --rdg-cell-frozen-box-shadow: calc(2px * var(--rdg-sign)) 0 5px -2px rgba(136, 136, 136, 0.3);
```
nstepien, line 34: We don't need to define this twice if it has the same value. We can move it next to `--rdg-selection-color` and `--rdg-font-size`

src/style/row.ts
```diff
@@ -17,6 +19,11 @@ export const row = css`
...
+      box-shadow: var(--rdg-cell-frozen-box-shadow);
+    }
```
nstepien, line 26: Does this need to be part of the (Header)Row styles?
We can move it to cell styles, like
```css
&:nth-... {}
```

src/GroupRow.tsx
```diff
@@ -16,7 +16,8 @@ const groupRow = css`
       background-color: var(--rdg-header-background-color);
     }
 
-    > .${cell}:not(:last-child):not(.${cellFrozenLast}) {
+    > .${cell}:not(:last-child, .${cellFrozen}),
+    > .${cellFrozen}:has(+ .${cellFrozen}) {
```
nstepien, line 20: Does this work?
```suggestion
    > :nth-last-child(n + 1 of .${cellFrozen}) {
```

amanmahajan7: Does not work

nstepien: `n + 2` then? 🤔

src/style/row.ts
amanmahajan7:
```
&:nth-last-child(1 of &) {
    box-shadow: var(--rdg-cell-frozen-box-shadow);
 }
```
This works

src/GroupRow.tsx
amanmahajan7: This works. How does it work :) ?

nstepien: For all the `nth` selectors, you can use `n` with some arithmetic to select specific elements.
https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child
I don't know why `+ 2` is needed to skip the first match though.

nstepien (approved):
:shipit:

PR #3454, opened 2024-02-13
Add `--rdg-scroll-height` variable

amanmahajan7:
This can be useful to set grid height based on the number of rows. For example we can do something like

```
block-size: 100%;
max-height: var(--rdg-scroll-height);
```


src/hooks/useGridDimensions.ts
```diff
@@ -17,11 +18,13 @@ export function useGridDimensions() {
 
     const { clientWidth, clientHeight, offsetWidth, offsetHeight } = gridRef.current!;
     const { width, height } = gridRef.current!.getBoundingClientRect();
-    const initialWidth = width - offsetWidth + clientWidth;
+    const horizontalScrollbarWidth = offsetWidth - clientWidth;
```
nstepien, line 21: Isn't that the vertical scrollbar width instead?

src/DataGrid.tsx
```diff
@@ -412,6 +413,8 @@ function DataGrid<R, SR, K extends Key>(
   const maxColIdx = columns.length - 1;
   const selectedCellIsWithinSelectionBounds = isCellWithinSelectionBounds(selectedPosition);
   const selectedCellIsWithinViewportBounds = isCellWithinViewportBounds(selectedPosition);
+  const rowsHeight =
+    headerRowHeight + totalRowHeight + summaryRowsHeight + horizontalScrollbarWidth;
```
nstepien, line 417: Why do we add the **vertical** scrollbar **width** to the height?

src/hooks/useGridDimensions.ts
amanmahajan7: 🤦‍♂️

src/DataGrid.tsx
```diff
@@ -1068,6 +1070,7 @@ function DataGrid<R, SR, K extends Key>(
           gridTemplateRows: templateRows,
           '--rdg-header-row-height': `${headerRowHeight}px`,
           '--rdg-summary-row-height': `${summaryRowHeight}px`,
+          '--rdg-rows-height': `${rowsHeight}px`,
```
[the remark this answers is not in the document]
amanmahajan7: Or `--rdg-scroll-height`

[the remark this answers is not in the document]
nstepien: scroll height is good

src/hooks/useGridDimensions.ts
```diff
@@ -7,6 +7,7 @@ export function useGridDimensions() {
   const gridRef = useRef<HTMLDivElement>(null);
   const [inlineSize, setInlineSize] = useState(1);
   const [blockSize, setBlockSize] = useState(1);
+  const [horizontalScrollbarWidth, setHorizontalScrollbarWidth] = useState(0);
```
nstepien, line 10: width -> height

nstepien (approved):

PR #3484, opened 2024-04-08
Fix styles on last selected row in TreeDataGrid

Fi1osof:
If used TreeDataGrid and select last row, have padding and scroll with :before element. 
image


src/DataGrid.tsx
```diff
@@ -1196,7 +1196,10 @@ function DataGrid<R, SR, K extends Key>(
               selectedPosition.rowIdx
             ),
             [rowSelected]: isGroupRowFocused,
-            [rowSelectedWithFrozenCell]: isGroupRowFocused && lastFrozenColumnIndex !== -1
+            [rowSelectedWithFrozenCell]:
+              isGroupRowFocused &&
+              lastFrozenColumnIndex !== -1 &&
+              selectedPosition.rowIdx !== maxRowIdx
```
amanmahajan7, line 1202: This change breaks the selected row style when columns are scrolled. `:before` is used to set the left border 
image

amanmahajan7:
Closing due to inactivity

PR #3539, opened 2024-07-09
Vitest@2.0.1

amanmahajan7:
https://vitest.dev/guide/migration.html#migrating-to-vitest-2-0
https://github.com/vitest-dev/vitest/releases/tag/v2.0.0

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -39,7 +39,6 @@ export default defineConfig({
       include: ['src/**/*.{ts,tsx}', '!src/types.ts'],
       reporter: ['text', 'json']
     },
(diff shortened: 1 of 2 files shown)


vite.config.ts
```diff
@@ -39,7 +39,6 @@ export default defineConfig({
       include: ['src/**/*.{ts,tsx}', '!src/types.ts'],
       reporter: ['text', 'json']
     },
-    pool: 'forks',
```
amanmahajan7, line 42: https://vitest.dev/guide/migration.html#default-pool-is-forks

PR #3556, opened 2024-07-26
title for columns

djayavardhanreddy:
we need title for rows when we hover any row data


amanmahajan7:
This feature is not generic and cannot be accepted. I would recommend using custom `renderRow` renderer.

PR #3557, opened 2024-07-26
Use the new `light-dark` color function

amanmahajan7:
https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark


nstepien (approved):
Nice!

PR #3561, opened 2024-07-30
Remove `--rdg-sign` variable and use `dir` pseudo-class

amanmahajan7:
https://developer.mozilla.org/en-US/docs/Web/CSS/:dir

### DIFF
--- a/src/style/core.ts
+++ b/src/style/core.ts
@@ -52,7 +52,7 @@ const root = css`
     --rdg-font-size: 14px;
     --rdg-cell-frozen-box-shadow: 2px 0 5px -2px rgba(136, 136, 136, 0.3);
 
-    :dir(rtl) {
+    &:dir(rtl) {
       --rdg-cell-frozen-box-shadow: -2px 0 5px -2px rgba(136, 136, 136, 0.3);
     }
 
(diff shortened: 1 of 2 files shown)


src/style/core.ts
```diff
@@ -50,7 +50,11 @@ const root = css`
     ${lightTheme}
     --rdg-selection-color: #66afe9;
     --rdg-font-size: 14px;
-    --rdg-cell-frozen-box-shadow: calc(2px * var(--rdg-sign)) 0 5px -2px rgba(136, 136, 136, 0.3);
+    --rdg-cell-frozen-box-shadow: 2px 0 5px -2px rgba(136, 136, 136, 0.3);
+
+    :dir(rtl) {
```
nstepien, line 55:
```suggestion
    &:dir(rtl) {
```

nstepien (commented):
Does this work in all browsers now?
Is this the only place where we use `--rdg-sign`?

src/style/core.ts
amanmahajan7: It is supported in all the browsers 
> Baseline 2023
Newly available
Since December 2023, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers.

This is the only place where `--rdg-sign` variable is used

nstepien (approved):

PR #3566, opened 2024-08-02
Tweak checkbox style to fix shift + select row selection

amanmahajan7:


src/cellRenderers/renderCheckbox.tsx
```diff
@@ -19,7 +19,12 @@ const checkboxLabelClassname = `rdg-checkbox-label ${checkboxLabel}`;
 
 const checkboxInput = css`
   @layer rdg.CheckboxInput {
-    all: unset;
```
amanmahajan7, line 22: When the checkbox has `all: unset` style then for some reason `onChange` is not getting called on shift click

nstepien (approved):

PR #3567, opened 2024-08-02
Use native checkbox and use `accent-color`

amanmahajan7:
image

image

image

### DIFF
--- a/src/cellRenderers/renderCheckbox.tsx
+++ b/src/cellRenderers/renderCheckbox.tsx
@@ -3,12 +3,18 @@ import clsx from 'clsx';
 
 import type { RenderCheckboxProps } from '../types';
 
+const checkboxContainer = css`
+  @layer rdg.checkboxContainer {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    position: absolute;
+    inset: 0;
+  }
+`;
+
 const checkbox = css`
   @layer rdg.CheckboxInput {
-    position: absolute;
-    inset-block-start: 4px;
-    inset-inline-start: 3px;
-
     inline-size: 20px;
     block-size: 20px;
     cursor: pointer;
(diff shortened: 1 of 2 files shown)


src/cellRenderers/renderCheckbox.tsx
```diff
@@ -3,86 +3,55 @@ import clsx from 'clsx';
...
-  @layer rdg.CheckboxLabel {
-    cursor: pointer;
+const checkboxContainer = css`
+  @layer rdg.checkboxContainer {
     display: flex;
     align-items: center;
     justify-content: center;
     position: absolute;
     inset: 0;
-    margin-inline-end: 1px; /* align checkbox in row group cell */
```
amanmahajan7, line 14: Does not look like this is needed anymore?

src/cellRenderers/renderCheckbox.tsx
```diff
@@ -3,86 +3,55 @@ import clsx from 'clsx';
...
-        [checkboxLabelDisabledClassname]: props.disabled
-      })}
-    >
+    <div className={checkboxContainer}>
       <input
         type="checkbox"
         {...props}
-        className={checkboxInputClassname}
+        className={clsx(checkboxClassname, {
+          [checkboxDisabledClassname]: props.disabled
```
nstepien, line 51: We don't need this anymore, we can use the disabled pseudo selector.

src/cellRenderers/renderCheckbox.tsx
```diff
@@ -1,88 +1,44 @@
...
-    >
-      <input
-        type="checkbox"
-        {...props}
-        className={checkboxInputClassname}
-        onChange={handleChange}
-      />
-      <div className={checkboxClassname} />
-    </label>
+    <div className={checkboxContainer}>
```
nstepien, line 40: I wonder if we should make `checkboxContainer` a column class name, instead of rendering an extra absolutely-positioned div

nstepien (approved):

src/cellRenderers/renderCheckbox.tsx
amanmahajan7: We probably can but it requires updating many columns. May be this is okay for now

PR #3568, opened 2024-08-02
Add ability to deselect multiple rows using `Shift + Click`

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -514,21 +514,27 @@ function DataGrid<R, SR, K extends Key>(
...
-          const row = rows[i];
+    } else {
+      newSelectedRows.delete(rowKey);
+    }
+
+    if (isShiftClick && previousRowIdx !== -1 && previousRowIdx !== rowIdx) {
+      const step = sign(rowIdx - previousRowIdx);
+      for (let i = previousRowIdx + step; i !== rowIdx; i += step) {
+        const row = rows.at(i);
+        if (row === undefined) continue;
```
amanmahajan7, line 531: Added this extra check in case the previous row does not exist. Is there a better way?

nstepien: If the previous row does not exist I'd assume we don't want to apply the shift+click behavior?
We could check `rows.length` no?

amanmahajan7: You are right. I made that change

nstepien (approved):

Issue #3535, opened 2024-06-24
Columns are sorted when space key is down during typing in filters


Kosssma:
## Describe the bug 

When user type key word in filter when he hit space instead of space char as next sign he trigger column sorting functionality. User can not enter space from keyboard, he can only paste it into filter to be able to use it.

## To Reproduce 

1. Create a header with a filter like the the [Header filters](example https://adazzle.github.io/react-data-grid/#/header-filters)
2. Enable sorting like in the [Columns reordering[(example https://adazzle.github.io/react-data-grid/#/columns-reordering)

Link to code example:

## Expected behavior 
when typing in filter sign user hit space this should not trigger column sorting 

## Environment

- `react-data-grid` version: ^7.0.0-beta.44
- `react`/`react-dom` version: ^18.2.0

## Additional context 

code line responsible for effect:
https://github.com/adazzle/react-data-grid/blob/b667cec545666308ec59d74580b695e344380379/src/HeaderCell.tsx#L280
https://github.com/adazzle/react-data-grid/blob/9aa46269793cda576f60d2c6fb03d250028444a3/src/HeaderCell.tsx#L202

ronneldavis:
This seems to be related to this issue as well: https://github.com/adazzle/react-data-grid/issues/3439

holaChaitanya:
![image](https://github.com/user-attachments/assets/cf8240ee-c308-46b0-9cea-0d7bd6dec449)

why not try something like - stop event propagation inside the `renderHeaderCell` itself?

amanmahajan7:
The above recommended approach (using `stopPropagation`) is the right way to resolve this issue
https://github.com/adazzle/react-data-grid/blob/main/website/routes/HeaderFilters.lazy.tsx#L72-L82

PR #3574, opened 2024-08-11
fix: disable sorting of rows from header if interacting with input elements

holaChaitanya:
fixes [this](https://github.com/adazzle/react-data-grid/issues/3535) issue

how to recreate the bug -
- let's say you're putting an input element in `HeaderCell` component
- this input element was being used for filtering functionality as shown in [this](https://adazzle.github.io/react-data-grid/#/header-filters) example
- now, you also implemented sorting of rows from the column header
- when you try to filter out by typing in the input element and put an empty space, this will trigger sorting
- the above behaviour also occurs when you click inside an input element

fix - 
- checking if the click or keyDown events happened inside an input element, if yes, then not triggering the `onSort` function in `HeaderCell` component


amanmahajan7:
Closing due to inactivity

PR #3576, opened 2024-08-14
Tweak Vite config

nstepien:


vite.config.ts
```diff
@@ -19,38 +20,39 @@ const resizeColumn: BrowserCommand<[resizeBy: number]> = async (context, resizeB
...
-  base: isCI ? '/react-data-grid/' : '/',
+export default defineConfig(({ command }) => ({
+  base: '/react-data-grid/',
   cacheDir: '.cache/vite',
+  clearScreen: false,
   build: {
-    emptyOutDir: true,
-    sourcemap: true
+    target: browserslist().map((version) => version.replace(' ', '')),
+    modulePreload: { polyfill: false },
```
nstepien, line 29: This will optimize the build output

vite.config.ts
```diff
@@ -19,38 +20,39 @@ const resizeColumn: BrowserCommand<[resizeBy: number]> = async (context, resizeB
   await page.mouse.up();
 };
 
-export default defineConfig({
-  base: isCI ? '/react-data-grid/' : '/',
+export default defineConfig(({ command }) => ({
+  base: '/react-data-grid/',
```
nstepien, line 24: Keeping it consistent between dev/prod

vite.config.ts
```diff
@@ -19,38 +20,39 @@ const resizeColumn: BrowserCommand<[resizeBy: number]> = async (context, resizeB
...
   },
   test: {
-    root: '.',
     globals: true,
     coverage: {
       provider: 'istanbul',
       enabled: isCI,
-      include: ['src/**/*.{ts,tsx}', '!src/types.ts'],
-      reporter: ['text', 'json']
+      include: ['src/**/*.{ts,tsx}'],
```
nstepien, line 54: Vitest should now handle type-only files well.

vite.config.ts
amanmahajan7: Which polyfills were getting included in the build?

amanmahajan7 (approved):

vite.config.ts
nstepien: Nothing in the RDG output, but I remember seeing differences in other projects.
I think it'd matter if we had lazy-loaded modules, so it'd use native methods to pre-load split chunks.

PR #3577, opened 2024-08-15
Add `isRowSelectionDisabled` prop

amanmahajan7:
This prop can be used to disable checkboxes 

image

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -371,19 +371,25 @@ function DataGrid<R, SR, K extends Key>(
 
   const headerSelectionValue = useMemo((): HeaderRowSelectionContextValue => {
     // no rows to select = explicitely unchecked
-    let isRowSelected = false;
-    let isIndeterminate = false;
-    const { length } = rows;
+    let hasSelectedRow = false;
+    let hasUnselectedRow = false;
 
-    if (length !== 0 && selectedRows != null && rowKeyGetter != null && selectedRows.size > 0) {
-      const selectedRowsCount = rows.reduce((count, row) => {
-        return selectedRows.has(rowKeyGetter(row)) ? count + 1 : count;
-      }, 0);
-      isRowSelected = selectedRowsCount === length;
-      isIndeterminate = !isRowSelected && selectedRowsCount > 0;
+    if (rowKeyGetter != null && selectedRows != null && selectedRows.size > 0) {
+      for (const row of rows) {
+        if (selectedRows.has(rowKeyGetter(row))) {
+          hasSelectedRow = true;
+        } else {
+          hasUnselectedRow = true;
+        }
+
+        if (hasSelectedRow && hasUnselectedRow) break;
+      }
     }
 
-    return { isRowSelected, isIndeterminate };
+    return {
+      isRowSelected: hasSelectedRow && !hasUnselectedRow,
+      isIndeterminate: hasSelectedRow && hasUnselectedRow
+    };
   }, [rows, selectedRows, rowKeyGetter]);
 
   const {
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -151,7 +151,7 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
   /** Set of selected row keys */
   selectedRows?: Maybe<ReadonlySet<K>>;
   /** Determines if a row can be selected */
-  isRowSelectable?: Maybe<(row: NoInfer<R>) => boolean>;
+  isRowSelectionDisabled?: Maybe<(row: NoInfer<R>) => boolean>;
   /** Function called whenever row selection is changed */
   onSelectedRowsChange?: Maybe<(selectedRows: Set<NoInfer<K>>) => void>;
   /** Used for multi column sorting */
@@ -230,7 +230,7 @@ function DataGrid<R, SR, K extends Key>(
     summaryRowHeight: rawSummaryRowHeight,
     // Feature props
     selectedRows,
-    isRowSelectable,
+    isRowSelectionDisabled,
     onSelectedRowsChange,
     sortColumns,
     onSortColumnsChange,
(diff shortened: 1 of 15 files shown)


src/DataGrid.tsx
```diff
@@ -363,16 +368,19 @@ function DataGrid<R, SR, K extends Key>(
...
-    return (
-      length !== 0 &&
-      selectedRows != null &&
-      rowKeyGetter != null &&
-      selectedRows.size >= length &&
-      rows.every((row) => selectedRows.has(rowKeyGetter(row)))
-    );
+
+    if (length !== 0 && selectedRows != null && rowKeyGetter != null && selectedRows.size > 0) {
+      const selectedRowsCount = rows.filter((row) => selectedRows.has(rowKeyGetter(row))).length;
```
nstepien, line 378: `reduce` would be a good fit here, instead of `filter`

src/DataGrid.tsx
```diff
@@ -149,6 +152,7 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
   selectedRows?: Maybe<ReadonlySet<K>>;
   /** Function called whenever row selection is changed */
   onSelectedRowsChange?: Maybe<(selectedRows: Set<NoInfer<K>>) => void>;
+  isRowSelectable?: Maybe<(row: NoInfer<R>) => boolean>;
```
nstepien, line 155: I wonder if this should be named `isRowSelectionDisabled/Enabled` instead.
For example, a row can be selected, but toggling it may be disabled/read-only.

src/DataGrid.tsx
```diff
@@ -1008,6 +1021,7 @@ function DataGrid<R, SR, K extends Key>(
           rowIdx,
           row,
           viewportColumns: rowColumns,
+          isRowSelectable: isRowSelectable?.(row),
```
nstepien, line 1024: Could enforce a boolean/default.
```suggestion
          isRowSelectable: isRowSelectable?.(row) ?? true,
```

src/cellRenderers/renderCheckbox.tsx
```diff
@@ -30,14 +30,28 @@ const checkbox = css`
...
 
-export function renderCheckbox({ onChange, ...props }: RenderCheckboxProps) {
+function Checkbox({ onChange, indeterminate, ...props }: RenderCheckboxProps) {
   function handleChange(e: React.ChangeEvent<HTMLInputElement>) {
-    onChange(e.target.checked, (e.nativeEvent as MouseEvent).shiftKey);
+    onChange(indeterminate || e.target.checked, (e.nativeEvent as MouseEvent).shiftKey);
   }
 
   return (
     <div className={checkboxContainer}>
```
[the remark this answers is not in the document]
amanmahajan7: Yeah, it is used to horizontally align checkbox

src/cellRenderers/renderCheckbox.tsx
```diff
@@ -30,14 +30,28 @@ const checkbox = css`
 
 const checkboxClassname = `rdg-checkbox-input ${checkbox}`;
 
-export function renderCheckbox({ onChange, ...props }: RenderCheckboxProps) {
+function Checkbox({ onChange, indeterminate, ...props }: RenderCheckboxProps) {
```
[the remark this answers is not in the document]
amanmahajan7: I was initially use using `useLayoutEffect` to set the `indeterminate`

src/cellRenderers/renderCheckbox.tsx
```diff
@@ -30,14 +30,28 @@ const checkbox = css`
...
   function handleChange(e: React.ChangeEvent<HTMLInputElement>) {
-    onChange(e.target.checked, (e.nativeEvent as MouseEvent).shiftKey);
+    onChange(indeterminate || e.target.checked, (e.nativeEvent as MouseEvent).shiftKey);
   }
 
   return (
     <div className={checkboxContainer}>
-      <input type="checkbox" {...props} className={checkboxClassname} onChange={handleChange} />
+      <input
+        ref={(el) => {
```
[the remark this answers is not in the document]
amanmahajan7: We can but then it needs to be changed to a component. I think it is okay to have an inline function but I can change it if you prefer

src/DataGrid.tsx
amanmahajan7: Maybe we rename it to `isRowSelectionDisabled`

test/rowSelection.test.tsx
```diff
@@ -161,6 +220,9 @@ test('extra keys are preserved when updating the selectedRows Set', async () =>
   await toggleSelection(1);
   expect(set).toStrictEqual(new Set([...initialSet, 2]));
 
+  await userEvent.click(headerCheckbox);
+  expect(set).toStrictEqual(initialSet);
```
amanmahajan7, line 224: All the rows are deselected when the indeterminate header checkbox is clicked. I think this behavior makes sense. Wdyt?

amanmahajan7: There are other options
1. We can keep track of number of selected rows and if all the selectable rows are selected then clicking on the header checkbox can unselect them otherwise we can select all selectable rows
2. We can also consider selectable rows to determine if the header checkbox is selected. So if all the selectable rows are selected then mark the header checkbox as selected instead of indeterminate

src/DataGrid.tsx
```diff
@@ -363,16 +369,21 @@ function DataGrid<R, SR, K extends Key>(
...
-      selectedRows.size >= length &&
-      rows.every((row) => selectedRows.has(rowKeyGetter(row)))
-    );
+
+    if (length !== 0 && selectedRows != null && rowKeyGetter != null && selectedRows.size > 0) {
+      const selectedRowsCount = rows.reduce((count, row) => {
+        return selectedRows.has(rowKeyGetter(row)) ? count + 1 : count;
+      }, 0);
+      isRowSelected = selectedRowsCount === length;
+      isIndeterminate = !isRowSelected && selectedRowsCount > 0;
```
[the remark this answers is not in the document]
amanmahajan7: Easier to understand 👍

nstepien (approved):

src/cellRenderers/renderCheckbox.tsx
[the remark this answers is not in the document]
nstepien: We didn't 😄

nstepien (approved):
:shipit:

PR #3578, opened 2024-08-16
Migrate from rollup to rolldown

nstepien:
https://rolldown.rs/
https://www.npmjs.com/package/rolldown
https://github.com/rolldown/rolldown


nstepien:
@amanmahajan7 rolldown seems to work great now!
I'll open a few issues upstream first.

package.json
```diff
@@ -55,21 +55,14 @@
     "clsx": "^2.0.0"
   },
   "devDependencies": {
-    "@babel/core": "^7.26.0",
```
amanmahajan7, line 58: nice!

tsconfig.js.json
```diff
@@ -4,8 +4,9 @@
     "allowJs": true,
     "module": "NodeNext",
     "moduleResolution": "NodeNext",
+    "resolveJsonModule": true,
     "skipLibCheck": true
   },
-  "include": ["**/*.js", ".github/**/*.js"],
+  "include": ["**/*.js", ".github/**/*.js", "package.json"],
```
amanmahajan7, line 10: Why do we include `package.json`?

rolldown.config.js
```diff
@@ -0,0 +1,49 @@
...
+    cssEntryFileNames: 'styles.css',
+    format: 'cjs',
+    sourcemap: true
+  }
+];
+
+export default outputs.map((output) =>
+  defineConfig({
+    input: './src/index.ts',
+    output,
```
amanmahajan7, line 31: output cannot be an array?

amanmahajan7 (commented):
Is the bundle same as before?

tsconfig.js.json
nstepien: `rolldown.config.js` imports it, and we typecheck it (via `// @ts-check`), so the file must be included for typechecking to work correctly.
Otherwise:
```
File 'd:/repos/react-data-grid/package.json' is not listed within the file list of project 'd:/repos/react-data-grid/tsconfig.js.json'. Projects must list all files or use an 'include' pattern.
```

rolldown.config.js
nstepien: It couldn't before, it can be now. (https://github.com/rolldown/rolldown/issues/3200)

nstepien:
>Is the bundle same as before?

I've compared the bundle output after running prettier on the output files:
- no changes in CSS output
- we lost PURE annotations (https://github.com/rolldown/rolldown/issues/3201) in js files
- we lost `babel-plugin-optimize-clsx`
- otherwise I couldn't find any really meaningful differences, just some braces or formatting changes and other smaller stuff.

amanmahajan7 (approved):
Maybe we can manually optimize clsx calls. Not sure how much difference it makes in performance though

nstepien:
Probably not worth the effort.

Issue #3580, opened 2024-08-21
When clicking on the table header, onSelectedCellChange is called with row as undefined


Bartmr:
## Describe the bug 

When clicking on the table header that is not sortable, `onSelectedCellChange` is called with `row` set as `undefined`

## To Reproduce 

1. Click on a table header cell that is not sortable
2. log the arguments coming in `onSelectedCellChange`

Link to code example:

![Screenshot from 2024-08-21 11-29-44](https://github.com/user-attachments/assets/99b34953-454d-470a-8953-67156638e664)

```tsx
const columns: Column<Accountant>[] = [
    {
      key: "name",
      name: "Name",
      resizable: true,
      renderCell: (args) => {
        return args.row.name;
      },
    },
    {
      key: "email",
      name: "Email",
      resizable: true,

      renderCell: (args) => {
        return args.row.email;
      },
    },
  ];

return <ReactDataGrid
      className={`rdg-light`}
      rows={props.accountants}
      columns={columns}
      rowKeyGetter={(r) => r.id}
      onSelectedCellChange={(args) => {
        props.onAccountantSelect(args.row.id);
      }}
    />
```

## Expected behavior 

Either:
- don't call `onSelectedCellChange` when a table header is selected
- type `row` argument in `onSelectedCellChange` as possibly `undefined`

## Environment

- `react-data-grid` version: 7.0.0-beta.46
- `react`/`react-dom` version: 18

amanmahajan7:
Can you create a reproducible example?

amanmahajan7:
Closing due to lack of response.

PR #3584, opened 2024-08-22
Conditionally Include Row Data in `onSelectedCellChange` Callback

balaji-sivasakthi:
## Description
- updated the row type for `CellSelectArgs`
- conditionally include row data on the `onSelectedCellChange` callback

Ref #3580


amanmahajan7:
Thanks for the PR. This was fixed in https://github.com/adazzle/react-data-grid/pull/3641

PR #3593, opened 2024-09-05
Do not pass `isRowSelectionDisabled` prop to the div

amanmahajan7:
Fixes 

```
Warning: React does not recognize the `%s` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `%s` instead. If you accidentally passed it from a parent component, remove it from the DOM element.%s isRowSelectionDisabled isrowselectiondisabled 
```


src/GroupRow.tsx
```diff
@@ -42,6 +42,7 @@ function GroupedRow<R, SR>({
   gridRowStart,
   groupBy,
   toggleGroup,
+  isRowSelectionDisabled,
   ...props
```
amanmahajan7, line 46: Spreading props can be dangerous sometimes. Wonder if we should avoid it if possible

nstepien (approved):

PR #3594, opened 2024-09-05
Column resizing: handle coalesced pointer move events

nstepien:
Our column resize test are [flaky](https://github.com/adazzle/react-data-grid/actions/runs/10724678950/job/29740854268?pr=3592) in CI, which lead me to rethink this part of the codebase.

I think what's happening is that when we run the following code
```js
  await page.mouse.down();
  await page.mouse.move(x + resizeBy + 5, y);
  await page.mouse.up();
```
playwright may not wait for pointer events to be dispatched before executing the next command,
so `mouse.up()` would run before the next render frame/before the `pointermove` event is dispatched,
so the `pointermove` event may be coalesced into the `lostpointercapture` event.

An alternative could be to add pauses between each mouse commands,
but handling the final pointer position does make sense, and is the right thing to do.

I couldn't reproduce the flaky test locally, so this is all theoretical.

https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/getCoalescedEvents

### DIFF
--- a/src/HeaderCell.tsx
+++ b/src/HeaderCell.tsx
@@ -122,22 +122,37 @@ export default function HeaderCell<R, SR>({
     const headerCell = currentTarget.parentElement!;
     const { right, left } = headerCell.getBoundingClientRect();
     const offset = isRtl ? event.clientX - left : right - event.clientX;
+    let hasDoubleClicked = false;
 
     function onPointerMove(event: PointerEvent) {
-      const { right, left } = headerCell.getBoundingClientRect();
-      const width = isRtl ? right + offset - event.clientX : event.clientX + offset - left;
-      if (width > 0) {
-        onColumnResize(column, clampColumnWidth(width, column));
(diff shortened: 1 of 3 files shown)


src/HeaderCell.tsx
```diff
@@ -122,22 +122,37 @@ export default function HeaderCell<R, SR>({
...
+      let newWidth = isRtl ? right + offset - event.clientX : event.clientX + offset - left;
+      newWidth = clampColumnWidth(newWidth, column);
+      if (width > 0 && newWidth !== width) {
+        onColumnResize(column, newWidth);
       }
     }
 
-    function onLostPointerCapture() {
+    function onDoubleClick() {
+      hasDoubleClicked = true;
```
nstepien, line 137: I could have kept the `onDoubleClick` prop, and used a ref, but I'm not sure that'd be 100% safe, what if we fail to reset the ref in time in the `onLostPointerCapture` handler?

amanmahajan7 (approved):

PR #3613, opened 2024-10-08
Implement download without `react-dom/server`

amanmahajan7:


website/routes/CommonFeatures.lazy.tsx
```diff
@@ -342,61 +345,54 @@ function CommonFeatures() {
...
-      direction={direction}
-    />
-  );
+  function handleExportToCsv() {
+    flushSync(() => {
+      setIsExporting(true);
+    });
+
+    exportToCsv(gridRef.current!.element!, 'CommonFeatures.csv');
+    setIsExporting(false);
```
nstepien, line 354: Should we `flushSync` here too? If the export is synchronous then we can potentially avoid a split second of displaying the grid with virtualization disabled. Not a huge concern though.

If the export is async then we should `await` it.

nstepien (approved):

PR #3615, opened 2024-10-08
Update eslint deps

amanmahajan7:
**New plugin**
-  `eslint-plugin-react-hooks-extra`. https://eslint-react.xyz/
- `react-compiler/react-compiler`. https://react.dev/learn/react-compiler#installing-eslint-plugin-react-compiler


src/DataGrid.tsx
```diff
@@ -1061,7 +1065,6 @@ function DataGrid<R, SR, K extends Key>(
   // Reset the positions if the current values are no longer valid. This can happen if a column or row is removed
   if (selectedPosition.idx > maxColIdx || selectedPosition.rowIdx > maxRowIdx) {
     setSelectedPosition({ idx: -1, rowIdx: minRowIdx - 1, mode: 'SELECT' });
-    setDraggedOverRowIdx(undefined);
```
amanmahajan7, line 1064: Is this needed? We do set it to undefined after the drag operation 
https://github.com/adazzle/react-data-grid/blob/main/src/DragHandle.tsx#L103

package.json
```diff
@@ -62,37 +62,39 @@
...
+    "@eslint/compat": "^1.2.1",
     "@faker-js/faker": "^9.0.0",
     "@ianvs/prettier-plugin-sort-imports": "^4.0.2",
     "@linaria/core": "^6.0.0",
     "@microsoft/api-extractor": "^7.23.0",
     "@rollup/plugin-babel": "^6.0.3",
     "@rollup/plugin-node-resolve": "^15.1.0",
-    "@tanstack/react-router": "^1.57.13",
-    "@tanstack/router-plugin": "^1.57.13",
+    "@tanstack/react-router": "^1.70.0",
```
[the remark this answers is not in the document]
nstepien: Looks like the plugin needs to export `Plugin`, manually tweaking the declaration files fixes it.
Composite projects require declaration files to be emitted, and TS can't emit a declaration file without access to the `Plugin` type from the `@tanstack/eslint-plugin-router`.
Should open an issue.

src/DataGrid.tsx
```diff
@@ -917,18 +917,22 @@ function DataGrid<R, SR, K extends Key>(
     );
   }
 
+  function cancelEditing() {
+    setSelectedPosition(({ idx, rowIdx }) => ({ idx, rowIdx, mode: 'SELECT' }));
+  }
+
+  function closeEditor(shouldFocusCell: boolean) {
+    shouldFocusCellRef.current = shouldFocusCell;
```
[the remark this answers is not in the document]
amanmahajan7: I think we can completely remove the ref and use `flushSync`. It works but it seems like the checkbox checked state is incorrect when cell is clicked. Maybe a bug in React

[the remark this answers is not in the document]
amanmahajan7: https://github.com/facebook/react/issues/31358

src/DataGrid.tsx
amanmahajan7: Probably. I added it back for now. I will look into alternatives

src/DataGrid.tsx
```diff
@@ -1182,6 +1183,7 @@ function DataGrid<R, SR, K extends Key>(
               );
             })}
             <RowSelectionChangeProvider value={selectRowLatest}>
+              {/* eslint-disable-next-line react-compiler/react-compiler */}
```
amanmahajan7, line 1186: Don't know which line it is complaining about

src/DataGrid.tsx
```diff
@@ -338,15 +340,13 @@ function DataGrid<R, SR, K extends Key>(
...
+  const [prevSelectedPosition, setPrevSelectedPosition] = useState(selectedPosition);
 
   /**
    * refs
    */
-  const prevSelectedPosition = useRef(selectedPosition);
   const latestDraggedOverRowIdx = useRef(draggedOverRowIdx);
-  const lastSelectedRowIdx = useRef(-1);
   const focusSinkRef = useRef<HTMLDivElement>(null);
-  const shouldFocusCellRef = useRef(false);
```
amanmahajan7, line 349: Removed a few refs

src/DataGrid.tsx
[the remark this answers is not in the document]
amanmahajan7: Using a state now

src/DataGrid.tsx
```diff
@@ -1061,6 +1061,7 @@ function DataGrid<R, SR, K extends Key>(
   // Reset the positions if the current values are no longer valid. This can happen if a column or row is removed
   if (selectedPosition.idx > maxColIdx || selectedPosition.rowIdx > maxRowIdx) {
     setSelectedPosition({ idx: -1, rowIdx: minRowIdx - 1, mode: 'SELECT' });
+    // eslint-disable-next-line react-compiler/react-compiler
     setDraggedOverRowIdx(undefined);
```
nstepien, line 1065: Mmh, setting ref during render?

src/DataGrid.tsx
nstepien: Probably the `setDraggedOverRowIdx` call?

nstepien (approved):

src/DataGrid.tsx
amanmahajan7: I tried removing it and it was still showing the same warning

src/DataGrid.tsx
amanmahajan7: Yeah, need a workaround. We need ref to access the latest value. I will investigate

PR #3633, opened 2024-11-18
Migrate more test to use `vitest-browser-react` and `browserUserEvent`

amanmahajan7:


test/browser/column/key.test.ts
```diff
@@ -10,6 +10,6 @@ test('key is escaped in query selectors', () => {
   ];
 
   expect(() => {
-    setup({ columns, rows: [] });
+    setupNew({ columns, rows: [] });
   }).not.toThrow();
```
nstepien, line 14: Does it throw if we don't escape the key?

nstepien (approved):

test/browser/column/key.test.ts
amanmahajan7: It does 

image

PR #3639, opened 2024-11-20
Fix flaky tests

amanmahajan7:
These tests fail almost everytime

### DIFF
--- a/vitest.workspace.ts
+++ b/vitest.workspace.ts
@@ -23,8 +23,7 @@ const dragFill: BrowserCommand<[from: string, to: string]> = async (context, fro
   await frame.locator('.rdg-cell-drag-handle').hover();
   await page.mouse.down();
   const toCell = frame.getByRole('gridcell', { name: to });
-  const { x, y } = (await toCell.boundingBox())!;
-  await page.mouse.move(x, y);
+  await toCell.hover();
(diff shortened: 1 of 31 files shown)


vitest.workspace.ts
```diff
@@ -15,6 +15,19 @@ const resizeColumn: BrowserCommand<[resizeBy: number]> = async (context, resizeB
...
+// TODO: remove when `userEvent.pointer` is supported
+const dragFill: BrowserCommand<[from: string, to: string]> = async (context, from, to) => {
+  const page = context.page;
+  const frame = await context.frame();
+  await frame.getByRole('gridcell', { name: from }).click();
+  await frame.locator('.rdg-cell-drag-handle').hover();
+  await page.mouse.down();
+  const toCell = frame.getByRole('gridcell', { name: to });
+  const { x, y } = (await toCell.boundingBox())!;
+  await page.mouse.move(x, y);
```
nstepien, line 27: Does this work?
```suggestion
  await toCell.hover();
```

nstepien (approved):

vitest.workspace.ts
amanmahajan7: It does 👍

nstepien (approved):

PR #3641, opened 2024-11-22
Fix `onSelectedCellChange` arguments

amanmahajan7:


src/DataGrid.tsx
```diff
@@ -771,7 +771,7 @@ function DataGrid<R, SR, K extends Key>(
     if (onSelectedCellChange && !samePosition) {
       onSelectedCellChange({
         rowIdx: position.rowIdx,
-        row,
+        row: isRowIdxWithinViewportBounds(position.rowIdx) ? rows[position.rowIdx] : undefined,
         column: columns[position.idx]
```
amanmahajan7, line 775: Maybe we can change the type to include row type similar to 
https://github.com/adazzle/react-data-grid/blob/main/src/types.ts#L284

nstepien (approved):

PR #3642, opened 2024-11-22
Add `aria-description`

amanmahajan7:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description


nstepien (approved):
Do we need to do the same for TreeDataGrid?

amanmahajan7:
It forwards the props so should be good. I can add a test in the next PR

PR #3643, opened 2024-11-22
Add `data-cy` prop

amanmahajan7:
Looks like this can be [useful](https://github.com/adazzle/react-data-grid/pull/3113). I tried to allow any `data-attributes` but that can be error prone as spreading props in general is not safe


nstepien (approved):
Is this for cypress?

amanmahajan7:
> Is this for cypress?

yeah

PR #3671, opened 2024-12-16
implement right frozen

robert-luoqing:
I'm not override frozen to left | right | null, I just add rightFrozen property. It will not break original version.

### DIFF
--- a/src/hooks/useCalculatedColumns.ts
+++ b/src/hooks/useCalculatedColumns.ts
@@ -52,13 +52,15 @@ export function useCalculatedColumns<R, SR>({
   const defaultResizable = defaultColumnOptions?.resizable ?? false;
   const defaultDraggable = defaultColumnOptions?.draggable ?? false;
 
-  const { columns, colSpanColumns, lastFrozenColumnIndex, headerRowsCount } = useMemo((): {
+  const { columns, colSpanColumns, lastFrozenColumnIndex, rightFrozenColumnCount, headerRowsCount } = useMemo((): {
     readonly columns: readonly CalculatedColumn<R, SR>[];
     readonly colSpanColumns: readonly CalculatedColumn<R, SR>[];
     readonly lastFrozenColumnIndex: number;
+    readonly rightFrozenColumnCount: number;
     readonly headerRowsCount: number;
   } => {
     let lastFrozenColumnIndex = -1;
+    let rightFrozenColumnCount = 0;
     let headerRowsCount = 1;
     const columns: MutableCalculatedColumn<R, SR>[] = [];
 
@@ -85,6 +87,7 @@ export function useCalculatedColumns<R, SR>({
         }
 
         const frozen = rawColumn.frozen ?? false;
+        const rightFrozen = rawColumn.rightFrozen ?? false;
 
         const column: MutableCalculatedColumn<R, SR> = {
           ...rawColumn,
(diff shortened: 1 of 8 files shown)


eshwaruppala:
this is a necessary feature

ksparakis:
Whats needed to get this across the finish line? willing to put resources on it

src/types.ts
```diff
@@ -39,6 +39,8 @@ export interface Column<TRow, TSummaryRow = unknown> {
   readonly colSpan?: Maybe<(args: ColSpanArgs<TRow, TSummaryRow>) => Maybe<number>>;
   /** Determines whether column is frozen or not */
   readonly frozen?: Maybe<boolean>;
+  /** Determines whether column is right frozen or not */
+  readonly rightFrozen?: Maybe<boolean>;
```
amanmahajan7, line 43: I would rather not add a new prop that can cause conflicts. What happens when both `rightFrozen` and `frozen` props are set? We can instead change `frozen` to `boolean | 'left' | 'right'`.

src/hooks/useCalculatedColumns.ts
```diff
@@ -52,13 +52,15 @@ export function useCalculatedColumns<R, SR>({
   const defaultResizable = defaultColumnOptions?.resizable ?? false;
   const defaultDraggable = defaultColumnOptions?.draggable ?? false;
 
-  const { columns, colSpanColumns, lastFrozenColumnIndex, headerRowsCount } = useMemo((): {
+  const { columns, colSpanColumns, lastFrozenColumnIndex, rightFrozenColumnCount, headerRowsCount } = useMemo((): {
```
amanmahajan7, line 55: All the new functionality needs extensive tests

amanmahajan7 (changes requested):

PaloSP:
Hi @amanmahajan7 

This PR has been stuck since July 2025. Right-side column pinning is a feature I need — happy to take it over with a fresh PR.

Before I invest time:

1. Still want `frozen: boolean | 'left' | 'right'` instead of a new `rightFrozen` prop?
2. Tests: unit + playwright e2e — any patterns you want beyond what's in the repo?
3. Review bandwidth — is this on your radar or effectively deprioritized?

If there's genuine interest, I'll open a PR within ~2 weeks. If not, please say so — I'll plan accordingly.

Thanks.

amanmahajan7:
We will be happy to review the PR 

> Still want frozen: boolean | 'left' | 'right' instead of a new rightFrozen prop?

Yes. @nstepien wdyt?

> Tests: unit + playwright e2e — any patterns you want beyond what's in the repo?

We use vitest browser tests. Here is an example of testing frozen columns
https://github.com/Comcast/react-data-grid/blob/main/test/browser/column/frozen.test.ts

> Review bandwidth — is this on your radar or effectively deprioritized?

We can prioritize the review 

Few important notes
- This will affect virtualization logic so we should add more tests (https://github.com/Comcast/react-data-grid/blob/main/test/browser/virtualization.test.ts)
- We want to prevent the shadow logic (https://github.com/Comcast/react-data-grid/blob/main/src/style/core.ts#L96)
- Ensure RTL mode works as expected

nstepien:
> > Still want frozen: boolean | 'left' | 'right' instead of a new rightFrozen prop?
> 
> Yes. @nstepien wdyt?

I'd go with `frozen: boolean | 'start' | 'end'` so it's compatible with `direction="rtl"`

PaloSP:
Hi @robert-luoqing — I've opened #4034 as a fresh attempt at this feature, using the `frozen: boolean | 'start' | 'end'` API agreed with @amanmahajan7 and @nstepien. Full credit to your original work here, which informed the approach.

Feel free to close this PR if you'd like, or keep it open for reference — either way, thanks for getting the ball rolling on right-edge pinning.

amanmahajan7:
Replaced by https://github.com/Comcast/react-data-grid/pull/4034

PR #3694, opened 2025-01-24
feat[renderHeaderCell]: pass props into renderHeaderCell for controll…

PudovAlexey:
… mutating outside the lib


PudovAlexey:
allows developers to controll onColumnResize outside the lib

PR #3708, opened 2025-02-13
Drop support for CJS build output

nstepien:
Now that Node.js support `require(esm)`, I think it's time we drop support for CJS.
If users still need CJS, they'll have to bundle/transform react-data-grid themselves.

I've also added the `DataGrid` named export, which will make it easier for auto-completion to import it from react-data-grid.
Should we even keep the `default` export?


amanmahajan7 (dismissed):
I am okay with removing default export

amanmahajan7:
Not sure if RDG can be ever be imported using `React.Lazy`

> This code relies on [dynamic import(),](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) which might require support from your bundler or framework. Using this pattern requires that the lazy component you’re importing was exported as the default export.

But users can always add a wrapper

test/browser/utils.tsx
```diff
@@ -2,7 +2,7 @@ import { act, render, screen } from '@testing-library/react';
 import { page, userEvent } from '@vitest/browser/context';
 import { css } from '@linaria/core';
 
-import DataGrid from '../../src';
+import { DataGrid } from '../../src';
 import type { DataGridProps } from '../../src';
```
amanmahajan7, line 6: Can combine these

amanmahajan7 (approved):

test/browser/utils.tsx
nstepien: Some places are combined, some not, dunno if we want to prefer one over the other. 🤔 
Should be handled by prettier or something.

PR #3711, opened 2025-02-17
docs(readme): update Quick Start section to reflect named export

ademtonay:
This commit updates README.md Quick Start section to align with current package exports. Remove default export example as the package now uses named export.


nstepien (approved):

nstepien:
Thanks for the PR!

Issue #3723, opened 2025-02-26
Column size reset not re-measured


Wroud:
If the column supports resizing and the table is not flexing, there are cases when the column will change size when scrolling (because of a change in the content)

## To Reproduce 

https://stackblitz.com/edit/react-data-grid-bug-max-content?file=src%2FApp.tsx,src%2Findex.css&terminal=dev
double-click on the resize of the first column, then scroll and  try to double-click twice

## Expected behavior 

Column width set to `max-content` must be measured (actually after column width set to anything except number)

## Environment

- `react-data-grid` version: v7.0.0-beta.48
- `react`/`react-dom` version: react 18+

Seems like it's related to this part of the code:
https://github.com/adazzle/react-data-grid/blob/main/src/hooks/useColumnWidths.ts#L73-L78

After column width is [set to max-content](https://github.com/adazzle/react-data-grid/blob/main/src/HeaderCell.tsx#L138C31-L138C42), it may not be replaced with a measured pixel size

Wroud:
I've also noticed that after the first render, the grid uses default values that lead to dynamic size changes in the grid until any interaction (for example, scrolling). After that, column sizes will be fixed.

https://github.com/user-attachments/assets/5eed8970-d92e-419b-9982-b0f94c721d9a

(to reproduce, use 'auto' or 'max-content' as column width)

(I'm working on that bug, but it's not fixed yet)

PR #3724, opened 2025-02-26
fix: columns resize sync

Wroud:
#3723

This PR will fix column resize sync. As a side effect, you don't need to pass maxWidth programmatically. You can apply it as a style to the `data-measuring-cell` elements (so you don't need to check if the resize amount in between max/min size; the browser will handle it)


Wroud:
I found that setting `gridRef.current!.style.gridTemplateColumns` will break the react reconciliation mechanism, so props changes in this style will not be applied

Issue #3725, opened 2025-02-28
Column size not re-measured when table get rows (change from empty with columns, to not empty with rows and columns)


Wroud:
## Describe the bug 

When the table is rendered first, it measures the column size. If the table is rendered with some columns but empty rows after rows are added, it will not re-measure the column size.

## To Reproduce 

1. render the table with columns but with no rows. Set columns width to 'max-content'
2. add rows

Link to code example: https://stackblitz.com/edit/react-data-grid-bug-re-measurement?file=src%2FApp.tsx,src%2Fmain.tsx&terminal=dev

## Expected behavior 

Column width is re-measured when the table becomes empty or has rows.

## Environment

- `react-data-grid` version: main
- `react`/`react-dom` version: 19

amanmahajan7:
Can we use a key to reset the grid? 

```
<DataGrid key={columns.length} ... />
```

https://stackblitz.com/edit/react-data-grid-bug-re-measurement-9w6wljna?file=src%2FApp.tsx

Wroud:
yes we can, but it's more like a workaround

ZachSARAO:
> Can we use a key to reset the grid?
> 
> ```
> <DataGrid key={columns.length} ... />
> ```
> 
> https://stackblitz.com/edit/react-data-grid-bug-re-measurement-9w6wljna?file=src%2FApp.tsx

The table will re-render fully, meaning that if you have scrolled down and resize the table it will jump back to the top row

Wroud:
now can be solved by using `columnWidths` (just reset it when you need)

PR #3726, opened 2025-02-28
fix: remeasure columns when rows added

Wroud:
#3725


amanmahajan7:
Thank you for the PR but I think this may not be the best solution. It is common to have workflows where users are creating multiple rows and it can be a bad UX if columns resize everytime a row is added.  Would it work if we expose a method like `gridRef.current.remeasureColumns()` which essentially clears `measuredColumnWidths`

amanmahajan7:
> my suggestion is to remeasure columns when the grid changes from 0 rows to >=1 row and when it changes from >=1 row to 0 rows

This can be a bad UX where there is a search functionality and grid keeps toggling b/w 0 and >=1 rows. I think we will work on providing props like `columnWidths/onColumnWidthsChange` so user can reset width outside the grid

PR #3733, opened 2025-03-11
feat(render-header-cell): Support override defaultRenderHeaderCell from

### LINKED ISSUE #3720: Override default column.renderHeaderCell
Looking through the docs I don't see a way to override the default `renderHeaderCell` property for a column. Is this possible?

I do so there is a `renderers` prop on the Table object - but as far as I can tell that only allows for overriding row-level cells?

hadyelzayady:
Fixes #3720 by give the ability to override default header render from defaultColumnOptions as the following
```tsx
      <DataGrid
        columns={columns}
        rows={rows}
        direction={direction}
        defaultColumnOptions={{
            renderHeaderCell:(props)=>{
            return <div>{props.column.name}</div>
        }
        }}
        onCellKeyDown={handleCellKeyDown}
      />


```


src/DataGrid.tsx
```diff
@@ -93,7 +93,7 @@ interface EditCellState<R> extends Position {
 
 export type DefaultColumnOptions<R, SR> = Pick<
   Column<R, SR>,
-  'renderCell' | 'width' | 'minWidth' | 'maxWidth' | 'resizable' | 'sortable' | 'draggable'
+  'renderCell' | 'width' | 'minWidth' | 'maxWidth' | 'resizable' | 'sortable' | 'draggable' | 'renderHeaderCell'
```
nstepien, line 96: nitpick
```suggestion
  'renderCell' | 'renderHeaderCell' | 'width' | 'minWidth' | 'maxWidth' | 'resizable' | 'sortable' | 'draggable'
```

nstepien (commented):
Hi, thanks for the PR. Looks like the build is failing.

nstepien (approved):
Thank you for the PR!

hadyelzayady:
You're welcome! Glad to contribute

PR #3744, opened 2025-03-19
Fix double-click to auto-size columns

nstepien:
I don't think we need `hasDoubleClickedRef` in practice as the double click event happens after `lostpointercapture`, but it does happen before `lostpointercapture` in tests for some reason.

### DIFF
--- a/test/browser/column/resizable.test.tsx
+++ b/test/browser/column/resizable.test.tsx
@@ -1,4 +1,4 @@
-import { commands, userEvent } from '@vitest/browser/context';
+import { commands, page } from '@vitest/browser/context';
 
 import type { Column } from '../../../src';
 import { resizeHandleClassname } from '../../../src/HeaderCell';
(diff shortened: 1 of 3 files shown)


test/browser/column/resizable.test.tsx
```diff
@@ -36,9 +36,9 @@ async function resize({ column, resizeBy }: ResizeArgs) {
 }
 
 async function autoResize(column: Element) {
-  const resizeHandle = getResizeHandle(column);
+  const resizeHandle = page.elementLocator(getResizeHandle(column));
 
-  await userEvent.dblClick(resizeHandle);
+  await resizeHandle.dblClick();
```
amanmahajan7, line 41: Any reason for this change?

amanmahajan7 (approved):

test/browser/column/resizable.test.tsx
nstepien: I wanted to see if the order of events would reflect actual browser behavior, as `userEvent` does not perform real browser/cursor actions.

PR #3746, opened 2025-03-19
Fix column auto resize

amanmahajan7:
Fixes https://github.com/adazzle/react-data-grid/issues/3723

### DIFF
--- a/src/hooks/useColumnWidths.ts
+++ b/src/hooks/useColumnWidths.ts
@@ -1,5 +1,4 @@
-import { useLayoutEffect, useRef } from 'react';
-import { flushSync } from 'react-dom';
+import { useLayoutEffect, useRef, useState } from 'react';
 
 import type { CalculatedColumn, StateSetter } from '../types';
 import type { DataGridProps } from '../DataGrid';
@@ -16,6 +15,7 @@ export function useColumnWidths<R, SR>(
   setMeasuredColumnWidths: StateSetter<ReadonlyMap<string, number>>,
   onColumnResize: DataGridProps<R, SR>['onColumnResize']
 ) {
+  const [columnToAutoresize, setColumnToAutoResize] = useState<string | null>(null);
   const prevGridWidthRef = useRef(gridWidth);
   const columnsCanFlex: boolean = columns.length === viewportColumns.length;
   // Allow columns to flex again when...
@@ -26,7 +26,10 @@ export function useColumnWidths<R, SR>(
   const columnsToMeasure: string[] = [];
 
   for (const { key, idx, width } of viewportColumns) {
-    if (
+    if (key === columnToAutoresize) {
+      newTemplateColumns[idx] = 'max-content';
+      columnsToMeasure.push(key);
+    } else if (
       typeof width === 'string' &&
       (ignorePreviouslyMeasuredColumns || !measuredColumnWidths.has(key)) &&
       !resizedColumnWidths.has(key)
@@ -62,40 +65,53 @@ export function useColumnWidths<R, SR>(
 
       return hasChanges ? newMeasuredColumnWidths : measuredColumnWidths;
     });
+
+    if (columnToAutoresize !== null) {
+      setColumnToAutoResize(null);
+      setResizedColumnWidths((resizedColumnWidths) => {
+        const newResizedColumnWidths = new Map(resizedColumnWidths);
+        newResizedColumnWidths.set(
+          columnToAutoresize,
+          measureColumnWidth(gridRef, columnToAutoresize)!
+        );
+        return newResizedColumnWidths;
+      });
+    }
   }
 
   function handleColumnResize(column: CalculatedColumn<R, SR>, nextWidth: number | 'max-content') {
     const { key: resizingKey } = column;
-    const newTemplateColumns = [...templateColumns];
     const columnsToMeasure: string[] = [];
 
-    for (const { key, idx, width } of viewportColumns) {
-      if (resizingKey === key) {
-        const width = typeof nextWidth === 'number' ? `${nextWidth}px` : nextWidth;
-        newTemplateColumns[idx] = width;
-      } else if (columnsCanFlex && typeof width === 'string' && !resizedColumnWidths.has(key)) {
-        newTemplateColumns[idx] = width;
+    for (const { key, width } of viewportColumns) {
+      if (
+        resizingKey !== key &&
+        columnsCanFlex &&
+        typeof width === 'string' &&
+        !resizedColumnWidths.has(key)
+      ) {
         columnsToMeasure.push(key);
       }
     }
 
-    gridRef.current!.style.gridTemplateColumns = newTemplateColumns.join(' ');
-    const measuredWidth =
-      typeof nextWidth === 'number' ? nextWidth : measureColumnWidth(gridRef, resizingKey)!;
-
-    // TODO: remove
-    // need flushSync to keep frozen column offsets in sync
-    // we may be able to use `startTransition` or even `requestIdleCallback` instead
-    flushSync(() => {
-      setResizedColumnWidths((resizedColumnWidths) => {
-        const newResizedColumnWidths = new Map(resizedColumnWidths);
-        newResizedColumnWidths.set(resizingKey, measuredWidth);
-        return newResizedColumnWidths;
-      });
-      updateMeasuredWidths(columnsToMeasure);
+    setMeasuredColumnWidths((measuredColumnWidths) => {
+      const newMeasuredColumnWidths = new Map(measuredColumnWidths);
+      for (const columnKey of columnsToMeasure) {
+        newMeasuredColumnWidths.delete(columnKey);
+      }
+      return newMeasuredColumnWidths;
     });
 
-    onColumnResize?.(column, measuredWidth);
+    if (typeof nextWidth === 'number') {
+      setResizedColumnWidths((resizedColumnWidths) => {
+        const newResizedColumnWidths = new Map(resizedColumnWidths);
+        newResizedColumnWidths.set(resizingKey, nextWidth);
+        return newResizedColumnWidths;
+      });
+      onColumnResize?.(column, nextWidth);
+    } else {
+      setColumnToAutoResize(resizingKey);
+    }
   }
 
   return {


src/hooks/useColumnWidths.ts
```diff
@@ -40,62 +43,81 @@ export function useColumnWidths<R, SR>(
...
-          newMeasuredColumnWidths.set(key, measuredWidth);
+    if (columnToAutoResize !== null) {
+      setColumnToAutoResize(null);
+      setResizedColumnWidths((resizedColumnWidths) => {
+        const oldWidth = resizedColumnWidths.get(columnToAutoResize);
+        const newWidth = measureColumnWidth(gridRef, columnToAutoResize);
+        if (newWidth !== undefined && oldWidth !== newWidth) {
+          const newResizedColumnWidths = new Map(resizedColumnWidths);
+          newResizedColumnWidths.set(columnToAutoResize, newWidth);
+          onColumnResize?.(viewportColumns.find((c) => c.key === columnToAutoResize)!, newWidth);
```
amanmahajan7, line 77: Not sure if calling prop like this can be an issue 🤔

src/hooks/useColumnWidths.ts
```diff
@@ -40,62 +43,81 @@ export function useColumnWidths<R, SR>(
...
+        !resizedColumnWidths.has(key)
+      ) {
         columnsToMeasure.push(key);
       }
     }
 
-    gridRef.current!.style.gridTemplateColumns = newTemplateColumns.join(' ');
-    const measuredWidth =
-      typeof nextWidth === 'number' ? nextWidth : measureColumnWidth(gridRef, resizingKey)!;
+    if (columnsToMeasure.length > 0) {
```
nstepien, line 101: Why only if we have columns to measure?

test/browser/column/resizable.test.tsx
```diff
@@ -66,48 +66,52 @@ test('cannot not resize or auto resize column when resizable is not specified',
 test('should resize column when dragging the handle', async () => {
   const onColumnResize = vi.fn();
   setup<Row, unknown>({ columns, rows: [], onColumnResize });
-  const [, col2] = getHeaderCells();
   const grid = getGrid();
   expect(onColumnResize).not.toHaveBeenCalled();
   await expect.element(grid).toHaveStyle({ gridTemplateColumns: '100px 200px' });
+  const [, col2] = getHeaderCells();
```
nstepien, line 72: I don't understand, surely the cell element should still be in the document if we get it earlier, no?

test/browser/column/resizable.test.tsx
```diff
@@ -66,48 +66,52 @@ test('cannot not resize or auto resize column when resizable is not specified',
...
+    ],
+    onColumnResize
   });
-  const [, col2] = getHeaderCells();
   const grid = getGrid();
   await expect.element(grid).toHaveStyle({ gridTemplateColumns: '100px 200px' });
+  const [, col2] = getHeaderCells();
   await autoResize(col2);
   await expect.element(grid).toHaveStyle({ gridTemplateColumns: '100px 327.703px' });
+  // This is called twice in strict mode
```
nstepien, line 113: Should it? Do we not check to only call the function if the width has changed?

nstepien (commented):
Found a bug, when I add `maxWidth: 150,` to the country column, the grid isn't filled, then I try to reduce its width, and the other columns fill the empty space.
Recording 2025-03-20 at 18 09 12

src/hooks/useColumnWidths.ts
```diff
@@ -40,62 +43,81 @@ export function useColumnWidths<R, SR>(
...
+        }
+        return newMeasuredColumnWidths;
+      });
+    }
 
-    // TODO: remove
-    // need flushSync to keep frozen column offsets in sync
-    // we may be able to use `startTransition` or even `requestIdleCallback` instead
-    flushSync(() => {
+    if (typeof nextWidth === 'number') {
```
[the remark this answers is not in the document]
amanmahajan7: It seems like it does. I will double check 
https://github.com/adazzle/react-data-grid/blob/main/src/HeaderCell.tsx#L128

test/browser/column/resizable.test.tsx
amanmahajan7: It does call it twice. May be the state updates after the `useLayoutEffect` has run twice 
https://github.com/adazzle/react-data-grid/blob/8a6bb2873ccf76cd340cd6139b0278d4426c1924/src/hooks/useColumnWidths.ts#L69

src/hooks/useColumnWidths.ts
```diff
@@ -40,62 +43,81 @@ export function useColumnWidths<R, SR>(
 
   useLayoutEffect(() => {
     prevGridWidthRef.current = gridWidth;
-    updateMeasuredWidths(columnsToMeasure);
+    updateMeasuredAndResizedWidths();
   });
 
-  function updateMeasuredWidths(columnsToMeasure: readonly string[]) {
-    if (columnsToMeasure.length === 0) return;
+  function updateMeasuredAndResizedWidths() {
```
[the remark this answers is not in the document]
amanmahajan7: Hmm, shows a warning. Looks like eslint only finds this issue when inlined 

image

src/hooks/useColumnWidths.ts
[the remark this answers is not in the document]
amanmahajan7: It works. We should put the min/max logic in one place

src/hooks/useColumnWidths.ts
amanmahajan7: Not sure I understand

nstepien: Me neither, I think I got confused

src/hooks/useColumnWidths.ts
```diff
@@ -16,6 +15,7 @@ export function useColumnWidths<R, SR>(
   setMeasuredColumnWidths: StateSetter<ReadonlyMap<string, number>>,
   onColumnResize: DataGridProps<R, SR>['onColumnResize']
 ) {
+  const [columnToAutoResize, setColumnToAutoResize] = useState<string | null>(null);
   const prevGridWidthRef = useRef(gridWidth);
```
Wroud, line 19: @amanmahajan7 https://github.com/adazzle/react-data-grid/pull/3746/files#diff-e002f181d41ce4fd6e070724347a44e4220d0332c37e5964d04cb3fc5d60d416R19
I think you better not use ref for the previous grid width. There is a bug with the initial render when:
1. table rendered with a different width from the previous
2. columns calculated + updated
3. table rendered with a different width from the previous
4. columns calculated but aren't updated -> there are no re-render -> cells will have "measurement" size (like "max-content") and will change size on content size

This issue is reflected in my video from this comment: https://github.com/adazzle/react-data-grid/issues/3723#issuecomment-2704255982

src/hooks/useColumnWidths.ts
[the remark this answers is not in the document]
Wroud: min / max can't overflow in case styles are applied to the measurement cells correctly, and grid re-render is synchronized

I also found it helpful to use styles instead of JS for this, especially when you can have different styles for cells while measuring (when we set its size to "max-content" for example, and will measure its size after):

You want to have a grid with auto-sized columns but with some rules for this auto-size mechanism:
min: 50px
max: 300px

But at the same time, you want to be able to resize columns to exceed these values manually.

In this situation, you will have different rules for measured cells and while measuring:
When measuring cells (like "max-content" or "auto"), we want to apply these rules to that cell
but when we resize manually, we don't

```css
.measuringCellClassname {
  min-width: 20px;
}
.measuringCellClassname[measuring] {
  min-width: 50px;
  max-width: 300px
}
```

This is why I've added this code also:
https://github.com/adazzle/react-data-grid/pull/3724/files#diff-62e848727f791df23e85ccc2140643630d80c20c5140f073aba8f581eb122019L13-R26

src/hooks/useColumnWidths.ts
```diff
@@ -40,62 +43,81 @@ export function useColumnWidths<R, SR>(
...
         const newResizedColumnWidths = new Map(resizedColumnWidths);
-        newResizedColumnWidths.set(resizingKey, measuredWidth);
+        newResizedColumnWidths.set(resizingKey, nextWidth);
         return newResizedColumnWidths;
       });
-      updateMeasuredWidths(columnsToMeasure);
-    });
-
-    onColumnResize?.(column, measuredWidth);
+      onColumnResize?.(column, nextWidth);
```
Wroud, line 117: I will suggest moving the onColumnResize call to the setter function to have the same possible behavior for resizing 
https://github.com/adazzle/react-data-grid/pull/3746/files#diff-e002f181d41ce4fd6e070724347a44e4220d0332c37e5964d04cb3fc5d60d416R77
(or vice-versa)

src/hooks/useColumnWidths.ts
amanmahajan7: Can you check if this issues still exists in my PR? I think the issue you are describing was related to https://github.com/adazzle/react-data-grid/pull/3724#issuecomment-2686081416

src/hooks/useColumnWidths.ts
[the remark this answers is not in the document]
amanmahajan7: I can add dependencies and `useMemo` for `columnsToMeasure` but not sure it is worth. Wdyt?

test/browser/column/resizable.test.tsx
amanmahajan7: It should be. I will investigate in a separate PR

amanmahajan7:
> Found a bug, when I add maxWidth: 150, to the country column, the grid isn't filled, then I try to reduce its width, and the other columns fill the empty space.

This exists in main also. I believe the reason is that we are not restricting width calculation using grid columns and instead we are using `min/max` width on measuring cell. I have created a separate PR to refactor width calculation logic 

https://github.com/adazzle/react-data-grid/pull/3747

src/hooks/useColumnWidths.ts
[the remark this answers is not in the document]
amanmahajan7: > But at the same time, you want to be able to resize columns to exceed these values manually.

We want the same rules for bot cases. If `maxWidth` is specified then users should not be able to increase the width past that value. 

> I also found it helpful to use styles instead of JS
We already set `min/maxWidth` on the measuring cell 
https://github.com/adazzle/react-data-grid/blob/main/src/utils/renderMeasuringCells.tsx#L18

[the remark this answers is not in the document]
Wroud: Yep, but I'm talking about cases when you have not specified maxWidth for the column, and this `max-width` only applies to the moment when we calculate the size of the column in cases when both are specified:
`column.maxWidth: 100px`
`measuringCellClassname[measuring] { max-width: 200px; }`
then it should not exceed `100px`
But in case:
`column.maxWidth: 100px`
`measuringCellClassname[measuring] { max-width: 50px; }`
the auto-sized width will not exceed 50px but can be re-sized up to 100px manually

src/hooks/useColumnWidths.ts
```diff
@@ -40,62 +43,81 @@ export function useColumnWidths<R, SR>(
 
   useLayoutEffect(() => {
     prevGridWidthRef.current = gridWidth;
-    updateMeasuredWidths(columnsToMeasure);
+    updateMeasuredAndResizedWidths();
   });
 
-  function updateMeasuredWidths(columnsToMeasure: readonly string[]) {
-    if (columnsToMeasure.length === 0) return;
```
nstepien, line 47: Since we add columnToAutoResize to columnsToMeasure, we can keep this early return guard if you want

test/browser/column/resizable.test.tsx
```diff
@@ -137,9 +141,58 @@ test('should use the minWidth if specified on auto resize', async () => {
...
+      }
+    ],
+    onColumnResize
+  });
+  const grid = getGrid();
+  await expect.element(grid).toHaveStyle({ gridTemplateColumns: '639.328px 639.328px 639.344px' });
+  const [col1] = getHeaderCells();
+  await autoResize(col1);
+  await expect.element(grid).toHaveStyle({ gridTemplateColumns: '79.1406px 919.422px 919.438px' });
+  expect(onColumnResize).toHaveBeenCalled();
```
nstepien, line 192: Can we check this?
```suggestion
  expect(onColumnResize).toHaveBeenCalledOnce();
```

test/browser/column/resizable.test.tsx
```diff
@@ -137,9 +141,58 @@ test('should use the minWidth if specified on auto resize', async () => {
...
+  onColumnResize.mockClear();
+  // onColumnResize is not called if width is not changed
+  await autoResize(col1);
+  await expect.element(grid).toHaveStyle({ gridTemplateColumns: '79.1406px 919.422px 919.438px' });
+  expect(onColumnResize).not.toHaveBeenCalled();
```
nstepien, line 197: Without clearing the mock if you want
```suggestion
  // onColumnResize is not called if width is not changed
  await autoResize(col1);
  await expect.element(grid).toHaveStyle({ gridTemplateColumns: '79.1406px 919.422px 919.438px' });
  expect(onColumnResize).toHaveBeenCalledOnce();
```

nstepien (dismissed):

test/browser/column/resizable.test.tsx
amanmahajan7: It gets called twice in strict mode

nstepien: Worth checking if the width has changed?

test/browser/column/resizable.test.tsx
amanmahajan7: Same here.

test/browser/column/resizable.test.tsx
amanmahajan7: Needs some work. I will check in the next PR

test/browser/column/resizable.test.tsx
amanmahajan7: Changed it to `toHaveBeenCalledOnce`

test/browser/utils.tsx
```diff
@@ -4,16 +4,13 @@ import { css } from '@linaria/core';
 import { DataGrid } from '../../src';
 import type { DataGridProps } from '../../src';
 
+export const testGridClassname = css`
```
nstepien, line 7: We can revert this if you want

src/hooks/useColumnWidths.ts
```diff
@@ -62,40 +70,52 @@ export function useColumnWidths<R, SR>(
...
-      updateMeasuredWidths(columnsToMeasure);
     });
 
-    onColumnResize?.(column, measuredWidth);
+    const previousWidth = resizedColumnWidths.get(resizingKey);
+    const newWidth =
+      typeof nextWidth === 'number' ? nextWidth : measureColumnWidth(gridRef, resizingKey);
+    if (newWidth !== undefined && newWidth !== previousWidth) {
+      onColumnResize?.(column, newWidth);
+    }
```
nstepien, line 118: Maybe we can do all this only when `onColumnResize` is defined

nstepien (approved):

PR #3747, opened 2025-03-21
Refactor column width calculation logic 

amanmahajan7:


src/hooks/useGridDimensions.ts
```diff
@@ -28,12 +27,9 @@ export function useGridDimensions() {
       const size = entries[0].contentBoxSize[0];
       const { clientHeight, offsetHeight } = gridRef.current!;
 
-      // we use flushSync here to avoid flashing scrollbars
-      flushSync(() => {
-        setInlineSize(size.inlineSize);
-        setBlockSize(size.blockSize);
-        setHorizontalScrollbarHeight(offsetHeight - clientHeight);
-      });
+      setInlineSize(size.inlineSize);
```
amanmahajan7, line 30: Does not look like this is needed anymore

src/utils/index.ts
```diff
@@ -17,19 +17,44 @@ export function assertIsValidKeyGetter<R, K extends React.Key>(
...
+
+  const hasMaxWidth = maxWidth != null;
+  const clampedWidth = hasMaxWidth
+    ? `clamp(${minWidth}px, ${widthWithUnit}, ${maxWidth}px)`
+    : `max(${minWidth}px, ${widthWithUnit})`;
 
-  // ignore maxWidth if it less than minWidth
-  if (typeof maxWidth === 'number' && maxWidth >= minWidth) {
-    return min(width, maxWidth);
+  // clamp() and max() do not handle all the css grid column width values
```
amanmahajan7, line 36: I tried `calc-size` but it does not work unfortunately 
https://developer.mozilla.org/en-US/docs/Web/CSS/calc-size

src/hooks/useGridDimensions.ts
amanmahajan7: nvm. it is needed

src/utils/index.ts
```diff
@@ -17,19 +17,44 @@ export function assertIsValidKeyGetter<R, K extends React.Key>(
     throw new Error('Please specify the rowKeyGetter prop to use selection');
   }
 }
-
-export function clampColumnWidth<R, SR>(
-  width: number,
+export function getColumnWidthForMeasurement<R, SR>(
```
Wroud, line 20: I believe that this function is not needed at all if we pass `min-width` and `max-width` to columns styles; then setting any `width` will not break the grid layout

src/hooks/useColumnWidths.ts
```diff
@@ -15,38 +15,41 @@ export function useColumnWidths<R, SR>(
...
     }
   }
 
   const gridTemplateColumns = newTemplateColumns.join(' ');
 
-  useLayoutEffect(() => {
-    prevGridWidthRef.current = gridWidth;
-    updateMeasuredAndResizedWidths();
-  });
+  useLayoutEffect(updateMeasuredAndResizedWidths);
```
amanmahajan7, line 48: Cannot inline this function other exhaustive deps rule complains. I can add `useMemo` but not sure it is worth

src/utils/index.ts
amanmahajan7: This is needed to fix this [bug](https://github.com/adazzle/react-data-grid/pull/3746#issuecomment-2744260592). We already set `min/max` width on the measuring cell

src/hooks/useColumnWidths.ts
```diff
@@ -131,3 +128,47 @@ function measureColumnWidth(gridRef: React.RefObject<HTMLDivElement | null>, key
...
+  { minWidth, maxWidth }: CalculatedColumn<R, SR>
+) {
+  const widthWithUnit = typeof width === 'number' ? `${width}px` : width;
+
+  // don't break in Node.js (SSR) and jsdom
+  if (typeof CSS === 'undefined') {
+    return widthWithUnit;
+  }
+
+  const hasMaxWidth = maxWidth != null;
```
nstepien, line 143: Do we need to handle `maxWidth` here if we set it on measuring cells?

test/browser/column/resizable.test.tsx
```diff
@@ -76,12 +76,15 @@ test('should resize column when dragging the handle', async () => {
...
-  setup<Row, unknown>({ columns, rows: [] });
+  const onColumnResize = vi.fn();
+  setup<Row, unknown>({ columns, rows: [], onColumnResize });
   const grid = getGrid();
+  expect(onColumnResize).not.toHaveBeenCalled();
   await expect.element(grid).toHaveStyle({ gridTemplateColumns: '100px 200px ' });
   const [, col2] = getHeaderCells();
   await resize({ column: col2, resizeBy: 1000 });
   await expect.element(grid).toHaveStyle({ gridTemplateColumns: '100px 400px' });
+  expect(onColumnResize).toHaveBeenCalledWith(expect.objectContaining(columns[1]), 400);
```
nstepien, line 87:
```suggestion
  expect(onColumnResize).toHaveBeenCalledExactlyOnceWith(expect.objectContaining(columns[1]), 400);
```

src/hooks/useColumnWidths.ts
amanmahajan7: Not 100% sure. My understanding is that grid computes the column widths and then apply min/max style from the measuring cell. This is the reason there was some left over space in the bug you found. I will check again

nstepien (approved):

src/hooks/useColumnWidths.ts
amanmahajan7: Also, resizing does not work properly without this as `clampColumnWidth` helper was removed

amanmahajan7: Found another edge case.

PR #3754, opened 2025-04-10
Resize column via keyboard

amanmahajan7:
Fixes https://github.com/adazzle/react-data-grid/issues/3320

Resize column on `Ctrl + ArrowRight/Left`.


src/HeaderCell.tsx
```diff
@@ -160,10 +161,25 @@ export default function HeaderCell<R, SR>({
...
-    if (event.key === ' ' || event.key === 'Enter') {
+    const { key } = event;
+    if (sortable && (key === ' ' || key === 'Enter')) {
       // prevent scrolling
       event.preventDefault();
       onSort(event.ctrlKey || event.metaKey);
+    } else if (
+      resizable &&
+      isCtrlKeyHeldDown(event) &&
+      (key === 'ArrowLeft' || key === 'ArrowRight')
```
amanmahajan7, line 172: Open to suggestion on which shortcut to use. I did not find any good example other than [this](https://react-spectrum.adobe.com/react-aria/Table.html#column-resizing). But the example uses different a different strategy where the resizer gets focused. I think we can keep out implementation simple

src/HeaderCell.tsx
```diff
@@ -160,10 +161,25 @@ export default function HeaderCell<R, SR>({
...
+    ) {
+      // prevent navigation
+      // TODO: check if we can use `preventDefault` instead
+      event.stopPropagation();
+      const { width } = event.currentTarget.getBoundingClientRect();
+      const offset = key === 'ArrowLeft' ? -10 : 10;
```
amanmahajan7, line 178: We can also add support to resize to `max-content`. Maybe `ctrl + m`?

src/HeaderCell.tsx
```diff
@@ -160,10 +161,25 @@ export default function HeaderCell<R, SR>({
...
       // prevent scrolling
       event.preventDefault();
       onSort(event.ctrlKey || event.metaKey);
+    } else if (
+      resizable &&
+      isCtrlKeyHeldDown(event) &&
+      (key === 'ArrowLeft' || key === 'ArrowRight')
+    ) {
+      // prevent navigation
+      // TODO: check if we can use `preventDefault` instead
```
amanmahajan7, line 175: I think it would be better to use `preventDefault` and `handleCellKeyDown` can ignore the event if it is prevented. I remember there was issue which editor though. I will check in a separate PR

src/HeaderCell.tsx
nstepien, line 178: Should we check for ltr/rtl?

amanmahajan7: good point 😱

nstepien (approved):

Issue #3403, opened 2023-12-11
Column widths should recalculate when columns are added/removed from the columns array


Artforge:
When I render a grid with a set of columns, then filter the dataset in a way that keeps some of the existing columns, the remaining columns do not resize to fill the grid.

I'm currently on the 7.0.0-beta40 version.

Likewise, if I replace the data in an existing grid with a dataset that adds columns (in addition to existing column keys) the additional columns are added outside the visible area of the grid (need to scroll right to view them)

I've seen earlier bug reports with variations of this issue - but none with a resolution.

Is there any way to force the data grid to recalculate column widths after a change to the columns array?

Any help is much appreciated.

KostyanWest:
I've faced with the same issue. I noticed that changing the grid area causes the column widths to be recalculated. But the grid virtualization excludes columns outside the screen from the width recalculation, so these steps worked for me:

Firstly, I disabled a column-only virtualization by making a patch following this comment: https://github.com/adazzle/react-data-grid/issues/3249#issuecomment-1675277771 .

Secondly, I patched a `lib/bundle.js` file by changing this line
`const ignorePreviouslyMeasuredColumns: boolean =`...
to that one
`const ignorePreviouslyMeasuredColumns: boolean = true;`
(https://github.com/adazzle/react-data-grid/blob/main/src/hooks/useColumnWidths.ts#L23C7-L23C7).

I can't say I 100% understand what my changes led to, but it works pretty well for me now, I haven't noticed any performance issues. I hope this workaround helps you too.

iamyoki:
> I've faced with the same issue. I noticed that changing the grid area causes the column widths to be recalculated. But the grid virtualization excludes columns outside the screen from the width recalculation, so these steps worked for me:
> 
> Firstly, I disabled a virtualization for the columns only by making a patch following this comment: [#3249 (comment)](https://github.com/adazzle/react-data-grid/issues/3249#issuecomment-1675277771) .
> 
> Secondly, I patched a `lib/bundle.js` file by changing this line `const ignorePreviouslyMeasuredColumns: boolean =`... to that one `const ignorePreviouslyMeasuredColumns: boolean = true;` (https://github.com/adazzle/react-data-grid/blob/main/src/hooks/useColumnWidths.ts#L23C7-L23C7).
> 
> I can't say I 100% understand what my changes led to, but it works pretty well for me now, I haven't noticed any performance issues. I hope this workaround helps you too.

[...]

Issue #3133, opened 2023-01-22
Recalculate column widths


voliva:
## Use case 

Columns set with `max-content` only calculate their width on the initial render. It would be nice to have the possibility to trigger a recalculation of the column sizes when needed, and have all columns' width update accordingly.

## Proposed solution 

Because it's a expensive operation, I think it's probably better to have it as part of the imperative ref.

If I'm not missing anything, the implementation should be straightforward: just update `setColumnWidths` with a map that excludes the columns with `max-content` (and similar). I did a patched version of react-data-grid that adds `recalculateColumns: () => setColumnWidths(() => new Map())` on the `useImperativeHandle` and it works in a way that would fit my needs.

tlehtimaki:
It would be super to see a patch for this. With `npm` we are able to patch this, but unfortunately we need to use `pnpm` and for that the patching did not work. Anyway forking or patching is not the most feasible option. 

Now I understand that making features or fixing regression is tough, so I would ask that in what parts of the code we should look into help into making this feature towards a pull-request?

Issue #2042, opened 2020-05-22
V7 column resize bugs


igoriok1994:
Hi,

In V7 canary 16 now supported column width in percents. That's nice.

After some testing I found bug.

Steps:
1. Dynamically set column width from props 
```
<DataGrid
    columns={...}
>
```
2. Do manual resize (adjust column width with mouse)
3. Update grid columns prop (like in Step 1.). Now columns will not react on width property...

Here is the video report:

https://youtu.be/gGtp1-TsTx8

Here is the code:

```javascript
require('!style!css!sass!./Temp_i_3.scss');

import React, { PureComponent } from 'react';
import _ from 'lodash';
import 'rdg.v7/dist/react-data-grid.css';
import DataGrid from 'rdg.v7';
import { AutoSizer } from 'react-virtualized';
import { Button } from '@c/button';

class Temp_i_3 extends PureComponent {
	constructor() {
		super();

		this.state = {
			cols: [
				{
					key: 'c1',
					name: 'c1',
					resizable: true,
					width: '15%'
				},
				{
					key: 'c2',
					name: 'c2',
					resizable: true
				},
				{
					key: 'c3',
					name: 'c3',
					resizable: true
				}
			]
		};
	}

	componentWillUnmount() {
		this._mounted = false;
	}

	componentDidMount() {
		this._mounted = true;
	}

	setRand1 = () => {
		let v = Math.floor(Math.random() * 400) + 50;

		let cols = _.cloneDeep(this.state.cols);
		cols[0].width = v;

		console.log("new width: ", v);

		this.setState({ cols });
	}

	setRand2 = () => {
		let v = Math.floor(Math.random() * 50) + 10;

		let cols = _.cloneDeep(this.state.cols);
		cols[0].width = `${v}%`;

		console.log("new width: ", `${v}%`);

		this.setState({ cols });
	}

	render() {
		return (
			<div className='ST_H100 ST_W100'>
				<AutoSizer>
					{({ height, width }) => (
						<React.Fragment>
							<DataGrid
								width={width}
								height={height - 35}
								columns={this.state.cols}
								rows={_.map(new Array(200), (v,i) => {
									return {
										c1: i,
										c2: i,
										c3: i
									};
								})}
							/>
							<div className='ST_ROW ST_MG-T' style={{width}}>
								<Button onClick={this.setRand1}>
									set random px width
								</Button>
								<Button onClick={this.setRand2}>
									set random % width
								</Button>
							</div>
[...]

PR #3761, opened 2025-04-16
Add `columnWidths` and `onColumnWidthsChange` props

amanmahajan7:
Fixes
https://github.com/adazzle/react-data-grid/issues/3725 
https://github.com/adazzle/react-data-grid/issues/3403
https://github.com/adazzle/react-data-grid/issues/3133
https://github.com/adazzle/react-data-grid/issues/2042

These new props can be used to reset widths when needed for example 

```
const [columnWidths, setColumnWidths] = useState((): ColumnWidths => new Map());

function addNewRow() {
  setRows(...);
  setColumnWidths(new Map());
}

...
function toggleColumnVisibility(..) {
  setVisibleColumns(...);
  setColumnWidths(new Map());
}

,,
return <DataGrid columnWidths={columnWidths} onColumnWidthsChange={setColumnWidths} ../>
```

If these props are not provided then the grid uses internal state. `onColumnWidthsChange` at the end of the resize operation whereas `onColumnResize` fires as the column is being resized

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -342,6 +342,8 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
   const columnWidths = isColumnWidthsControlled ? columnWidthsRaw : columnWidthsInternal;
   const onColumnWidthsChange = isColumnWidthsControlled
     ? (columnWidths: ColumnWidths) => {
+        // we need to keep the columnWidths state and prop in sync otherwise it leads to bugs like
+        // resize column, auto resize column, and resize column again will use the internal width.
         setColumnWidthsInternal(columnWidths);
         onColumnWidthsChangeRaw(columnWidths);
       }
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -109,6 +109,7 @@ export interface DataGridHandle {
   element: HTMLDivElement | null;
   scrollToCell: (position: PartialPosition) => void;
   selectCell: (position: Position, enableEditor?: Maybe<boolean>) => void;
+  resetMeasuredColumnWidths: () => void;
 }
 
 type SharedDivProps = Pick<
@@ -540,7 +541,10 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
         setScrollToPosition({ idx: scrollToIdx, rowIdx: scrollToRowIdx });
       }
     },
-    selectCell
+    selectCell,
+    resetMeasuredColumnWidths() {
+      setMeasuredColumnWidths(new Map());
+    }
   }));
 
   /**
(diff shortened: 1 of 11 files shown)


src/DataGrid.tsx
```diff
@@ -320,25 +327,32 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
...
+  const [isColumnResizing, setColumnResizing] = useState(false);
   const [isDragging, setDragging] = useState(false);
   const [draggedOverRowIdx, setOverRowIdx] = useState<number | undefined>(undefined);
   const [scrollToPosition, setScrollToPosition] = useState<PartialPosition | null>(null);
   const [shouldFocusCell, setShouldFocusCell] = useState(false);
   const [previousRowIdx, setPreviousRowIdx] = useState(-1);
 
+  const isColumnWidthsControlled =
+    columnWidthsRaw != null && onColumnWidthsChangeRaw != null && !isColumnResizing;
+  const columnWidths = isColumnWidthsControlled ? columnWidthsRaw : columnWidthsInternal;
```
amanmahajan7, line 342: Internal state serves two purposes 
1. It is used when the props are not provided
2. It is used for setting temporary until the resize operation is complete

test/browser/column/resizable.test.tsx
```diff
@@ -240,3 +240,26 @@ test('should remeasure flex columns when resizing a column', async () => {
...
+  await expect.element(grid).toHaveStyle({ gridTemplateColumns: '101px 201px' });
+  const [, col2] = getHeaderCells();
+  await autoResize(col2);
+  expect(onColumnWidthsChange).toHaveBeenCalledExactlyOnceWith(
+    new Map([
+      ['col1', { width: 101, type: 'measured' }],
+      ['col2', { width: 100, type: 'resized' }]
+    ])
+  );
+});
```
nstepien, line 265: Can we add a test that tests that we sync the internal state with the external state when the external state changes. We can try to resize after an external change, to make sure the external change remains even while resizing

amanmahajan7: Done

test/browser/column/resizable.test.tsx
```diff
@@ -240,3 +241,92 @@ test('should remeasure flex columns when resizing a column', async () => {
...
 });
+
+test('should use columnWidths and onColumnWidthsChange props when provided', async () => {
+  const onColumnWidthsChangeSpy = vi.fn();
+  const onColumnResizeSpy = vi.fn();
+
+  function TestGrid() {
+    const [columnWidths, setColumnWidths] = useState(
+      (): ColumnWidths =>
+        new Map([
```
nstepien, line 252: FYI the return type isn't 100% enough to make `new Map(...)` type safe, for example you can add random properties in the object
image
https://github.com/microsoft/TypeScript/issues/55023
Might be good to export a type for the `type/width` object

amanmahajan7: Done

nstepien (approved):

PR #3773, opened 2025-04-25
Fix focus logic in Firefox

amanmahajan7:
https://developer.chrome.com/blog/keyboard-focusable-scrollers
https://bugzilla.mozilla.org/show_bug.cgi?id=1069739

Scrollable containers without tabIndex are keyboard focusable in Chrome only if there is no focusable element inside whereas they are always focusable in Firefox. This causes issues when grid receives focus as FF has an extra tab stop. 

image
image

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -1196,7 +1196,7 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
       aria-colcount={columns.length}
       aria-rowcount={ariaRowCount}
       // Scrollable containers without tabIndex are keyboard focusable in Chrome only if there is no focusable element inside
-      // whereas they are are always focusable in Firefox. We need to set tabIndex to have a consistent behavior across browsers.
+      // whereas they are always focusable in Firefox. We need to set tabIndex to have a consistent behavior across browsers.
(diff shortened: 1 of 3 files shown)


src/DataGrid.tsx
```diff
@@ -1185,6 +1195,9 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
       aria-multiselectable={isSelectable ? true : undefined}
       aria-colcount={columns.length}
       aria-rowcount={ariaRowCount}
+      // Scrollable containers without tabIndex are keyboard focusable in Chrome only if there is no focusable element inside
+      // whereas they are always focusable in Firefox. We need to set tabIndex to have a consistent behavior across browsers.
```
nstepien, line 1199: Should we do that?
Maybe Firefox users want to preserve the default behavior, wdyt?

nstepien (dismissed):

src/DataGrid.tsx
amanmahajan7: Not sure if it is helpful. It is good to have consistent behavior across browsers though. I checked a few libraries and they all work similarly where the first cell receives focus. Also shift + tab does not focus the grid 🤔 in FF. I think we can merge this and revisit if there any issues.

nstepien (approved):

PR #3774, opened 2025-04-27
Use mouse down event to select cell

amanmahajan7:
I checked a few popular grids and all select the cell on pointer down but they do not scroll the cell into view which we do. Not sure which behavior is better

🆕 `onCellMouseDown` event


src/SummaryCell.tsx
```diff
@@ -37,7 +37,7 @@ function SummaryCell<R, SR>({
     typeof summaryCellClass === 'function' ? summaryCellClass(row) : summaryCellClass
   );
 
-  function onClick() {
+  function onPointerDown() {
     selectCell({ rowIdx, idx: column.idx });
```
nstepien, line 41: I wonder if we should have unified event handlers at the DataGrid level instead of having the same implementations in various components.
Something like
```ts
function onCellEvent(event: any event, pointerArgs: which row/cell) {
  // if pointer down...
  // if click...
}
```

test/browser/events.test.tsx
```diff
@@ -103,19 +103,20 @@ describe('Events', () => {
...
-      />
-    );
+    const onCellContextMenu = vi.fn();
+    page.render(<EventTest onCellContextMenu={onCellContextMenu} />);
+    expect(onCellContextMenu).not.toHaveBeenCalled();
     await userEvent.click(getCellsAtRowIndex(0)[0], { button: 'right' });
-    expect(getCellsAtRowIndex(0)[0]).toHaveAttribute('aria-selected', 'false');
-    await userEvent.click(getCellsAtRowIndex(0)[1], { button: 'right' });
-    expect(getCellsAtRowIndex(0)[1]).toHaveAttribute('aria-selected', 'true');
+    expect(onCellContextMenu).toHaveBeenLastCalledWith(
```
nstepien, line 110:
```suggestion
    expect(onCellContextMenu).toHaveBeenCalledExactlyOnceWith(
```

src/Cell.tsx
```diff
@@ -46,28 +47,35 @@ function Cell<R, SR>({
     selectCell({ rowIdx, idx: column.idx }, openEditor);
   }
 
+  function handlePointerDown(event: React.PointerEvent<HTMLDivElement>) {
```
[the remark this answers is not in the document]
amanmahajan7: Switched to mousedown which works better

src/SummaryCell.tsx
amanmahajan7: I was thinking the when I added https://github.com/adazzle/react-data-grid/pull/3774#discussion_r2067467182. I will check

src/DataGrid.tsx
```diff
@@ -525,23 +533,17 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
   /**
    * effects
    */
-  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
amanmahajan7: I don't believe so

nstepien (approved):

PR #3775, opened 2025-04-28
PoC: Add grid.rowStyle and col.cellStyle options

leeoniya:
hey @nstepien @amanmahajan7 

image

i wanted to open this up to see how y'all feel about allowing users to style rows and cells _without generating classNames_.

our use case is being able to style a row or cell using something like [d3 color scheme gradients](https://observablehq.com/@d3/color-schemes), based on cell value. there can be potentially 256 different colors _per column_, and generating a class for each color on demand is not great. additionally we may want to style both text and background, leading to a possible combinatorial explosion of classes.

i know we can always throw an additional div into `renderCell` and `renderRow` to do this, but that leads to extra DOM and worse performance. we can use renderers, but this increases complexity a lot in user code that seems silly just to be able to add colors and change little else.

would you be open to adding arbitrary style callbacks like this?

thanks for considering!

### DIFF
--- a/src/Cell.tsx
+++ b/src/Cell.tsx
@@ -31,7 +31,7 @@ function Cell<R, SR>({
 }: CellRendererProps<R, SR>) {
   const { tabIndex, childTabIndex, onFocus } = useRovingTabIndex(isCellSelected);
 
-  const { cellClass, cellStyle } = column;
+  const { cellClass } = column;
   className = getCellClassname(
     column,
     {
@@ -40,9 +40,6 @@ function Cell<R, SR>({
     typeof cellClass === 'function' ? cellClass(row) : cellClass,
     className
   );
-
-  const _cellStyle = typeof cellStyle === 'function' ? cellStyle(row) : cellStyle;
-
   const isEditable = isCellEditableUtil(column, row);
 
   function selectCellWrapper(openEditor?: boolean) {
@@ -91,8 +88,7 @@ function Cell<R, SR>({
       className={className}
       style={{
         ...getCellStyle(column, colSpan),
-        ...style,
-        ..._cellStyle
+        ...style
       }}
       onClick={handleClick}
       onDoubleClick={handleDoubleClick}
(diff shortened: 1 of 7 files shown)


amanmahajan7:
Thank you for the PR. You mentioned `renderRow` and `renderCell` and I think we can use them without needing an extra div 

```
function renderCell(props) {
  return defaultRenderCell({ ...props, style={customStyle} />
}

function renderRow(props) {
  return defaultRenderRow({ ...props, style={customStyle} />
}

function Component() {
  return <DataGrid  renderers={{ renderCell, renderRow}} ...
}
```

For this to work we need to make a few changes 
1. Export `defaultRenderCell` and `defaultRenderRow` (https://github.com/adazzle/react-data-grid/blob/main/src/index.ts#L2)
2. Update `Row` component to accept `style` prop (
https://github.com/adazzle/react-data-grid/blob/main/src/types.ts#L222)

wdyt?

leeoniya:
@amanmahajan7 yep that sounds great, go for it :)

amanmahajan7:
Do you want to make these changes in your PR?

leeoniya:
if you prefer, sure. i don't need the credit, so if you feel like there would be less review burden / faster, feel free to close this and do the PR on your end.

leeoniya:
@amanmahajan7 i pushed the suggested approach, but not seeing anything render. what am i missing?

amanmahajan7 (approved):
Thank you for the PR

leeoniya:
@amanmahajan7 thank you!

one comment about the demo, though...

i think the scenario shown in the demo and my simple examples are probably better done with className, since the amount of possible states is limited/known.

the d3-gradient/color-by-value better illustrates a real situation when className wont work well.

anyways, looking forward to next beta. also excited to finally see a columnResizeEnd equivalent, instead of  having to hack that in as a userland custom hook with click/mouseup handlers.

amanmahajan7:
> the d3-gradient/color-by-value better illustrates a real situation when className wont work well.

I think the demo illustrate the idea but feel free to update the example

PR #3779, opened 2025-05-08
Add an example for smooth scrolling

amanmahajan7:


src/utils/domUtils.ts
```diff
@@ -4,6 +4,6 @@ export function stopPropagation(event: React.SyntheticEvent) {
   event.stopPropagation();
 }
 
-export function scrollIntoView(element: Maybe<Element>) {
-  element?.scrollIntoView({ inline: 'nearest', block: 'nearest' });
+export function scrollIntoView(element: Maybe<Element>, behavior: ScrollBehavior = 'instant') {
```
amanmahajan7, line 7: This way behavior cannot be overridden by an outside class.

website/routes/ScrollToCell.tsx
```diff
@@ -77,6 +88,29 @@ function ScrollToCell() {
             }}
           />
         </label>
+        <fieldset>
+          <legend>Scroll behavior</legend>
+          <label>
+            Auto
+            <input
+              type="radio"
```
nstepien, line 96: Don't we need a `name` for radio buttons?

nstepien (dismissed):

nstepien (approved):

PR #3780, opened 2025-05-09
Remove `memo` and use `react-compiler`

amanmahajan7:
https://react.dev/learn/react-compiler


amanmahajan7:
RC does not remove `memo` and this change severely affects performance

PR #3783, opened 2025-05-14
Set drag column image

amanmahajan7:
image

### DIFF
--- a/src/HeaderCell.tsx
+++ b/src/HeaderCell.tsx
@@ -44,8 +44,7 @@ const cellDraggableClassname = 'rdg-cell-draggable';
 
 const cellDragging = css`
   @layer rdg.HeaderCell {
-    // opacity: 0.5;
-    // TODO
+    background-color: var(--rdg-header-draggable-background-color);
   }
 `;
 
@@ -63,6 +62,7 @@ const dragImageClassname = css`
   @layer rdg.HeaderCell {
     border-radius: 4px;
     width: fit-content;
+    outline: 2px solid hsl(207, 100%, 50%);
   }
 `;
 
@@ -82,7 +82,8 @@ export interface HeaderCellProps<R, SR> extends SharedHeaderRowProps<R, SR> {
   colSpan: number | undefined;
   rowIdx: number;
   isCellSelected: boolean;
-  dragDropKey: string;
+  draggedColumnKey: string | undefined;
+  setDraggedColumnKey: (draggedColumnKey: string | undefined) => void;
 }
 
 export default function HeaderCell<R, SR>({
@@ -97,10 +98,11 @@ export default function HeaderCell<R, SR>({
   onSortColumnsChange,
   selectCell,
   direction,
-  dragDropKey
+  draggedColumnKey,
+  setDraggedColumnKey
 }: HeaderCellProps<R, SR>) {
-  const [isDragging, setIsDragging] = useState(false);
   const [isOver, setIsOver] = useState(false);
+  const isDragging = draggedColumnKey === column.key;
   const rowSpan = getHeaderCellRowSpan(column, rowIdx);
   const { tabIndex, childTabIndex, onFocus } = useRovingTabIndex(isCellSelected);
   const sortIndex = sortColumns?.findIndex((sort) => sort.columnKey === column.key);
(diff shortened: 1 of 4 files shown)


src/HeaderCell.tsx
```diff
@@ -43,17 +43,29 @@ export const resizeHandleClassname = css`
 const cellDraggableClassname = 'rdg-cell-draggable';
 
 const cellDragging = css`
-  opacity: 0.5;
+  @layer rdg.HeaderCell {
+    background-color: var(--rdg-header-draggable-background-color);
```
amanmahajan7, line 47: Kept the same style as cell over for now. We cannot use opacity as we are using a custom drag image

src/HeaderCell.tsx
```diff
@@ -180,13 +195,18 @@ export default function HeaderCell<R, SR>({
   }
 
   function onDragStart(event: React.DragEvent<HTMLDivElement>) {
-    event.dataTransfer.setData(dragDropKey, column.key);
+    const dragImage = event.currentTarget.cloneNode(true) as HTMLDivElement;
+    dragImage.classList.add(dragImageClassname);
+    dragImage.id = dragImageId;
+    event.currentTarget.parentElement!.insertBefore(dragImage, event.currentTarget);
```
amanmahajan7, line 201: This way the element is behind the column

src/HeaderCell.tsx
```diff
@@ -223,19 +234,23 @@ export default function HeaderCell<R, SR>({
...
-      onDragEnd,
-      /* events fired on the drop targets */
-      onDragOver,
-      onDragEnter,
-      onDragLeave,
-      onDrop
+      onDragEnd
     };
+
+    if (draggedColumnKey !== undefined && draggedColumnKey !== column.key) {
```
amanmahajan7, line 246: Now we won't fire these events if external item is dragged over

src/HeaderRow.tsx
```diff
@@ -60,7 +60,7 @@ function HeaderRow<R, SR, K extends React.Key>({
   selectCell,
   direction
 }: HeaderRowProps<R, SR, K>) {
-  const dragDropKey = useId();
+  const [draggedColumnKey, setDraggedColumnKey] = useState<string>();
```
amanmahajan7, line 63: We can move this to the parent columns and reorder columns internally as user is dragging.  I will check

website/routes/ColumnsReordering.tsx
```diff
@@ -106,18 +106,27 @@ function ColumnsReordering() {
...
+        );
+        const sourceColumnOrder = columnsOrder[sourceColumnOrderIndex];
+        const newColumnsOrder = columnsOrder.toSpliced(sourceColumnOrderIndex, 1);
+        newColumnsOrder.splice(targetColumnOrderIndex, 0, sourceColumnOrder);
+        return newColumnsOrder;
+      });
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (document.startViewTransition) {
```
amanmahajan7, line 125: FF does not support this

src/HeaderCell.tsx
```diff
@@ -180,13 +195,18 @@ export default function HeaderCell<R, SR>({
...
+    event.dataTransfer.setDragImage(dragImage, 0, 0);
     event.dataTransfer.dropEffect = 'move';
-    setIsDragging(true);
+    setDraggedColumnKey(column.key);
   }
 
   function onDragEnd() {
-    setIsDragging(false);
+    setDraggedColumnKey(undefined);
+    document.getElementById(dragImageId)!.remove();
```
nstepien, line 209: What if React removes it during a re-render?
```suggestion
    document.getElementById(dragImageId)?.remove();
```

src/HeaderRow.tsx
```diff
@@ -84,11 +84,11 @@ function HeaderRow<R, SR, K extends React.Key>({
         sortColumns={sortColumns}
         selectCell={selectCell}
         direction={direction}
-        dragDropKey={dragDropKey}
+        draggedColumnKey={draggedColumnKey}
+        setDraggedColumnKey={setDraggedColumnKey}
       />
     );
   }
-
```
nstepien, line 91: Could keep this newline

src/HeaderCell.tsx
amanmahajan7: I don't think it should happen but cannot say for sure. I will check if there is a way to render another element instead of cloning

PR #3787, opened 2025-05-19
Update rolldown, adopt rolldown-plugin-dts

nstepien:
- Updated rolldown
  - https://github.com/rolldown/rolldown/blob/main/CHANGELOG.md
  - Updated the rolldown config
    - js -> ts
    - removed `output.entryFileNames`
      - Had to change to a function to properly output the `.d.ts` file
      - It was simpler to just remove it, use the defaults, and update the paths in `package.json` instead.
    - removed `external`
      - Not sure if it was still needed before, but it seems to work fine now without it
- Replaced `api-extractor` with [`rolldown-plugin-dts`](https://github.com/sxzz/rolldown-plugin-dts), simplifying/speeding up the build process


package.json
```diff
@@ -17,12 +17,12 @@
...
-      "default": "./lib/bundle.js"
+      "default": "./lib/index.js"
     }
   },
-  "browser": "./lib/bundle.js",
-  "main": "./lib/bundle.js",
-  "module": "./lib/bundle.js",
+  "browser": "./lib/index.js",
+  "main": "./lib/index.js",
+  "module": "./lib/index.js",
```
amanmahajan7, line 25: Do we need all three?

vite.config.ts
```diff
@@ -65,7 +65,7 @@ export default defineConfig(({ command }) => ({
       exclude: ['./.cache/**/*']
     }),
     wyw({
-      exclude: ['./.cache/**/*'],
+      exclude: ['./.cache/**/*', '**/*.d.ts', '**/*.gen.ts'],
```
amanmahajan7, line 68: Do we not need to excluded `'**/*.d.ts', '**/*.gen.ts'` in `rolldown.config`?

nstepien: I did ad `'**/*.d.ts'`, as otherwise the build fails with the new plugin.
`'**/*.gen.ts'` is probably not needed as rolldown won't touch the website files.

package.json
nstepien: Probably not at this point, it's just there for older bundlers, but we could try removing those.

amanmahajan7 (approved):

PR #3790, opened 2025-05-21
Fix cell stealing focus on click

amanmahajan7:
This was introduced in https://github.com/adazzle/react-data-grid/pull/3774

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -57,6 +57,7 @@ import type {
   Position,
   Renderers,
   RowsChangeData,
+  SelectCellOptions,
   SelectHeaderRowEvent,
   SelectRowEvent,
   SortColumn
@@ -109,7 +110,7 @@ export type DefaultColumnOptions<R, SR> = Pick<
 export interface DataGridHandle {
   element: HTMLDivElement | null;
   scrollToCell: (position: PartialPosition) => void;
-  selectCell: (position: Position, enableEditor?: Maybe<boolean>) => void;
+  selectCell: (position: Position, options: SelectCellOptions) => void;
 }
 
 type SharedDivProps = Pick<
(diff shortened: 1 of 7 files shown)


src/DataGrid.tsx
```diff
@@ -952,7 +953,7 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
       isCellWithinBounds: isCellWithinSelectionBounds
     });
 
-    selectCell(nextSelectedCellPosition);
+    selectCell(nextSelectedCellPosition, { shouldFocusCell: true });
```
amanmahajan7, line 953: We only need to set `shouldFocusCell` flag on cell navigation using keyboard.

src/hooks/useRovingTabIndex.ts
```diff
@@ -12,6 +12,15 @@ export function useRovingTabIndex(isSelected: boolean) {
   function onFocus(event: React.FocusEvent<HTMLDivElement>) {
```
amanmahajan7, line 12: When a cell is clicked it is automatically focused as the cell has a `tabIndex: -1`. We can focus the child here

test/browser/column/renderCell.test.tsx
```diff
@@ -116,6 +116,35 @@ describe('Custom cell renderer', () => {
...
+        </>
+      );
+    }
+  };
+
+  page.render(<DataGrid columns={[column]} rows={[{ id: 1 }]} />);
+
+  const [cell] = getCells();
+  const button = page.getByRole('button', { name: 'value: 1' });
+  expect(cell).toHaveTextContent('value: 1External Text');
```
nstepien, line 139: Text content doesn't matter for this test no?
```suggestion
```

nstepien (approved):

test/browser/column/renderCell.test.tsx
```diff
@@ -116,6 +116,34 @@ describe('Custom cell renderer', () => {
...
+          </button>
+          <span>External Text</span>
+        </>
+      );
+    }
+  };
+
+  page.render(<DataGrid columns={[column]} rows={[{ id: 1 }]} />);
+
+  const [cell] = getCells();
```
amanmahajan7, line 137:
```suggestion
```

nstepien (approved):

PR #3798, opened 2025-06-09
`vitest@3.2.3`, `eslint@9.28.0`

amanmahajan7:
https://vitest.dev/blog/vitest-3-2.html


vite.config.ts
```diff
@@ -89,7 +89,7 @@ export default defineConfig(({ command }) => ({
     sequence: {
       shuffle: true
     },
-    workspace: [
+    projects: [
       {
         extends: true,
```
nstepien, line 94: Do we still need `extends` with `projects`?

amanmahajan7: probably. 

>  add "extends: true" to inherit the options from the root config

PR #3804, opened 2025-06-12
Do not focus grid when scrolled

amanmahajan7:
This reverts some changes from https://github.com/adazzle/react-data-grid/pull/3773

Currently the first cell is selected when the grid is scrolled. This is undesirable IMO. I have changed to logic back to previous one and added `tabIndex` to have consistent focus behavior in FF

Recording 2025-06-12 at 13 58 22

### DIFF
--- a/src/hooks/useRovingTabIndex.ts
+++ b/src/hooks/useRovingTabIndex.ts
@@ -10,16 +10,21 @@ export function useRovingTabIndex(isSelected: boolean) {
   }
 
   function onFocus(event: React.FocusEvent<HTMLDivElement>) {
-    const elementToFocus = event.currentTarget.querySelector<Element & HTMLOrSVGElement>(
-      '[tabindex="0"]'
-    );
+    // Do not steal focus if event originates from a focusable child
+    if (event.target === event.currentTarget) {
+      const elementToFocus = event.currentTarget.querySelector<Element & HTMLOrSVGElement>(
+        '[tabindex="0"]'
+      );
 
-    // Focus cell content when available instead of the cell itself
-    if (elementToFocus !== null) {
-      elementToFocus.focus({ preventScroll: true });
-      setIsChildFocused(true);
+      // Focus cell content when available instead of the cell itself
+      if (elementToFocus !== null) {
+        elementToFocus.focus({ preventScroll: true });
+        setIsChildFocused(true);
+      } else {
+        setIsChildFocused(false);
+      }
     } else {
-      setIsChildFocused(false);
+      setIsChildFocused(true);
     }
   }
 


nstepien (approved):

nstepien:
A test failed

src/hooks/useRovingTabIndex.ts
```diff
@@ -1,23 +1,26 @@
...
   if (isChildFocused && !isSelected) {
     setIsChildFocused(false);
   }
 
   function onFocus(event: React.FocusEvent<HTMLDivElement>) {
-    const elementToFocus = event.currentTarget.querySelector<Element & HTMLOrSVGElement>(
-      '[tabindex="0"]'
-    );
+    // Check if the focus event is originated from the cell and not from a focusable child
+    if (event.target === event.currentTarget) {
```
amanmahajan7, line 14: I not 100% sure if this is needed but grid probably should not steal focus in a complex formatter where user may want to manage focus manually. I will do more investigation and try to add a complex formatter

src/hooks/useRovingTabIndex.ts
```diff
@@ -1,23 +1,26 @@
...
 
   if (isChildFocused && !isSelected) {
     setIsChildFocused(false);
   }
 
   function onFocus(event: React.FocusEvent<HTMLDivElement>) {
-    const elementToFocus = event.currentTarget.querySelector<Element & HTMLOrSVGElement>(
-      '[tabindex="0"]'
-    );
+    // Check if the focus event is originated from the cell and not from a focusable child
```
nstepien:
```suggestion
    // Check if the focus event originated from the cell and not from a focusable child
```

src/hooks/useRovingTabIndex.ts
nstepien, line 14: Should we call `setIsChildFocused(true)` in `else`?

test/browser/column/renderCell.test.tsx
```diff
@@ -124,23 +124,45 @@ test('Focus child if it sets tabIndex', async () => {
       return (
         <>
           <button type="button" tabIndex={props.tabIndex}>
-            value: {props.row.id}
+            Button 1
+          </button>
+          <span>Text</span>
+          <button type="button" tabIndex={-1}>
```
nstepien, line 130: Do we need `tabIndex={-1}` here?
Not sure how roving focus works

test/browser/column/renderCell.test.tsx
```diff
@@ -124,23 +124,45 @@ test('Focus child if it sets tabIndex', async () => {
...
+  const button1 = page.getByRole('button', { name: 'Button 1' });
+  const button2 = page.getByRole('button', { name: 'Button 2' });
+  const cell = page.getByRole('gridcell', { name: 'Button 1 Text Button 2' });
+  expect(button1).toHaveAttribute('tabindex', '-1');
+  expect(cell).toHaveAttribute('tabindex', '-1');
+  await userEvent.click(page.getByText('Text'));
+  expect(button1).toHaveFocus();
+  expect(button1).toHaveAttribute('tabindex', '0');
+  await userEvent.tab({ shift: true });
+  expect(button1).not.toHaveFocus();
```
nstepien, line 149: What is focused instead?

amanmahajan7: Header cell. There grid has 1 column and 1 row (2 rows with header)

test/browser/column/renderCell.test.tsx
amanmahajan7: There should only be one element with `tabIndex: 0`. In this particular case the user should set the `childTabIndex` on `button1` and `button2` as needed

test/browser/column/renderCell.test.tsx
nstepien: Might be good to assert that so we don't regress

PR #3805, opened 2025-06-12
Try gap style

amanmahajan7:
https://developer.chrome.com/blog/gap-decorations?hl=en
https://microsoftedge.github.io/Demos/css-gap-decorations/playground.html


amanmahajan7:
There are 2 issues so it may not work for us
- sticky cells do not have border when stuck. May be a bug
- gaps adds to height so virtualization logic breaks

PR #3816, opened 2025-06-23
Fix a few accessibility issues

amanmahajan7:
- [x] Add `aria-label` to all grids
- [x] Restructure markup. I tested with Narrator and found a few issues. 
- [x] Apply Lighthouse recommendations

### DIFF
--- a/website/Nav.tsx
+++ b/website/Nav.tsx
@@ -4,23 +4,23 @@ import { css } from '@linaria/core';
 
 import type { Direction } from '../src/types';
 
-const containerClassname = css`
+const headerClassname = css`
   border-inline-start: 4px solid light-dark(hsl(210deg 50% 80%), hsl(210deg 50% 40%));
-`;
 
-const h1Classname = css`
-  font-size: 24px;
-  margin: 8px;
+  h1 {
+    font-size: 24px;
+  }
+
+  h1,
+  h2 {
+    margin: 8px;
+  }
 `;
 
 const navClassname = css`
   display: flex;
   flex-direction: column;
   white-space: nowrap;
-
-  h2 {
-    margin: 8px;
-  }
 
   a {
     color: inherit;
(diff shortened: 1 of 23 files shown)


website/Nav.tsx
```diff
@@ -44,62 +51,73 @@ const rtlCheckboxClassname = css`
...
-      <Link to="/NoRows">No Rows</Link>
-      <Link to="/ResizableGrid">Resizable Grid</Link>
-      <Link to="/RowsReordering">Rows Reordering</Link>
-      <Link to="/ScrollToCell">Scroll To Cell</Link>
-      <Link to="/TreeView">Tree View</Link>
-      <Link to="/VariableRowHeight">Variable Row Height</Link>
-      <Link to="/Animation">Animation</Link>
+  return (
+    <header className={headerClassname}>
+      <h1 id={headerId}>react-data-grid</h1>
```
amanmahajan7, line 65: It is generally recommended to use `h1` as the first child of `main` but in our case `h1` is on the left section. Using `aria-labelledby` fixes the issue

website/Nav.tsx
```diff
@@ -44,62 +51,73 @@ const rtlCheckboxClassname = css`
...
+      <nav aria-labelledby={demosNavId} className={navClassname}>
```
amanmahajan7, line 65: Added accessible name so screen reader can announce it like `Demos navigation landmark`

website/Nav.tsx
```diff
@@ -44,62 +51,73 @@ const rtlCheckboxClassname = css`
...
-        href="https://github.com/adazzle/react-data-grid/discussions"
-        target="_blank"
-        rel="noreferrer"
-      >
-        Discussions
-      </a>
-      <a href="https://github.com/adazzle/react-data-grid/issues" target="_blank" rel="noreferrer">
-        Issues
-      </a>
+      <nav aria-labelledby={linksNavId} className={navClassname}>
```
amanmahajan7, line 89: This is not part of demos so I have split these navs

nstepien (approved):

PR #3833, opened 2025-08-01
Try tanstack start

amanmahajan7:
https://tanstack.com/start/latest/docs/framework/react/build-from-scratch


nstepien:
Not worth it?
I was wondering if we could pre-render some of the content into HTML

amanmahajan7:
It is probably worth. I will take a look in a separate when I get a chance.

PR #3834, opened 2025-08-01
Set aria-hidden on div in ResizeHandle

maniekpl00:
Added aria-hidden="true" to the inner div used for column resizing.
This prevents VoiceOver from reading the column header twice while keeping resize functionality available for mouse users.

**Tested environment**
- MacOS, Voiceover

**Before** 

https://github.com/user-attachments/assets/1f03df0d-54fd-44bc-8a69-0c416662aa70

**After**

https://github.com/user-attachments/assets/11f197e2-34ab-4f56-ad38-54d8aab31623

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,8 +87,8 @@
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
     "rolldown": "^1.0.0-beta.28",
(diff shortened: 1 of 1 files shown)


amanmahajan7 (approved):
Thank you for the PR

PR #3842, opened 2025-08-11
Allow passing children to data grid

geovie:
This allows passing custom children to the data grid container which is useful if some kind of overlay or scroll buttons are needed which should be aligned with the grid or columns of the grid.

### DIFF
--- a/src/DataGrid.tsx
+++ b/src/DataGrid.tsx
@@ -123,6 +123,7 @@ type SharedDivProps = Pick<
   | 'aria-rowcount'
   | 'className'
   | 'style'
+  | 'children'
 >;
 
 export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends SharedDivProps {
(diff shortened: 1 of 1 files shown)


amanmahajan7:
Can you elaborate on your use case?

geovie:
Sure, my use case is displaying some control buttons inside the grid which are aligned with the sticky columns on the left and vertically centered.

https://github.com/user-attachments/assets/4d91c5bb-c98c-48b0-bae3-b321834c23d0


Here is the commit for the above example: https://github.com/Timeular/react-data-grid/commit/d06ada4ec575751b2c18a3cee25890f4c1d3969c

geovie:
Then I'd need to check that continusoly so that the left is updated when resizing the column 😅 and with the children it would get handled automatically by the browser/css

```tsx
  const [left, setLeft] = useState('');

  useLayoutEffect(() => {
    const interval = setInterval(() => {
      setLeft(
        `calc(${gridRef.current.element.style.getPropertyValue('--rdg-frozen-left-2')} + 128px)`
      );
    }, 100);
    return () => clearInterval(interval);
  }, []);
```

amanmahajan7:
You can use onColumnResize.

PR #3865, opened 2025-09-16
Optimize grid template rows generation with CSS repeat() function

nickolasKrayn:
Fix performance issue with large grid row template generation

## Problem
The previous implementation generated excessively long CSS strings when dealing with large datasets containing rows of identical heights. For 5000+ rows with same height, it would create a string like '10px 10px 10px ...' (5000 times), causing performance degradation in both JavaScript execution and CSS parsing.

## Solution
Implemented an optimized algorithm that uses CSS `repeat()` function to compact consecutive rows with identical heights into efficient notation.

## Key Changes
- Added intelligent grouping of consecutive rows with same height values
- Utilizes CSS `repeat()` function for optimized output

## Performance Impact
- **Before**: '10px '.repeat(5000) → ~25KB string
- **After**: 'repeat(5000, 10px)' → 18 characters

## Example Output
**Before**: '10px 10px 10px 10px 5px 20px 20px 10px 10px 10px 10px 10px'
**After**: 'repeat(4, 10px) 5px repeat(2, 20px) repeat(5, 10px)'

This fix prevents memory issues and improves rendering performance for grids with large numbers of rows, especially when row heights are consistent.

### DIFF
--- a/src/hooks/useViewportRows.ts
+++ b/src/hooks/useViewportRows.ts
@@ -32,19 +32,17 @@ export function useViewportRows<R>({
     // and we can consider using a similar approach as react-window
     // https://github.com/bvaughn/react-window/blob/b0a470cc264e9100afcaa1b78ed59d88f7914ad4/src/VariableSizeList.js#L68
     let totalRowHeight = 0;
-    const rowPositions: { top: number; height: number }[] = [];
     let gridTemplateRows = '';
     let currentHeight: number | null = null;
     let repeatCount = 0;
 
-    rows.forEach((row, index) => {
+    const rowPositions = rows.map((row, index) => {
       const currentRowHeight = rowHeight(row);
 
       const position = {
         top: totalRowHeight,
         height: currentRowHeight
       };
-      rowPositions.push(position);
       totalRowHeight += currentRowHeight;
 
       if (currentHeight === null) {
@@ -71,6 +69,8 @@ export function useViewportRows<R>({
           gridTemplateRows += `${currentHeight}px`;
         }
       }
+
+      return position;
     });
 
     const validateRowIdx = (rowIdx: number) => {


src/hooks/useViewportRows.ts
```diff
@@ -28,19 +28,52 @@ export function useViewportRows<R>({
       };
     }
 
-    let totalRowHeight = 0;
-    let gridTemplateRows = ' ';
     // Calcule the height of all the rows upfront. This can cause performance issues
     // and we can consider using a similar approach as react-window
     // https://github.com/bvaughn/react-window/blob/b0a470cc264e9100afcaa1b78ed59d88f7914ad4/src/VariableSizeList.js#L68
-    const rowPositions = rows.map((row) => {
+    let totalRowHeight = 0;
```
amanmahajan7, line 34: Can you add some tests for the new change and I can review after that 
https://github.com/adazzle/react-data-grid/blob/main/test/browser/rowHeight.test.ts#L39

nickolasKrayn: Yes, I conducted performance testing on my project and observed significant improvements.

With the previous implementation, handling ~50000 rows caused noticeable UI lag and performance issues due to the massive CSS string generation. I had to implement temporary workarounds to handle these large datasets.

After applying this optimization, the solution now handles even 100,000+ records smoothly without any performance degradation. The key insight is that in my typical use case, only 2 rows have unique sizes while all remaining rows share identical dimensions.

nickolasKrayn: > Can you add some tests for the new change and I can review after that https://github.com/adazzle/react-data-grid/blob/main/test/browser/rowHeight.test.ts#L39

Sure! I've added tests. All tests pass and the optimization is working correctly. Ready for review when you are.

src/hooks/useViewportRows.ts
```diff
@@ -28,19 +28,52 @@ export function useViewportRows<R>({
...
+        repeatCount = 1;
+      }
+
+      if (index === rows.length - 1) {
+        if (repeatCount > 1) {
+          gridTemplateRows += `repeat(${repeatCount}, ${currentHeight}px)`;
+        } else {
+          gridTemplateRows += `${currentHeight}px`;
+        }
+      }
```
amanmahajan7, line 71: What happens if the last row has a different height? Maybe we can add 2 more tests where the first and last rows heights are different.

nickolasKrayn: I've added those test cases. Thanks for the suggestion!

src/hooks/useViewportRows.ts
```diff
@@ -28,19 +28,52 @@ export function useViewportRows<R>({
...
     // and we can consider using a similar approach as react-window
     // https://github.com/bvaughn/react-window/blob/b0a470cc264e9100afcaa1b78ed59d88f7914ad4/src/VariableSizeList.js#L68
-    const rowPositions = rows.map((row) => {
+    let totalRowHeight = 0;
+    const rowPositions: { top: number; height: number }[] = [];
+    let gridTemplateRows = '';
+    let currentHeight: number | null = null;
+    let repeatCount = 0;
+
+    rows.forEach((row, index) => {
```
amanmahajan7:
```suggestion
    for (const row of rows) {
```

src/hooks/useViewportRows.ts
```diff
@@ -28,19 +28,52 @@ export function useViewportRows<R>({
...
+      };
+      rowPositions.push(position);
       totalRowHeight += currentRowHeight;
-      return position;
+
+      if (currentHeight === null) {
+        currentHeight = currentRowHeight;
+        repeatCount = 1;
+      } else if (currentHeight === currentRowHeight) {
+        repeatCount++;
```
amanmahajan7, line 53:
```suggestion
        // If the current row height is the same as the previous one, increment the repeat count
        repeatCount++;
```

src/hooks/useViewportRows.ts
```diff
@@ -28,19 +28,52 @@ export function useViewportRows<R>({
...
+      } else {
+        if (repeatCount > 1) {
+          gridTemplateRows += `repeat(${repeatCount}, ${currentHeight}px) `;
+        } else {
+          gridTemplateRows += `${currentHeight}px `;
+        }
+
+        currentHeight = currentRowHeight;
+        repeatCount = 1;
+      }
```
amanmahajan7, line 63:
```suggestion
      }
   }
```

src/hooks/useViewportRows.ts
amanmahajan7, line 71:
```suggestion
      if (repeatCount > 1) {
        gridTemplateRows += `repeat(${repeatCount}, ${currentHeight}px)`;
      } else {
        gridTemplateRows += `${currentHeight}px`;
      }
```

src/hooks/useViewportRows.ts
```diff
@@ -28,19 +28,52 @@ export function useViewportRows<R>({
...
+      if (index === rows.length - 1) {
+        if (repeatCount > 1) {
+          gridTemplateRows += `repeat(${repeatCount}, ${currentHeight}px)`;
+        } else {
+          gridTemplateRows += `${currentHeight}px`;
+        }
+      }
     });
 
+    gridTemplateRows = gridTemplateRows.trim();
```
amanmahajan7:
```suggestion
```
Is this needed?

src/hooks/useViewportRows.ts
amanmahajan7: Actually ignore this comment. My suggestion can result in setting null height

src/hooks/useViewportRows.ts
amanmahajan7: Same here

src/hooks/useViewportRows.ts
amanmahajan7: And here

src/hooks/useViewportRows.ts
nickolasKrayn: You're right! Removed the unnecessary trim(). Thanks for the sharp eye! 👀

src/hooks/useViewportRows.ts
nickolasKrayn: Done!

amanmahajan7 (commented):
LGTM. I will do some testing and merge this PR

nickolasKrayn:
> LGTM. I will do some testing and merge this PR

Great! Thank you!

src/hooks/useViewportRows.ts
```diff
@@ -28,17 +28,49 @@ export function useViewportRows<R>({
...
+    rows.forEach((row, index) => {
       const currentRowHeight = rowHeight(row);
-      const position = { top: totalRowHeight, height: currentRowHeight };
-      gridTemplateRows += `${currentRowHeight}px `;
+
+      const position = {
+        top: totalRowHeight,
+        height: currentRowHeight
+      };
+      rowPositions.push(position);
```
nstepien: We can keep using `rows.map()` if we move this to the end and `return position;`

test/browser/rowHeight.test.ts
```diff
@@ -51,3 +60,47 @@ test('rowHeight is function', async () => {
   await userEvent.keyboard('{Control>}{end}');
   expect(grid.scrollTop + grid.clientHeight).toBe(grid.scrollHeight);
 });
+
+test('rowHeight with repeat pattern - multiple identical heights', () => {
+  expectGridRows(() => 40, 'repeat(1, 35px) repeat(50, 40px)');
```
nstepien, line 65: Why do we always have `repeat(1, 35px)`?
I understand it's for the header row, but can't it just be `35px`?

amanmahajan7: https://github.com/adazzle/react-data-grid/blob/main/src/DataGrid.tsx#L1161
We are using `repeat` for header rows

src/hooks/useViewportRows.ts
nickolasKrayn: Thank you. I've updated the code.

nstepien (approved):
LGTM

PR #3866, opened 2025-09-16
Vitest: add WebKit browser instance

amanmahajan7:


nstepien:
Opened a related Playwright issue: https://github.com/microsoft/playwright/issues/39246

vite.config.ts
```diff
@@ -31,12 +31,21 @@ const resizeColumn: BrowserCommand<[name: string, resizeBy: number | readonly nu
...
+  to
+) => {
   await iframe.getByRole('gridcell', { name: from, exact: true }).click();
   await iframe.locator('.rdg-cell-drag-handle').hover();
   await page.mouse.down();
-  const toCell = iframe.getByRole('gridcell', { name: to, exact: true });
-  await toCell.hover();
+  await iframe.getByRole('gridcell', { name: to, exact: true }).hover();
+  if (project.name.includes('webkit')) {
+    // let React re-render after handleDragHandlePointerMove calls setDraggedOverRowIdx()
```
nstepien, line 44: Alternative would be to use `flushSync` perhaps

test/browser/utils.tsx
```diff
@@ -1,8 +1,11 @@
-import { page, userEvent, type Locator } from 'vitest/browser';
+import { page, server, userEvent, type Locator } from 'vitest/browser';
 
 import { DataGrid } from '../../src';
 import type { DataGridProps } from '../../src';
 
+// copy/paste do not work in webkit in CI
```
amanmahajan7, line 6: Worth testing in webdriver instead of playwright?

nstepien: Probably not, we'll catch any regressions locally, and I'd rather not set up yet another provider.

.github/workflows/publish.yml
```diff
@@ -49,7 +49,13 @@ jobs:
         run: git diff --exit-code website/routeTree.gen.ts
 
       - name: Install Playwright Browsers
-        run: npx playwright install chromium firefox
+        run: |
+          sudo tee /etc/apt/apt.conf.d/99-ci >/dev/null <<'EOF'
+          Acquire::Retries "0";
+          Acquire::http::Timeout "10";
+          Acquire::https::Timeout "10";
+          EOF
```
amanmahajan7, line 57: Fails without these commands?

src/DataGrid.tsx
```diff
@@ -694,6 +695,9 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
     }
 
     if (isCellEditable(activePosition) && isDefaultCellInput(event, onCellPaste != null)) {
+      // ensure cell is fully visible
+      scrollIntoView(cell);
```
amanmahajan7, line 699: Was this a bug in safari?

test/browser/column/renderCell.test.tsx
```diff
@@ -176,7 +176,12 @@ test('Cell should not steal focus when the focus is outside the grid and cell is
 
     return (
       <>
-        <button type="button" onClick={onClick}>
+        <button
+          type="button"
+          // tabIndex needed otherwise webkit will set focus on the <body> after clicking the button
```
amanmahajan7, line 181: 😕

vite.config.ts
amanmahajan7: How did you even find this bug 😄

vite.config.ts
```diff
@@ -169,6 +178,10 @@ export default defineConfig(({ isPreview }): ViteUserConfig => ({
           }),
           // TODO: remove when FF tests are stable
```
amanmahajan7, line 179: Wonder if this is needed anymore. We can check in a separate PR

amanmahajan7 (commented):
Thanks for fixing the issues. You can approve and merge :)

.github/workflows/publish.yml
nstepien: apt-get times out without these for some reason

src/DataGrid.tsx
nstepien: Yes and no, seems to be a different behavior in webkit when typing on a focused element

I think it's like this:
- element is out of view
- press enter
- chrome/firefox scroll to the cell
- render editor, focus input
- scroll to the input
  - since it's already in view in chrome/firefox, no issue
  - maybe webkit hasn't scrolled to the cell yet, so instead of scrolling to the cell it scrolls to the input only?

Just my interpretation

vite.config.ts
nstepien: with tests failing locally 🙃

vite.config.ts
nstepien: Still needed when I last tried

src/DataGrid.tsx
amanmahajan7: 👍 Does not hurt to keep it

nstepien (approved):

PR #3871, opened 2025-09-23
Rewrite some tests to reduce flakiness: part 1

amanmahajan7:
https://vitest.dev/guide/browser/assertion-api.html#assertion-api
> We recommend to always use expect.element when working with page.getBy* locators to reduce test flakiness.

https://vitest.dev/guide/browser/locators.html#nth
> Before resorting to nth, you may find it useful to use chained locators to narrow down your search. Sometimes there is no better way to distinguish than by element position; although this can lead to flake, it's better than nothing.

I think we should find `cell/headercell/row` using name instead of relying on index when possible

### DIFF
--- a/test/browser/headerRowClass.test.ts
+++ b/test/browser/headerRowClass.test.ts
@@ -1,21 +1,23 @@
+import { page } from '@vitest/browser/context';
+
 import type { Column } from '../../src';
 import { headerRowClassname } from '../../src/HeaderRow';
-import { getRow, setup } from './utils';
+import { setup } from './utils';
 
 interface Row {
   id: number;
 }
 
 const columns: readonly Column<Row>[] = [{ key: 'id', name: 'ID' }];
-const rows: readonly Row[] = [{ id: 0 }];
+const rows: readonly Row[] = [];
 
 test('headerRowClass is undefined', async () => {
   setup({
     columns,
     rows,
-    rowClass: undefined
+    headerRowClass: undefined
   });
-  const header = getRow('ID');
+  const header = page.getByRole('row');
   await expect.element(header).toHaveClass(headerRowClassname, { exact: true });
 });
 
@@ -25,6 +27,6 @@ test('headerRowClass is a string', async () => {
     rows,
     headerRowClass: 'my-header-row'
   });
-  const header = getRow('ID');
+  const header = page.getByRole('row');
   await expect.element(header).toHaveClass(`${headerRowClassname} my-header-row`, { exact: true });
 });


test/browser/rowSelection.test.tsx
```diff
@@ -49,120 +52,122 @@ function setup(initialRows = defaultRows) {
...
 test('should not select row when isRowSelectionDisabled returns true', async () => {
   page.render(
     <RowSelectionTest initialRows={defaultRows} isRowSelectionDisabled={(row) => row.id === 2} />
   );
   await toggleSelection(0);
-  testSelection(0, true);
-  await toggleSelection(1);
-  testSelection(1, false);
+  await testSelection(0, true);
+  await toggleSelection(1, false, true); // force click even if disabled
```
amanmahajan7, line 131: forcing click causes flaky test in FF.

vite.config.ts
```diff
@@ -41,6 +41,24 @@ const dragFill: BrowserCommand<[from: string, to: string]> = async (context, fro
   await page.mouse.up();
 };
 
+const scrollGrid: BrowserCommand<[{ scrollLeft?: number; scrollTop?: number }]> = async (
```
amanmahajan7, line 44: https://playwright.dev/docs/input#scrolling

test/browser/column/renderEditCell.test.tsx
```diff
@@ -258,20 +258,19 @@ describe('Editor', () => {
...
       }
 
       page.render(<EditorTest gridRows={rows} />);
 
-      await userEvent.dblClick(getCellsAtRowIndex(0)[1]);
+      await userEvent.dblClick(getCell('name0'));
       await userEvent.keyboard('abc');
 
-      await scrollGrid({ scrollTop: 1500 });
-      expect(getCellsAtRowIndex(40)[1]).toHaveTextContent(/^40$/);
```
amanmahajan7, line 270: For some reason current scrolling logic is flaky in FF 
https://github.com/Comcast/react-data-grid/actions/runs/17952854675/job/51056628087

Hopefully the new `commands.scrollGrid` will help

test/browser/column/renderEditCell.test.tsx
```diff
@@ -258,20 +258,19 @@ describe('Editor', () => {
...
-      await userEvent.click(getCellsAtRowIndex(40)[1]);
```
amanmahajan7, line 271: This can still fail when the grid is properly scrolled. Using locators fixes it

vite.config.ts
```diff
@@ -41,6 +41,24 @@ const dragFill: BrowserCommand<[from: string, to: string]> = async (context, fro
...
+  context,
+  { scrollLeft, scrollTop }
+) => {
+  const frame = await context.frame();
+  await frame.getByRole('grid').evaluate(
```
nstepien, line 49: Does this need to be a command? Can we not run `.evaluate()` in tests scope?

test/browser/utils.tsx
```diff
@@ -40,6 +40,34 @@ export function getTreeGrid() {
...
+export function getHeaderCell(name: string) {
+  return page.getByRole('columnheader', { name });
+}
+
+export function getHeaderCellsNew(...names: readonly string[]) {
+  return names.map(getHeaderCell);
+}
+
+export function getRow(name: string) {
+  return page.getByRole('row', { name });
```
nstepien, line 52: Do rows even have a "name"?

test/browser/utils.tsx
```diff
@@ -40,6 +40,34 @@ export function getTreeGrid() {
   return page.getByRole('treegrid');
 }
 
+export function getHeaderCell(name: string) {
+  return page.getByRole('columnheader', { name });
```
nstepien: Should we add `exact: true`?

vite.config.ts
amanmahajan7: It is exposed 
image

test/browser/utils.tsx
amanmahajan7: Yeah. 
image

PR #3873, opened 2025-09-24
use `useSyncExternalStore` to subscribe to `ResizeObserver`

amanmahajan7:
https://www.epicreact.dev/use-sync-external-store-demystified-for-practical-react-development-w5ac0

### DIFF
--- a/src/hooks/useGridDimensions.ts
+++ b/src/hooks/useGridDimensions.ts
@@ -23,7 +23,7 @@ export function useGridDimensions() {
 
     const resizeObserver = new ResizeObserver((entries) => {
       const size = entries[0].contentBoxSize[0];
-      const { clientHeight, offsetHeight } = gridRef.current!;
+      const { clientHeight, offsetHeight } = element;
       dimensionsRef.current = [size.inlineSize, size.blockSize, offsetHeight - clientHeight];
       callback();
     });
--- a/src/hooks/useGridDimensions.ts
+++ b/src/hooks/useGridDimensions.ts
@@ -1,4 +1,4 @@
-import { useCallback, useRef, useState, useSyncExternalStore } from 'react';
+import { useCallback, useLayoutEffect, useRef, useSyncExternalStore } from 'react';
 
 type Dimensions = [inlineSize: number, blockSize: number, horizontalScrollbarHeight: number];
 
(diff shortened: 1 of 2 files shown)


src/hooks/useGridDimensions.ts
```diff
@@ -1,46 +1,48 @@
...
+  return defaultDimensions;
+}
+
+export function useGridDimensions() {
+  const gridRef = useRef<HTMLDivElement>(null);
+  const dimensionsRef = useRef(defaultDimensions);
 
-    const { clientWidth, clientHeight, offsetWidth, offsetHeight } = gridRef.current!;
+  const subscribe = useCallback((callback: () => void) => {
+    const element = gridRef.current!;
```
nstepien: How do we know `.current` is not nullish here?

src/hooks/useGridDimensions.ts
```diff
@@ -1,46 +1,48 @@
...
-    resizeObserver.observe(gridRef.current!);
+    resizeObserver.observe(element);
 
     return () => {
       resizeObserver.disconnect();
     };
   }, []);
 
+  const getSnapShot = useCallback(() => {
+    return dimensionsRef.current;
```
nstepien: Don't we read `.current` during render since we call `getSnapShot()` to intialize the `useSyncExternalStore`?

src/hooks/useGridDimensions.ts
amanmahajan7: This is the place to setup subscription so I assumed `.current` is not null. I can double check

src/hooks/useGridDimensions.ts
amanmahajan7: Not sure 🤔 `react-complier` did not complain. I can add a class
https://github.com/Comcast/react-data-grid/pull/3873/commits/d13b83fef0f30c44bcc047e8e0653996147159b3#diff-2dbc2c9db9fe9009d9fe02a7a37f929629e684c70a2d02fff405bb30499ccb24R11

src/hooks/useGridDimensions.ts
amanmahajan7: image

`subscribe` is called after `useLayoutEffect`

src/hooks/useGridDimensions.ts
```diff
@@ -1,46 +1,63 @@
...
-export function useGridDimensions() {
-  const gridRef = useRef<HTMLDivElement>(null);
-  const [inlineSize, setInlineSize] = useState(1);
-  const [blockSize, setBlockSize] = useState(1);
-  const [horizontalScrollbarHeight, setHorizontalScrollbarHeight] = useState(0);
+type Dimensions = [inlineSize: number, blockSize: number, horizontalScrollbarHeight: number];
 
-  useLayoutEffect(() => {
-    const { ResizeObserver } = window;
+const defaultDimensions: Dimensions = [1, 1, 0];
```
amanmahajan7, line 5: Wonder if we should change the default to a higher value and get rid of initial width/height logic in the `subscribe` function

src/hooks/useGridDimensions.ts
```diff
@@ -1,46 +1,63 @@
...
-
-      // we use flushSync here to avoid flashing scrollbars
-      flushSync(() => {
-        setInlineSize(size.inlineSize);
-        setBlockSize(size.blockSize);
-        setHorizontalScrollbarHeight(offsetHeight - clientHeight);
-      });
+      const { clientHeight, offsetHeight } = element;
+      this.#dimensions = [size.inlineSize, size.blockSize, offsetHeight - clientHeight];
+      callback();
```
nstepien, line 22: Should we avoid calling `callback()` if all three values have not changed?
Overall I feel like this is might not be a good fit, the layout effect seems simpler to me.

amanmahajan7: Yeah, I am on the fence too leaning towards keeping `useSyncExternalStore `

Pros
- `flushSync` is no longer needed
- `getServerSnapshot` handles server side rendering so we do not need to check `if (ResizeObserver == null);`
- Seems like `useSyncExternalStore` is designed for this specific use case

Cons
- Seems a bit more complex 
- Not entirely sure if there can be issues mixing `ref` and `useSyncExternalStore `
- Need to manually check values to avoid re-render but I don't know if resizer observer is fired if the dimensions have not changed.

amanmahajan7: I will close this PR for now. We can revisit if needed

PR #3874, opened 2025-09-25
Added options for enableVirtualization

### LINKED ISSUE #3619: Allow to specify enableVirtualization for rows or columns
## Use Case 

It would be highly beneficial to allow more control over the `enableVirtualization` option. Currently, enabling virtualization affects both rows and columns automatically. However, it would be great to have the ability to specify the scope of virtualization, such as rows, columns, or all, instead of just a simple true/false toggle.

This feature could also address the issue with frozen columns on the right (#2873), enabling the freezing of columns through CSS alone. Now, applying a CSS-based solution conflicts with virtualization, as enabling it prevents the last column from rendering correctly, thereby breaking the freezing behaviour.

## Proposed Solution

The proposal is straightforward: allow the `enableVirtualization` option to accept something like `columns | rows | all` as value, and adjust the behaviour accordingly. This would provide more flexibility in configuring virtualization while addressing current limitations.

apcqa:
Changed type of enableVirtualization property on DataGrid to accept either a boolean, undefiend or a VirtualizationOptions object
which allows the user to enable or disable virtualization for rows and or columns and also specify the overscan threshold for rows (default 4).

Change maintains compatibility with existing boolean type and defaults.

Fixes #3619


src/DataGrid.tsx
```diff
@@ -219,7 +221,9 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
    * Toggles and modes
    */
   /** @default true */
-  enableVirtualization?: Maybe<boolean>;
+  enableVirtualization?: Maybe<boolean | VirtualizationOptions>;
```
amanmahajan7, line 224:
```suggestion
  enableVirtualization?: Maybe<'row' | 'column' | boolean>;
```
I think we can keep it simple for now.

amanmahajan7 (commented):
We need to add extensive tests for the new functionality

PR #3877, opened 2025-09-29
Update eslint dependencies 

amanmahajan7:


biome.json
```diff
@@ -385,6 +385,18 @@
           }
         }
       }
+    },
+    {
+      "assist": { "actions": { "source": { "organizeImports": "on" } } },
+      "includes": ["**/*.js"]
```
amanmahajan7: Should we not have this turned on for `*.js` files? Also, should we remove https://github.com/Comcast/react-data-grid/blob/main/.prettierignore#L1-L3 and let prettier handle everything. Wdyt?

eslint.config.js
```diff
@@ -1,3 +1,5 @@
+import markdown from '@eslint/markdown';
+import reactX from '@eslint-react/eslint-plugin';
```
amanmahajan7, line 1: https://eslint-react.xyz/docs/changelog#v200-2025-09-26

PR #3880, opened 2025-10-03
Rewrite some tests to reduce flakiness: part 3

amanmahajan7:


test/browser/column/resizable.test.tsx
```diff
@@ -349,9 +325,11 @@ async function testGridTemplateColumns(chrome: string, firefox: string, firefoxC
   if (navigator.userAgent.includes('Chrome')) {
     await expect.element(grid).toHaveStyle({ gridTemplateColumns: chrome });
   } else {
-    expect((grid.element() as HTMLDivElement).style.gridTemplateColumns).toBeOneOf([
-      firefox,
-      firefoxCI
-    ]);
+    await vi.waitFor(() => {
```
amanmahajan7, line 328: fails if I comment https://github.com/Comcast/react-data-grid/blob/main/src/hooks/useGridDimensions.ts#L17-L25

test/browser/renderers.test.tsx
```diff
@@ -116,21 +116,21 @@ function setupContext<R, SR, K extends React.Key>(props: DataGridProps<R, SR, K>
 test('fallback defined using renderers prop with no rows', async () => {
   setup({ columns, rows: noRows, renderers: { noRowsFallback: <NoRowsFallback /> } });
 
-  expect(getRows()).toHaveLength(0);
+  await testLength(getRowsNew(), 1);
```
amanmahajan7: I think using `page.getByRole('row').elements().slice(1)` causes confusion IMO. We can just check the total row count. Also, we can await now

test/setupBrowser.ts
```diff
@@ -1,9 +1,16 @@
...
 // vitest-browser-react also automatically injects render method on the page
 // need to import it so TypeScript can pick up types
 import 'vitest-browser-react';
 
-vitestReactConfigure({
+configure({
   reactStrictMode: true
 });
+
+locators.extend({
```
amanmahajan7, line 12: https://main.vitest.dev/guide/browser/locators.html#custom-locators

test/browser/utils.tsx
```diff
@@ -48,12 +48,8 @@ export function getHeaderCellsNew(...names: readonly string[]) {
...
 
-export function getRow(name: string) {
-  return page.getByRole('row', { name, exact: true });
-}
-
-export function getRowsNew(...names: readonly string[]) {
-  return names.map(getRow);
+export function getRowByCellOrCellName(cell: string | Locator) {
+  return page.getByRole('row').filter({ has: typeof cell === 'string' ? getCell(cell) : cell });
 }
```
nstepien: wdyt?
```suggestion
export function getRowByCell(cell: Locator) {
  return page.getByRole('row').filter({ has: cell });
}

export function getRowByCellName(cell: string) {
  return getRowByCell(getCell(cell));
}
```

nstepien (approved):

PR #3883, opened 2025-10-07
Remove `clsx` dependency

amanmahajan7:
`clsx` is quite a small utility and we do not need all the features so I think we can write our own simplified version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,9 +48,6 @@
     "prepublishOnly": "npm install && node --run build",
     "postpublish": "git push --follow-tags origin HEAD"
   },
-  "dependencies": {
-    "clsx": "^2.0.0"
-  },
   "devDependencies": {
     "@babel/preset-typescript": "^7.27.1",
     "@biomejs/biome": "2.2.5",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -72,6 +69,7 @@
     "@vitest/eslint-plugin": "^1.3.4",
     "@wyw-in-js/rollup": "^0.7.0",
     "@wyw-in-js/vite": "^0.7.0",
+    "clsx": "^2.1.1",
```
amanmahajan7, line 72: Kept this for website

src/utils/styleUtils.ts
```diff
@@ -45,11 +44,33 @@ export function getCellStyle<R, SR>(
...
+        for (const key in arg) {
+          if (arg[key]) {
+            classname += ` ${key}`;
+          }
+        }
+      }
+    }
+  }
+
+  return classname.trimStart();
```
nstepien, line 66: also valid
```suggestion
  return classname.slice(1);
```

nstepien (approved):

PR #3886, opened 2025-10-09
Setup React Compiler

amanmahajan7:


amanmahajan7:
We can try https://npmx.dev/package/oxc-transform-react

PR #3915, opened 2025-11-21
Publish via workflow with trusted publishing

nstepien:
https://docs.npmjs.com/trusted-publishers

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
       id-token: write
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
 


.github/workflows/publish.yml
```diff
@@ -0,0 +1,77 @@
...
+          git config user.name "${{ github.actor }}"
+          git config user.email "${{ github.actor_id }}+${{ github.actor }}@users.noreply.github.com"
+          npm version ${{ inputs.version }}
+
+      # https://docs.npmjs.com/trusted-publishers
+      - name: Publish to npm
+        run: npm publish
+
+      - name: Push commit and tag
+        run: git push origin main --follow-tags
```
nstepien, line 72: I will have to double-check our branch rules, we need to allow directly pushing to `main`.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Create GitHub Release
+        run: gh release create v${{ inputs.version }} --title "v${{ inputs.version }}" --generate-notes
```
nstepien, line 75: We'll create immutable releases now, we'll have to write changelogs manually.
We can improve this in the future to automatically add changelogs from files, but this will do for now.

Image

.github/workflows/publish.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: node --run test
+        timeout-minutes: 4
+
+      - name: Visual regression test
+        run: node --run visual
+
+      - name: Update version
+        run: |
+          git config user.name "${{ github.actor }}"
+          git config user.email "${{ github.actor_id }}+${{ github.actor }}@users.noreply.github.com"
```
amanmahajan7, line 64: `@users.noreply.github.com` can this be any email id?

amanmahajan7 (approved):
LGTM. Have we restricted token access?
https://docs.npmjs.com/trusted-publishers#recommended-restrict-token-access-when-using-trusted-publishers

.github/workflows/publish.yml
nstepien: It has to be `@users.noreply.github.com`

PR #3927, opened 2025-12-06
Expose unselectCell on DataGrid handle

marcellofuschi:
Working with `DataGrid` I found myself needing to unselect the selected cell (when the user moves focus to an external element)


marcellofuschi:
The intention is to clear the selected position because we base some other UI behavior on the selected cell, and when not focused we'd like to avoid triggering that behavior

PR #3967, opened 2026-02-21
Multiple renames for clarity and consistency

nstepien:
I've renamed a few components/type for clarity and consistency.
- Removed "formatter" from names, it's legacy nomenclature
- Removed "renderer" from some names, virtually all React components are "renderers" so it's redundant.
  It's like naming components `*Component`.
- To help differentiate between the `renderers.renderCell` and `column.renderCell`,
  I've changed some names from `*CellProps` to `*CellContentProps` for example
- Removed some `default` exports so we use the same named import everywhere

Breaking changes:
- cell content:
  - `RenderCellProps` -> `RenderCellContentProps`
  - `RenderHeaderCellProps` -> `RenderHeaderCellContentProps`
  - `RenderSummaryCellProps` -> `RenderSummaryCellContentProps`
  - `RenderGroupCellProps` -> `RenderGroupCellContentProps`
  - `RenderEditCellProps` -> `RenderEditCellContentProps`
- `CellRendererProps` -> `RenderCellProps`
  - while this is consistent with everything else now (`Render*Props` pattern), it'll conflict with the older `RenderCellProps` type name, need to be careful when updating RDG
- `SelectCellFormatter` -> `SelectCheckbox`


src/TreeDataGrid.tsx
```diff
@@ -115,27 +115,29 @@ export function TreeDataGrid<R, SR = unknown, K extends Key = Key>({
   const [groupedRows, rowsCount] = useMemo(() => {
     if (groupBy.length === 0) return [undefined, rawRows.length];
 
-    const groupRows = (
+    function groupRows(
       rows: readonly R[],
-      [groupByKey, ...remainingGroupByKeys]: readonly string[],
+      groupByIndex: number,
```
nstepien, line 120: Refactored this so we don't create temp arrays

src/hooks/useLatestFunc.ts
```diff
@@ -3,18 +3,25 @@ import { useCallback, useLayoutEffect, useRef } from 'react';
...
+
+export function useLatestFunc<Args extends unknown[]>(
+  fn: (...args: Args) => void
+): (...args: Args) => void;
+
+export function useLatestFunc<Args extends unknown[]>(
+  fn: Maybe<(...args: Args) => void>
+): Maybe<(...args: Args) => void>;
+
+export function useLatestFunc<Args extends unknown[]>(fn: Maybe<(...args: Args) => void>) {
```
nstepien, line 15: We can implement this without `any`/`@ts-expect-error`!

src/utils/eventUtils.ts
```diff
@@ -4,17 +4,17 @@ export function createCellEvent<E extends React.SyntheticEvent<HTMLDivElement>>(
...
+      value() {
+        return defaultPrevented;
+      }
     }
-  };
-
-  Object.setPrototypeOf(cellEvent, Object.getPrototypeOf(event));
-
-  return cellEvent;
+  });
```
nstepien, line 19: This creates a new object that inherits from the `event` object, and adds `preventGridDefault`/`isGridDefaultPrevented` property descriptors

WDYT?

src/DataGrid.tsx
```diff
@@ -469,15 +469,15 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
    */
   const handleColumnResizeLatest = useLatestFunc(handleColumnResize);
   const handleColumnResizeEndLatest = useLatestFunc(handleColumnResizeEnd);
-  const onColumnsReorderLastest = useLatestFunc(onColumnsReorder);
```
nstepien, line 501: "Lastest" 🙈

src/DataGrid.tsx
```diff
@@ -962,7 +962,8 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
     const isLastRow = rowIdx === maxRowIdx;
     const columnWidth = getColumnWidth(column);
     const colSpan = column.colSpan?.({ type: 'ROW', row: rows[rowIdx] }) ?? 1;
-    const { insetInlineStart, ...style } = getCellStyle(column, colSpan);
+    const style = getCellStyle(column, colSpan);
```
nstepien, line 960: -1 object

PR #3968, opened 2026-02-22
useGridDimensions: useSyncExternalStore implementation

nstepien:
`useSyncExternalStore` seems like the better choice to handle this, especially when concurrent rendering is involved.


test/failOnConsole.ts
```diff
@@ -5,13 +5,6 @@ beforeAll(() => {
   globalThis.console = {
     ...console,
     error(...params) {
-      if (
-        params[0] instanceof Error &&
-        params[0].message === 'ResizeObserver loop completed with undelivered notifications.'
-      ) {
-        return;
```
nstepien: Just hoping this will be resolved 🤞

src/hooks/useGridDimensions.ts
```diff
@@ -1,38 +1,99 @@
...
+const idToTargetMap = new Map<string, HTMLDivElement>();
+// use an unmanaged WeakMap so we preserve the cache even when
+// the component partially unmounts via Suspense or Activity
+const sizeMap = new WeakMap<HTMLDivElement, ResizeObserverSize>();
+const subscribers = new Map<string, () => void>();
+
+// don't break in Node.js (SSR), jsdom, and environments that don't support ResizeObserver
+const resizeObserver =
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  globalThis.ResizeObserver == null ? null : new ResizeObserver(resizeObserverCallback);
```
nstepien, line 17: We'll have 1 RO instance for all rendered grids on the page, so all resize events will be batched together -> improved perf?

src/hooks/useGridDimensions.ts
```diff
@@ -1,38 +1,99 @@
...
+  sizeMap.set(target, size);
+  subscribers.get(id)?.();
+}
+
+function getServerSnapshot(): ResizeObserverSize {
+  return initialSize;
+}
 
 export function useGridDimensions() {
+  const id = useId();
```
nstepien: Using an id to sync things.
I tried using `gridRef.current` at first, but `getSnapshot` gets called during render, so it would probably not be safe.

src/hooks/useGridDimensions.ts
```diff
@@ -1,38 +1,99 @@
...
-    const { clientWidth, clientHeight } = gridRef.current!;
+  const getSnapshot = useCallback((): ResizeObserverSize => {
+    if (idToTargetMap.has(id)) {
+      const target = idToTargetMap.get(id)!;
+      if (sizeMap.has(target)) {
+        return sizeMap.get(target)!;
+      }
+    }
+    return initialSize;
+  }, [id]);
```
nstepien: `getSnapshot` gets called multiple times during render, so it has to be as fast as possible, and must avoid returning different values.

amanmahajan7 (commented):
I wonder if it is worth using a single RO. Most of the pages should have 1 grid and it would be nice if everything is contained. I am just little concerned about odd bugs because of the new approach. We can go ahead and review if you prefer a single RO

src/hooks/useGridDimensions.ts
```diff
@@ -1,38 +1,86 @@
...
+import { useCallback, useLayoutEffect, useRef, useSyncExternalStore, type RefObject } from 'react';
+
+const initialSize: ResizeObserverSize = {
+  inlineSize: 1,
+  blockSize: 1
+};
+
+// use an unmanaged WeakMap so we preserve the cache even when
+// the component partially unmounts via Suspense or Activity
+const sizeMap = new WeakMap<RefObject<HTMLDivElement | null>, ResizeObserverSize>();
```
nstepien, line 10: I took great care to ensure this works great even if the components unmounts but the div isn't deleted.
The updated hook is more complicated than before, but should behave great with concurrent rendering :v:

src/hooks/useGridDimensions.ts
```diff
@@ -1,38 +1,86 @@
...
+    return () => {
+      subscribers.delete(ref);
+    };
+  }, []);
+
+  const getSnapshot = useCallback((): ResizeObserverSize => {
+    // ref.current is null during the initial render, when suspending, or in <Activity mode="hidden">.
+    // We use ref as key instead to access stable values regardless of rendering state.
+    return sizeMap.has(ref) ? sizeMap.get(ref)! : initialSize;
+  }, []);
```
nstepien: `getSnapshot` gets called multiple times during render, so it has to be as fast as possible, and must avoid returning different values.

nstepien (commented):
Re-applying outdated comments

src/hooks/useGridDimensions.ts
```diff
@@ -1,39 +1,85 @@
...
+    },
+    [gridRef]
+  );
 
-    setInlineSize(clientWidth);
-    setBlockSize(clientHeight);
+  const getSnapshot = useCallback((): ResizeObserverSize => {
+    // ref.current is null during the initial render, when suspending, or in <Activity mode="hidden">.
+    // We use ref as key instead to access stable values regardless of rendering state.
+    return sizeMap.has(gridRef) ? sizeMap.get(gridRef)! : initialSize;
```
amanmahajan7:
```suggestion
    return sizeMap.get(gridRef) ?? initialSize;
```
I am fine with either

src/hooks/useGridDimensions.ts
```diff
@@ -1,39 +1,85 @@
...
+        inlineSize: target.clientWidth,
+        blockSize: target.clientHeight
       });
-    });
-    resizeObserver.observe(gridRef.current!);
+    }
 
     return () => {
-      resizeObserver.disconnect();
+      resizeObserver?.unobserve(target);
```
amanmahajan7, line 82: should we call `targetToRefMap.remove(target)`; to be explicit

nstepien: Mmh, in general I'd not manage WeakMaps too much, that way if the code becomes more complicated we won't run into an bug where we expect the mapping to still exist.

FYI layout effects will clean up when suspending or when it's under `<Activity mode='hidden'>`, but the actually HTML elements will not be removed, and they will not be re-created when the component re-mounts/renders.

src/hooks/useGridDimensions.ts
```diff
@@ -1,39 +1,85 @@
...
+// don't break in Node.js (SSR), jsdom, and environments that don't support ResizeObserver
+const resizeObserver =
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  globalThis.ResizeObserver == null ? null : new ResizeObserver(resizeObserverCallback);
+
+function resizeObserverCallback(entries: ResizeObserverEntry[]) {
+  for (const entry of entries) {
+    const target = entry.target as HTMLDivElement;
+
+    if (targetToRefMap.has(target)) {
```
amanmahajan7, line 23: Is it worth removing this map and just looping through `sizeMap`?

amanmahajan7 (approved):

src/hooks/useGridDimensions.ts
nstepien: Cannot loop through WeakMaps/WeakSets
image

amanmahajan7: TIL

nstepien: Yup, it's because weak references cannot be held strongly -> cannot find all the weak references -> cannot iterate through them

PR #3986, opened 2026-02-26
DataGrid: split scroll/scroll to position/should focus position handling into hooks

nstepien:
`src/DataGrid.tsx` is quite fat, so I'm trying to slim it down.
This improves co-locality too!


src/hooks/useScrollState.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  const getSnapshot = useCallback((): ScrollState => {
+    // gridRef.current is null during initial render, suspending, or <Activity mode="hidden">
+    // to avoid returning a different state in those cases,
+    // we key the ref object instead of the element itself
+    return scrollStateMap.get(gridRef) ?? initialScrollState;
+  }, [gridRef]);
+
+  return useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot);
```
nstepien, line 65: Scroll state now uses `useSyncExternalStore` instead of 2x`useState`+`flushSync`

src/hooks/useShouldFocusPosition.ts
```diff
@@ -0,0 +1,26 @@
...
+import { focusCell, focusRow } from '../utils';
+
+export function useShouldFocusPosition({
+  gridRef,
+  selectedPosition
+}: {
+  gridRef: React.RefObject<HTMLDivElement | null>;
+  selectedPosition: { idx: number; rowIdx: number };
+}) {
+  const shouldFocusPositionRef = useRef(false);
```
nstepien: This used to be a `useState`, but an eslint rule complained about setting state in effect.
Strangely it did not complain when the same code was in `DataGrid`, maybe because `DataGrid` is too big and complicated so it confuses static analysis algos.
Though to make sure the effect works, I had to change the `selectedPosition.idx` dependency to `selectedPosition`.
But now we 1 fewer state that can trigger re-renders!

src/hooks/useScrollToPosition.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export function useScrollToPosition({ gridRef }: Props) {
+  const [scrollToPosition, setScrollToPosition] = useState<PartialPosition | null>(null);
+
+  return {
+    setScrollToPosition,
+    scrollToPositionElement: scrollToPosition && (
+      <div
+        ref={(div) => {
```
nstepien, line 22: before: `useRef`+`useLayoutEffect`
now: 1 function
Seems to work just as well.
Since it doesn't need any hooks anymore, I've also inlined the component

src/hooks/useScrollState.ts
```diff
@@ -0,0 +1,66 @@
...
+const initialScrollState: ScrollState = {
+  scrollTop: 0,
+  scrollLeft: 0
+};
+
+function getServerSnapshot() {
+  return initialScrollState;
+}
+
+const scrollStateMap = new WeakMap<React.RefObject<HTMLDivElement | null>, ScrollState>();
```
amanmahajan7, line 19: Is there a benefit of using a map instead of setting a local state inside the component?

src/hooks/useShouldFocusPosition.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { Position } from '../types';
+
+export function useShouldFocusPosition({
+  gridRef,
+  activePosition
+}: {
+  gridRef: React.RefObject<HTMLDivElement | null>;
+  activePosition: Position;
+}) {
+  const shouldFocusPositionRef = useRef(false);
```
amanmahajan7: Is it better to use state? little concerned that we have to manually handle the dependencies in the effect.

src/DataGrid.tsx
```diff
@@ -1034,7 +1019,7 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
     const closeOnExternalRowChange = column.editorOptions?.closeOnExternalRowChange ?? true;
 
     const closeEditor = (shouldFocus: boolean) => {
-      setShouldFocusPosition(shouldFocus);
+      shouldFocusPositionRef.current = shouldFocus;
       setActivePosition(({ idx, rowIdx }) => ({ idx, rowIdx, mode: 'ACTIVE' }));
```
amanmahajan7: We are already setting state so react will batch should focus state. Wdyt?

src/hooks/useScrollState.ts
nstepien: Are you suggesting we keep using `useState` instead of `useSyncExternalStore`?

src/hooks/useShouldFocusPosition.ts
nstepien: That's what tests are for!
We have a few tests that fail when the dependency is `activePosition.idx` instead of `activePosition`.
I think `ref` makes sense here.

src/DataGrid.tsx
nstepien: But then we'd have to set the state to `false` again once the layout effect runs.

src/hooks/useScrollState.ts
nstepien:
```tsx
function Test() {
  const [resolvers, setResolvers] = useState<PromiseWithResolvers<void>>(() =>
    Promise.withResolvers()
  );

  return (
    <>
      <button type="button" onClick={() => setResolvers(Promise.withResolvers())}>
        New promise
      </button>
      <button type="button" onClick={() => resolvers.resolve()}>
        Resolve
      </button>

      <Suspense fallback="Loading...">
        <Inner promise={resolvers.promise} />
      </Suspense>
    </>
  );
}


function Inner({ promise }: { promise: Promise<void> }) {
  useState(() => {
    console.log('init state');
  });

  use(promise);

  return 'ok';
}
```

In this example the `useState` init function in `<Inner>` is called twice, then once more after the promise resolves. (multiply by two in StrictMode)
Once React commits a render tree to the DOM though, then the init function is not called again.

This also happens when the suspense is triggered deeper in the tree:
```tsx
function Inner({ promise }: { promise: Promise<void> }) {
  useState(() => {
    console.log('init state');
  });

  // use(promise);

  return (
    <>
      <InnerSide />
      <InnerDeep promise={promise} />
    </>
  );
}

function subscribe() {
  console.log('subscribe');

  return () => {
    console.log('unsubscribe');
  };
}

function getSnapshot() {
  console.log('get snapshot');
}

function InnerSide() {
  useState(() => {
    console.log('init state side');
  });

  useSyncExternalStore(subscribe, getSnapshot);

  return 'side';
}

function InnerDeep({ promise }: { promise: Promise<void> }) {
  useState(() => {
    console.log('init state deep');
  });

  use(promise);

  return 'ok';
}
```
React only stops re-initializing states when it renders the tree under `<Suspense>` to the DOM (as I understand it).
So `<DataGrid>` could be affected if anything suspends the render tree it is in.
Also, `subscribe` is called just after `useLayoutEffect`s, but it is not cleaned up if the component suspends again.
`useTransition` and `useDeferredValue` also trigger concurrent rendering which may affect all this.

All that to say... I guess it's slightly better to have a single WeakMap?

src/hooks/useShouldFocusPosition.ts
nstepien: Ended up moving this into `useActivePosition`, and found a more elegant solution

src/DataGrid.tsx
amanmahajan7: That is a good point.

src/hooks/useActivePosition.ts
```diff
@@ -123,14 +127,25 @@ export function useActivePosition<R, SR>({
...
       };
       setActivePosition(newPosition);
-      setShouldFocusPosition(false);
+      setPositionToFocus(null);
       ({ resolvedActivePosition, validatedPosition } = getResolvedValues(newPosition));
     }
   }
 
+  useLayoutEffect(() => {
+    if (positionToFocus !== null) {
```
amanmahajan7: Do we not need to reset `positionToFocus` after focus?

src/hooks/useScrollState.ts
amanmahajan7: Does this mean a single WeakMap would be more performant or prevent some edge cases?

src/hooks/useScrollToPosition.tsx
amanmahajan7: So this works because we recreate ref callback  on each render?

src/DataGrid.tsx
```diff
@@ -189,8 +193,6 @@ export interface DataGridProps<R, SR = unknown, K extends Key = Key> extends Sha
   >;
   /** Function called whenever the active position is changed */
   onActivePositionChange?: Maybe<(args: PositionChangeArgs<NoInfer<R>, NoInfer<SR>>) => void>;
-  /** Callback triggered when the grid is scrolled */
-  onScroll?: Maybe<(event: React.UIEvent<HTMLDivElement>) => void>;
```
amanmahajan7, line 193: 👍

amanmahajan7 (commented):
A few more tiny comments

src/hooks/useActivePosition.ts
nstepien: No, to avoid 1 re-render.
I've added a ref tho to avoid re-focusing elements after the layout effect re-mounts

src/hooks/useScrollState.ts
nstepien: It's simple and potentially avoids states reinitialization due to suspense/activity.
How would you rather this be implemented?

src/hooks/useScrollToPosition.tsx
nstepien: Correct.
Although it might not work 100% well with auto-sized columns + smooth scrolling.
With smooth scrolling it may take >1 frame before the grid starts scrolling, so the condition passes immediately.
We can revisit later or I can revert. Not sure this worked well before either.

src/hooks/useActivePosition.ts
amanmahajan7: Should we add a test? I though we had test to prevent focus stealing

src/hooks/useScrollState.ts
amanmahajan7: Not sure. Let's go with what we have.

src/hooks/useActivePosition.ts
nstepien: We don't have any tests for suspense/activity.
I can add one if you want

amanmahajan7: I think it is okay. I got little confused on why we were resetting the state previously. The previous deps were `[shouldFocusPosition, activePositionIsRow, gridRef])` so it would rerun when `activePositionIsRow` is changed. But now we only have `[positionToFocus, gridRef])` which is more robust

amanmahajan7 (approved):

PR #4005, opened 2026-03-16
Refactor column width and resizing handling

### LINKED ISSUE #4120: [Bug] Duplicate frozen column elements in DOM and key collision warning when enableVirtualization={false}
**Describe the bug**
When enableVirtualization is set to false, frozen/pinned columns are rendered twice in the DOM structure.

While the second set of frozen columns may not be visibly apparent on the screen in regular view, it causes duplicate key warnings (Warning: Encountered two children with the same key) in the console. Furthermore, when exporting the grid to PDF (where enableVirtualization={false} is typically used to render all items), these duplicated frozen columns appear twice side-by-side in the resulting PDF document.

**To Reproduce**
1. Render a grid containing frozen/pinned columns.
2. Set enableVirtualization={false} on the grid component.
3. Open the browser DevTools and inspect the DOM structure for the grid header/row elements.
4. Check the browser console for React key collision warnings.
5. Export to PDF (or trigger print view) with enableVirtualization={false}.

**Expected behavior**
1. Frozen columns should only be rendered once in the DOM.
2. No React key duplicate warnings should occur in the console.
3. Exported PDFs should only display frozen columns once.

**Actual behavior**
1. Inspecting the DOM reveals that the frozen column elements are duplicated.
2. Console logs a React warning about duplicate keys for the frozen column elements.
3. Exported PDF output shows the frozen columns printed twice side-by-side.

**Environment**
1. Library Version: 7.0.0-beta.61
2. React Version: 19.2.0
3. Browser/OS: Chrome / Windows 11

**Additional Context**
It appears that when virtualization is disabled (enableVirtualization={false}), the internal rendering logic duplicates the frozen column container/cells alongside the main column list, resulting in duplicate DOM nodes and key conflicts.

nstepien:
- Consolidated column widths management into the reworked `useColumnWidths` hook
- measuring cell widths are measured using a `ResizeObserver`
- all cells now have `minWidth` and `maxWidth` set to resolve issues when some columns have a `maxWidth`
  - that way all `auto` columns will correctly fill the grid's width when one or more columns are constrained by their `maxWidth`
- less computations during render
- -1 layout effect


src/hooks/useGridDimensions.ts
```diff
@@ -1,10 +1,12 @@
-import { useLayoutEffect, useRef, useState } from 'react';
+import { useDeferredValue, useLayoutEffect, useRef, useState } from 'react';
 import { flushSync } from 'react-dom';
 
 export function useGridDimensions() {
   const gridRef = useRef<HTMLDivElement>(null);
   const [inlineSize, setInlineSize] = useState(1);
   const [blockSize, setBlockSize] = useState(1);
+  const deferredInlineSize = useDeferredValue(inlineSize, -1);
+  const isResizingWidth = inlineSize !== deferredInlineSize;
```
nstepien, line 73: I found a simple way to let columns flex when resizing the grid, without having to update any states.
Works surprisingly well.

I've also set the initial deferred value to -1 so columns can flex for a little longer during initial render to let the layout stabilize before we commit to measured widths.
Though it probably doesn't make much of a difference.

src/HeaderCell.tsx
```diff
@@ -205,6 +207,13 @@ export default function HeaderCell<R, SR>({
     }
   }
 
+  function onKeyUp() {
+    if (resizingRef.current) {
+      resizingRef.current = false;
+      onColumnResizeEnd();
```
nstepien, line 209: Fixed an issue where we were not resetting the resizing state when resizing via the keyboard.

test/browser/column/renderEditCell.test.tsx
```diff
@@ -18,54 +18,67 @@ interface Row {
...
-    // TODO: await userEvent.keyboard('123{enter}'); fails in FF
-    await userEvent.keyboard('{enter}123{enter}');
-    await expect.element(getCellsAtRowIndex(0).nth(0)).toHaveTextContent(/^1123$/);
+    const cell = getCellsAtRowIndex(0).nth(0);
+    await userEvent.click(cell);
+    await userEvent.keyboard('123{enter}');
+    await expect.element(cell).toHaveTextContent(/^1123$/);
+  });
+
+  it.fails('should open editor when test runs userEvent.type() on unfocused cell', async () => {
```
nstepien, line 53: Added a failing test, I need to open that vitest issue

src/DataGrid.tsx
```diff
@@ -695,21 +668,17 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
...
-  }
-
-  function handleColumnResizeEnd() {
-    // This check is needed as double click on the resize handle triggers onPointerMove
-    if (isColumnResizing) {
-      onColumnWidthsChangeRaw?.(columnWidths);
-      setIsColumnResizing(false);
+      // ensure we render the editor quickly enough,
+      // otherwise the user input might be lost in Firefox
+      flushSync(() => {
```
nstepien, line 682: resolved a TODO in tests

src/DataGrid.tsx
```diff
@@ -947,8 +916,9 @@ export function DataGrid<R, SR = unknown, K extends Key = Key>(props: DataGridPr
     }
 
     const isLastRow = rowIdx === maxRowIdx;
-    const columnWidth = getColumnWidth(column);
-    const colSpan = column.colSpan?.({ type: 'ROW', row: getActiveRow() }) ?? 1;
+    const columnWidth = columnMetrics.get(column)?.width ?? 0;
+    const colSpan =
+      getColSpan(column, lastFrozenColumnIndex, { type: 'ROW', row: getActiveRow() }) ?? 1;
```
nstepien: Any reason we were not using `getColSpan` before?

test/browser/column/resizable.test.tsx
```diff
@@ -272,12 +274,14 @@ test('should use columnWidths and onColumnWidthsChange props when provided', asy
...
-    new Map([
-      ['col1', { width: 101, type: 'measured' }],
-      ['col2', { width: 100, type: 'resized' }]
-    ])
-  );
+  await expect
+    .poll(() => onColumnWidthsChangeSpy)
+    .toHaveBeenCalledExactlyOnceWith(
+      new Map([
+        ['col1', { width: 100, type: 'measured' }],
```
nstepien, line 296: The initial state has `{ width: 101, type: 'measured' }`, since it's "measured" we let it re-flow, though we then use the initial `width: 100`, which is why it changes to 100 now.

Isn't that the expected behavior?
When would a column with a specific initial width have a different measured width?

test/browser/column/resizable.test.tsx
```diff
@@ -303,11 +307,12 @@ test('should use columnWidths and onColumnWidthsChange props when provided', asy
...
   expect(onColumnResizeSpy).not.toHaveBeenCalled();
   await expect.element(grid).toHaveStyle({ gridTemplateColumns: '120px 120px' });
+
   await resize('col2', [5, 5]);
   expect(onColumnWidthsChangeSpy).toHaveBeenCalledExactlyOnceWith(
     new Map([
-      ['col1', { width: 120, type: 'measured' }],
-      ['col2', { width: 130, type: 'resized' }]
+      ['col1', { width: 100, type: 'measured' }],
+      ['col2', { width: 150, type: 'resized' }]
```
nstepien, line 330: Likewise here, both columns have `120` widths.
The first column resets to its initial 100px width, which means the next column receives that 20px for free.

test/browser/scrollToCell.test.tsx
```diff
@@ -29,56 +29,56 @@ test('scrollToCell', async () => {
     ref,
     columns,
     rows,
-    topSummaryRows: summaryRows,
+    bottomSummaryRows: summaryRows,
```
nstepien, line 32: It was confusing to me why the top summary rows had higher row indexes than the viewport rows.

test/browser/scrollToCell.test.tsx
```diff
@@ -29,56 +29,56 @@ test('scrollToCell', async () => {
...
     rowHeight: 60
   });
 
   expect(ref.current).toBeDefined();
 
   await validateCellVisibility('0×0', true);
-  await validateCellVisibility('40×30', false);
+  await validateCellVisibility('45×30', false);
```
nstepien, line 39: Not entirely sure why, but the `40x*` column stayed visible when scrolling back to the 6th column, so I picked a different column index.

test/browser/scrollToCell.test.tsx
```diff
@@ -29,56 +29,56 @@ test('scrollToCell', async () => {
...
 
   // should not scroll if scroll to column is frozen
   ref.current!.scrollToCell({ idx: 2 });
-  await validateCellVisibility('40×30', true);
+  await validateCellVisibility('45×30', true);
 
   // should not scroll if rowIdx is header row
-  ref.current!.scrollToCell({ idx: -1 });
-  await validateCellVisibility('40×30', true);
+  ref.current!.scrollToCell({ rowIdx: -1 });
```
nstepien, line 68: We were testing the wrong index here and below

amanmahajan7:
Is it possible to add a test for https://github.com/Comcast/react-data-grid/issues/4120?

nstepien:
> Is it possible to add a test for #4120?

image

this test covers it

PR #4032, opened 2026-04-21
target baseline, lightningcss tweaks, eslint updates, `eslint.config.ts`, remove 3 `@ts-expect-error`

nstepien:
- `eslint.config.js` -> `eslint.config.ts`
  - this was the last .js file in the repo
  - https://eslint.org/docs/head/use/configure/configuration-files#typescript-configuration-files
  - https://nodejs.org/api/typescript.html


.vscode/settings.json
```diff
@@ -11,6 +11,10 @@
   "js/ts.tsdk.promptToUseWorkspaceVersion": true,
   "js/ts.tsdk.path": "node_modules/typescript/lib",
   "js/ts.experimental.useTsgo": true,
+  "eslint.options": {
+    "flags": ["unstable_native_nodejs_ts_config"]
+  },
+  "eslint.problems.shortenToSingleLine": true,
```
nstepien, line 17: Now lints won't span multiple lines anymore, which was inconvenient.
https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

test/browser/TreeDataGrid.test.tsx
```diff
@@ -123,8 +123,7 @@ function TestGrid({
 }
 
 function rowGrouper(rows: readonly Row[], columnKey: string) {
-  // @ts-expect-error
-  return Object.groupBy(rows, (r) => r[columnKey]) as Record<string, readonly R[]>;
+  return Object.groupBy(rows, (r) => r[columnKey as keyof Row]) as Record<string, readonly Row[]>;
```
nstepien, line 126: We used an incorrect `R` type, which these 3 `@ts-expect-error`s hid from us

eslint.config.ts
```diff
@@ -255,10 +255,10 @@ export default defineConfig([
       'unicode-bom': 1,
 
       // React Hooks
-      // https://www.npmjs.com/package/eslint-plugin-react-hooks
+      // https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks
+      // https://react.dev/reference/eslint-plugin-react-hooks
       'react-hooks/rules-of-hooks': 1,
       'react-hooks/exhaustive-deps': 1,
-      'react-hooks/component-hook-factories': 1,
```
nstepien, line 261: This rule is deprecated and a no-op
https://github.com/facebook/react/commit/bc249804d3c2ea6ee95bc5543c3735a65d1239b5

eslint.config.ts
```diff
@@ -365,9 +365,11 @@ copy(
       '@eslint-react/jsx-no-children-prop': 1,
       '@eslint-react/jsx-no-children-prop-with-children': 1,
       '@eslint-react/jsx-no-comment-textnodes': 1,
-      '@eslint-react/jsx-no-useless-fragment': [1, { allowExpressions: false }],
       '@eslint-react/jsx-no-key-after-spread': 1,
+      '@eslint-react/jsx-no-leaked-dollar': 1,
+      '@eslint-react/jsx-no-leaked-semicolon': 1,
```
nstepien, line 370: https://github.com/Rel1cx/eslint-react/issues/1685 💪

src/Cell.tsx
```diff
@@ -45,7 +45,7 @@ function Cell<R, SR>({
   );
   const isEditable = isCellEditableUtil(column, row);
 
-  function setActivePositionWrapper(enableEditor?: boolean) {
+  function setActivePositionWrapper(enableEditor = false) {
```
nstepien, line 48: Caught by `sonarjs/bool-param-default`, decided to keep this rule enabled.

tsdown.config.ts
```diff
@@ -1,17 +1,23 @@
 import { ecij } from 'ecij/plugin';
+import { Features } from 'lightningcss';
 import { defineConfig } from 'tsdown';
 
 import pkg from './package.json' with { type: 'json' };
 
 export default defineConfig({
   outDir: 'lib',
+  target: ['baseline-widely-available', 'node24.0.0'],
```
nstepien, line 9: They've added support for `baseline-widely-available` so I've added it.
Also added node 24 as it's the current LTS version.
There's 0 difference in the output of `index.js`/`index.js.map`/`index.d.ts`.
https://github.com/rolldown/tsdown/releases

```
ℹ targets: chrome111, edge111, firefox114, safari16.4, ios16.4, node24.0.0
```

tsdown.config.ts
```diff
@@ -1,17 +1,23 @@
...
   deps: {
     skipNodeModulesBundle: true
   },
   css: {
-    fileName: 'styles.css'
+    fileName: 'styles.css',
+    lightningcss: {
+      // https://github.com/parcel-bundler/lightningcss/issues/873
+      exclude: Features.Nesting | Features.LightDark
+    }
```
nstepien, line 20: Instead of changing the `css.target` to not break `light-dark()`, I've set this `lightningcss` configuration.

https://tsdown.dev/options/css#css-target
https://tsdown.dev/options/css#lightning-css

The only before/after difference is this:

Image

Safari only supports the prefixed property, so this is actually a bug fix!

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/user-select#browser_compatibility

vite.config.ts
```diff
@@ -56,9 +57,24 @@ export default defineConfig(
     build: {
       modulePreload: { polyfill: false },
       sourcemap: true,
-      reportCompressedSize: false,
```
nstepien, line 59: Vite is fast now, so I removed `reportCompressedSize: false`

vite.config.ts
```diff
@@ -56,9 +57,24 @@ export default defineConfig(
...
-      // https://github.com/parcel-bundler/lightningcss/issues/873
-      cssTarget: 'esnext'
+      rolldownOptions: {
+        output: {
+          codeSplitting: {
+            groups: [
+              {
+                name: 'faker',
+                test: '@faker-js/faker'
```
nstepien, line 66: Faker is the biggest dependency we have, and it doesn't get updates too often, might be worth splitting it into its own chunk.
```
dist/assets/faker-DefjNDsx.js                   410.66 kB │ gzip: 153.89 kB │ map:   696.97 kB
```

https://rolldown.rs/in-depth/manual-code-splitting#improve-loading-performance

Wonder if we should split react/react-dom together too 🤔

vite.config.ts
```diff
@@ -56,9 +57,24 @@ export default defineConfig(
...
+            ]
+          }
+        }
+      }
+    },
+    css: {
+      lightningcss: {
+        // https://github.com/parcel-bundler/lightningcss/issues/873
+        exclude: Features.Nesting | Features.LightDark
+      }
```
nstepien, line 77: Applied the same lightningcss strategy here to fix the missing `-webkit-user-select` property.

test/browser/TreeDataGrid.test.tsx
```diff
@@ -123,8 +123,7 @@ function TestGrid({
 }
 
 function rowGrouper(rows: readonly Row[], columnKey: string) {
```
amanmahajan7, line 125: Worth deduping?
https://github.com/Comcast/react-data-grid/blob/main/test/browser/utils.tsx

tsdown.config.ts
amanmahajan7: https://github.com/rolldown/tsdown/issues/893 💪

amanmahajan7 (approved):
Nice 👍

test/browser/TreeDataGrid.test.tsx
nstepien: We have 2 in tests, and 1 in a website route, not sure it's worth it 🤷‍♂️

PR #4040, opened 2026-05-12
support dynamic height rows for non-virtualized rendering

fastfrwrd:
### Summary

- adds an option to use a string for rowHeight, which can be set to keyword values like `auto`, `fit-content`, `stretch`, etc. 
- updates `useViewportRows` to handle this, which adds new required new props `element` and `gridHeight` when using a string-based height.
- updates page up and page down to handle this.
- when using a string-based height, obviously the rowHeight calculations required for virtualization are not possible, so virtualization is disabled by default when using string rowHeights, and an error is thrown if you attempt to force virtualization on with rowHeights set to a string value.

### Background

At Grafana, our react-data-grid fork has been powering our Table panel for the past 9 months or so. One of the features we really need in our table panel is a mode where dynamic content can be supplied by a datasource (see [the Grafana docs](https://grafana.com/docs/grafana/latest/visualizations/panels-visualizations/visualizations/table/#markdown--html)).

We forked from the package when the decision was made in this project to [drop support for React 18 on a non-breaking version update](https://github.com/Comcast/react-data-grid/pull/3503), but now that Grafana 13 is released and we are getting closer to having React 19 as the only supported version for Grafana going forward, we want to get back on the main package. But to do that we need dynamic row height support since we have added that in our fork and we now depend on that being available in some form. We're open to the specific form it takes, though, so please let us know if you have concerns with the approach in here.

### DIFF
--- a/src/hooks/useViewportRows.ts
+++ b/src/hooks/useViewportRows.ts
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useMemo, type RefObject } from 'react';
 
 import { floor, max, min } from '../utils';
 
@@ -12,7 +12,7 @@ interface ViewportRowsBaseArgs<R> {
 
 interface ViewportRowsArgsStringHeight {
   rowHeight: string;
-  element: HTMLElement | null;
+  gridRef: RefObject<HTMLDivElement | null>;
   gridHeight: number;
 }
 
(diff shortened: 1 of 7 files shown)


src/hooks/useViewportRows.ts
```diff
@@ -28,6 +41,71 @@ export function useViewportRows<R>({
       };
     }
 
+    if (typeof rowHeight === 'string') {
+      const { element, gridHeight } = rest as ViewportRowsArgsStringHeight;
```
fastfrwrd: i expected this typeguard to work but it didn't, which led to the cast. I'm gonna dig into whether there's something I could do differently to make it work.

PR #4075, opened 2026-06-19
TypeScript 7.0.1-rc

nstepien:
https://devblogs.microsoft.com/typescript/announcing-typescript-7-0-rc/
I couldn't figure out how to make the vscode extension to use the locally installed rc version instead of the bundled version, but not a showstopper as they'd behave the same anyway.


package.json
```diff
@@ -73,7 +72,8 @@
     "react": "^19.2.5",
     "react-dom": "^19.2.5",
     "tsdown": "^0.22.0",
-    "typescript": "~6.0.3",
+    "typescript": "npm:@typescript/typescript6@~6.0.1",
```
amanmahajan7, line 75: time to migrate to oxlint?

nstepien: Yup, although I think it supports tsgo but not the rc just yet

PR #4076, opened 2026-06-23
Vite 8.1

nstepien:
- https://vite.dev/blog/announcing-vite8-1
  - enable imports maps (`build.chunkImportMap`), tried it and it seems to work well 👍 
    - https://vite.dev/guide/features#chunk-import-map-optimization
    - https://vite.dev/config/build-options#build-chunkimportmap
    - image
  - I've set `css.transformer: 'lightningcss'`, seems to work fine, no build output change
  - I tried enabling `experimental.bundledDev` but it did not work.
    - It seems to work even worse when `build.chunkImportMap` is enabled.
    - It also breaks tests, although we could disable it with `!isTest` if it worked fine with the dev server
- https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/CHANGELOG.md

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -55,6 +55,7 @@ export default defineConfig(
     cacheDir: '.cache/vite',
     clearScreen: false,
     build: {
+      chunkImportMap: true,
       modulePreload: { polyfill: false },
       sourcemap: true,
       rolldownOptions: {
@@ -71,6 +72,7 @@ export default defineConfig(
       }
     },
     css: {
+      transformer: 'lightningcss',
       lightningcss: {
         // https://github.com/parcel-bundler/lightningcss/issues/873
         exclude: Features.Nesting | Features.LightDark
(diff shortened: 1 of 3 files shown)


vite.config.ts
```diff
@@ -71,6 +72,7 @@ export default defineConfig(
       }
     },
     css: {
+      transformer: 'lightningcss',
```
amanmahajan7, line 75: I thought this was the default 😕 
https://vite.dev/guide/migration#css-minification-by-lightning-css

amanmahajan7 (commented):
Maybe we need to update tanstack to make `bundledDev` work.

vite.config.ts
nstepien: Only as the minifier, not as the transformer
https://vite.dev/config/shared-options#css-transformer

nstepien:
> Maybe we need to update tanstack to make `bundledDev` work.

It's already updated, I think it's just bugs in the experimental implementation

PR #4119, opened 2026-07-27
Migrate from ESLint to oxlint

nstepien:
Replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter/) for JS/TS linting. ESLint stays only for markdown, since `@eslint/markdown` is built on ESLint's language plugin API, which oxlint does not support.

`.oxlintrc.json` was generated with `@oxlint/migrate --type-aware --with-nursery` and carries over **545 rules** at their original `warn` severity. `@eslint-react` and `sonarjs` load as `jsPlugins`; `typescript`, `react` and `vitest` use oxlint's native implementations, so `typescript-eslint`, `eslint-plugin-react-hooks` and `@vitest/eslint-plugin` are dropped.

### Scripts

| before | after |
| --- | --- |
| `node --run eslint` | `node --run lint` (oxlint) + `node --run lint:md` (markdown) |
| `node --run eslint:fix` | `node --run lint:fix` / `node --run lint:md:fix` |

`--report-unused-disable-directives` is passed to oxlint to match ESLint 9's default behaviour. CI, `publish.yml`, `AGENTS.md` and `.vscode/settings.json` are updated to match.

### Rules that could not be migrated

- **The 13 React Compiler rules** (`react-hooks/purity`, `preserve-manual-memoization`, `set-state-in-effect`, …). oxlint only ships these as a single experimental bundled `react/react-compiler` rule, and will split them once they stabilise. This is the most substantive loss — happy to keep a narrow ESLint config for them alongside markdown if we'd rather not go without.
- **`@eslint-react/no-implicit-key`, `no-leaked-conditional-rendering`, `no-unused-props`.** oxlint's JS plugin API cannot supply `@typescript-eslint` parser services, so these throw on every file rather than running.
- `@typescript-eslint/naming-convention`, `prefer-destructuring`, `no-unused-private-class-members`, `require-atomic-updates`, `one-var`.

Everything else is covered, including all 40 type-aware `@typescript-eslint` rules via `oxlint-tsgolint`.

### Suppressions added

Three, each a behavioural difference rather than a real defect:

- `sonarjs/no-redundant-optional` in `src/types.ts` — cannot see `exactOptionalPropertyTypes` without type information, so it flags the intentional `?: boolean | undefined`.
- `vitest/no-conditional-expect` in `scrollToCell.test.tsx` — oxlint flags a helper function that is not a test block.
- The existing directive in `src/globals.d.ts` moved down one line, because oxlint reports the index signature where ESLint reported the interface.

The other 15 `eslint-disable` comments work unchanged — oxlint accepts the `@typescript-eslint/` prefix, and the unused-directive check confirms all of them are still live.

### Incidental

`@eslint-react/component-hook-factories` was removed: it does not exist in that plugin and ESLint silently ignored it because it was `off`. It was most likely meant to be `react-hooks/component-hook-factories`.

`.agents` is ignored by both linters, mirroring the existing `.claude` entry.

### Known issue: `--type-aware` on Windows

oxlint 1.76.0 only looks for `node_modules/.bin/tsgolint.exe`, but npm writes `.cmd`/`.ps1` shims there, so `node --run lint` fails locally on Windows with *"Failed to find tsgolint executable"*. `TSGOLINT_PATH` does not help and 1.76.0 is the latest release. CI on `ubuntu-latest` is unaffected, and oxlint exits `1` rather than silently skipping the rules. Local workaround:

```powershell
Copy-Item node_modules\@oxlint-tsgolint\win32-x64\tsgolint.exe node_modules\.bin\tsgolint.exe -Force
```

The oxc VS Code extension has an `oxc.path.tsgolint` setting that avoids this for editor diagnostics.

### Verification

`node --run lint`, `node --run lint:md`, `node --run typecheck`, `node --run format:check` all pass, and the full suite is green (353 passed, 2 skipped across 65 files).


.oxlintrc.json
```diff
@@ -0,0 +1,955 @@
+{
+  "$schema": "./node_modules/oxlint/configuration_schema.json",
+  "plugins": [],
+  "categories": {
+    "correctness": "off"
+  },
+  "options": {
+    "typeAware": true
```
amanmahajan7: Worth enabling `typeCheck`? Would that cover all the files checked by the `typecheck` command?

.oxlintrc.json
```diff
@@ -0,0 +1,955 @@
...
+        "@eslint-react/web-api-no-leaked-event-listener": "warn",
+        "@eslint-react/web-api-no-leaked-fetch": "warn",
+        "@eslint-react/web-api-no-leaked-intersection-observer": "warn",
+        "@eslint-react/web-api-no-leaked-interval": "warn",
+        "@eslint-react/web-api-no-leaked-resize-observer": "warn",
+        "@eslint-react/web-api-no-leaked-timeout": "warn",
+        "@eslint-react/naming-convention-context-name": "warn",
+        "@eslint-react/naming-convention-id-name": "warn",
+        "@eslint-react/naming-convention-ref-name": "warn",
+        "sonarjs/anchor-precedence": "warn",
```
amanmahajan7: Wonder if we should look into unicorn rules. Seems like it can replaces some of the the sonarjs rules

nstepien: I wouldn't mind removing the sonarjs rules

amanmahajan7:
This will be a great PR to finish and merge now that oxlint supports react compiler rules. I think we can get rid of eslint completely. For markdown, we can wait for oxlint to support it https://github.com/oxc-project/oxc/issues/18407

.vscode/settings.json
```diff
@@ -7,10 +7,13 @@
   },
   "editor.defaultFormatter": "oxc.oxc-vscode",
   "js/ts.experimental.useTsgo": true,
-  "eslint.options": {
-    "flags": ["unstable_native_nodejs_ts_config"]
-  },
+  // ESLint is only used for markdown; JS/TS is linted by the oxc extension
+  "eslint.validate": ["markdown"],
   "eslint.problems.shortenToSingleLine": true,
+  "oxc.path.tsgolint": "node_modules/.bin/tsgolint.cmd",
```
nstepien, line 13: need this otherwise type-aware rules don't work, but that path won't work outside windows... should we only set this path in our own individual settings?

CLAUDE.md
```diff
@@ -0,0 +1 @@
+@AGENTS.md
```
nstepien, line 1: https://code.claude.com/docs/en/memory#agents-md

.oxlintrc.json
nstepien: Done.
I've kept the `typecheck` command since it's very useful with `--watch`.
oxlint does not support `--watch`

.oxlintrc.json
nstepien: I've removed the sonarjs rules. We can look into enabling [unicorn rules](https://oxc.rs/docs/guide/usage/linter/rules.html?sort=name&dir=asc&scope=unicorn) in a separate PR

nstepien:
Ready for review.
There's still some work to be done, like enabling the react compiler rules, but we can leave that for a separate PR.

> I think we can get rid of eslint completely. For markdown, we can wait for oxlint to support it [oxc-project/oxc#18407](https://github.com/oxc-project/oxc/issues/18407)

I've kept eslint to lint the markdown files. It's fast and shouldn't be much of an issue.

.vscode/settings.json
amanmahajan7: They don't work with the extension or when running the npm command?

.oxlintrc.json
```diff
@@ -0,0 +1,592 @@
...
+  "plugins": ["eslint", "react", "typescript", "vitest"],
+
+  "jsPlugins": ["@eslint-react/eslint-plugin"],
+
+  "options": {
+    "denyWarnings": true,
+    "reportUnusedDisableDirectives": "warn",
+    "respectEslintDisableDirectives": false,
+    "typeAware": true,
+    "typeCheck": true
```
amanmahajan7, line 13: Does this check all the files previously checked by typescript?

.oxlintrc.json
```diff
@@ -0,0 +1,592 @@
...
+      row.children[1].textContent,
+      row.querySelector(':any-link').textContent
+    ])
+    .filter(([group]) => groups.has(group))
+    .map(([group, rule]) => `"${group}/${rule}": "warn",`)
+    .toArray()
+    .sort()
+    .join('\n')
+);
+  */
```
amanmahajan7, line 48: worth keeping if agents can crawl and copy the rules for us?

amanmahajan7 (approved):

.vscode/settings.json
nstepien: both
`oxc.path.tsgolint` fixes the extension
the OXLINT_TSGOLINT_PATH env variables fixes the cli

.oxlintrc.json
nstepien: Seems like it

.oxlintrc.json
nstepien: Maybe, maybe not, at least this will consistently order rules the same way

PR #4146, opened 2026-08-19
Vitest 5

nstepien:
- https://vitest.dev/blog/vitest-5.html
- https://main.vitest.dev/guide/migration
- https://github.com/vitest-dev/vitest/releases/tag/v5.0.0

TODO
- [ ] ~remove `overrides` in `package.json` if possible~


test/setupBrowser.ts
```diff
@@ -5,6 +5,12 @@ import 'vitest-browser-react';
 import { configure } from 'vitest-browser-react/pure';
 import { locators, userEvent, type Locator, type LocatorByRoleOptions } from 'vitest/browser';
 
+// avoid priting huge stack traces
+// chromium defaults to 10
+// firefox defaults to 128
+// @ts-expect-error
+Error.stackTraceLimit = 10;
```
nstepien, line 13: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stackTraceLimit

vite.config.ts
```diff
@@ -174,7 +161,6 @@ export default defineConfig(({ isPreview }): ViteUserConfig => ({
     },
     projects: [
       {
-        extends: true,
```
amanmahajan7: Is this not needed anymore?

vite.config.ts
```diff
@@ -105,6 +105,7 @@ export default defineConfig(({ isPreview }): ViteUserConfig => ({
   test: {
     dir: 'test',
     globals: true,
+    injectCjsGlobals: false,
```
amanmahajan7, line 108: Is this needed?

nstepien: No, but it seems good to have to validate that we don't depend on them
https://vitest.dev/config/injectcjsglobals

.github/workflows/ci.yml
```diff
@@ -56,10 +56,7 @@ jobs:
         uses: actions/upload-artifact@v7
         with:
           name: test-artifacts
-          path: |
-            test/**/__screenshots__/**
-            test/**/__traces__/**
-            .vitest-attachments/test/**
+          path: .vitest/attachments/**
```
amanmahajan7, line 59: Do the failed screenshots go under `vitest/attachment`. Wonder if we should archive the whole `.vitest` folder?

nstepien: > Do the failed screenshots go under `vitest/attachment`.

Yes
https://main.vitest.dev/guide/migration/#generated-reports-and-artifacts-use-the-vitest-directory

>Wonder if we should archive the whole `.vitest` folder?

I don't think we generate anything else under `.vitest` at the moment.
Traces are also saved under `.vitest/attachments/`.
We can change this later if we need to.

PR #4157, opened 2026-08-27
oxlint: add unicorn rules

nstepien:
https://oxc.rs/docs/guide/usage/linter/rules.html?sort=name&dir=asc&scope=unicorn


amanmahajan7 (commented):
Nice! I will review soon. What is your take on unicorn rules? Also, do we need to disable `correctness`?

nstepien:
> What is your take on unicorn rules?

Most are good rules, reminds me of the sonarjs rules.
Some other rules may be a bit too pedantic though. I've kept them enabled since they don't affect the repo as it stands, but I wonder if it'd be better to disable them.

> Also, do we need to disable `correctness`?

Maybe. Since we configure all the rules ourselves, and list all the plugins we use, it doesn't really affect us, until oxlint adds new rules.

amanmahajan7 (approved):

PR #4160, opened 2026-08-29
Move .cache, tweak eslint

nstepien:
- I've moved the common cache folder to `node_modules/.cache`
  - the location of the `.cache` folder was convenient, but we rely on it less and less, so we rarely need to manually delete it nowadays
  - we have to manually ignore it in all tools
    - vite plugins handle files under `.cache/vite` if we forget to configure all the `exclude`s
    - this caused issues with the react compiler, which is why I've opened this PR
- I've also tweaked the eslint config
  - much like oxfmt/oxlint, it now respects our `.gitignore`, so we don't need to keep these in sync anymore
    - https://eslint.org/docs/latest/use/configure/ignore#include-gitignore-files
  - keeping `globalIgnores(['**/*.{js,ts,tsx}'])` made no difference so I've removed it as well
    - FYI we could speed up eslint with `globalIgnores(['src/', 'test/', 'website/'])`, but it's negligible
  - `linterOptions` is now inlined in the only config object left

### DIFF
--- a/tsconfig.node.json
+++ b/tsconfig.node.json
@@ -2,7 +2,8 @@
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "lib": ["ESNext"],
-    "module": "NodeNext"
+    "module": "NodeNext",
+    "types": ["node"]
   },
   "include": ["eslint.config.ts"]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/.cache
 /.claude
 /dist
 /lib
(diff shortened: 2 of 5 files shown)


tsconfig.node.json
```diff
@@ -2,7 +2,8 @@
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "lib": ["ESNext"],
-    "module": "NodeNext"
+    "module": "NodeNext",
+    "types": ["node"]
```
amanmahajan7, line 6: Were the types failing?

amanmahajan7 (approved):

tsconfig.node.json
nstepien: Needed to import `node:url`, and use `import.meta.resolve`

PR #4168, opened 2026-09-03
Remove tsgolint path settings

nstepien:
We probably don't need to keep these, pretty sure I'm the only one affected because of my environment
https://github.com/oxc-project/oxc/pull/26234

Please check locally


amanmahajan7 (approved):
Works for me.

PR #4171, opened 2026-09-05
website: various router tweaks

nstepien:


src/EditCell.tsx
```diff
@@ -27,6 +27,7 @@ import type {
  * We must also rely on React's event capturing/bubbling to handle elements rendered in a portal.
  */
 
+// TODO: remove when all browsers support the scheduler APIs
```
nstepien, line 30: Still no support from webkit...

website/routes/AllFeatures.tsx
```diff
@@ -10,38 +10,30 @@ import { textEditorClassname } from '../../src/editors/renderTextEditor';
 import { useDirection } from '../directionContext';
 
 export const Route = createFileRoute('/AllFeatures')({
+  // split the loader to avoid importing faker in the site's entry js
+  codeSplitGroupings: [['component', 'loader']],
```
nstepien, line 14: Alternatively we could `import(faker)` in the loader itself, but this is cleaner IMO.

website/routes/AllFeatures.tsx
```diff
@@ -10,38 +10,30 @@ import { textEditorClassname } from '../../src/editors/renderTextEditor';
...
   component: AllFeatures,
-  loader() {
```
nstepien, line 14: Turns out the router's vite plugin DOES NOT split functions when using the method shorthand syntax, so we can't use `loader() {}`.
The same applies to `component`, so `component() { return <AllFeatures /> }` would keep the entire component in the entry bundle 😬 

I'll open an issue/PR in the TSR repo later.

website/routes/AllFeatures.tsx
```diff
@@ -10,38 +10,30 @@ import { textEditorClassname } from '../../src/editors/renderTextEditor';
...
-    rows ??= createRows();
-    return rows;
-  }
+  loader,
+  staleTime: Number.POSITIVE_INFINITY
```
nstepien, line 17: That way we don't re-generate the rows unnecessarily.
The rows can still be gc'd, `defaultGcTime` defaults to 5 minutes so it should be fine.

website/router.tsx
```diff
@@ -6,21 +6,28 @@ export const router = createRouter({
...
+  defaultPendingComponent: PendingComponent,
   defaultErrorComponent: ErrorComponent,
   defaultNotFoundComponent: NotFound,
+  defaultPendingMs: 0,
   defaultPendingMinMs: 0,
   defaultPreload: 'intent',
   defaultStructuralSharing: true,
-  scrollRestoration: true
+  scrollRestoration: true,
+  scrollToTopSelectors: ['.rdg']
```
nstepien, line 17: https://comcast.github.io/react-data-grid/#/ColumnSpanning
Right now if you open the "Column Spanning" page for example, scroll the grid, and navigate to the "Column Grouping", then the second grid will have the same scroll position as the previous grid, that's probably because the two components are essentially the same so React re-uses the existing DOM/components.

Scroll restoration still works.

PR #4181, opened 2026-09-13
website: adopt `<ViewTransition>` in a couple places

nstepien:
https://react.dev/reference/react/ViewTransition
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:active-view-transition


amanmahajan7 (approved):
👍

PR #4182, opened 2026-09-14
Run Firefox test files in parallel

nstepien:
Removes `fileParallelism: false` from the Firefox browser instance.

Vitest runs parallel test files as pages of the same browser, but Firefox pages opened in the same browser share focus and mouse states, so test files interfere with each other:

- Playwright's Firefox raises the page's window before dispatching mouse events, which blurs the focused element of all the other pages: `document.hasFocus()` returns `false` in their iframes, editors close, etc.
- a click in one page dispatches `pointermove`/`pointerup` to the pointer capture target of other pages, ending ongoing drags (column resize, drag fill).

With file parallelism enabled, 149/177 Firefox tests failed.

Changes:

- add `PlaywrightBrowserPerSessionProvider`, which extends Vitest's Playwright provider: every session after the first one gets its own browser. It creates the session's context in a new browser before calling `openPage()`, and the base provider reuses that context.
- remove `fileParallelism: false` from the Firefox instance
- set `maxWorkers: 8`

Notes:

- the provider relies on the base provider reusing a session's existing context, which isn't a documented behavior.
- `fileParallelism: false` also made Vitest run Firefox test files after the other browsers' tests. Firefox now runs at the same time as Chromium and WebKit.
- locally, `node --run test` went from ~60s to ~24s.


amanmahajan7:
Worth opening an issue upstream?

PR #4183, opened 2026-09-14
EditCell: support `:modal` editors

nstepien:
- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:modal
- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal


website/routes/CommonFeatures.tsx
```diff
@@ -78,177 +59,191 @@ interface Row {
...
+    key: 'country',
+    name: 'Country',
+    renderEditCell: (p) => (
+      <select
+        autoFocus
+        className={textEditorClassname}
+        value={p.row.country}
+        onChange={(e) => p.onRowChange({ ...p.row, country: e.target.value }, true)}
+        onFocus={(event) => {
+          event.target.showPicker();
```
nstepien: Somehow it's not opening automatically anymore, had to add this.
I tried calling `showPicker()` in the `ref` prop, but that throws an error as it requires an user action.

website/routes/CommonFeatures.tsx
```diff
@@ -78,177 +59,191 @@ interface Row {
...
-        return dateFormatter.format(props.row.endTimestamp);
-      }
-    },
-    {
-      key: 'budget',
-      name: 'Budget',
-      renderCell(props) {
-        return currencyFormatter.format(props.row.budget);
-      }
+            <li>
```
nstepien, line 172: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/menu

website/routes/CommonFeatures.tsx
```diff
@@ -78,177 +59,191 @@ interface Row {
...
+              <button type="button" command="close" commandfor={dialogId}>
```
nstepien, line 173: https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#commandfor

website/routes/CommonFeatures.tsx
```diff
@@ -78,177 +59,194 @@ interface Row {
...
+    name: 'Country',
+    renderEditCell: (p) => (
+      <select
+        autoFocus
+        className={textEditorClassname}
+        value={p.row.country}
+        onChange={(e) => p.onRowChange({ ...p.row, country: e.target.value }, true)}
+        onFocus={(event) => {
+          try {
+            event.target.showPicker();
```
nstepien, line 105: Somehow it's not opening automatically anymore, had to add this.
I tried calling `showPicker()` in the `ref` prop, but that throws an error as it requires an user action.

website/routes/CommonFeatures.tsx
```diff
@@ -78,177 +59,194 @@ interface Row {
...
+              display: flex;
+              justify-content: end;
+              gap: 8px;
+              margin: 0;
+              padding: 0;
+            `}
           >
-            <dialog open>
-              <input
-                autoFocus
```
nstepien, line 167: The dialog automatically focuses the first focusable element

PR #4189, opened 2026-09-16
Enforce expect.element() for DOM assertions

amanmahajan7:
Did not find any issues in our codebase but probably good to enforce it. Thoughts?

image


nstepien (commented):
`toMatchTextContent` and `toHaveRole` are missing, maybe more
https://vitest.dev/api/browser/assertions.html#tomatchtextcontent
https://vitest.dev/api/browser/assertions.html#tohaverole

